PR #360, opened 2015-01-14
Fixed italian 'Keywords' translation + BUG in date format

lrkwz:
Corrected the translation for "Keywords"


lrkwz:
Although the translation is somehow only cosmetic stuff, the date format is a real issue: in admin's listing using A instead of Y cause AM/PM visualization, g instead of j outputs hour in place of day.

mikejolley:
Hey can you do this on https://www.transifex.com/projects/p/wp-job-manager/ please? Otherwise it will be overwritten.

lrkwz:
Ok, done

PR #459, opened 2015-05-06
Remove unnecessary closing tag </a> in admin settings

calvincanas:


mikejolley:
Should be a closing label.

PR #475, opened 2015-06-03
Fix translate

damienlejart:
In fr_FR file is missing a variable into the translation string (line 1292)
"%s listed successfully. To view your listing <a href=\"%s\">click here</a>."

Therefore in the template job-submited.php the Post's link is wrong.


mikejolley:
I've fixed this on https://www.transifex.com/projects/p/wp-job-manager/translate/#fr_FR/wp-job-managerpot/39185211?q=To%20view%20your%20listing

PR #483, opened 2015-06-12
Turkish Translation

emirsavran:
There was only 3 items that was translated into Turkish. Now this is
fully translated.


mikejolley:
Please add it to https://www.transifex.com/projects/p/wp-job-manager/ :) Then it will be pulled into the next update.

PR #518, opened 2015-08-20
Account password wording support for WordPress >= 4.3

tripflex:
With the release of WordPress 4.3 passwords are no longer sent via the `wp_new_user_notification` function, instead a link to set a password is sent.

This commit just updates the wording in the `account-signin.php` to output

`A link to set a password will be automatically emailed to you.`
instead of
`A password will be automatically emailed to you.`

If the WordPress install is version greater than or equal to 4.3, otherwise it will still show the standard wording.

### DIFF
--- a/templates/account-signin.php
+++ b/templates/account-signin.php
@@ -13,7 +13,7 @@
 	</fieldset>
 
 <?php else :
-
+	global $wp_version;
 	$account_required             = job_manager_user_requires_account();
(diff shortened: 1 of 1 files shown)


mikejolley:
I don't think this is worth a conditional check. The wording should be ambiguous enough to meet both.

tripflex:
Yeah wasn't sure which way you wanted to go with that, this works for me though, thanks!

PR #520, opened 2015-08-26
Added HTML5 history management

lindsaymacvean:
* store each search as a url in the History of the browser. 
* This means each url can be shared on social media etc and routes back to the specific search query which was intended by the original user.
* The search queries can be tracked with a custom google analytics script
* The user can click back through searches. 

Tested:
Firefox 39.0.3
Chrome Version 44.0.2403.155 (64-bit)

More battle testing is recommended, if there are any unit tests I can run I would be happy to do so.


lindsaymacvean:
Thanks @aheckler I'll wait for mike to get back on the 5th. Warmest Lindsay

assets/js/ajax-filters.js
```diff
@@ -175,15 +183,13 @@ jQuery( document ).ready( function ( $ ) {
...
 
-	$( '#search_keywords, #search_location, .job_types :input, #search_categories, .job-manager-filter' ).change( function() {
-		var target   = $( this ).closest( 'div.job_listings' );
-		target.triggerHandler( 'update_results', [ 1, false ] );
-		job_manager_store_state( target, 1 );
-	} )
+	//end of the initial update_results listener
 
-	.on( "keyup", function(e) {
+	// Changed this to just listen for enter button instead of listening for change event which was calling the loop twice.
```
mikejolley: I think this will be less reliable. Users don't need to hit enter to get results. They could for example just enter text and tab out of the field. Change is better IMO

assets/js/ajax-filters.js
```diff
@@ -255,26 +262,67 @@ jQuery( document ).ready( function ( $ ) {
...
 
-	var location = document.location.href.split('#')[0];
-
 	function job_manager_store_state( target, page ) {
+		// changed this to ? because this is the standard query string identifier in php
+		// and changed to window.location for better cross browser support
+		// and moved into the function so it rechecks everytime the function is called
+		var location = window.location.href.split('?')[0];
+		var query = window.location.href.split('?')[1];
+		query = query.split('&');
```
mikejolley: Won't this affect WP queries if the querystring conflicts with something used by wp query?

lindsaymacvean: Ok so thought about this and the real issue is that we need to isolate GET variables specific to WP job manager. Using ? parameters does not affect WP_Query as per the [final comment on this thread](https://core.trac.wordpress.org/ticket/16373). Its when we push or replace HTML5 window.history on circa line [302](https://github.com/lindsaymacvean/WP-Job-Manager/blob/master/assets/js/ajax-filters.js#L302) that the problem occurs because this may confuse the WP engine by removing WP query parameters necessary to direct to the correct page.

Whether we switch back to a # or not does not change the fundamental challenge of isolating just the three parameters relevant to our api, using them and if necessary adding updated parameters into the HTML5 history all while keeping any other parts of the complete URL string intact, both before and after the parameters relevant to us.

I will write something to update this.


[1](http://superuser.com/questions/498617/does-an-anchor-tag-come-before-the-query-string-or-after)

assets/js/ajax-filters.js
lindsaymacvean: Yes you are right, my comment is misleading, it was not the change event that was causing the error but the invocation of `job_manager_store_state` on line [181](https://github.com/Automattic/WP-Job-Manager/blob/master/assets/js/ajax-filters.js#L181), I will replace the change event listener and instead comment out the invocation and rely instead on line [31](https://github.com/lindsaymacvean/WP-Job-Manager/blob/master/assets/js/ajax-filters.js#L31) to call the HTML5 state management function only once

lindsaymacvean: further to my last comments the reason the form was submitting twice was because of implicit submission on the enter key I opened [a question about this on stackoverflow](http://stackoverflow.com/questions/33026415/keyup-13-event-repeats-implied-submission-submit-form-on-enter-issue) and have now fixed the issue by preventing default on the keypress 13 event.

assets/js/ajax-filters.js
lindsaymacvean: So I can either write a function to get and set our specific query strings or I can use this plugin https://github.com/medialize/URI.js which adds extra over head what do you think @mikejolley ??

lindsaymacvean: So I can either write a url query setter and getter myself to surgically alter our specific queries whilst leaving the rest of the URI string intact, or I can use this plugin [https://github.com/medialize/URI.js](https://github.com/medialize/URI.js) but perhaps that adds unneccessary overhead? What do you think @mikejolley ??

mikejolley: Would rather not add more libs if possible - I'd stick with the hash which avoids the issue completely.

mikejolley:
Closing this due to lack of traction. Worried about the comments above causing other issues and conflicts.

PR #533, opened 2015-09-24
Update wp_new_user_notification

### LINKED ISSUE #532: "wp_new_user_notification" still trying to send password
We're using a deprecated argument in `wp_new_user_notification()` here:

https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L402

Should be something like:

`wp_new_user_notification( $user_id, null, 'both' );`

Or perhaps the third argument could be filterable? ¯\\_(ツ)_/¯ More info:

https://core.trac.wordpress.org/ticket/33654

Causing PHP notices in Kayako ticket 2354997.

kraftbj:
4.3 stopped sending out e-mailed passwords and changed this function up a bit. In order to send notifications to the user (and not only the admin), we need to add a third arg. 

Fixes #532

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -327,6 +327,7 @@ if ( ! function_exists( 'job_manager_create_account' ) ) :
  */
 function wp_job_manager_create_account( $args, $deprecated = '' ) {
 	global $current_user;
+	global $wp_version;
 
 	// Soft Deprecated in 1.20.0
 	if ( ! is_array( $args ) ) {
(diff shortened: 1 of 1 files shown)


kraftbj:
From some conversation in Slack about it:

We should actually have a conditional to check for pre-4.3.1. Previously to 4.3, it would expect the password as the 2nd arg, so we maintain BC until our minimum version is 4.3.1+.

In 4.3, the 2nd arg was repurposed to the now 3rd arg. It's okay to send the password over—it'll act as "both" functionally. Part of this mess is because wp_new_user_notification is a pluggable function, so leaving it in the pre-4.3 state for 4.3 will maintain functionality in un-updated plugged functions.

wp-job-manager-functions.php
```diff
@@ -399,7 +400,12 @@ function wp_job_manager_create_account( $args, $deprecated = '' ) {
     }
 
     // Notify
-    wp_new_user_notification( $user_id, $password );
+    if ( version_compare( $wp_version, '4.3.0', '<=' ) ) {
+    	wp_new_user_notification( $user_id, $password );
+    }
+    elseif {
```
aheckler: To nitpick: `elseif` should be on the same line as the `}` above it per WP coding standards: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style 

Also, not sure why it's `elseif` instead of just `else`?

kraftbj: Noting that you're right on this and that the version check should have been `if ( version_compare( $wp_version, '4.3.1', '<' ) ) {`

PR #543, opened 2015-10-15
remove RP4WP template override because of alternative RP4WP template …

barrykooij:
…setup


barrykooij:
@mikejolley because of RP4WP's new layout configurator, the default WPJM content listing looks pretty bad. Using RP4WP's default's template files would be better. Let me know if you want to talk this over further.

barrykooij:
Gave this more thought and with all the relevant meta data WPJM adds, I rather keep it like it already is.

PR #551, opened 2015-10-30
Allow different new user notification method

jamesjonesphoenix:
Allows us to add our own new user notification in place of the default
one. This is by placing an action hook in wp_job_manager_create_account.
The default notification code is placed in it's own function and hooked
in.

I added this because on my site I'd rather use the WooCommerce mailer. I
imagine I wouldn't be the only one who would want a nice looking email
for the notification. Same issue as this post: https://wordpress.org/support/topic/new-user-email-notification-1

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -1,139 +1,139 @@
 <?php
 if ( ! function_exists( 'get_job_listings' ) ) :
-/**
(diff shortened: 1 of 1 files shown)


mikejolley:
Can you re-do this without changing the entire file's indentation in the same commit - impossible to see your mod.

PR #557, opened 2015-11-05
Fix job_manager_user_can_edit_job()

adamhurlebaus:
Users who have created job listings are unable to make changes (edit, mark filled, delete) from the page with the [job_dashboard] shortcode after the most recent update (1.23.10).  

Changing $job->post_author !== get_current_user_id()  to $job->post_author != get_current_user_id() solved the problem.  There may be a better fix if the variable data types need to match by using !== but it seems like it should be fine.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -449,7 +449,7 @@ function job_manager_user_can_edit_job( $job_id ) {
 	} else {
 		$job      = get_post( $job_id );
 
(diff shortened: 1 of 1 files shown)


adamhurlebaus:
Related Bug/Issue noted on the support forum: https://wordpress.org/support/topic/manage-jobs-page-invalid-listing

mikejolley:
Whoops missed this PR - already committed.

adamhurlebaus:
No worries!  Thanks for taking care of it so quickly.

PR #561, opened 2015-11-10
New user notification action

jamesjonesphoenix:
Allows us to add our own new user notification in place of the default
one. This is by placing an action hook in wp_job_manager_create_account.
The default notification code is placed in it's own function and hooked
in.

I added this because on my site I'd rather use the WooCommerce mailer. I
imagine I wouldn't be the only one who would want a nice looking email
for the notification.
Same issue as this post:
https://wordpress.org/support/topic/new-user-email-notification-1

Now without indent change.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -332,7 +332,6 @@ if ( ! function_exists( 'wp_job_manager_notify_new_user' ) ) :
 		}
 	}
 endif;
-add_action( 'job_manager_new_user_notification', 'wp_job_manager_notify_new_user', 10, 2);
 
 if ( ! function_exists( 'job_manager_create_account' ) ) :
 /**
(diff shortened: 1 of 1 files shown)


kraftbj:
Devil's advocate: `wp_new_user_notification` is a pluggable function, so any plugin could declare their own `wp_new_user_notification` and core's wouldn't run. What about defining it that way?

fourleafed:
-- Please reply above this line --

Hi Automattic/wp-Job-Manager We have received your email. We will
respond to you as soon as possible.Kind regards!

-----------------------------------------------------------
## Automattic/wp-Job-Manager sent a message, on Nov 12 @ 0:00 (CET):



	Devil's advocate: wp_new_user_notification is a pluggable function,
so any plugin could declare their own wp_new_user_notification and
core's wouldn't run. What about defining it that way? 

	-
 Reply to this email directly or view it on GitHub [1].

Links:
------
[1]
https://github.com/Automattic/WP-Job-Manager/pull/561#issuecomment-155936765


-----------------------------------------------------------

jamesjonesphoenix:
Yeah you're right there's no need for an action really

fourleafed:
-- Please reply above this line --

Hi Automattic/wp-Job-Manager We have received your email. We will
respond to you as soon as possible.Kind regards!

-----------------------------------------------------------
## Automattic/wp-Job-Manager sent a message, on Nov 12 @ 6:00 (CET):



	Yeah you're right there's no need for an action really 

	-
 Reply to this email directly or view it on GitHub [1].

Links:
------
[1]
https://github.com/Automattic/WP-Job-Manager/pull/561#issuecomment-155998883


-----------------------------------------------------------

PR #565, opened 2015-11-14
Allow optional taxonomy term-select fields

tripflex:
Currently the way that taxonomy fields are validated is in a `foreach` loop through the array.  If the field is a `term-select` field the value will be either empty or an integer for the selected taxonomy. 

The problem is if a `term-select` field is set as optional, and the user does not select a value, when processing through validation the value is an empty string, and is placed into an array, creating an array with one member, that is empty.  

This causes the `term_exists()` to fail and returns a `WP_Error` saying the term is invalid.  

This PR adds an inline if statement to check if the value is empty, and if so, return an empty array, instead of an array with one member that is empty.

https://github.com/tripflex/wp-job-manager-field-editor/issues/289

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -241,7 +241,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 					if ( is_array( $values[ $group_key ][ $key ] ) ) {
 						$check_value = $values[ $group_key ][ $key ];
 					} else {
-						$check_value = empty($values[ $group_key ][ $key ]) ? array() : array($values[ $group_key ][ $key ]);
+						$check_value = empty( $values[ $group_key ][ $key ] ) ? array() : array( $values[ $group_key ][ $key ] );
 					}
 					foreach ( $check_value as $term ) {
 						if ( ! term_exists( $term, $field['taxonomy'] ) ) {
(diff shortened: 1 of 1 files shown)


fourleafed:
-- Please reply above this line --

Hi Automattic/wp-Job-Manager We have received your email. We will
respond to you as soon as possible.Kind regards!

-----------------------------------------------------------
## Automattic/wp-Job-Manager sent a message, on Nov 14 @ 1:26 (CET):

Currently the way that taxonomy fields are validated is in a foreach
loop through the array. If the field is a term-select field the value
will be either empty or an integer for the selected taxonomy.
The problem is if a term-select field is set as optional, and the user
does not select a value, when processing through validation the value
is an empty string, and is placed into an array, creating an array
with one member, that is empty.
This causes the term_exists() to fail and returns a WP_Error saying
the term is invalid.
This PR adds an inline if statement to check if the value is empty,
and if so, return an empty array, instead of an array with one member
that is empty.
tripflex/wp-job-manager-field-editor#289

You can view, comment on, or merge this pull request online at:
  https://github.com/Automattic/WP-Job-Manager/pull/565
Commit Summary

allow optional term-select fields

File Changes

M includes/forms/class-wp-job-manager-form-submit-job.php (2)

Patch Links:

https://github.com/Automattic/WP-Job-Manager/pull/565.patch
https://github.com/Automattic/WP-Job-Manager/pull/565.diff

-
Reply to this email directly or view it on GitHub.

-----------------------------------------------------------

PR #567, opened 2015-11-16
Fix bug in user editing permission function

jamesjonesphoenix:
I'm experiencing a bug where when a user gets 'invalid listing' when they got to edit one of their listings. The problem is that $job->post_author is a string and the if statement is strict. One line code change fixes it.


tripflex:
You may need to rebase your fork as there are conflicts preventing merge (if this PR is approved), I think it may be the new user notification merge you have in that branch

jamesjonesphoenix:
Did that do the trick? I'm still a bit new to the ol' github

spencerfinnell:
Pretty sure this was fixed with https://github.com/Automattic/WP-Job-Manager/commit/264c375aa440b46144dd3826943a65f919ae1fca already.

mikejolley:
'twas

PR #571, opened 2015-11-18
New user notification email will never send on WP 4.3.1+

tripflex:
The PR to add the pluggable function `wp_job_manager_notify_new_user` does not have `$wp_version` global declaration, and as such `wp_new_user_notification` will always send the legacy arguments. 

If someone can give me a :+1: so I can merge this, will be a huge issue if this patch does not go in before next release


kraftbj:
Bug introduced in https://github.com/Automattic/WP-Job-Manager/commit/8a3a9cfd169b65e33f424f883f82e6dd208f0dd2

Good catch. :+1:

tripflex:
Thanks, yeah I just happened to run across it while testing my Registration Use Email plugin, glad I got it before it went into a release :+1: 

https://github.com/tripflex/wp-job-manager-registration-use-email

PR #585, opened 2016-01-04
Use HTTPS for Google Maps Geocoding API calls

aheckler:


mikejolley:
This one should not be required; you can use either, and the request is non-sensitive.

PR #608, opened 2016-02-16
Handed over size parameter to get company image.

jonasvogel:
The `$size` is missing as parameter when calling a company image for a job with a company logo as attachment (since 1.24.0).
Otherwise the default of `get_the_company_logo()` is used, which is _thumbnail_.

### DIFF
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -311,7 +311,7 @@ function get_the_job_location( $post = null ) {
  * @return void
  */
 function the_company_logo( $size = 'thumbnail', $default = null, $post = null ) {
-	$logo = get_the_company_logo( $post );
+	$logo = get_the_company_logo( $post, $size );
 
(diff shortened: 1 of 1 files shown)


kraftbj:
@mikejolley - This looks good to me. Just traced a ticket in the queue back to this too. What do you think?

cwbartholomew:
Does the size need to be specified in the function.php file? Or is there something else I'm missing, because adding $size isn't fixing the size issue for me.

jonasvogel:
@cwbartholomew: `$size` is a parameter in the [function call](https://github.com/jonasvogel/WP-Job-Manager/blob/88365d58b71d0b8045cb4747e4aec6877292d479/wp-job-manager-template.php#L313).  
The size had to be registered before with [``add_image_size()``](https://developer.wordpress.org/reference/functions/add_image_size/)

PR #611, opened 2016-02-22
Add jQuery datepicker to custom date-type input fields

aapis:
I added a new field by following the [tutorial](https://wpjobmanager.com/document/tutorial-adding-a-salary-field-for-jobs/) which was is a date type and was hoping it would automatically run datepicker on my new field but, alas, it did not.

This PR makes the datepicker functionality more generic by binding to any element with class `wpjm-date`.  The datepicker functionality only appears to be enabled on the admin "post a job" page so it won't do anything new on the front end forms.

**Example**

```php
// /path/to/my_theme/functions.php

function admin_wp_job_manager_configure_fields($fields){
	$fields['_job_closing_date'] = array(
    'label'       => __( 'Closing Date', 'job_manager' ),
    'type'        => 'text',
    'placeholder' => 'YYYY-MM-DD',
    // added new 'classes' key
    'classes'     => ['wpjm-date'], // note that this is an array, not a string
    'required'    => true,
    'priority'    => 7,
  );

  $fields['_job_posting_date'] = array(
    'label'       => __( 'Posting Date', 'job_manager' ),
    'type'        => 'text',
    'placeholder' => 'YYYY-MM-DD',
    'value'       => date('Y-m-d'),
    // added new 'classes' key
    'classes'     => ['wpjm-date'], // note that this is an array, not a string
    'required'    => true,
    'priority'    => 7
  );

  return $fields;
}
add_filter( 'job_manager_job_listing_data_fields', 'admin_wp_job_manager_configure_fields' );
```


includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -83,6 +83,7 @@ public function job_listing_fields() {
 			$fields['_job_expires'] = array(
 				'label'       => __( 'Listing Expiry Date', 'wp-job-manager' ),
 				'priority'    => 11,
+				'classes'     => ['wpjm-date'],
```
tripflex: May not be good to be using shorthand array syntax as that is only supported on PHP 5.4+

aapis: Ah yes, good point.

Issue #617, opened 2016-02-24
[jobs] shortcode is not querying jobs on non default languages using WPML. 


vukvukovich:
On multilingual site using WPML plugin, [jobs] shortcode is only listing jobs on default language. On other languages "There are no listings matching your search." is shown.

mikejolley:
This is a WPML issue - they said they would come by here and comment/suggest workarounds but nothing yet..

vukvukovich:
It looks like your AJAX calls are not following WP codex standards, and it is related to that?

mikejolley:
Custom ajax endpoints is not against standards. It's a more efficient way of doing ajax calls. The code for WPML still sets lang:

https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L754

mikejolley:
And https://github.com/Automattic/WP-Job-Manager/blob/5d96bb4ebcaf3106cc5c90c01951e851b8700254/wp-job-manager.php#L158

mikejolley:
Maybe you can check your jobs page to see if ICL_LANGUAGE_CODE is set. It's apparently one of their constants.

vukvukovich:
I did breakpoint on it, and it looks that later is reverted to default language an not showing content. Do you have any clue where it can happen?

mikejolley:
I'd only guess that WPML is not setting it to the posted LANG. Maybe it changed the way it picks up language codes.

kochhase:
hi vukukovich, that was the reason why i have changed to polylang. since polylang 1.8 it is working fast an you could use https://de.wordpress.org/plugins/lingotek-translation/ for any kind of translation.

PR #626, opened 2016-03-04
Fixed WPML AJAX issue.

vukvukovich:
Was causing jobs not to be listed on non-default languages. #617

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -141,6 +141,9 @@ You can view (and contribute) translations via the [translate.wordpress.org](htt
 
 == Changelog ==
 
+= 1.25.1 =
+* Fix - Refactored WPML workaround, which was causing no job listings on non-default languages. (https://github.com/Automattic/WP-Job-Manager/issues/617)
+
 = 1.25.0 =
 * Feature - Ability to duplicate listings from job dashboard.
 * Fix - Ensure logo is displayed on edit, before submission.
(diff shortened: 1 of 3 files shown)


mikejolley:
@vukvukovich are you really suggesting this is the preferred options vs actually **looking** for $_GET/$_POST 'lang' on all requests?

This makes no sense. You support ?lang=x on some permalink structures but not all. Please do developers a favour and standardise how this works. Firing actions just for WPML seems off (the previous workaround was just as bad and I don't want either ideally).

WPML seems to have the 'all plugins need to support us' approach. In my humble opinion WPML should do more to support plugins without hacks. 

I'll merge this if we need to to get basic support in, but please consider my $_GET lang request first. It's just simpler for all parties!

mikejolley:
Awaiting your feedback before merge..

vukvukovich:
Hi Mike, sorry for waiting. WPML dev team is trying their best to standardise things by following WP codex, which is best and probably only way to achieve overall plugin compatibility. 

Regarding this case, WPML has standardised approach to permalink structures also following WP codex standards. We have short document about it here: https://wpml.org/tutorials/language-urls/

About action firing you are right, I have refactored code and removed unnecessary "lang" parameter if WPML is not present, so that action is firing properly.

Please review code and tell me if there is anything else I can help and edit.

Cheers!

Issue #623, opened 2016-03-02
Plus sign stripped from URLs


aheckler:
If I enter a URL like `https://plus.google.com/+Chrome` as my company website, the `+` gets stripped out on save:

![screen shot 2016-03-02 at 11 15 41 am](https://cloud.githubusercontent.com/assets/5512652/13466479/1da6c8e8-e068-11e5-968b-62b9aabd4d96.png)

From: 2556725-t

PR #629, opened 2016-03-04
Fix issue #623

s4ad:
Encoding company website url in job listing form.


mikejolley:
This doesn't fix the saved value sadly.

PR #636, opened 2016-03-17
Adds a job_feed_item hook

kraftbj:
Provides a hook for site admins to add additional items to the RSS entry for each job listings.


edd677:
Also be awesome!

PR #643, opened 2016-03-29
Start using `grunt-wp-i18n`

georgestephanis:
This will not only generate pot files without relevant command line utilities installed, it will also add textdomains to strings.

Deprecate grunt-shell, we don't need it any longer.


includes/admin/class-wp-job-manager-settings.php
```diff
@@ -322,7 +322,7 @@ public function output() {
 								break;
 								case "password" :
 
-									?><input id="setting-<?php echo $option['name']; ?>" class="regular-text" type="password" name="<?php echo $option['name']; ?>" value="<?php esc_attr_e( $value ); ?>" <?php echo implode( ' ', $attributes ); ?> <?php echo $placeholder; ?> /><?php
+									?><input id="setting-<?php echo $option['name']; ?>" class="regular-text" type="password" name="<?php echo $option['name']; ?>" value="<?php echo esc_attr( $value ); ?>" <?php echo implode( ' ', $attributes ); ?> <?php echo $placeholder; ?> /><?php
```
mikejolley: Just a note, not 100% if this is the case here (may need to monitor support after) but plugins like WPML can translate settings values and this would show the localised version depending on admin access. 

On the flip side, there are other ways to translate WPML settings using their XML files, so this fix is probably fine. Just thought I'd mention it.

templates/job-submit.php
```diff
@@ -56,7 +56,7 @@
 			<input type="hidden" name="job_manager_form" value="<?php echo $form; ?>" />
 			<input type="hidden" name="job_id" value="<?php echo esc_attr( $job_id ); ?>" />
 			<input type="hidden" name="step" value="<?php echo esc_attr( $step ); ?>" />
-			<input type="submit" name="submit_job" class="button" value="<?php esc_attr_e( $submit_button_text ); ?>" />
+			<input type="submit" name="submit_job" class="button" value="<?php echo esc_attr( $submit_button_text ); ?>" />
```
mikejolley: WPML comment above.

mikejolley:
LG just the note on WPML, particularly for output of settings on the frontend (submit buttons).

templates/job-submit.php
georgestephanis: $submit_button_text is already passed through translation calls here: https://github.com/Automattic/WP-Job-Manager/blob/master/includes/forms/class-wp-job-manager-form-edit-job.php#L90 so this should be fine.

includes/admin/class-wp-job-manager-settings.php
georgestephanis: As these are form inputs, I don't see any point in translating a form value that the user then can edit and save -- wouldn't they just change the form field directly and then save it?  If the user wanted to use WPML to translate the option for different front-end views, that's fine I guess, but that would have nothing to do with how it's displayed in a field on the back-end, as it'd be like ... the original value gets translated on the first save, then the translation gets passed in to get translated on the second?  I don't think it's a workable use case in this instance.

mikejolley: If its translated on frontend/when retrieved, it should continue to work. WPML logic ¯\_(ツ)_/¯

chaselivingston:
@georgestephanis Looks like some conflicts need to be resolved before we can merge this in.

dbtlr:
@georgestephanis I'm looking to clear up some of the PRs here. Is this one ready to go? Can you please clear up these conflicts and rebase master?

georgestephanis:
@dbtlr -- Conflicts resolved.  Should be ready to go.

dbtlr:
Sweet, thanks @georgestephanis!

Issue #631, opened 2016-03-09
Add "View" button for attachments in backend


aheckler:
Would be nice to have a button that just opens the attachment in a new tab. Like so:

![screen shot 2016-03-09 at 2 05 11 pm](https://cloud.githubusercontent.com/assets/5512652/13646984/f753ca9a-e5ff-11e5-888d-1b7949bb4e19.png)

From #2568879-t.

mikejolley:
@aheckler not really ideal since:

1. This is an edit interface. The value would be changed and thus view link would not be 'current'
2. Multiple attachments are supported in some cases
3. It depends if the URL entered is relative or absolute which would make linking more difficult.

Links should be displayed when 'viewing' applications only IMO - thats in the list table and on the frontend views for employers.

PR #650, opened 2016-04-07
Add view button for file input fields in admin (w/ dynamic support)

tripflex:
Mike,
I know you mentioned the possible issues with this in the other closed issue, #631 ... I actually had already started working on this, and the only potential issue I see that you mentioned is a relative URL (but I have yet to see those anywhere).

**Example**
example

Basically this adds a View button that uses jQuery to get the URL from the input, and open it in a new window if the value has `://` in it.  If it does not, will add the `file_no_url` class to the input, then remove it after a second.  That class has some CSS animation to flash the background the same red color as core WordPress error class (in admin notices).

I went ahead and used the same jQuery `.live` even though it's depreciated, just to try and keep everything as similar as possible.

If you don't want this in core that's fine, just let me know and I'll add it to my plugin ... just already had it completed for the most part for core before I saw your reply on the issue.


mikejolley:
Looks like it could be useful :+1: I would however avoid live() and use .on() - WPJM should be updated to follow suit, cc @georgestephanis

tripflex:
@mikejolley no problem, I can go ahead and update the PR to use `.on()` instead (updating existing ones as well) ... let me know if you want me to and i'll get it taken care of today.

tripflex:
@mikejolley lmk and i'll update this PR 👍

georgestephanis:
@tripflex ^^ just sent it as a PR in #664

tripflex:
@mikejolley @georgestephanis Updated my PR to include the updates, as well as merge the update from your PR #664

tripflex:
@mikejolley think this can be merged?  I can close and implement in my plugin, just figured it would be good to have in core

kraftbj:
@chaselivingston tested, @mikejolley approved, @kraftbj merged.

PR #658, opened 2016-04-19
Remove WPJM taxes from tag cloud

kraftbj:
The Tag Cloud with WPJM taxonomies doesn't work. They display, but direct you to non-functional links. We should remove ourselves from tag cloud consideration.


mikejolley:
Did not realise this existed. Makes perfect sense.

PR #664, opened 2016-04-29
Stop using jQuery.live

georgestephanis:
By changing over to $(document).on('click', 'selector', function(){});
we maintain the same functionality without .live

Mentioned in #650


mikejolley:
👍

PR #669, opened 2016-05-09
Don't Show Job Type Drop Down if There Are No Job Types.

KZeni:
I have a site which doesn't use Job Types, but they want to use the front-end submission form. Unfortunately, the current code forces the job type field to be required even if there aren't any job types to choose from!

_The Fix:_ I've made this edit to the form field code so it only includes the job type field if there's job types to choose from. The job type field remains required, but only if there are job types to be shown.

I'd love to see this included in a future version of the plugin since I'm sure there's plenty of people out there who share the need for this use case, and I want them to be able to use this great plugin. Thanks!

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -116,7 +116,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 			break;
 		}
 
-		$this->fields = apply_filters( 'submit_job_form_fields', array(
+		$job_form_fields = array(
 			'job' => array(
 				'job_title' => array(
 					'label'       => __( 'Job Title', 'wp-job-manager' ),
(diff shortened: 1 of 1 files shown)


spencerfinnell:
Definitely would be nice, and would allow some code in Listify to be removed as well.

But `wp_count_terms('job_listing_type')` would need to be cached for sure.

Also would be great if this check was added elsewhere, like the output on the filters, etc.

https://github.com/Automattic/WP-Job-Manager/blob/c838d6ee3a3d0fd224d666bfee55f58517e10cf6/templates/job-filter-job-types.php

KZeni:
Sure, I'm not familiar with the code for this plugin so I'd definitely turn to your recommendation for proper implementation. This was just one method I saw that would address the form's conflict.

KZeni:
Okay, new commit added. I just noticed how there's already conditional code which unsets the `job_category` form field when needed so I updated my patch to have this follow suit for the `job_type` form field.

kraftbj:
I think this looks good. I'd like to test it a bit more before merge.

tripflex:
👍 from my review this should work fine without any issues.  I know there are plenty of users that use my field editor plugin to disable this field, and checking through the code it looks like almost everywhere the core checks if `get_the_job_type` returns a value before using it.

**_This however could cause issues with people who are using the filter on that specific field._**

My recommendation is to **NOT** unset the value if there aren't any taxonomy values ... IMO if anything we should just add another setting just the same as categories, and unset it from the fields if the user disabled it in the settings.  That way it will work exactly the same at categories ...

tripflex (commented):
IMO this should be done just the same as categories, add something to settings to allow users to disable it

tripflex:
Just noticed the categories method checks the # of taxonomies as well, in that case this should work fine (may still want to think of adding option to disable types though?)

donnchawp:
I think we'll need a setting for this. It's neater than deleting the types and matches what happens with categories.

Also, if you delete all the job types and unset the job_type as in this patch the Job Listing metabox shows on the admin page when you edit a job.

donnchawp:
I created a branch to address this. It adds a new setting and changes some of the UI. What do you think?

PR #674, opened 2016-05-12
add widget() method to WP_Job_Manager_Widget

### LINKED ISSUE #673: WP_Widget convert to abstract class, widget() must  be defined
Found this fatal error issue when testing `4.6-alpha-37427` where `WP_Widget` class is converted to an abstract class.  Because `widget()` is not defined, fatal error shows up

See:
https://core.trac.wordpress.org/ticket/35981

Needs to be resolved before 4.6 is released

tripflex:
fixes #673

### DIFF
--- a/includes/class-wp-job-manager-widgets.php
+++ b/includes/class-wp-job-manager-widgets.php
@@ -131,6 +131,19 @@ class WP_Job_Manager_Widget extends WP_Widget {
 			}
 		}
 	}
+
+	/**
+	 * widget function.
+	 *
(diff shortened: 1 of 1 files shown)


kraftbj:
Confirmed bug and confirmed restored previously behavior. I'm going to 👍  and merge it.

kraftbj:
Core change: https://core.trac.wordpress.org/changeset/37425
Ticket: https://core.trac.wordpress.org/ticket/35981

PR #684, opened 2016-06-01
Include a filter to disable the Job Listings cache

benjaminprojas:
This will allow plugin developers to temporarily disable the cache while using the get_job_listings() function.


kraftbj:
I might want to combine those conditionals to make it a bit easier to read with one fewer levels of conditionals, but not a blocker, can be done post-merge.

PR #689, opened 2016-06-10
update/fix visibility declarations

hulet:
Make property and method visibility declarations more clear and explicit.

Per http://docs.php.net/manual/en/language.oop5.visibility.php: 

"Class properties must be defined as public, private, or protected. If declared using var, the property will be defined as public. ... The PHP 4 method of declaring a variable with the var keyword is still supported for compatibility reasons (as a synonym for the public keyword).  ... Class methods may be defined as public, private, or protected. Methods declared without any explicit visibility keyword are defined as public."


tripflex:
👍

PR #708, opened 2016-06-29
Adding additional contributors

kraftbj:
Waiting on a couple of names. p6r3EZ-8K-p2


chaselivingston:
@kraftbj Looks like you may have what you need now :)

Issue #740, opened 2016-08-18
0day: Arbitrary file upload


JeffMatson:
It looks like a 0-day just hit.  Here are the details:

[removed the link to make it a tiny bit harder to find]

JeffMatson:
The proposed solution is to only allow images that match instead of any recognized mime type.  See pull request #07cd7e4

tripflex:
So the way it's supposed to work, is by pulling the allowed mime types from the field configuration, and if that is not set, falling back to `get_allowed_mime_types`, as almost always the mime type configuration is set in the field config

https://github.com/Automattic/WP-Job-Manager/blob/master/includes/abstracts/abstract-wp-job-manager-form.php#L326-L330

But it looks like he's not passing the `$allowed_mime_types` anywhere ... i'll check the ajax in a minute and check that

tripflex:
Ajax uploading details:
https://github.com/Automattic/WP-Job-Manager/blob/master/assets/js/ajax-file-upload.js#L19

`data-file_types` output:
https://github.com/Automattic/WP-Job-Manager/blob/master/templates/form-fields/file-field.php#L24-L24

Upload handling:
https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-ajax.php#L213-L213

tripflex:
So with the fix of adding `$allowed_mime_types` with `allowed_mime_types` key in my first comment, should fix normal uploading, and the potential patch that Jeff mentioned should fix the ajax as uploading files with ajax doesn't have field config available

kraftbj:
Quick update to say I have a patch I'm running through internal testing now. Strictly speaking, this isn't an exploit as much as unexpected behavior. In any case, it would only allow file types already allowed by WordPress to be uploaded, excluding htm(l) unless the user is already logged in with a role allowed to do so.

That said, still taking it very seriously and will have it hardened right away.

PR #741, opened 2016-08-18
Possible solution to issue #740

JeffMatson:
Possible solution to issue #740 and arbitrary file uploads.


kraftbj:
Thanks! I'm closing this since we're going to do something slightly different

PR #742, opened 2016-08-19
Additional MIME checks to further reduce options from what WP already provides

### LINKED ISSUE #740: 0day: Arbitrary file upload

kraftbj:
Fixes #740

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -698,7 +698,7 @@ function job_manager_prepare_uploaded_files( $file_data ) {
 function job_manager_upload_file( $file, $args = array() ) {
 	global $job_manager_upload, $job_manager_uploading_file;
 
-	include_once( ABSPATH . 'wp-admin/includes/file.php'  );
+	include_once( ABSPATH . 'wp-admin/includes/file.php' );
 	include_once( ABSPATH . 'wp-admin/includes/media.php' );
 
 	$args = wp_parse_args( $args, array(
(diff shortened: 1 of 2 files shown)


wp-job-manager-functions.php
```diff
@@ -738,6 +743,46 @@ function job_manager_upload_file( $file, $args = array() ) {
...
+ */
+function job_manager_get_allowed_mime_types( $field = '' ){
+	if ( 'company_logo' === $field ) {
+		$allowed_mime_types = array(
+			'jpg|jpeg|jpe' => 'image/jpeg',
+			'gif'          => 'image/gif',
+			'png'          => 'image/png',
+		);
+	}
+	else {
```
chaselivingston: I think the else should be next to the closing `}` above.

wp-job-manager-functions.php
```diff
@@ -704,18 +704,23 @@ function job_manager_upload_file( $file, $args = array() ) {
...
 		'file_key'           => '',
 		'file_label'         => '',
-		'allowed_mime_types' => get_allowed_mime_types()
+		'allowed_mime_types' => '',
 	) );
 
 	$job_manager_upload         = true;
 	$job_manager_uploading_file = $args['file_key'];
 	$uploaded_file              = new stdClass();
+	if ( '' == $args['allowed_mime_types'] ) {
```
mikejolley: Use `===` here instead of `==`

PR #751, opened 2016-09-08
Fix jobs shortcode so the random orderby is actually random. #714

### LINKED ISSUE #714: Orderby Rand not working
When you try using the shortcode [jobs orderby="rand"], it doesn't display job listings in a random order as it should.

Tried using different themes, the issue persists. Brought up in [this forum thread](https://wordpress.org/support/topic/orderbyrand-doenst-work?replies=1).

donnchawp:
This fixes #714 by shuffling the posts returned from WP_Query, which may have been cached.


pgk:
Looks Good!

PR #753, opened 2016-09-08
Set job_id to 0 instead of reading from the request array. See #690

### LINKED ISSUE #690: Preview of a listing can be viewed and edited by anybody

The preview of a listing on the frontend should be restricted to the user who created it.
#### What Happened Instead

However, it can be viewed and edited by anyone.

Create a new listing and click on preview. Copy the URL of that page. Open page source and look for job_id and its value. Append ?job_id=xxxx (xxxx being the number of the job_id from the page source) and visit the URL in an incognito window/another browser. You'll be able to view and edit the preview.

donnchawp:
This hopefully fixes #690 and it means that whoever wants to edit a job listing must have the job editing cookies to do so.


tripflex:
I don't think this was the correct way to fix this, and this may cause issues down the line

PR #768, opened 2016-09-29
Fix the_company_video unnecessary call to wp_oembed_get causing additional site load time

tripflex:
Even though the `the_company_video` field is not a required field, the core code base still assumes that a value will be returned each time for this field, causing an unnecessary call to `wp_oembed_get` causing a call to `wp_safe_remote_get` which ends up in a GET timeout, adding additional load time to the site.

This PR fixes this issue by only calling `wp_oembed_get` if `$video_embed` has a value, while still maintaining the ability to use the filter and still have the video output.

### DIFF
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -413,15 +413,16 @@ function job_manager_get_resized_image( $logo, $size ) {
  * Output the company video
  */
 function the_company_video( $post = null ) {
-	$video    = get_the_company_video( $post );
-	$filetype = wp_check_filetype( $video );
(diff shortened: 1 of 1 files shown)


tripflex (commented):
It's better to only set `$video_embed` to a value other than false, if `$video` has a value.  This will prevent the output of the `company_video` DIV if there isn't any possible video to output.  First commit in this PR would still set `$video_embed` value to flowplayer shortcode (without src having a value).

kraftbj (approved):
Makes sense. Apologies for the delay.

PR #769, opened 2016-09-30
Stop setup.css loading except on the setup page

### LINKED ISSUE #728: Admin Setup CSS is Bleeding In All Admin Pages

setup.css is only needed on "setup page". 

#### What Happened Instead

currently it's loaded in all admin pages.

here's the code from master branch:
https://github.com/Automattic/WP-Job-Manager/blob/master/includes/admin/class-wp-job-manager-setup.php#L76

current latest tag:
https://github.com/Automattic/WP-Job-Manager/blob/1.25.0/includes/admin/class-wp-job-manager-setup.php#L76

install plugin and check the admin page.

donnchawp:
Fixes #728
The file setup.css was enqueued on every Job Manager admin page
but this checks that the setup page is loaded before enqueuing it.


pgk:
Looks OK! :)

PR #771, opened 2016-10-01
check job_manager_user_can_edit_job b4 set job_id to 0

tripflex:
check `job_manager_user_can_edit_job` before setting job_id to zero #690 #753


tripflex:
Just the same as https://github.com/Automattic/WP-Job-Manager/blob/master/includes/forms/class-wp-job-manager-form-edit-job.php#L40-L44

PR #776, opened 2016-10-07
Simplify the search message on the job listing page.

### LINKED ISSUE #647: Translation error in "Showing" text box
When filtering with -for example- job types in French, the showing text is wrong because of the way it is coded.
#### What Happens

It shows: 

> Afficher tous les **postes** Temps plein & Temps partiel **postes**

(poste = job, displayed twice)

It should show: 

> Afficher tous les **postes** Temps plein & Temps partiel
#### Code causing the problem

includes/class-wp-job-manager-ajax.php:178
`$result['showing'][] = $post_type_label;

if ( $search_location ) {
    $result['showing'][] = sprintf( __( 'located in &ldquo;%s&rdquo;', 'wp-job-manager' ), $search_location );
}

if ( 1 === sizeof( $result['showing'] ) ) {
    $result['showing_all'] = true;
}

$result['showing'] = apply_filters( 'job_manager_get_listings_custom_filter_text', sprintf( __( 'Showing all %s', 'wp-job-manager' ), implode( ' ', $result['showing'] ) ) );`
#### Solution

Do not add the Job label at the end of the text and change the sprintf() text:
`//$result['showing'][] = $post_type_label; //REMOVE THIS LINE

if ( $search_location ) {
    $result['showing'][] = sprintf( __( 'located in &ldquo;%s&rdquo;', 'wp-job-manager' ), $search_location );
}

if ( 0 === sizeof( $result['showing'] ) ) { //CHANGE HERE
    $result['showing_all'] = true;
}

$result['showing'] = apply_filters( 'job_manager_get_listings_custom_filter_text', sprintf( __( 'Showing all %sjobs', 'wp-job-manager' ), implode( ' ', $result['showing'] ) ) ); //ADD THE LABEL IN A TRANSLATABLE STRING`

donnchawp:
Fixes #647
The original string was translated incorrectly to French but because of the way it was constructed it couldn't be easily translated correctly anyway. Since the search terms are already on the page I modified the string so it would display how many records had been found.

### DIFF
--- a/includes/class-wp-job-manager-ajax.php
+++ b/includes/class-wp-job-manager-ajax.php
@@ -185,7 +185,13 @@ class WP_Job_Manager_Ajax {
 			$result['showing_all'] = true;
 		}
 
-		$result['showing'] = apply_filters( 'job_manager_get_listings_custom_filter_text', sprintf( __( 'Showing all %s', 'wp-job-manager' ), implode( ' ', $result['showing'] ) ) );
+		if ( $jobs->post_count && ( $search_location || $search_keywords || $search_categories ) ) {
+			$message = sprintf( __( '%1$s search completed. Found %2$d matching records.', 'wp-job-manager' ), __( $post_type_label, 'wp-job-manager' ), $jobs->post_count );
+			$result['showing_all'] = true;
+		} else {
+			$message = "";
+		}
+		$result['showing'] = apply_filters( 'job_manager_get_listings_custom_filter_text', $message );
 
 		// Generate RSS link
 		$result['showing_links'] = job_manager_get_filtered_links( array(


donnchawp:
It's not ideal to have a variable passed to __(), but by default the string is "Jobs" and is already a string in the pot file.

includes/class-wp-job-manager-ajax.php
```diff
@@ -185,7 +185,13 @@ public function get_listings() {
 			$result['showing_all'] = true;
 		}
 
-		$result['showing'] = apply_filters( 'job_manager_get_listings_custom_filter_text', sprintf( __( 'Showing all %s', 'wp-job-manager' ), implode( ' ', $result['showing'] ) ) );
+		if ( $jobs->post_count && ( $search_location || $search_keywords || $search_categories ) ) {
+			$message = sprintf( __( '%1$s search completed. Found %2$d matching records.', 'wp-job-manager' ), __( $post_type_label, 'wp-job-manager' ), $jobs->post_count );
```
akirk: `xgettext` will probably choke on this, as it extracts translatable strings from plain text. What about changing this to:
if ( 'Jobs' === $post_type_label ) {
$message = sprintf( _n( 'Jobs search completed. Found %d matching record.', 'Jobs search completed. Found %d matching records.', $jobs->post_count, 'wp-job-manager' ), $jobs->post_count );
} else {
$message = sprintf( _n( 'Search completed. Found %d matching record.', 'Jobs search completed. Found %d matching records.', $jobs->post_count, 'wp-job-manager' ), $jobs->post_count );
}

PR #781, opened 2016-10-11
Only allow one job type per job in the job editor

### LINKED ISSUE #549: Editing job in backend allows for multiple types
... but we enforce a " one type per job" policy on the default submission form:

screen shot 2015-10-27 at 10 02 43 am

This leads to confusion:

https://wordpress.org/support/topic/bug-single-job-listing-page-only-displays-first-type

https://wordpress.org/support/topic/multiple-job-types

https://wordpress.org/support/topic/multiple-job-types-1

We should either allow multiple job types per job or limit the backend meta box in such as way as to enforce the "one type per job" rule.

donnchawp:
Fixes #549
The post editor metabox uses checkboxes which allows more than one job type, but replacing them with radio buttons only allows one. Code from https://code.tutsplus.com/articles/how-to-use-radio-buttons-with-taxonomies--wp-24779


eppul:
I think it should be optional to have only one or to have multiple job types.

Issue #688, opened 2016-06-10
Translated Job-category permalink not working when translated (WPML) - 404 Error 


Lens40:
#### What I Expected To Happen

I run WP Job Manager in three languages using Jobify theme. I use WPML and the .MO-Files to translate the strings. When I click the category link to display all jobs of the category I expect to see all jobs of the category.

EN: /job-category/marketing
DE: de/jobkategorie/marketing-de
FR: fr/categorie-emploi/marketing-fr
#### What Happened Instead

In the English version it works and I get the the job listings displayed. In the DE and FR version I get an 404 page. The slug is correctly translated (see URL above) but the archive is not existing under this URL. Instead I only can access the DE & FR archive via the following links with the base slug in English:

DE: de/**job-category**/marketing-de
FR: fr/**job-category**/marketing-fr

The PO file

`#: includes/class-wp-job-manager-post-types.php:74
msgctxt "Job category slug - resave permalinks after changing this"
msgid "job-category"
msgstr "jobkategorie"`

The code that takes care of the slug
`
/**
         * Taxonomies
         */
        if ( get_option( 'job_manager_enable_categories' ) ) {
            $singular  = __( 'Job category', 'wp-job-manager' );
            $plural    = __( 'Job categories', 'wp-job-manager' );
            if ( current_theme_supports( 'job-manager-templates' ) ) {
                $rewrite   = array(
                    'slug'         => _x( 'job-category', 'Job category slug - resave permalinks after changing this', 'wp-job-manager' ),
                    'with_front'   => false,
                    'hierarchical' => false
                );
                $public    = true;
            } else {
                $rewrite   = false;
                $public    = false;
            }
`

When accessing the English correctly working archive I noticed that the hreflang-URLs in page header are also wrong:

**What is displayed:**
`
<link rel="alternate" hreflang="en-US" href="http://example.com/job-category/retail/" />
<link rel="alternate" hreflang="fr-FR" href="http://example.com/fr/job-category/vente-au-detail/" />
<link rel="alternate" hreflang="de-DE" href="http://example.com/de/job-category/einzelhandel/" />`

**What should be displayed**

`<link rel="alternate" hreflang="en-US" href="http://example.com/job-category/retail/" />
<link rel="alternate" hreflang="fr-FR" href="http://example.com/fr/categorie-emploi/vente-au-detail/" />
<link rel="alternate" hreflang="de-DE" href="http://example.com/de/jobkategorie/einzelhandel/" />`
[...]

Issue #786, opened 2016-10-17
Multilangual site (WPML) - Setting Pages only one page for all languages can be set


Lens40:
#### What I Expected To Happen

When I set up WP Job Manager I expect to be able to set the pages in the settings for each language.
![set_page](https://cloud.githubusercontent.com/assets/19168364/19446221/1482b2d8-9498-11e6-8a79-9ccb887092e2.PNG)
#### What Happened Instead

Unfortunately I can only set one page for all languages which breaks some functionalities in the languages that couldn't be set which rely on these settings.

Currently I have to decide for which language I want to do these settings as it can only be done for one.
#### Steps to Reproduce the Bug

WP Job Manager + WPML

spencerfinnell:
There is Polylang support for this already: https://github.com/Automattic/WP-Job-Manager/blob/53efe87050c205984d6e7b96e09ef9b015590a64/wp-job-manager-functions.php#L635

WPML support would require https://wpml.org/wpml-hook/wpml_object_id/

spencerfinnell:
@donnchawp I think this would be worth adding. Would fix a good chunk of WPML issues.

donnchawp:
That "page settings" page only creates a new page with a shortcode in it. It doesn't do anything special to them, other than that.

I installed WPML this morning and added the wpml_object_id filter as @spencerfinnell suggested and it looks like I was able to create landing pages for different languages through the standard page editor interface.
WPML lets you create other language versions of any page so I created French and Italian (I had those .mo files for the plugin downloaded already) "versions" of my jobs listings page on my test server.
Both pages only have the "jobs" and "submit_job_form" shortcodes but when I viewed them the pages were mostly translated!
The URLs looked like this too:
http://local.wordpress.dev/it/jobs-list/
http://local.wordpress.dev/fr/french-jobs-list/
The permalink to each page can be different strings because they're completely different pages to the English page.

Unfortunately the jobs types list wasn't translated but we're getting somewhere.

![screen shot 2016-10-18 at 13 11 01](https://cloud.githubusercontent.com/assets/5656673/19477403/dcc66a26-9535-11e6-803f-980feab3f79a.png)

![screen shot 2016-10-18 at 13 09 08](https://cloud.githubusercontent.com/assets/5656673/19477447/102b1d6c-9536-11e6-816e-2db9bca4fc02.png)

[...]

PR #787, opened 2016-10-18
Add the wpml_object_id filter to aid multi language use of the plugin.

donnchawp:
See #786. If the site owner has WPML installed this filter should help redirect the visitor to the translated page they need. Props @spencerfinnell


donnchawp:
This also probably fixes #688

PR #845, opened 2016-12-20
Add search values to the job_manager_get_listings_custom_filter_text filter

kraftbj:
Adds an array of the search terms to be passed to the message filter so they can be included in any modified output.

### DIFF
--- a/includes/class-wp-job-manager-ajax.php
+++ b/includes/class-wp-job-manager-ajax.php
@@ -191,7 +191,13 @@ class WP_Job_Manager_Ajax {
 		} else {
 			$message = "";
 		}
-		$result['showing'] = apply_filters( 'job_manager_get_listings_custom_filter_text', $message );
+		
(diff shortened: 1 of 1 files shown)


spencerfinnell:
Not exactly sure when it this was introduced but the message now collects a bunch of info then overrides it at the end with a single string. I actually prefer this approach. 

https://github.com/Automattic/WP-Job-Manager/blob/a69b61d7c72302a27318fd021f5967efb0290b38/includes/class-wp-job-manager-ajax.php#L189

The values added to the message filter make all of this reproducible if needed, but we see the majority of people wanting to remove the message altogether, so the simpler message is preferred anyway.

dbtlr:
It was introduced as part of 1.25.2 -- https://github.com/Automattic/WP-Job-Manager/compare/1.25.1...1.25.2#diff-364b7a907b7409cda9ed115f463f23ba

LGTM

Issue #835, opened 2016-12-14
BUG - Posting two jobs in a row overwrites the previous pending job offer


Lens40:
#### What Happens

When I publish two jobs in a row which are not publish right away (pending payement e.g.) the second job overwrites the first one. Woocommerce creates two customer orders but in both cases the Post-ID for both jobs are the same. The first job is lost.

#### Steps to Reproduce the Bug

Use Wp Jobmanager with WC-Paid-Listing Addon. As a guest create a first new job -> Preview it -> Publish Listing -> Fill Order Form and save. Then do the same again -> create a new job -> Preview -it -> Publish listing -> Fill Order Form and Save. Now this new jobs overwrites the existing one. 

It gets even crazier as it is not related to one user. Means I can create a job as User A which is pending and then a User B creates also a job and it simply overwrites the pending job offer of User A.

So its quite an important issue...as it seems that the system allows only one pending job offer at a time.

BR Lens

eppul:
It seems to me that Automattic should go another round of hard core bug hunting/fixing.
It has taken too long since the last update.

Lens40:
After some research we found possible reasons for the issue. In the submit form we have following code:  
`if ( ( 'preview' === $job_status || 'pending_payment' === $job_status ) && get_post_meta( $job_id, 'submitting_key', true ) === $_COOKIE['wp-job-manager-submitting-job-key'] ) {
$this->jobid = $job_id;
}`

Means that when a user starts creating a job and does not publish it the cookie will save the infos so that when he returns to the submit form he edits the same job. The idea is good but there is an issue:

When you use WC-Paid-Listing the job status stays in "preview" after submitting the job even though I would expect it to change to "pending_payement" (bug?). So even though the user has submitted the job as long as the job has not been changed to published he will always edit its current job even though he thinks he is creating a second one. Unfortunatley WC-paid-Listings doesn't make a difference and makes the user pay a second time even though he just edited his current order by mistake.

I guess that without WC-Paid-Listings you might have the same issue when the admin needs to confirm the job publication if the status stays in preview.

kochhase:
Hi there you are right. this will only work, when you pay via paypal etc. That the job is paid, and the user is been able to take the next job to enter. but at the BtoB business you use an invoice and the paying going later. so I think it would better if the job status would thange to "pending_payement" after submitting the job even though.
I hope that Mike would look here in this area. last year there was all better...

donnchawp:
[...]

PR #847, opened 2016-12-22
Warn the user if they're editing an existing job.

donnchawp:
See #835

The plugin allows the user to edit jobs before they're published, but if the paid listings plugin is installed a new job may be in "pending_payment" status even after the user has submitted it. This patch warns the user and displays a link   to delete the cookie. The link will allow users to start a new job submission even without the paid listing plugin installed.


dbtlr (changes requested):

dbtlr (approved):

kylemabaso:
Hi Guys,

I never get a warning on my installation if a user is editing an existing job. Do I need to configure something else?

Thanks

PR #855, opened 2017-01-12
Implicit whitelisting for API requests

muddletoes:
To mitigate the security risk of exposing arbitrary class construction methods to execution via a GET request, the `api_requests()` method now only instantiates those classes for which an `job_manager_api_$api` action hook has been registered.


dbtlr:
LGTM

PR #859, opened 2017-01-23
Fixed taxonomy search condition for keyword search

jonasvogel:
The taxonomy search condition for searching keywords was not set correctly. The database query compared `term_id` with `taxonomy_term_id`, which isn't the same (see http://wordpress.stackexchange.com/a/230585) and resulted in false positives for the search query. The fix adds an additional join for the `wp_term_taxonomy` table.


dbtlr:
Great catch! LGTM

PR #873, opened 2017-02-16
Adding WordPress function has_shortcode() instead of strstr

chinmeh:
The php function strstr() provided small check which could be easily skipped by adding the string ['job_dashboard'  to the post content. By adding the function has_shortcode() it does a more strict search using regex for searching the shortcode in post content and it couldn't be skipped.


dbtlr:
`has_shortcode` is a fairly complex regex, which could have some performance ramifications. However, you're right that it is more exact. I'm personally fine with this change, since this is limited to only running on pages though, it should be okay.

LGTM

dbtlr (approved):
LGTM

Issue #821, opened 2016-12-06
Page Error on User Deletion


stephenbair:
#### What I Expected To Happen
User to be deleted without error.

#### What Happened Instead
After clicking 'Confirm Deletion' I received an error to which I had to change the URL path back to the base domain. The user was deleted, but the error appeared. 

Warning: Missing argument 2 for job_manager_add_post_types() in /home/content/80/13050580/html/wp-content/plugins/wp-job-manager/wp-job-manager.php on line 174

Warning: Cannot modify header information - headers already sent by (output started at /home/content/80/13050580/html/wp-content/plugins/wp-job-manager/wp-job-manager.php:174) in /home/content/80/13050580/html/wp-includes/pluggable.php on line 1174

#### Steps to Reproduce the Bug
Go to Users, delete an existing user, confirm deletion, error appears, page does not load.

jom:
Fixed in #875

PR #875, opened 2017-02-17
$id unused, no need to expect both params passed

tripflex:
This is related to this support request:
https://wordpress.org/support/topic/error-on-user-deletion/#post-8818607

Since `$id` is not even used, no need to expect both args


spencerfinnell:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/821

Issue #887, opened 2017-03-06
Allow AJAX Results to be Returned as Pure JSON


spencerfinnell:
PR incoming.

Instead of forcing AJAX results to build the HTML themselves we should allow plugin/theme authors to have total control over the JSON object based on the jobs found. This will allow people to build out their JSON with the relevant info they need and easily feed it to JS based templates instead of PHP.

jom:
Fixed by #888

PR #888, opened 2017-03-06
Allow AJAX searching to return JSON early

spencerfinnell:
See #887

- Also removes “showing” message building that is no longer used.
- Uses proper brace styling for HTML generation.


includes/class-wp-job-manager-ajax.php
```diff
@@ -214,6 +149,27 @@ public function get_listings() {
 
 		$result['max_num_pages'] = $jobs->max_num_pages;
 
+		// Give plugins the opportunity to control what JSON is returned
+		if ( $result = apply_filters( 'wp_job_manager_ajax_get_jobs', null, $result, $jobs ) ) {
```
dbtlr: The majority of filters use `job_manager_*` as their name, we should keep this consistent.

dbtlr (changes requested):
Regarding the changes to the 'showing' message. We've had recent pull requests regarding this as well. #845 

If you would like to remove that, please split into a new pull request and make a case for why this would be better without it, rather than just bundle it into another PR.

includes/class-wp-job-manager-ajax.php
```diff
@@ -121,6 +121,11 @@ public function get_listings() {
 
 		$result['found_jobs'] = false;
 
+		// Give plugins the opportunity to control what JSON is returned
+		if ( $result = apply_filters( 'job_manager_ajax_get_jobs', null, $result, $jobs ) ) {
```
dbtlr: This is a little confusing to me, I see 2 things happening here.

1) If not overridden, this will return null. This will make the conditional fail and it will move on.
2) In order to make this succeed, you are expected to define something like this:

```php
add_filter( 'job_manager_ajax_get_jobs', 'my_func_override_ajax_get_job`, 1, 3 );
my_func_override_ajax_get_job( $serves_no_purpose, $result, $jobs ) {
   return $result;
}
```

This seems overly complex and partially redundant.

Instead of using the `job_manager_ajax_get_jobs` filter to return the results and perhaps provide another avenue to change what is in the `$result` I think it would be better to maintain that functionality with the `job_manager_get_listings_result` filter, which is meant for this.

My suggestion is two-fold.

1) Rename the filter to be something more specific to the intent here, like: `job_manager_ajax_get_jobs_html_results`

then 

2) Change the definition to be more like this: `if ( true !== apply_filters( 'job_manager_ajax_get_jobs_html_results', '__return_true', $result, $jobs ) ) {`

This will allow you to still programmatically decide whether or not to override the results, but would also prevent result tampering from too many sources.

The resulting add_filter would then look more like this:

```php
add_filter( 'job_manager_ajax_get_jobs_html_results', 'my_func_override_ajax_get_job_html_results`, 1, 3 );
my_func_override_ajax_get_job_html_results( $should_return_html, $result, $jobs ) {
    if ( $result /* has something good in it */ ) {
        $should_return_html = false;
    }

    return $should_return_html;
}
```

Without this, it seems like we're adding more hard to understand filters into WPJM. One thing I'd like to do is start auditing and documenting the existing filters, so making sure new ones are well thought through and make sense is also a priority.

dbtlr (changes requested):

spencerfinnell:
`job_manager_ajax_get_jobs_html_results` is too specific I think. The main use case I see for this is avoiding HTML completely, so not sure that filter name is more obvious. 

I do agree checking for null probably doesn't make the most sense since this method doesn't return a value by default. 

Can't think of a filter name that satisfies both though...

spencerfinnell:
Updated the check, but still not 100% on filter name.

spencerfinnell:
That actually doesn't work since $result needs to be assigned the value.

jom:
Hi @spencerfinnell! Why does `$result` need to be assigned the value of the first filter? I agree with @dbtlr. That seems to be the purpose of the `job_manager_get_listings_result` filter. The first filter should just be a `boolean` check for whether or not to skip HTML generation entirely in Ajax search requests. In which case, a filter name like the one Drew recommended seems appropriate.

Additionally, I'm hesitant to pass `$result` to either filter at this point in the function. The only attribute it has at that point is `found_jobs` which hasn't been properly set. I say either move this code above `$result['found_jobs'] = false;` line or pass it an empty `array()` just to be explicit.

As @dbtlr mentioned, we are going to be working on documenting all filters and actions in core. Please document the new (first) filter [according to WordPress documentation specs](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters). Go ahead and set it at `@since 1.26` and we can adjust if needed.

spencerfinnell:
- Brought up to date with master.
- Introduced new filter with documentation to halt and return AJAX response.
- Moved all non-HTML generation above introduced filter.

jom:
@spencerfinnell I've made a couple of fixes. Default behavior was broken. Mind checking my commit? Let me know once you've done that and I'll merge.

spencerfinnell:
Nice, looks good to me.

Issue #795, opened 2016-11-01
Use Registered Taxonomy Label for Singular Job Type Metabox Title


spencerfinnell:
https://github.com/Automattic/WP-Job-Manager/blob/master/includes/admin/class-wp-job-manager-writepanels.php#L129

When creating the custom meta box the title that is set when the taxonomy is registered should be used: https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-post-types.php#L137

PR #893, opened 2017-03-08
Types metabox should say Job Types not Job Listings

tripflex:
Job Types metabox title should not say Job Listings, it should say Job Types to match the appropriate taxonomy (this has confused users who are opening support tickets with me about it)

### DIFF
--- a/includes/admin/class-wp-job-manager-writepanels.php
+++ b/includes/admin/class-wp-job-manager-writepanels.php
@@ -128,7 +128,8 @@ class WP_Job_Manager_Writepanels {
 			remove_meta_box( 'job_listing_typediv', 'job_listing', 'side');
 		} elseif ( false == job_manager_multi_job_type() ) {
 			remove_meta_box( 'job_listing_typediv', 'job_listing', 'side');
-			add_meta_box( 'job_listing_type', __( 'Job Types', 'wp-job-manager' ), array( $this, 'job_listing_metabox' ),'job_listing' ,'side','core');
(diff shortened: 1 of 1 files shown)


tripflex:
Related #795

dbtlr (dismissed):
Look good, but please wrap the label in ucwords() to make sure that it maintains a title case like the rest of the meta boxes

```
ucwords( $job_listing_type->label )
```

spencerfinnell:
@dbtlr I think forcing a case defeats the purpose. The idea is to respect whatever the label was set to originally, or filtered to.

spencerfinnell:
I think it would be better to change the cases in the original labels.

dbtlr:
@spencerfinnell I'm requesting this because by default it says 'Post types' which looks awkward compared to the rest of the interface, which has all the meta box titles cased like ucwords().

For me, this is about making sure that the words were correct from the label, not the casing. The casing is a style issue and should remain consistent with the rest of the interface.

spencerfinnell:
What about using the menu label then? 

https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-post-types.php#L143

Which has it applied as well.

dbtlr:
@spencerfinnell I am 100% fine with that solution. :)

tripflex:
Ok no problem, i'll update the PR

tripflex:
Updated the PR to use `menu_name` now

dbtlr:
Thanks for the update. LGTM!

Issue #901, opened 2017-03-22
Remove Unused Code in AJAX Response


spencerfinnell:
https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-ajax.php#L141-L178

All of that is overwritten in https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-ajax.php#L200

And none of those values are passed to any other filters.

PR #902, opened 2017-03-22
Remove unused code in AJAX response.

spencerfinnell:
See #901


dbtlr:
My main trepidation with making this change is that if there is another plugin or theme using these variables, then they will break, as this essentially changes the result of the API in a subtractive way.

Can you provide me with some evidence that this is not the case and also with evidence that removing this code is an enhancement in some way, other than code readability.

spencerfinnell:
The change that introduced the new static message already made those breaking changes. The code removed here is used nowhere else and passed to no other filters. 

https://github.com/Automattic/WP-Job-Manager/pull/845

Allows for the message to still be changed though.

Issue #869, opened 2017-02-14
Review Expired Listing Functionality


spencerfinnell:
Related #805, #822, #843, #860, #858, #865 

Tracking ticket to contain things about job expirations. 

There a couple of expectations on how expirations should work -- which will require adding additional options in the settings:

- A listing expires and it is no longer visible or accessible on the site at all.
  - This should set a status code of 404 like normal.
- A listing expires and still appears in listing results. Clicking the listing displays an "Expiration" notice.
  - This should set a status code of 401
- A listing expires and does not appear in listing results but the listing content itself is fully accessible (without the ability to apply).
  - This should set a status code of 200

Beyond that we need to ensure:

1. Dates input remain the same when saving/updating.
2. When the date occurs the item is properly expired baed on the chosen style above.

Draxeiro:
Please note: the issue described in #858 is the same as issue #697. This bug was previously solved but since the update to Wordpress 4.7.1 it is back again.

Picking the date via the date picker will result in the listing date not being added properly (the date will show correctly until the listing is saved). The summary page will indeed show 1-1-1970 as the date and on the job listing page itself the date will be in the past (so the listing will not show any more).

Filling in the date by hand will work as a temporary workaround. Please be sure to fix this annoying bug which occurs for users on a non-English language version

annezazuu:
affecting 3066029-t too

rachelsquirrel:
Also reported in #3099049-t

I can reproduce this. I've got a job which has expired but still appears on my jobs page. I'm using the WC Paid Listings plugin as well.

![job](https://cloud.githubusercontent.com/assets/7946521/23895651/68b2d5d4-089e-11e7-9d0f-917c312a8290.png)

![job2](https://cloud.githubusercontent.com/assets/7946521/23895654/6bd5a264-089e-11e7-8cf4-3d3ceb39730e.png)

bikedorkjon:
#3111459-t

rachelsquirrel:
Also reported that when a payment fails the listing remains public. Reported in: #3102235-t

spencerfinnell:
@rachelsquirrel Referring to a subscription payment failing?

rachelsquirrel:
@spencerfinnell I believe so but not 100% sure. I'm checking.

aaronjharms:
[...]

Issue #823, opened 2016-12-06
Feature: Add ability for a redirect on an expired job single


kraftbj:
When visiting an expired job, allowed via option or filter, a redirect to another URL.

2944087-t

spencerfinnell:
Related #869

jom:
Right now WPJM doesn't hook into the loop early enough, but users can easily do this with a filter:
```
add_action( 'wp', function() {
	global $post;
	if ( ! is_singular( 'job_listing' ) || empty( $post ) || 'job_listing' !== $post->post_type ) {
		return;
	}
	if ( 'expired' === $post->post_status ) {
		wp_redirect( 'https://www.example.com' );
		exit;
	}
} );
```

Issue #865, opened 2017-02-04
“Apply For Job” Button Still Displayed Even Listing has Expired


lordzden:
#### What I Expected To Happen
Expired listings should disable or show any indicators that the listing is expired.
Apply for Job should be replaced with word “Expired”.
I dont want to hide expired jobs because it will hurt my SEO. I want to keep the expired listing but I dont want the applicants to apply using the button.

#### What Happened Instead
The Apply for Job is still displayed after listing has expired.

#### Steps to Reproduce the Bug
https://dumjobs.com/job/citymall-dumaguete-job-hiring/
![](https://content.screencast.com/users/BaloydiLloydi/folders/Jing/media/2b4893ee-6a20-4cd9-b714-1c0c822eef77/2017-02-04_1218.png)
`https://www.screencast.com/t/87XNf9oC

**More Information:**

- **Hide content within expired listings is disabled.**
- Notes: “If enabled, the content within expired listings will be hidden. Otherwise, expired listings will be displayed as normal (without the application area).”
- I have updated WordPress
- Updated Job Manager Plugin
- Theme is updated to latest version
**- Jobs are set to expired after 30 days.**
-

spencerfinnell:
Closing for #869 -- will ensure this is hidden and listings are properly expired.

lordzden:
thanks!

Issue #860, opened 2017-01-24
Expired jobs are being listed, expiry date field bugging out


jerwre:
Wordpress 4.7.1 (sv_SE - swedish if that matters)
WP Job Manager 1.25.2
Tested with default themes and all other plugins disabled.

#### What I Expected To Happen

Expired jobs should not be listed. 

Expiry date field should display saved expiry date, and allow you to enter a date earlier than today.

#### What Happened Instead

Expired jobs are being listed. 

In the database the post_status is set to expired, and the _job_expires in postmeta has the original expiry date set.

When I go to edit the post, the expiry date field shows the old date for a second before changing to the current date. If I try to set an earlier date than today and Update post, it'll be updated to todays date, and the post status will be changed to active. 

Saves normally if you set a future date.

_My current "work around" for not displaying expired jobs:_
Expired jobs are no longer being listed when the following option is checked:

> Hide content within expired listings
> If enabled, the content within expired listings will be hidden. Otherwise, expired listings will be displayed as normal (without the application area).

However I'd like the expired jobs to still be accessible via the URL, like before, without being listed.

#### Steps to Reproduce the Bug

Post/edit a job, set expiry date/status earlier than today to have it be expired.

Lens40:
Check out: https://github.com/Automattic/WP-Job-Manager/issues/843. At least for the second half of your issue it is the same. In my opinion the development went itno a wrong direction.

spencerfinnell:
Closing for https://github.com/Automattic/WP-Job-Manager/issues/869

PR #903, opened 2017-03-22
Expired listing inconsistency issue and additional settings

turtlepod:
There are several issues in how WPJM handle expired listing.
You can read it here: https://github.com/Automattic/WP-Job-Manager/issues/869

### This PR try to solve this by.
- Fix exp date consistency by setting listing status on post save.
- Do not change the exp listing date via JS (in meta box): also allow flexibility in setting expire date.
- Separate setting in handling expired listing in archive/job search vs single listing content

There also an interesting feature suggestion: Ability to redirect expired listing/set to 404:
https://github.com/Automattic/WP-Job-Manager/issues/823
I think it can easily added via add-on. I create an example plugin to do this:
https://github.com/turtlepod/wp-job-manager-redirect-expired-listing

Note: Please test before merge.


dbtlr:
Can you please provide more detailed testing instructions?

turtlepod:
1. First, switch to master branch
2. You will need to have an expired listing. (e.g listing with expired status, and expired date for example two weeks ago, etc)

eg-expired-listing

3. Now Edit it, and you will have this (it shows todays date, and not the actual expired date)

eg-expired-listing-edit


This will cause a lot of issue, user experience, data submit/save, etc.. (just messy)
Here's several issue that i think cause by this: #805, #822, #843, #860, #858, #865

I think you need to fix that.

There are other feature also added in this PR, but, I need you to check the above first and see if you can replicate the issue.

turtlepod:
And try to update that expired listing too (try to keep the expired date as the correct expired date, not todays date)

so you'll understand what I mean by data submit/save issue and user experience issue.

includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -85,8 +85,8 @@ public function job_listing_fields() {
 				'priority'    => 11,
 				'classes'     => array( 'job-manager-datepicker' ),
 				/* translators: date format placeholder, see https://secure.php.net/date */
-				'placeholder' => _x( 'yyyy-mm-dd', 'Date format placeholder.', 'wp-job-manager' ),
-				'value'       => metadata_exists( 'post', $post->ID, '_job_expires' ) ? get_post_meta( $post->ID, '_job_expires', true ) : calculate_job_expiry( $post->ID ),
+				'placeholder' => metadata_exists( 'post', $post->ID, '_job_expires' ) ? esc_attr_x( 'yyyy-mm-dd', 'Date format placeholder.', 'wp-job-manager' ) : calculate_job_expiry( $post->ID ),
```
jom: No need to escape attribute (`_x()` is fine here). Attributes are escaped in templates.

wp-job-manager-functions.php
```diff
@@ -23,7 +23,7 @@ function get_job_listings( $args = array() ) {
 		'fields'            => 'all'
 	) );
 
-	if ( false == get_option( 'job_manager_hide_expired_content', 1 ) ) {
+	if ( false == get_option( 'job_manager_hide_expired', get_option( 'job_manager_hide_expired_content' ) ) ) {
```
jom: Do we want `get_option( 'job_manager_hide_expired_content' )` to have the original default value (`1`)?

includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -556,6 +556,20 @@ public function save_job_listing_data( $post_id, $post ) {
 				}
 			}
 		}
+
+		/* Set Post Status To Expired If Already Expired */
+		$expiry_date = get_post_meta( $post_id, '_job_expires', true );
+		$today_date  = date( 'Y-m-d', current_time( 'timestamp' ) );
+		$post_status = $expiry_date && $today_date > $expiry_date ? 'expired' : false;
+		if( $post_status ){
```
jom: Mind [adding space around parenthesis](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage) of this `if ( ... ) {}` statement?

turtlepod:
I added the fix as requested 👍 nice catch!

jom:
Tested. Looks good to me! Thanks, @turtlepod!

PR #905, opened 2017-03-25
Update class-wp-job-manager-form-submit-job.php

mreding:
Added contact name and telephone numbers

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -209,6 +209,41 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 					'required'    => true,
 					'placeholder' => __( 'Enter the name of the company', 'wp-job-manager' ),
 					'priority'    => 1
+					
+				'contact_first_name' => array(
+					'label'       => __( 'contact first name', 'wp-job-manager' ),
+					'type'        => 'text',
+					'required'    => true,
+					'placeholder' => __( 'Enter the contact first name', 'wp-job-manager' ),
+					'priority'    => 1
+					
+				'contact_last_name' => array(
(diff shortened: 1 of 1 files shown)


tripflex:
Have you discussed adding these fields with anybody on the WPJM team?  I'm not sure if we want to add additional default fields as this would mean anybody that doesn't want these fields would need to add their own custom PHP to remove them.

@kraftbj what do you think?

mreding:
Hello;
I am not concerned with what WPJM wants, these fields I need are for my website alone. Besides, it makes no sense to have an Add Job screen without any company contact names or phone numbers. 

Furthermore, the WPJM Candidate Application screen assumes that everyone is applying for a job on the Current Jobs list, which is a flawed assumption and that is why there is not a shortcode for an Application screen. I had to use Gravity Forms to create one for the front end of my website. I am not impressed with the WPJM plugin family even though I bought entire collection.
Thanks for the prompt reply
Malcolm Reding

spencerfinnell:
> I am not concerned with what WPJM wants, these fields I need are for my website alone. 

You should add these as a plugin to your website then. Or use @tripflex's plugin to do so. 

-1 from me.

mreding:
I do not need a plugin, I just need to add some code to the form:

 class-wp-job-manager-form-submit-job.php

dbtlr:
@mreding: You are more than welcome to add this to your own version of the plugin. That is the beauty of open-source after all, however, we have thousands of others whom this change will effect. As you say:

> I am not concerned with what WPJM wants

I am very concerned with this, however.

For the time being, I'm going to close this PR. If you would like to create a discussion for a better default form, please do so in the [issues](https://github.com/Automattic/WP-Job-Manager/issues). As for now, the ability to extend this WPJM and modify it to your own needs is our primary purpose and the use of more specific plugins and @spencerfinnell pointed out is the correct path to solve this.

Issue #798, opened 2016-11-04
Define a Primary Column in CPT List Table


spencerfinnell:
#### What I Expected To Happen

Responsive list table in "Job Listings"

#### What Happened Instead

Jumbled mess! https://cloudup.com/cBNKItlLgQq

#### Steps to Reproduce the Bug

Visit "Job Listings" in the WordPress dashboard on a mobile device.

https://make.wordpress.org/core/2015/08/08/list-table-changes-in-4-3/

jom:
Fixed in #906

Issue #735, opened 2016-08-08
Remove Padding From "Job Listing Data" Metabox


spencerfinnell:
https://github.com/Automattic/WP-Job-Manager/blob/c838d6ee3a3d0fd224d666bfee55f58517e10cf6/assets/css/admin.less#L198

Seems a bit random and makes the UI look inconsistent compared to other metaboxes.

PR #906, opened 2017-03-27
Responsive Admin Fix

turtlepod:
Related Issue: https://github.com/Automattic/WP-Job-Manager/issues/798

### Fixes in this PR:

- Set "Job Position" as primary columns
- Set "Job Position" as 1st column
- Remove Action Links in Primary Columns.
- Add "Show more details" button in "Job Position" custom columns callback
- Fix CSS for action buttons in mobile


dbtlr (changes requested):
There appears to be a few style bugs still.

screen shot 2017-03-27 at 10 41 33 am

As you can see the bottom margin is not being held properly. It appears to be coming from this rule:

```css
/* list-tables.css?ver=4.7.3:1793 */
@media screen and (max-width: 782px)
.wp-list-table tr:not(.inline-edit-row):not(.no-items) td:not(.check-column) {
    display: block;
}
```

Removing this `display: block;` reverts to the default style: `display: table-cell;`

Which appears to solve the issue:

screen shot 2017-03-27 at 10 41 47 am

Otherwise, there are a couple of style bits that would be nice to fix with the open panel: 

screen shot 2017-03-27 at 10 48 52 am

turtlepod:
@dbtr: in https://github.com/Automattic/WP-Job-Manager/pull/906/commits/d7d38f4376e7fd0f1f80cf11c7860a3b390f5f59
I also make job data meta box responsive and address #735

jom (approved):
LGTM

Issue #794, opened 2016-10-28
Set a placeholder for Job Types Multiselect


spencerfinnell:
When multiple job types are enabled the default "Choose a category..." placeholder is used instead of a relevant one.

jom:
Fixed in #907

PR #907, opened 2017-03-27
Job Type placeholder

turtlepod:
Issue: https://github.com/Automattic/WP-Job-Manager/issues/794

Note: I didn't update the pot file. Please add it after merge.
(not sure why but grunt makepot did not work in my install)

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -173,7 +173,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 					'label'       => __( 'Job type', 'wp-job-manager' ),
 					'type'        => $job_type,
 					'required'    => true,
-					'placeholder' => '',
+					'placeholder' => __( 'Choose job type&hellip;', 'wp-job-manager' ),
 					'priority'    => 3,
 					'default'     => 'full-time',
 					'taxonomy'    => 'job_listing_type'
--- a/languages/wp-job-manager.pot
+++ b/languages/wp-job-manager.pot
@@ -4,7 +4,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: WP Job Manager 1.25.3\n"
 "Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/wp-job-manager\n"
-"POT-Creation-Date: 2017-03-08 15:09:11+00:00\n"
+"POT-Creation-Date: 2017-03-27 04:41:25+00:00\n"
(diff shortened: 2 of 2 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -173,7 +173,7 @@ public function init_fields() {
 					'label'       => __( 'Job type', 'wp-job-manager' ),
 					'type'        => $job_type,
 					'required'    => true,
-					'placeholder' => '',
+					'placeholder' => __( 'Choose job type&hellip;', 'wp-job-manager' ),
```
dbtlr: Adding the HTML entity `&hellip;` into the translation might make translations more difficult, as translators need to know what this means AND it is inconsistent, as no other placeholders in this form use them.

I would suggest ditching the `&hellip;` but at bare minimum, move them outside of the translation like so: `__( 'Choose job type', 'wp-job-manager' ) . '&hellip;'`

dbtlr (changes requested):

includes/forms/class-wp-job-manager-form-submit-job.php
turtlepod: That hellip is to override this one:
https://github.com/Automattic/WP-Job-Manager/blob/v1.14.0/wp-job-manager-functions.php#L442

Which also using hellip (so, it's consistent)
Do you want me to change it to something else too?

turtlepod: Plus, you need to translate hellip. Your method might not work for non-latin language.
It's how WP do it too.
https://github.com/WordPress/WordPress/blob/7ad554f955a1eac48c0361c9997af5cbe8c49dd8/wp-content/themes/twentyseventeen/searchform.php#L19

dbtlr: Fair enough :)

dbtlr (approved):

includes/forms/class-wp-job-manager-form-submit-job.php
turtlepod: Just info: more hellip in WP core:
https://github.com/WordPress/WordPress/search?utf8=%E2%9C%93&q=%26hellip&type=

dbtlr:
LGTM: Still needs updated POT file though. Will merge this in later.

turtlepod:
Yeah, I cannot make the "makepot" works in my local dev.
I'm not sure why.

Please update the pot as needed/after merge.

turtlepod:
makepot is finally working in my local dev install. YEY!
(Sorry for the delay)

Issue #900, opened 2017-03-17
Searching for certain keywords return all results


chinmeh:
When searched for keywords **wp-content, tent, wp, content, uploads,etc** all listings are displayed instead of those listings which have the keyword in their post title, post content or valid post meta.

Steps to Reproduce the Bug:
- Visit the Jobs Page.Make sure you have listings containing attachments (image, doc)
- Search with keywords wp, wp-content,uploads, content,tent, etc. you would get all listings which have an attachment associated in the results.

Approach for solving the issue:
-Meta keys containing paths to attachments should be replaced by attachment ID or be excluded from the search query.

turtlepod:
I just want to mention that this happen for regular WordPress search too.
Add image in a post and search for "uploads" or "wp" and it will return in search results.

So, this is WP behavior not really WP Job Manager.

spencerfinnell:
There could be a blacklist of `meta_key`s that are not checked here: https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L182

Not sure what performance impact it would have though. 

There are no default fields that store a URL in meta anymore, however old data can potentially exist in the original `_company_logo` field, and themes/plugins can introduce their own fields (both of our themes do). It would be nice to have a way to easily exclude these from searching when registering fields for the submission form.

jom:
Fixed in #910

PR #910, opened 2017-03-28
Add filterable search-able meta keys 

turtlepod:
Currently WPJM search all meta value,
This will cause issue explained in https://github.com/Automattic/WP-Job-Manager/issues/900

> When searched for keywords wp-content, tent, wp, content, uploads,etc all listings are displayed

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -177,7 +177,7 @@ if ( ! function_exists( 'get_job_listings_keyword_search' ) ) :
 	function get_job_listings_keyword_search( $args ) {
 		global $wpdb, $job_manager_keyword;
 
-		/* Searchable Meta Keys: set to empty to search all meta keys */
+		// Searchable Meta Keys: set to empty to search all meta keys
 		$searchable_meta_keys = array(
 			'_job_location',
 			'_company_name',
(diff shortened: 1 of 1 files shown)


wp-job-manager-functions.php
```diff
@@ -177,9 +177,38 @@ function get_job_listings( $args = array() ) {
...
+		/* Search Post Title  */
 		$conditions[] = "{$wpdb->posts}.post_title LIKE '%" . esc_sql( $job_manager_keyword ) . "%'";
-		$conditions[] = "{$wpdb->posts}.ID IN ( SELECT post_id FROM {$wpdb->postmeta} WHERE meta_value LIKE '%" . esc_sql( $job_manager_keyword ) . "%' )";
+
+		/* Search Post Meta */
+		if( apply_filters( 'job_listing_search_post_meta', true ) ){
+
+			/* Only selected meta keys */
+			if( $searchable_meta_keys ){
+				$conditions[] = "{$wpdb->posts}.ID IN ( SELECT post_id FROM {$wpdb->postmeta} WHERE meta_key IN ( '" . implode( "','", $searchable_meta_keys ) . "' ) AND meta_value LIKE '%" . esc_sql( $job_manager_keyword ) . "%' )";
```
jom: Results from the new filter must be escaped before putting it in a SQL statement.

Like: `implode( "','", array_map( 'esc_sql', $searchable_meta_keys ) )`

jom (changes requested):
Thanks for this PR! This will be a big improvement.

I've added one change request. In addition, we're going to start enforcing WordPress coding and documentation standards. We'll be cleaning up core slowly, but new code should conform. Only wee things I can see are single line comments should use `//` format (https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#5-1-single-line-comments) and spaces around `if (...) {}` (https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage).

jom:
Tested. Looks good!

tripflex:
This is going to cause a huge issue with searching custom fields that users have created manually via PHP or with my plugin ... i'll have to figure out how to handle this, maybe just have to add all custom fields via filter to make it work.

... but those who added their own custom fields via PHP, will have issues when this is released into the wild.

tripflex:
Also going to cause issues for themes or other plugins that have added custom fields they want searchable through keywords, which will stop working after this is released.

spencerfinnell:
I'm not sure people really realized meta was being searched to begin with. WordPress doesn't look in meta with its built in search. 

@tripflex You could filter it like you said or you could disable it completely with the `job_listing_searchable_meta_keys` filter.

tripflex:
@spencerfinnell yeah i'll probably just disable it by returning **false** in the `job_listing_searchable_meta_keys` filter, and then handling the actual issue as you mentioned in #900 to blacklist certain meta values, or something ... hmm not sure yet, but will have to figure something out as I know this will cause tons of support requests about keywords not working anymore.

tripflex:
Major concern is going to be other themes or plugins that specifically expect all meta to be searched through keywords, so it would be best @kraftbj to include something in changelog that specifically states that ONLY those meta keys will be searched through keywords if the user updates

spencerfinnell:
Are there other themes that rely on that you think? I've never really thought about it/expected it with Listify. But you certainly have better insight on expectations/how people use custom fields...

jom:
They should be able to hook into `job_listing_searchable_meta_keys` and merge their list of custom fields with the default array. This is only fired during the query generation in `get_job_listings()` and I think also the RSS feed.

Issue #732, opened 2016-08-04
Duplicating a Job Listing Breaks Multiple File Upload Fields


spencerfinnell:
#### What I Expected To Happen

Duplicate a job listing that has a file upload field set to `multiple` and all images be copied.
#### What Happened Instead

The images do not appear copied.
#### Steps to Reproduce the Bug
1. Set a file upload field to accept multiple images.
2. Submit a listing with multiple images in the field.
3. Duplicate the listing.
#### Recommended Fix

Unserialize any data before sending it back through `update_post_meta`

`update_post_meta( $new_post_id, $meta_key, maybe_unserialize( $meta_value ) );`

https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L813

chaselivingston:
@spencerfinnell How are you adding multiple file upload fields exactly? Is it the Listing Gallery feature in Listify?

spencerfinnell:
Yes.

Just filtering the fields:

```
public function gallery_images( $fields ) {
	$fields[ 'job' ][ 'gallery_images' ] = array(
		'label'       => __( 'Gallery Images', 'listify' ),
		'type'        => 'file',
		'multiple'    => true,
		'required'    => false,
		'placeholder' => '',
		'priority'    => 4.999,
		'ajax'        => true,
		'allowed_mime_types' => array(
			'jpg'  => 'image/jpeg',
			'jpeg' => 'image/jpeg',
			'gif'  => 'image/gif',
			'png'  => 'image/png'
		)
	);
	return $fields;
}
```

tripflex:
Users can also add multiple upload fields by using my field editor plugin as well, which there are probably at least 500+ doing so already (if not more) ... using clone, probably not, but custom multiple file upload fields, yes.

spencerfinnell:
Definitely very common. Not theme specific. Any multi upload field will fail on duplication.

chaselivingston:
@spencerfinnell @tripflex Thanks for the information!

turtlepod:
Hi All,
How to duplicate a listing. I don't think WPJM have this functionality.
Is it using a plugin? like https://wordpress.org/plugins/post-duplicator/ ?

jom:
Fixed in #911

PR #911, opened 2017-03-28
Serialize Data Issue on Duplication

turtlepod:
Problem with multiple fields: https://github.com/Automattic/WP-Job-Manager/issues/732
Because multiple is serialize data. So, we need to unserialize it before duplicating the data.

Note: I think it's a bug in WP that it try to serialize serialized data using `maybe_serialize()` (?)

#### CMIIW:
The duplicate functionality (job dashboard shortcode) works well for string data,
But if the data is already serialize `update_metadata` will perform `maybe_serialize` that cause the data fail to be unserialized correctly.
https://github.com/WordPress/WordPress/blob/4.7.3/wp-includes/meta.php#L212

#### Example Code:
```
$data = array(
	'1' => 'one',
	'2' => 'two',
	'3' => 'three',
);
$data = serialize( $data ); // a:3:{i:1;s:3:"one";i:2;s:3:"two";i:3;s:5:"three";}
$data = maybe_serialize( $data ); // s:50:"a:3:{i:1;s:3:"one";i:2;s:3:"two";i:3;s:5:"three";}";
$data = maybe_unserialize( $data ); // a:3:{i:1;s:3:"one";i:2;s:3:"two";i:3;s:5:"three";}
```

The correct procedure:
```
$data = array(
	'1' => 'one',
	'2' => 'two',
	'3' => 'three',
);
$data = serialize( $data ); // a:3:{i:1;s:3:"one";i:2;s:3:"two";i:3;s:5:"three";}
$data = maybe_unserialize( $data ); // THIS STEP MISSING! back to array()
$data = maybe_serialize( $data ); // a:3:{i:1;s:3:"one";i:2;s:3:"two";i:3;s:5:"three";}
$data = maybe_unserialize( $data ); // back to array()
```

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -910,7 +910,7 @@ function job_manager_duplicate_listing( $post_id ) {
 			if ( in_array( $meta_key, apply_filters( 'job_manager_duplicate_listing_ignore_keys', array( '_filled', '_featured', '_job_expires', '_job_duration', '_package_id', '_user_package_id' ) ) ) ) {
 				continue;
 			}
-			update_post_meta( $new_post_id, $meta_key, $meta_value );
+			update_post_meta( $new_post_id, $meta_key, maybe_unserialize( $meta_value ) );
 		}
 	}
 


jom (approved):
LGTM

PR #931, opened 2017-04-26
Base setup of test suite with Travis support

jom:
This adds PHP unit test skeleton with PHPUnit and Travis.

Borrowed from what I liked in Jetpack and Sensei :)


dbtlr (approved):
Great to see this coming together!

pgk:
Good Stuff! Sensei uses circleci for tests,  @jom @dbtlr should we switch that back to travis too to keep things consistent?

PR #932, opened 2017-04-26
Add REST API Version Endpoint

pgk:

* Add REST API Version Endpoint to wpjm

#### Testing instructions:

* add `define( 'WPJM_REST_API_ENABLED', true );` to wp-config
* If you GET (or curl) `<yoursite>/wp-json/wpjm/v1/version` you get a HTTP 200 response with this body: `{"job_manager_version":"1.25.3"}`


jom (changes requested):
It'd be good to go over the architecture of Mixtape a bit more before putting it in production. Mixtape also seems to have a few Sensei artifacts in there. I'm curious about the decision to use the class factories that are duplicating the library files. Is it less a library and more of a code generator?

pgk:
Closed in favor of #987

PR #934, opened 2017-04-27
Update docblock of WP_Job_Manager_API::add_query_vars

pgk:
It now correctly states the return type and args

### DIFF
--- a/includes/class-wp-job-manager-api.php
+++ b/includes/class-wp-job-manager-api.php
@@ -24,7 +24,8 @@ class WP_Job_Manager_API {
 	 * add_query_vars function.
 	 *
 	 * @access public
-	 * @return void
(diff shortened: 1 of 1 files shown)


includes/class-wp-job-manager-api.php
```diff
@@ -24,7 +24,8 @@ public function __construct() {
 	 * add_query_vars function.
 	 *
 	 * @access public
-	 * @return void
+	 * @param $vars array the query vars
```
jom: Need to swap the type and name of param:
```
@param array $vars  the query vars
```

pgk: Fixed, and amended. I also removed the ref to a git submodule which slipped in

PR #960, opened 2017-05-10
Use WP_Query's search capability for core post fields

### LINKED ISSUE #894: Search with more than one keyword - zero results
Im using jobify for my job listings. When I use the search bar and type in more then one keyword in the search-field, zero jobs get found, even when I use two exact keywords that appear in the headline of a certain job_posting. The only way it works, when I type the words in the exact order of the title of a job listing.

Example:
Title of Job LIsting: Manager for research-project
Search : Manager research - 0 Results
Search: research-project for Manager - 0 Results
Search: Manager for research - 1 Result

I changed the code like in #859 suggested but it didnt help.

Can anyone help me with that issue. Thanks a lot!

jom:
Closes #894

Addresses issue where search queries would do exact string searches within title and content. WP core breaks up words in a search string and also searches the excerpt. This PR uses this functionality in WP core for the core post fields. Original issue still exists within taxonomy and meta fields.

#### Testing instructions:

* Create several published job listings. For example: "Senior UX Designer" and "Senior Web Developer"
* From the `[jobs]` shortcode, search for "senior designer" and verify that "Senior UX Designer" appears. 
* Try to also verify the location and job type filters still work.

Enhancement - Improves job filtering by breaking up search terms.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -114,8 +114,8 @@ function get_job_listings( $args = array() ) {
 	$job_manager_keyword = sanitize_text_field( $args['search_keywords'] );
 
 	if ( ! empty( $job_manager_keyword ) && strlen( $job_manager_keyword ) >= apply_filters( 'job_manager_get_listings_keyword_length_threshold', 2 ) ) {
-		$query_args['_keyword'] = $job_manager_keyword; // Does nothing but needed for unique hash
-		add_filter( 'posts_clauses', 'get_job_listings_keyword_search' );
+		$query_args['s'] = $job_manager_keyword;
+		add_filter( 'posts_search', 'get_job_listings_keyword_search' );
 	}
 
 	$query_args = apply_filters( 'job_manager_get_listings', $query_args, $args );
@@ -162,7 +162,7 @@ function get_job_listings( $args = array() ) {
 
 	do_action( 'after_get_job_listings', $query_args, $args );
 
-	remove_filter( 'posts_clauses', 'get_job_listings_keyword_search' );
+	remove_filter( 'posts_search', 'get_job_listings_keyword_search' );
 
 	return $result;
 }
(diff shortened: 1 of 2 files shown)


Bertsi:
Hi jom, thanks for your solution. I am testing it right now and theres one problem left:
My title is:    professor for computerscience
search:         professor for computer  (correct match)
search:         prof for computerscience (correct match)
search:         computerscience prof (correct match)
search:         prof for computer (no match)

So the think is, one of the words I am searching for has to be exact, otherwise the search finds nothing. Is there a way finding results without having one exact keyword? For instance: "prof for computer"

Thanks a lot!

jom:
@Bertsi Looking at WordPress core's code, that should produce results. I tried it with my environment:
screen shot 2017-05-12 at 10 12 25 am

This is still limited by the restrictions of WordPress core's search capability. The "next level" of search is quite a big jump. In the future, I'm interested in looking at how we can improve compatibility with plugins that augment WordPress' search functionality (like FacetWP and SearchWP, who advertises WPJM compatibility).

wp-job-manager-functions.php
```diff
@@ -212,15 +212,29 @@ function get_job_listings_keyword_search( $args ) {
...
+		if ( empty( $conditions ) ) {
+			return $search;
 		}
 
-		$args['where'] .= " AND ( " . implode( ' OR ', $conditions ) . " ) ";
+		$conditions_str = implode( ' OR ', $conditions );
 
-		return $args;
+		if ( ! empty( $search ) ) {
+			$search = preg_replace( '/^ AND /', '', $search );
```
dbtlr: I'm not sure it is clear what this preg replace is doing. I mean, I can guess, but it would be nice to know what the assumption of what the `$search` variable has in it and why clearing out the AND keyword works here. 

Is the assumption that `$search` will always have just one clause in it at this point?

dbtlr (approved):
Overall, I agree with your assessment that the next level of search is a huge jump that is beyond the scope of what core WordPress can really do, without specialized software like elastic search. 

It looks good overall, just would like the one point of clarity with the mentioned search clause.

wp-job-manager-functions.php
jom: WP Query starts its `WHERE` query segment with `1=1`. The `posts_search` filter prefaces its passed default value with ` AND `. I'm guessing they do this for more predictable chaining, but it still makes using the filter a bit clunky. Other hooked filters may have added other things after the default search query from WP core, but the assumption is that whatever is passed will start with ` AND ` or will be an empty string.

jom:
Holding pending @jchristopher's review for SearchWP. Hoping to get into core next week.

jchristopher:
Just took a pass through this PR, all good on my end 👍

Again, thank you so much for the proactive action letting me know about this upcoming change!

PR #963, opened 2017-05-11
Move WPML and Polylang code into separate files

### LINKED ISSUE #962: Segregate WPML compatibility code into its own file
Move all WPML related code into one file and tie to `wpml_loaded` action (see [Jetpack's implementation as an example](https://github.com/Automattic/jetpack/blob/master/3rd-party/wpml.php)).

Found code:
- [/wp-job-manager.php:#153](https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager.php#L153)
- [/wp-job-manager-functions.php:#49](https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L49)

jom:
Clean up code by moving tweaks for WPML and Polylang to separate files.
Used Jetpack for inspiration.

Added the filter `wpjm_lang` for retrieving the current language. This is used when creating cache key signatures.

Added the action `get_job_listings_init` which fires at the start of the `get_job_listings()` function. Used for updating WPML's language when it is an Ajax request.

Fixes #962

@vukvukovich If you wouldn't mind, please review the code changes related to WPML. Thank you!

#### Testing instructions:

* Separately for WPML and Polylang:
1) Create job listings and pages with WPJM shortcodes in two languages.
2) View job listing in one language and then switch language. Verify correct job listing language shows up on page refresh.
3) View list of job listings (`[jobs]`) in one language, switch language, refresh to verify you're given correct content.
4) View other WPJM shortcode pages. Verify the correct version shows up for different languages.

Moved third-party multi-language (WPML and Polylang) tweaks into separate files.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -4,7 +4,7 @@ if ( ! function_exists( 'get_job_listings' ) ) :
  * Queries job listings with certain criteria and returns them.
  *
  * @since 1.0.5
- * @param array $args
+ * @param string|array|object $args Arguments used to retrieve job listings.
  * @return array
  */
 function get_job_listings( $args = array() ) {
@@ -23,6 +23,15 @@ function get_job_listings( $args = array() ) {
 		'filled'            => null,
 		'fields'            => 'all'
 	) );
+
+	/**
+	 * Perform actions that need to be done prior to the start of the job listings query.
+	 *
+	 * @since 1.26.0
+	 *
+	 * @param array $args Arguments used to retrieve job listings.
+	 */
+	do_action( 'get_job_listings_init', $args );
 
 	if ( false == get_option( 'job_manager_hide_expired', get_option( 'job_manager_hide_expired_content', 1 ) ) ) {
 		$post_status = array( 'publish', 'expired' );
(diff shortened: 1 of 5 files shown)


spencerfinnell:
Are the `function_exists()` calls necessary since it only happens after being hooked in to a Polylang action? Doesn't hurt how it is now but might make more sense to do a check before loading the file to begin with.

jom:
Checks are not completely necessary, but just in the unlikely case they change their functions or someone creates a conflicting `pll_init` action. Just a tiny bit safer.

spencerfinnell:
👍 Looks good to me then.

vukvukovich:
All pages look neat. Code is executing well, you can merge it. 👍🏻

wp-job-manager-functions.php
```diff
@@ -638,12 +640,8 @@ function job_manager_dropdown_categories( $args = '' ) {
...
-	// WPML & Polylang caching per language
-	if ( defined( 'ICL_SITEPRESS_VERSION' ) ) {
-		$r['lang'] = apply_filters( 'wpml_current_language', NULL );
-	} elseif ( function_exists( 'pll_current_language' ) ) {
-		$r['lang'] = pll_current_language();
-	}
+	/** This filter is documented in wp-job-manager.php */
+	$r['lang'] = apply_filters( 'wpjm_lang', null );
 
 	extract( $r );
```
dbtlr: I know this is standard WordPress to use `extract`, but this makes me want to hulk smash all over this code (not that you added it) :)

dbtlr (approved):
Overall, these are smart changes. I like the abstractions used. Makes it feel easy to add in new providers later on. :)

PR #975, opened 2017-05-17
Shifting from expired to published extends expiration date

### LINKED ISSUE #974: Adjust Listing Expiration When Changing Status

Changing listing status to "Active" from "Expired"  should actually set the listing as active by adjusting the existing listing expiration based on the original duration set. 

#### What Happened Instead

The listing remains expired because its current expiration date has already passed.

1. Create a listing with an expiration date in the past.
2. Publish listing. 
3. Set listing to active.

jom:
Fixes #974

* Extend expiration date if post is expired and is changed to published

#### Testing instructions:

* See #974

### DIFF
--- a/includes/admin/class-wp-job-manager-writepanels.php
+++ b/includes/admin/class-wp-job-manager-writepanels.php
@@ -585,16 +585,32 @@ class WP_Job_Manager_Writepanels {
 		/* Set Post Status To Expired If Already Expired */
 		$expiry_date = get_post_meta( $post_id, '_job_expires', true );
 		$today_date  = date( 'Y-m-d', current_time( 'timestamp' ) );
-		$post_status = $expiry_date && $today_date > $expiry_date ? 'expired' : false;
-		if( $post_status ) {
+		$is_job_listing_expired = $expiry_date && $today_date > $expiry_date;
+		if( $is_job_listing_expired ) {
 			remove_action( 'job_manager_save_job_listing', array( $this, 'save_job_listing_data' ), 20, 2 );
-			$job_data = array(
-				'ID'          => $post_id,
-				'post_status' => $post_status,
-			);
-			wp_update_post( $job_data );
+			if ( $this->is_job_listing_being_reactivated() ) {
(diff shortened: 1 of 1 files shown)


includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -585,14 +585,23 @@ public function save_job_listing_data( $post_id, $post ) {
...
+		$expire_date_in_past = $expiry_date && $today_date > $expiry_date;
+		if( $expire_date_in_past ) {
 			remove_action( 'job_manager_save_job_listing', array( $this, 'save_job_listing_data' ), 20, 2 );
-			$job_data = array(
-				'ID'          => $post_id,
-				'post_status' => $post_status,
-			);
-			wp_update_post( $job_data );
+			if (
+				isset( $_POST['post_status'] )
```
pgk: This can be made more intention-revealing if we extract method (something like `is_job_listing_reactivated` or similar)

includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -585,14 +585,23 @@ public function save_job_listing_data( $post_id, $post ) {
 		/* Set Post Status To Expired If Already Expired */
 		$expiry_date = get_post_meta( $post_id, '_job_expires', true );
 		$today_date  = date( 'Y-m-d', current_time( 'timestamp' ) );
-		$post_status = $expiry_date && $today_date > $expiry_date ? 'expired' : false;
-		if( $post_status ) {
+		$expire_date_in_past = $expiry_date && $today_date > $expiry_date;
```
pgk: not necessary, but maybe this could be called `$is_job_listing_expired`

includes/admin/class-wp-job-manager-writepanels.php
jom: Agreed!

jom:
@pgk ✅

pgk:
Sweet as Sugar @jom!

Issue #924, opened 2017-04-14
Job position and to address emails are not passed to email services


kumaresandev:
#### What I Expected To Happen

When i try to apply for the job it shows the email services google,outlook,AOL.
When i click on gmail it passing the to email address and job title to the gmail, but for other services (outlook,AOL) it’s not working.

#### What Happened Instead

Gmail only passing the arguments and not working in outlook, AOL, Yahoo 

#### Steps to Reproduce the Bug

Click on apply for job button in job details page and click on gmail.

Now the admin email paased to to addresss and job position passed to subject.

Then click on outlook, it's only redirect to outlook mail but not passing any arguments to mail server.

Website url : http://bocacosmeticsgroup.com/bocacosmetics-sub/job/senior-programmer/

jom:
Need to see if these services still support this or if they have a workaround. Might need to be taken out of WPJM if they no longer support it.

spencerfinnell:
+1 for removing the direct email links. Always seemed out of place to me anyway.

jom:
Agreed. I've removed them in #977.

PR #977, opened 2017-05-18
Remove webmail service links

jom:
Closes: #924 

Out of place and most no longer work. In template file so folks can add
back in if necessary.


spencerfinnell:
👍

PR #988, opened 2017-05-23
Introduce REST API settings endpoint take 2

### LINKED ISSUE #986: Create WPJM REST API Settings Endpoint
placeholder

pgk:
Closes #986

N.B. Reviewers: 

Most file changes are additions of generated files, crammed into one commit. You can Ignore that commit. 

The rest of the commits are focused, each introducing/altering one thing. Makes it easier to review each separately.

(Mixtape generated files with: `./scripts/new_project.sh WPJM_REST ./../wp-job-manager/lib/wpjm_rest`)

* Introduces `wpjm/v1` REST API Namespace
* Introduces `wpjm/v1/settings` Endpoint for viewing/updating Settings

#### Testing instructions:

* On your `wp-config.php` add `define( 'WPJM_REST_API_ENABLED', true );`
* `GET wpjm/v1/settings`: returns settings
* `PUT,POST,PATCH wpjm/v1/settings`: updates one or more settings
* You will need admin privileges to view/update settings using the rest api

Optional, we might want to include an entry for the whole REST API instead.

* REST API Settings Endpoint (Beta)


wp-job-manager.php
```diff
@@ -66,6 +71,9 @@ public function __construct() {
 		include( 'includes/class-wp-job-manager-forms.php' );
 		include( 'includes/class-wp-job-manager-geocode.php' );
 		include( 'includes/class-wp-job-manager-cache-helper.php' );
+		include( 'includes/rest-api/class-wp-job-manager-rest-api.php' );
+
+		add_action( 'init', array( $this, 'rest_api' ) );
```
jom: Can we hook this into `rest_api_init` so we don't have the load overhead on non-REST requests?

includes/rest-api/class-wp-job-manager-models-settings.php
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @param string                                               $field_name The fields name.
+	 * @param WPJM_REST_Model_Field_Declaration_Builder            $field_builder The field builder.
+	 * @param array                                                $field_data The field data.
+	 * @param WPJM_REST_Model_Field_Declaration_Collection_Builder $def The definition.
+	 */
+	protected function on_field_setup( $field_name, $field_builder, $field_data, $def ) {
+		$field_builder->dto_name( str_replace( 'job_manager_' , '', $field_name ) );
+
+		if ( in_array( $field_name, $this->fields_requiring_page_id_validation, true ) ) {
```
jom: I'm wondering if we should either pass `$this->fields_requiring_page_id_validation` through a filter (at some point... ideally once) or sniff this from the `type` attribute (`'page' === type`) passed with the settings?

includes/rest-api/class-wp-job-manager-models-settings.php
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class WP_Job_Manager_Models_Settings
+ */
+class WP_Job_Manager_Models_Settings extends WPJM_REST_Model_Declaration_Settings
```
jom: Is the way WPJM defines settings universal enough to warrant the abstraction (`WPJM_REST_Model_Declaration_Settings`/`MT_Model_Declaration_Settings`) for defining its schema in Mixtape?

includes/rest-api/class-wp-job-manager-models-settings.php
```diff
@@ -0,0 +1,93 @@
...
+	 * Adds validations to fields requiring page ids.
+	 * Sets field's dto name.
+	 *
+	 * @param string                                               $field_name The fields name.
+	 * @param WPJM_REST_Model_Field_Declaration_Builder            $field_builder The field builder.
+	 * @param array                                                $field_data The field data.
+	 * @param WPJM_REST_Model_Field_Declaration_Collection_Builder $def The definition.
+	 */
+	protected function on_field_setup( $field_name, $field_builder, $field_data, $def ) {
+		$field_builder->dto_name( str_replace( 'job_manager_' , '', $field_name ) );
```
jom: I can't imagine a serious issue with this causing collisions, but it is a possibility I suppose. One thing to note is I don't believe that prefix isn't enforced and WPJM plugins that hook into `job_manager_settings` do use different prefixes.

tests/php/tests/includes/rest-api/test_class.wp-job-manager-models-settings.php
```diff
@@ -0,0 +1,43 @@
...
+        $settings = $setting_definition->get_data_store()->get_entity(null);
+        $this->assertNotNull( $settings );
+        $dto = $setting_definition->model_to_dto( $settings );
+        $this->assertInternalType( 'array', $dto );
+        $this->assertArrayHasKey( 'per_page', $dto );
+        foreach ( $dto as $key => $value ) {
+            $this->assertNotContains( 'job_manager_', $key );
+        }
+    }
+}
```
jom: Missing line ending on test files.

tests/php/includes/class-wpjm-rest-testcase.php
```diff
@@ -0,0 +1,71 @@
...
+    function setUp() {
+        parent::setUp();
+        $this->environment = WPJM()->rest_api()->get_bootstrap()->environment();
+        /** @var WP_REST_Server $wp_rest_server */
+        global $wp_rest_server;
+        $this->rest_server = $wp_rest_server = new WP_REST_Server;
+        do_action( 'rest_api_init' );
+    }
+
+    function assertClassExists( $cls) {
```
jom: Before merge, it'd be nice to add visibility and docs to test functions.

includes/rest-api/class-wp-job-manager-models-settings.php
```diff
@@ -0,0 +1,93 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class WP_Job_Manager_Models_Settings
+ */
+class WP_Job_Manager_Models_Settings extends WPJM_REST_Model_Declaration_Settings
+	implements WPJM_REST_Interfaces_Permissions_Provider {
```
jom: As a huge fan of interfaces, I ❤️ their abundant usage in Mixtape.

jom (changes requested):
I just started a review of non-Mixtape stuff. Just to break this down, I wonder if there is a way to review Mixtape separately. I'm going to try to do some digging around to see what is going on with tests this afternoon.

tests/php/includes/class-wpjm-rest-testcase.php
pgk: Good call!

tests/php/tests/includes/rest-api/test_class.wp-job-manager-models-settings.php
pgk: Fantastic!

includes/rest-api/class-wp-job-manager-models-settings.php
pgk: I am not sure we can rely on settings having type `page` set. +1 for Filtering! 👍

includes/rest-api/class-wp-job-manager-models-settings.php
pgk: another option is to not assign a different DTO name and just let it be exactly the same as the setting name! Happy with that too!

includes/rest-api/class-wp-job-manager-models-settings.php
pgk: I was thinking it would suffice, as long as the override of `WPJM_REST_Model_Declaration_Settings::get_settings` returns a similarly shaped array. Maybe this is a documentation thing?

wp-job-manager.php
pgk: Good point! We can, I'll have to change some things around on MT, definitely doable and sensible.

includes/rest-api/class-wp-job-manager-models-settings.php
jom: It seems like the logic that translates the settings array into the schema is mostly in `MT_Model_Declaration_Settings`. Do you think that will be used anywhere else besides WPJM?

includes/rest-api/class-wp-job-manager-models-settings.php
jom: That helps render the page select. It should always be set. We could also add a validation attribute to the settings.

jom: If not the validation field, I wonder if we could at least build the default list with fields that are `'page' === type`... that would allow addons who add pages (I know a couple) to not have to hook to the filter.

includes/rest-api/class-wp-job-manager-models-settings.php
pgk: I think it will, under two conditions (which might not hold, but I'd rather they did):

1. Plugins will want the settings endpoint to return a flat object
2. Classes override `MT_Model_Declaration_Settings ::get_settings` and return a correctly shaped array.

pgk: We can definitely push `MT_Model_Declaration_Settings` logic into our class :)

pgk:
@jom, @dbtlr as discussed, an initial version of the `build_mixtape` functionality is now part of this pr. All mixtape-specific things are now excluded from the repo. Makes for less files to review ;)

dbtlr (approved):
LGTM

jom (approved):
🌮 💯 🌮

PR #998, opened 2017-05-25
Only add the API key if available. 

### LINKED ISSUE #951: Don't Return Empty String for Google Maps URL
```
		// Google Maps API no longer accepts requests without an API key
		if ( '' === $api_key ) {
			return false;
		}
```

This is not true -- sites who have pinged the service before the requirement was put in place do not require a key. 

Also totally resetting the URL to nothing doesn't really make sense. The request should still be made and the API error should be shown to the user when saving the listing.

spencerfinnell:
Fixes #951

* Only add the `key` parameter to geocode request if available.

#### Testing instructions:

* Add a location to a job listing and publish.

Fix: Do not automatically stop geocoding when no Google API key is available. Legacy sites still have no restrictions.

### DIFF
--- a/includes/class-wp-job-manager-geocode.php
+++ b/includes/class-wp-job-manager-geocode.php
@@ -148,14 +148,13 @@ class WP_Job_Manager_Geocode {
 	 * @return string|bool
 	 */
 	public function add_geolocation_endpoint_query_args( $geocode_endpoint_url, $raw_address ) {
+		// Add an API key if available.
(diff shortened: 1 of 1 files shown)


spencerfinnell:
Doesn't completely cover the original issue (no notice is set) but this causes issues for lots and lots of our customers so a quick fix would be appreciated.

jom:
This works for me... I'd like to get that notice in ASAP.

PR #999, opened 2017-05-26
Allow linking to specific settings tab via hash in URL

### LINKED ISSUE #993: Deep Linkable Settings Tabs

I should be able to link directly to a "Settings" tab.

#### What Happened Instead

I can only ever link to the first tab.

---

It would be nice to have some JS in place to detect a hash and switch tabs automatically when the page loads.

tripflex:
Fixes #993 

* Allow linking to specific settings tab via hash in URL

#### Testing instructions:

* Goto settings page, copy URL from one of the tabs, go to some other non-settings page, enter the URL with the hash, voila!

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -431,7 +431,17 @@ class WP_Job_Manager_Settings {
 				jQuery(this).addClass('nav-tab-active');
 				return false;
 			});
-			jQuery('.nav-tab-wrapper a:first').click();
+			var goto_hash = window.location.hash;
+			if ( goto_hash ) {
+				var the_tab = jQuery( 'a[href="' + goto_hash + '"]' );
+				if ( the_tab ) {
+					the_tab.click();
+				} else {
+					jQuery( '.nav-tab-wrapper a:first' ).click();
+				}
+			} else {
+				jQuery( '.nav-tab-wrapper a:first' ).click();
+			}
 			jQuery('#setting-job_manager_enable_registration').change(function(){
 				if ( jQuery( this ).is(':checked') ) {
 					jQuery('#setting-job_manager_registration_role').closest('tr').show();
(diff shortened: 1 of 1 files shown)


pbearne:
Is a click() call the right way do this

We did it this way in a plugin we are working on

```
			jQuery(function(){

              var tabsIds = jQuery('.job-manager-settings-wrap div').toArray().map(function(value) { return '#' + value.id } );

              if((tabsIds.indexOf(window.location.hash) > -1)) {

                jQuery('.nav-tab-wrapper a[href="#settings-job_listings"]').removeClass('nav-tab-active');
                jQuery('.settings_panel').css('display', 'none');

                jQuery('.nav-tab-wrapper a[href="' + window.location.hash + '"]').addClass('nav-tab-active');
                jQuery('div' + window.location.hash).css('display', 'block');
              }
			}); 
```

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -431,7 +431,17 @@ public function output() {
 				jQuery(this).addClass('nav-tab-active');
 				return false;
 			});
-			jQuery('.nav-tab-wrapper a:first').click();
+			var goto_hash = window.location.hash;
+			if ( goto_hash ) {
+				var the_tab = jQuery( 'a[href="' + goto_hash + '"]' );
+				if ( the_tab ) {
```
jom: This should be:
`if ( the_tab.length > 0 ) {`

jQuery's selector will always return an object that would evaluate `true` in the current line.

jom (changes requested):
This seems like a sensible solution. One wee fix and then it is good to merge. Thanks, Myles!

includes/admin/class-wp-job-manager-settings.php
tripflex: Yup you're correct, updated the PR

jom (approved):

PR #1011, opened 2017-06-02
Solves submit_job_form_required_label bug when filtering.

### LINKED ISSUE #1010: Can't filter submit_job_form_required_label without throwing a warning
When debug is turned to true and a user is not logged in, trying to submit a job and filter on submit_job_form_required_label will result in a warning.

No warnings

#### What Happened Instead

screen shot 2017-06-02 at 14 02 11

- Create fresh install
- Install plugin
- Add the following filter to functions.php

```
add_filter( 'submit_job_form_required_label', 'test_label_function', 10, 2 );

function test_label_function( $tag, $field ) {

    // do some things

	return $tag;
}
```

Pull request with fix to follow.

piersb:
Fixes #1010 

* Adds new $field array to the email textbox which is fed into the submit_job_form_required_label filter
* Field is basically just array ( 'label' => 'Email', 'type' => 'text', ) so that it can be filtered as standard; I haven't tried to do anything cleverer than that.

#### Testing instructions:

* Follow bug replication instructions in #1010 
* Patch
* Refresh screen

Adds $field to call to submit_job_form_required_label when filtering on email.

### DIFF
--- a/templates/account-signin.php
+++ b/templates/account-signin.php
@@ -17,6 +17,8 @@
 	$account_required             = job_manager_user_requires_account();
 	$registration_enabled         = job_manager_enable_registration();
 	$generate_username_from_email = job_manager_generate_username_from_email();
+    $field                        = array ( 'label' => 'Email', 'type' => 'text', );
+
 	?>
 	<fieldset>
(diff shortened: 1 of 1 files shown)


tripflex:
👍

PR #1014, opened 2017-06-06
Better Support for Multiple Job Types

### LINKED ISSUE #854: Job type treated as string instead of array
There are many places where job type is treated as it would have only one value and it causes some problems. Here are some of my findings, but there are more i believe:

**/includes/forms/class-wp-job-manager-form-submit-job.php, row 519**
-> Returns array of job type id:s ( array(1) { [0]=> int(xxx) } ) and causes problem in row 523 where it sanitizes the title. I assume that something has changed and $values['job']['job_type']; in row 519 does not return string as it used to.

**/wp-job-manager-template.php, rows 232-269**
-> the_job_type & get_the_job_type functions only return one job type value and it's random one. You cannot get all the selected job types without creating your own function.

**Admin -view**
-> In "Job listings" view "Type" column only shows one random job type, instead of all the selected types

**Job url is generated badly**
-> Job url has "array" -text instead of job type. Example: /company-address-array-position-city/

I don't have any code fixes because this goes too deep and affects many places.

### LINKED ISSUE #928: Display Multiple Job Types
@donnchawp Created a cool [feature](https://github.com/Automattic/WP-Job-Manager/commit/3f37d98d635995f06d5f6a38edaf011f5c57055a) to add multiple job types to a listing. However, it would also be cool to have the labels for the multiple types displayed on the frontend inside both the jobs and single job pages - http://cld.wthms.co/Vy6G

Came up in 3169876-t

spencerfinnell:
Fixes #854 
Fixes #928

* Add the ability to output and fetch all job types for a listing.
* Output multiple job types in template files (if enabled).
* Output multiple job slugs in job listing URL (if enabled).
* Output multiple job types in WordPress admin view.

#### Testing instructions:

* Enable multiple job types in Job Listings > Settings
* Create a listing with multiple job types.

* Fix: Full support for job listings with multiple job types.

### DIFF
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -418,9 +418,13 @@ class WP_Job_Manager_CPT {
 
 		switch ( $column ) {
 			case "job_listing_type" :
-				$type = get_the_job_type( $post );
-				if ( $type )
-					echo '<span class="job-type ' . $type->slug . '">' . $type->name . '</span>';
+				$types = get_the_job_types( $post );
+
+				if ( $types && ! empty( $types ) ) {
+					foreach ( $types as $type ) {
+						echo '<span class="job-type ' . $type->slug . '">' . $type->name . '</span>';
+					}
+				}
 			break;
 			case "job_position" :
 				echo '<div class="job_position">';
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -418,7 +418,7 @@ class WP_Job_Manager_CPT {
 
 		switch ( $column ) {
 			case "job_listing_type" :
-				$types = get_the_job_types( $post );
+				$types = wpjm_get_the_job_types( $post );
 
 				if ( $types && ! empty( $types ) ) {
 					foreach ( $types as $type ) {
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -437,16 +437,17 @@ class WP_Job_Manager_Post_Types {
 	 * Adds custom data to the job feed.
 	 */
 	public function job_feed_item() {
-		$post_id  = get_the_ID();
-		$location = get_the_job_location( $post_id );
-		$job_type = get_the_job_type( $post_id );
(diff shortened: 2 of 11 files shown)


spencerfinnell:
These need to be updated to use the new `wpjm_` prefix now.

jom:
@spencerfinnell Thanks so much for this! I updated the version numbers and function/filter prefix.

I also moved the `job_manager_multi_job_type` check from `wpjm_the_job_types()` to `wpjm_get_the_job_types()`. I thought if in implementation we always use `wpjm_get_the_job_types()` we can just loop, assuming it is an array (possibly with one item). If you're aware of any negative consequences to this let me know and we can pluck those commits out.

To do:
- [ ] URL slug generation support in `WP_Job_Manager_Form_Submit_Job::save_job()`
- [ ] Remaining use of `the_job_type()`: templates/content-summary-job_listing.php
- [ ] Remaining use of `the_job_type()`: templates/content-widget-job_listing.php
- [ ] Job feed still assuming 1 job type: `WP_Job_Manager_Post_Types::job_feed_item()`

spencerfinnell:
Those changes look good to me and I'm fine with deprecating `the_job_type()` in favor of always looping through the job type(s) with the new functions.

jom:
@spencerfinnell Rebased with master. If making additional changes, reset your local branch (`git reset --hard origin/issue/854`).

spencerfinnell:
Only thing I would say is to move the deprecated functions to a separate file #1035 to help promote their deprecation and usage of new functions. Good to go for me otherwise.

jom:
@spencerfinnell Agreed! Done.

jom (approved):
👍

fitzg2:
Sorry for opening this. What's the deal with multiple types? Seeing plenty of code that allows for multiple types yet we can't find the option to actually enable it in settings.

jom:
@fitzg2 
Screen Shot 2020-05-07 at 7 39 30 PM

fitzg2:
That seems to be the case. Thanks for the help. We went directly to the DB and changed that setting to 1. Going to let the owners know about what you said. 

For anyone wondering.. it's ''job_manager_multi_job_type'' in wp_options.

Thanks again.

PR #1020, opened 2017-06-09
Prevents Ajax upload from non-logged in users

jom:
See: p6r3EZ-my-p2

### DIFF
--- a/templates/form-fields/file-field.php
+++ b/templates/form-fields/file-field.php
@@ -4,7 +4,7 @@ $allowed_mime_types = array_keys( ! empty( $field['allowed_mime_types'] ) ? $fie
 $field_name         = isset( $field['name'] ) ? $field['name'] : $key;
 $field_name         .= ! empty( $field['multiple'] ) ? '[]' : '';
 
-if ( ! empty( $field['ajax'] ) ) {
+if ( ! empty( $field['ajax'] ) && job_manager_user_can_upload_file_via_ajax() ) {
 	wp_enqueue_script( 'wp-job-manager-ajax-file-upload' );
 	$classes[] = 'wp-job-manager-file-upload';
 }
--- a/includes/class-wp-job-manager-ajax.php
+++ b/includes/class-wp-job-manager-ajax.php
@@ -263,6 +263,10 @@ class WP_Job_Manager_Ajax {
 	 * No nonce field since the form may be statically cached.
 	 */
 	public function upload_file() {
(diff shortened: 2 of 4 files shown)


templates/form-fields/file-field.php
```diff
@@ -4,7 +4,7 @@
 $field_name         = isset( $field['name'] ) ? $field['name'] : $key;
 $field_name         .= ! empty( $field['multiple'] ) ? '[]' : '';
 
-if ( ! empty( $field['ajax'] ) ) {
+if ( ! empty( $field['ajax'] ) && job_manager_user_can_upload_file_via_ajax() ) {
```
tripflex: Only comment I have is that if theme devs and myself update our template overrides to include this, if a user installs or is using an older version of WPJM it will throw fatal error, so we will have to check if function exists first before calling, besides that looks good to me

tripflex: So my solution was to just add this function and call it from my template overrides:

```php
if( ! function_exists( 'field_editor_user_can_upload_file_via_ajax' ) ){

	/**
	 * Checks if the user can upload a file via the Ajax endpoint.
	 *
	 * This is a function added for compatibility with older versions of WP Job Manager, as this function
	 * has to be called in the file-field.php template file to prevent unauthorized ajax uploads.
	 *
	 * @since @@since
	 * @return bool
	 */
	function field_editor_user_can_upload_file_via_ajax() {

		// Added in core @since 1.26.2
		if( function_exists( 'job_manager_user_can_upload_file_via_ajax' ) ){
			return job_manager_user_can_upload_file_via_ajax();
		}

		$can_upload = is_user_logged_in() && job_manager_user_can_post_job();

		/**
		 * Override ability of a user to upload a file via Ajax.
		 *
		 * @since @@since
		 *
		 * @param bool $can_upload True if they can upload files from Ajax endpoint.
		 */
		return apply_filters( 'job_manager_user_can_upload_file_via_ajax', $can_upload );
	}

}
```

tripflex:
👍  LGTM

templates/form-fields/file-field.php
jom: That works. Since Ajax uploads would technically be allowed in previous versions, you could also just do a WPJM version check. Either works, though!

tripflex:
@razwan --- going off @jom comment, that is true, but ONLY if you're using the core template file, and not using a theme with a template override, your own template override, or any plugin that overrides that `file-field.php` file

PR #1024, opened 2017-06-09
cache post counts and fix counts not showing for custom menu labels

### LINKED ISSUE #984: Cache the pending job listings count for admin pages
Counting of pending job listings is only done when job listings post status' are changed.

#### What Happened Instead
Job listings are counted on every admin page load.

### LINKED ISSUE #1023: Pending counts do not show in admin for custom menu names
Found this bug while working on the PR i'm about to submit for caching counts, which would be fixed in that PR as well, but it appears that while searching through the menu for the menu name, core is using standard `__( 'Job Listings', 'wp-job-manager' );` instead of the actual menu name set in post type.

What happens is, when themes customize the menu name (which most of them do), counts will not show in the admin dashboard as technically that menu name is never "found".

This is easily fixed by simply pulling the post object and searching for the `menu_name` instead

https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-post-types.php#L280

tripflex:
Fixes #1023 Fixes #984 

Here's my first stab at caching the counts in transients to prevent unnecessary queries on every admin page load for counts.

* For #1023 this PR uses the post type object to get the `menu_name` when looping through the menu to look for the correct one to update (with fallback to standard translation)

* For #984 this PR adds two static methods in cache helper class, one to return listing counts, with support for multiple post types, and statuses, and one to clear the transients when statuses are updated

So initially I was going to write this code to be specifically for Job Listings, but while doing so I realized that Resumes also uses counts (exactly the same as core), and it would prob be beneficial to write the code so caching could work with any other related post types or statuses as needed (ie, adding a count for say the Applications addon menu item)

Let me know what you guys think about this

### DIFF
--- a/includes/class-wp-job-manager-cache-helper.php
+++ b/includes/class-wp-job-manager-cache-helper.php
@@ -23,6 +23,7 @@ class WP_Job_Manager_Cache_Helper {
 		add_action( 'create_term', array( __CLASS__, 'edited_term' ), 10, 3 );
 		add_action( 'delete_term', array( __CLASS__, 'edited_term' ), 10, 3 );
 		add_action( 'job_manager_clear_expired_transients', array( __CLASS__, 'clear_expired_transients' ), 10 );
+		add_action( 'transition_post_status', array( __CLASS__, 'maybe_clear_count_transients' ), 10, 3 );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/class-wp-job-manager-post-types.php
```diff
@@ -277,17 +277,21 @@ public function register_post_types() {
...
-				if ( strpos( $menu_item[0], $plural ) === 0 ) {
-					if ( $order_count = $count_jobs->pending ) {
-						$menu[ $key ][0] .= " <span class='awaiting-mod update-plugins count-$order_count'><span class='pending-count'>" . number_format_i18n( $count_jobs->pending ) . "</span></span>" ;
-					}
-					break;
-				}
+		$pending_jobs = WP_Job_Manager_Cache_Helper::get_listings_count();
+
+		// No need to go further if no pending jobs, menu is not set, or is not an array
+		if( empty( $pending_jobs ) || empty( $menu ) || ! is_array( $menu ) ){
```
tripflex: I moved these checks up as there's no need to loop through all menu items if there aren't any counts to show

includes/class-wp-job-manager-cache-helper.php
```diff
@@ -128,6 +129,105 @@ public static function clear_expired_transients() {
...
+		}
+
+		// Sometimes transients are not in the DB, so we have to do this too:
+		wp_cache_flush();
+	}
+
+	/**
+	 * Get Listings Count from Cache
+	 *
+	 *
```
jom: Remove line

includes/class-wp-job-manager-cache-helper.php
```diff
@@ -128,6 +129,105 @@ public static function clear_expired_transients() {
...
+	 * @param string  $old_status Old post status.
+	 * @param WP_Post $post       Post object.
+	 */
+	public static function maybe_clear_count_transients( $new_status, $old_status, $post ) {
+		global $wpdb;
+
+		/**
+		 * Get supported post types for count caching
+		 *
+		 * @since 1.26.2
```
jom: Mind updating these `@since` to 1.26.3? Did a bit of a milestone shuffle with the last release.

includes/class-wp-job-manager-cache-helper.php
```diff
@@ -128,6 +129,105 @@ public static function clear_expired_transients() {
...
+
+		// Get user based cache transient
+		$user_id   = get_current_user_id();
+		$transient = "jm_{$status}_{$post_type}_count_user_{$user_id}";
+
+		// Set listings_count value from cache if exists, otherwise set to 0 as default
+		$status_count = ( $cached_count = get_transient( $transient ) ) ? $cached_count : 0;
+
+		// $cached_count will be false if transient does not exist
+		if ( $cached_count === FALSE || $force ) {
```
jom: Small formatting fix: Yoda; lowercase `false` (same with function signature's default for `$force`).

includes/class-wp-job-manager-cache-helper.php
```diff
@@ -128,6 +129,105 @@ public static function clear_expired_transients() {
...
+		/**
+		 * Get supported post types for count caching
+		 *
+		 * @since 1.26.2
+		 *
+		 * @param string  $new_status New post status.
+		 * @param string  $old_status Old post status.
+		 * @param WP_Post $post       Post object.
+		 */
+		$post_types = apply_filters( 'job_manager_count_cache_supported_post_types', array( 'job_listing' ), $new_status, $old_status, $post );
```
jom: As of this release, it seems like the general direction is using the `wpjm_` prefix for all new hooks. (See: #1034)

includes/class-wp-job-manager-cache-helper.php
```diff
@@ -128,6 +129,105 @@ public static function clear_expired_transients() {
...
+		/**
+		 * Get supported post statuses for count caching
+		 *
+		 * @since 1.26.2
+		 *
+		 * @param string  $new_status New post status.
+		 * @param string  $old_status Old post status.
+		 * @param WP_Post $post       Post object.
+		 */
+		$valid_statuses = apply_filters( 'job_manager_count_cache_supported_statuses', array( 'pending' ), $new_status, $old_status, $post );
```
jom: Same as above.

tripflex:
@jom no problem, just updated pretty sure i handled all the comments on everything.

What are your thoughts on using `@@since` and replacing with grunt on deploy?  That's how I handle all my plugins, just tag them with `@since @@since` and then in grunt deploy I have it replace all instances of @@since with the version release number (to eliminate things like this, where a release is pushed before another branch can be merged)

jom:
LGTM! We could add a grunt task and then add a commit on the version bump PR.

Issue #886, opened 2017-03-03
WooCommerce Paid Listings – multilingual expiry dates not synchron after monthly Stripe payment


nz352:
In multilingual LISTABLE DIRECTORY Theme I have English, German and French "Paid Listings" by monthly subscriptions. On February 28th, after Stripe Payment is succesfully processed, only the Original English Listing, gets a new expiry date of March 31, 2017 - the French and German counterparts of the Listing do not get a new Expiry Date.

I expected to see: The Original EN Listing and the FR & DE Listings should always have the same expiry date - otherwhise EN Listing is visible, but FR & DE are expired, and not visible.

- See screenshot

I checked with WPML support, and they informed me that Paid Listings is not fully compatible with WPML - is there any quickfix for this?

thanks and best regards,
Georges

![20170302_2151_54_wpml_date_syncro](https://cloud.githubusercontent.com/assets/6474997/23553139/c312d1aa-001f-11e7-9e73-5a66ec93134c.jpg)

madeincosmos:
Problem reported in 552734-zd-woothemes

kraftbj:
Can you confirm if you're using WPJM Simple Paid Listings or the WooCommerce Paid Listings add-on?

nz352:
I'm using the "Woocommerce Paid Listings"

thanks
Georges

----- Original Message -----

PR #1027, opened 2017-06-12
Adds WPML configuration

jom:
Partially Fixes: #886


jom:
@vukvukovich Mind checking this out?

PR #1033, opened 2017-06-15
Introduces `wpjm_the_job_title` filter and implements in WP admin and templates

jom:
Now that we're escaping `post_title` in WPJM, I wanted to add this filter to allow for plugins / themes to change the title in certain contexts. Our Applications plugin was affected by this. 

I'm using this filter in most places, including WP Admin. Be sure to check the context before changing the title.

### DIFF
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -162,10 +162,10 @@ class WP_Job_Manager_CPT {
 				$approved_jobs = array_map( 'absint', $approved_jobs );
 				$titles        = array();
 				foreach ( $approved_jobs as $job_id )
-					$titles[] = get_the_title( $job_id );
+					$titles[] = wpjm_get_the_job_title( $job_id );
 				echo '<div class="updated"><p>' . sprintf( __( '%s approved', 'wp-job-manager' ), '&quot;' . implode( '&quot;, &quot;', $titles ) . '&quot;' ) . '</p></div>';
 			} else {
-				echo '<div class="updated"><p>' . sprintf( __( '%s approved', 'wp-job-manager' ), '&quot;' . get_the_title( $approved_jobs ) . '&quot;' ) . '</p></div>';
+				echo '<div class="updated"><p>' . sprintf( __( '%s approved', 'wp-job-manager' ), '&quot;' . wpjm_get_the_job_title( $approved_jobs ) . '&quot;' ) . '</p></div>';
 			}
 		}
 	}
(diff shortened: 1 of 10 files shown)


jom:
@tripflex @adampickering @spencerfinnell (and other community contributors) I'd really appreciate feedback on this.

dbtlr (approved):
Looks good to me 👍

spencerfinnell:
I get why it's done/was initially done this way but my vote would be to prefix any new functions. Otherwise it looks solid.

turtlepod:
### Prefix.
WPJM have prefix issue. A plugin should have single prefix. this plugin have multiple prefixes for functions, class name, constants, hooks, etc. (if code is poetry, this definitely not a pretty poetry, LOL) 
and i really hope Automattic will eventually standardize them and deprecate all properly.

I'd prefer going forward it will use only 2 prefix `WP_Job_Manager_` for classes (for back compat) and `wpjm_` prefix (for easier typing)
So I'll vote for `wpjm_get_title()`/`wpjm_title()` or `wpjm_get_listing_title()`, `wpjm_listing_title()` for the function name.

And please standardize naming "job_listing", "job_listings", "job_manager" etc. 

### Param Note 
```
$titles[] = get_the_job_title( $job_id );
```
The param is $post_id (most of the time), not $post, or at least add not in the docs.

jom:
I agree the uniqueness of the functions aren't great. Until we come up with new consitent way to format our helper functions, I'm tempted to hold off and then do a slow deprecation phase-in/out in an upcoming major release. 

I did a quick code search on Github and didn't see any conflicts...obviously that only searches the repos I have access to. Are we aware of any known conflicts?

jom:
@tripflex I think the overall prefix discussion should be moved to another issue. I agree it needs to be done.

I didn't quite follow your param note. I think sometimes we're following the docs when we use `get_post()` and sometimes we're explicit that it is only `job_listing` post types.

spencerfinnell:
If that's the plan in the future I think it should start with this. Not a fan of introducing functions that we know will have to be deprecated.

turtlepod:
@jom I think you mention wrong user.
Sorry about the param, it's actually fine as it is.

I probably add note that dev can use $post_id instead of $post (WP_Post) object (just better inline docs).
```
/**
 * Gets the job title for the listing.
 *
 * @since 1.26.3
 * @param int|WP_Post|null $post Optional. Post ID or post object. Defaults to global $post.
 * @return string|bool|null
 */
+function get_the_job_title( $post = null ) {
    // ....
}
```

jom:
@turtlepod Sorry abou that... changed my comment.

@spencerfinnell I'm not opposed to doing it now and am happy with the `wpjm_` prefix, if that sounds good with y'all. I'll change it and update the PR.

jom:
@tripflex @turtlepod @dbtlr I've updated the PR with the new naming.

PR #1036, opened 2017-06-15
REST: Add Status Endpoint

### LINKED ISSUE #1032: WPJM REST API: Configuration/Status Endpoint
```
Endpoints

Index:

    wpjm/v1/plugin/configuration
    wpjm/v1/plugin/configuration?keys="foo,bar,baz"

Show:

    wpjm/v1/plugin/configuration/<key:string>

responses:

GET wpjm/v1/plugin/configuration:

{
    "foo": true
}

GET wpjm/v1/plugin/configuration/foo:

true

```

pgk:
closes #1032


tests/php/tests/includes/rest-api/test_class.wp-job-manager-controllers-status.php
```diff
@@ -0,0 +1,116 @@
...
+				'rest_api_admin_user@test.com' );
+			$admin = get_user_by( 'ID', $this->admin_id );
+			$admin->set_role( 'administrator' );
+		}
+
+		$this->default_user_id = get_current_user_id();
+		$this->login_as_admin();
+	}
+
+	function login_as_admin() {
```
jom: Is it worth adding any of these helper functions and/or `setUp` logic to a parent class?

includes/rest-api/class-wp-job-manager-filters-status.php
```diff
@@ -0,0 +1,47 @@
...
+class WP_Job_Manager_Filters_Status extends WPJM_REST_Model_Declaration {
+
+	/**
+	 * Declare our fields
+	 *
+	 * @param  WPJM_REST_Model_Field_Declaration_Collection_Builder $def Def.
+	 * @return array
+	 * @throws WPJM_REST_Exception Exc.
+	 */
+	function declare_fields( $def ) {
```
dbtlr: I'm just going to put this review in once, but it should apply across all the classes here (and everywhere!)

You should always explicitly declare method visibility.Aka, make it explicitly `public` :)

includes/rest-api/class-wp-job-manager-controllers-status.php
```diff
@@ -0,0 +1,131 @@
...
+
+
+	/**
+	 * Setup
+	 */
+	function setup() {
+		$this->add_route( '/' )
+			->handler( 'index', 'index' );
+
+		$this->add_route( '/(?P<key>[a-zA-Z_]+)' )
```
dbtlr: I like that you're doing a more explicit REGEX here. I have a tendency to forget to add the key names into my regex statements and this is far more readable!

includes/rest-api/class-wp-job-manager-data-stores-status.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class WP_Job_Manager_Data_Stores_Status
+ */
+class WP_Job_Manager_Data_Stores_Status extends WPJM_REST_Data_Store_Abstract
```
dbtlr: One thing I'm noticing, why the different prefixes `WP_Job_Manager` vs `WPJM` we should decide on a standard here and stick with it.

includes/rest-api/class-wp-job-manager-filters-status.php
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	/**
+	 * Explode keys
+	 *
+	 * @param  MT_Interfaces_Model $model Model.
+	 * @param  mixed               $keys  The keys.
+	 * @return array
+	 */
+	function explode_keys( $model, $keys ) {
```
dbtlr: This seems like a pretty standard way of doing things ... shouldn't this be in the abstract and then you can just override it if you need something more unique?

includes/rest-api/class-wp-job-manager-data-stores-status.php
```diff
@@ -0,0 +1,69 @@
...
+		return new WPJM_REST_Model_Collection( array( $this->get_entity() ) );
+	}
+
+	/**
+	 * Get a Model Using it's unique identifier
+	 *
+	 * @param  mixed $id The id of the entity.
+	 * @return WPJM_REST_Interfaces_Model
+	 */
+	public function get_entity( $id = null ) {
```
jom: When I ran the tests in PHP 5.2, I got an error that this signature does not match `WPJM_REST_Interfaces_Data_Store`. Looks like pre~5.3.9 was more strict about interface signature matching.

includes/rest-api/class-wp-job-manager-data-stores-status.php
pgk: @dbtlr `WPJM_REST` is the unique prefix of Mixtape. `WP_Job_Manager` is the prefix of all our "own" classes and was there before mixtape. I would prefer to keep those separate, but ok with putting them on a similar prefix.

We could rename the Mixtape one to e.g. `WP_Job_Manager_REST` (if we do this, this will be done on a separate pr)

includes/rest-api/class-wp-job-manager-filters-status.php
pgk: This whole Model class is for explicitly modelling the query params accepted by the get / of the controller, which in our case accepts a `keys` query param with comma-separated status keys to return.

 I am leaning towards leaving it here, but definitely possible to put it in the abstract declaration class (possibly with a more generic name `array_from_comma_separated_string`)

includes/rest-api/class-wp-job-manager-filters-status.php
pgk: :) Sure thing!

includes/rest-api/class-wp-job-manager-data-stores-status.php
pgk: Good catch!! Yay tests pass!!!!

includes/rest-api/class-wp-job-manager-data-stores-status.php
pgk: @jom what do you think about ^^?

jom: I'm +1 for prefixing with `WP_Job_Manager_REST_` in a separate PR. I miss PHP namespaces.

Issue #1043, opened 2017-06-19
Guest ajax file upload issues after 1.26.2 update


tripflex:
With the patch added in #1020, any themes, plugins, or users that have their own `file-field.php` template override files (which is a lot of them, if not the majority of users), **_will have problems with ajax uploads for any guest users (not-logged in users)_**.

The problem stems from relying on the function call to check in the template file, if it should be loaded as an ajax file upload field.

Almost every theme and plugin defaults to ajax upload as it has the best user experience, and because of that, if they have not updated their template overrides (which I don't think any have yet), it will still load as ajax for any guest user.

When the guest user attempts to upload a file it will fail silently in the background, as the jQuery handling only handles ajax errors, not an error sent by WP.

_IMO this is a major issue as chances are it will be very widespread_, i'll submit my PR to fix this in just a minute

spencerfinnell:
What about instead of adding the function check to the template file a filter is added that loops through registered fields and adjusts the `ajax` parameter of the registered `file` fields accordingly?

tripflex:
@spencerfinnell yup did you see my comment in the Slack chan?  That's what I think would probably be the best way to handle it, just need to test and figure out to make sure.

The PR I just submitted fixes the error failing silently in the background, but I do agree, I think we need to come up with a solution for this, outside of the template override file.

jom:
@tripflex did #1044 fix this issue? If not, mind providing more info on what issues remain in WPJM core?

I know there could be issues with custom `file-field.php` templates, but hopefully those get updated.

jom:
Closing for now. Feel free to comment if it is still an issue.

PR #1044, opened 2017-06-19
Guest AJAX upload error alerts and handling

tripflex:
#1043 

* Return standard string error instead of `WP_Error` in `wp_send_json_error`
* Show alert window when `success` is false

This doesn't exactly solve the entire problem (which could be through template files or filter on fields), but at least gives the user some type of feedback instead of failing silently in the background.

Because the entire core WP Job Manager never actually calls `wp_send_json_success` the JS for file uploads doesn't have any handling for when `success` is false, which is what is returned by `wp_send_json_error`

When sending a `WP_Error` response it then responds with an array of errors, even though ours is only a single error, and as such, I remove it from using `WP_Error` to send just a simple string so we can alert the user with the error (without a bunch of code to handle arrays)

### DIFF
--- a/assets/js/ajax-file-upload.js
+++ b/assets/js/ajax-file-upload.js
@@ -41,7 +41,7 @@ jQuery(function($) {
 			fail: function (e, data) {
 				var $file_field     = $( this );
 				var $form           = $file_field.closest( 'form' );
-
+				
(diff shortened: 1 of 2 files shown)


jom (approved):
LGTM

PR #1051, opened 2017-06-21
Add ability to filter by `post_status` in `[jobs]`; persist `job_types`

### LINKED ISSUE #788: Add ability to pass post status through to get_job_listings
Right now, there is no way to use `get_job_listings` to return only expired jobs. We can extend `get_job_listings` to include that as an optional argument and extend the jobs shortcode to be able to pass that as an option.

### LINKED ISSUE #837: The jobs shortcode does not persist job type
When using a shortcode to set the job type, e.g. `[jobs per_page="2" job_types="full-time" show_tags=false show_filters="false"]`, this works for the initial load of jobs.

After clicking to load more, all jobs are loaded, not those matching the `job_types` attr.

jom:
Fixes #837 
Fixes #788 

* `[jobs]` shortcode: Fixes issue with persisting job_types when `show_filters` is 0.
* `[jobs]` shortcode: Adds ability to filter by `post_status`

#### Testing instructions:

* With a `per_page="1"`, verify `post_status` and `job_types` persist (e.g. `[jobs job_types="cool" post_status="publish,expired" per_page="1" show_filters="false"]`)

* Enhancement: Adds the ability to filter `[jobs]` by `post_status` with new shortcode attribute. (#788)
* Fix: Persist `job_types` attribute in `[jobs]` shortcode between pages. (#837)


includes/class-wp-job-manager-ajax.php
```diff
@@ -133,7 +134,8 @@ public function get_listings() {
 			'search_location'    => $search_location,
 			'search_keywords'    => $search_keywords,
 			'search_categories'  => $search_categories,
-			'job_types'          => is_null( $filter_job_types ) || sizeof( $types ) === sizeof( $filter_job_types ) ? '' : $filter_job_types + array( 0 ),
+			'job_types'          => is_null( $filter_job_types )   || sizeof( $types ) === sizeof( $filter_job_types ) ? '' : $filter_job_types + array( 0 ),
```
spencerfinnell: Extra space here.

Issue #1034, opened 2017-06-15
Standardize WPJM function and hook names


jom:
The prefixes used and internal naming of things is inconsistent or nonexistent. In a major release, we should come up with a standard naming scheme, deprecate usage of the old functions/hooks, and phase in the new standard.

spencerfinnell:
Related #1035

jom:
@turtlepod's comment for his vote: https://github.com/Automattic/WP-Job-Manager/pull/1033#issuecomment-308751644

spencerfinnell:
My vote also goes towards `wpjm_` for all functions, actions, and filters and `WP_Job_Manager_` for all classes.

turtlepod:
Same vote here!

tripflex:
If this does end up happening **PLEASE** make sure this isn't done until a major release like @jom mentioned, I know my plugins as well as numerous themes and other plugins specifically call a number of core functions for correct handling, and IIRC the majority of them start with `job_manager_XXX`

Shouldn't be an issue as long as we keep it to a major release version

jom:
@tripflex It'll be a major version and they'll be deprecated slowly.

PR #1063, opened 2017-06-29
Allow users to specify user password when creating first job listing

### LINKED ISSUE #637: Allow new user to set password
When WPJM is set to create an user account automatically when submitting a job listing, we do not request a password from the new user. Since WordPress no longer e-mails out a pregenerated password, this would require all users to go through the Lost Password flow to determine their password.

The ability for the user to set a password initially would make the process more streamlined.

jom:
Fixes #637 

#### Guiding Principle
Using WordPress' standard behavior of emailing link to set password is best practice. 

* Introduces a new setting (`job_manager_use_standard_password_setup_email`) that allows admins to force users to use the standard WordPress flow of getting an email to set their password (default) _or_ be presented with password fields. 
* If the setting to automatically generate usernames is enabled, the `job_manager_use_standard_password_setup_email` setting value will default to `true`. Otherwise, users might not know their username.
* Introduced a few additional filters `wpjm_validate_new_password`, `wpjm_password_rules_hint`, and `wpjm_get_registration_fields` to help support this new functionality. I set the default minimum length to 8. We could put this in a filter or they could just override it in the validation/password rules filters.
* I kept the new naming prefix `wpjm_` for filters/functions, but I propose keeping the `job_manager_` prefix for settings. Way less hassle later on. (#1034)
* Bumped minimum supported WordPress version from 4.1 to 4.3.1 to better support the user creation flow. 4.1 and 4.2 have about 5.1% of all WordPress installs at the time of this PR.

Screencast of the settings: https://cloudup.com/cGa_M-F3THC
Password entry, when enabled: 
screen shot 2017-06-29 at 2 48 23 pm

#### Testing instructions:

* Try the first time job creation flows with all different combinations of user creation settings.

* Added the ability for admins to allow users to specify a user password when creating first job listing.


spencerfinnell:
I think a lot of the logic added to the `account-signin.php` template should be done before and sent to the template. Lots going on there that theme author's don't care about.

Other than I've never been a fan of password verification. It's easy to reset and it's not used anywhere else in WP.

jom:
@spencerfinnell Agreed with first bit. I'll move the field array building somewhere else.

Password verification is easy to remove. I usually have it in when it is obfuscated, but don't have strong opinions about it.

annezazu:
Tested this flow out (both with the setting on and off). Looks great to me!

tripflex:
@jom i've had a few users ask about being able to set password from a password field with my plugin, so i know there are a few out there who want this feature.

We just need to make sure they understand that doing this will completely bypass having the user confirm their email address on registration.

Regarding the password confirmation field, unfortunately I think that is a must ... mainly because a lot of users will end up fat fingering a key or entering it incorrectly and then complaining or getting frustrated because the password doesn't "work" when really they just hit an extra key or didn't enter it correctly.

Besides that, I like the new feature, only thing I could think potential issue wise would be themes that override the template file and need to update it to support this feature

jom:
Thanks, @tripflex. Agreed about email verification. Might be something to add to the help text to the setting.

dbtlr (approved):
The changes look okay. I'm feeling a bit fazed by all the different naming schemes in use, but obviously, that has nothing to do with this PR. Just looking forward to seeing that fixed/updated soon. :)

jom:
@dbtlr I agree. It bugs me a lot.

MrCenty:
Hey there! I've been waiting like crazy for this! When will it be published in the next update or how can I implement it in WP Job Manager on my site? :)

davoraltman:
@MrCenty The new version together with this enhancement should go out by the end of next week. Cheers :)

digitallawyer:
Thank you!

MrCenty:
Works nicely! But you should include a translation string for the textfield placeholder of the username, password & verify password as there will be errors with the wp job manager field editor plugin because there is no placeholder ! (And it's always nice to be able to edit the placeholders)

Maudcr:
Hi, I am also hoping desperately for this. I bought jobify (including wp job manager) few days ago. How could I have access to this update? Thanks :)

jom:
@Maudcr It is in the most recent release.

Maudcr:
Thanks,

It dosen't seem to be there for me. I bought the theme a few days ago.... Would you have any advice?


Maud Comtois-Rouillard
maud.cr@hotmail.com
(587) 437-3453




________________________________

LukeBichard1:
Hi, will this be included in the resume manager submission form?

dipaksaraf:
Is there any possibilities for adding it in Resume Manager Submission Form?

toleabivol:
How can we add this feature to Resume Manager Submission ?

thedigitalcircle:
Yes, I have also the same problem too. How Can I add this feature with Resume Manager Submission.
This is not yet working with RM.

Shuvashish

jom:
Resume Manager is a separate repository and we are tracking this enhancement request there. We don't have an ETA, but I've added your +1s to that issue.

PR #1065, opened 2017-06-29
Removes job posting resume ID and key cookies at logout

### LINKED ISSUE #826: Delete wp-job-manager-submitting-job-id cookie
Upon logging out of the WordPress user (and perhaps while logging in), we should clear the `wp-job-manager-submitting-job-id` cookie. If I'm reading things right, we could load a previous job if, on the same computer, someone logs in as another user and tries to submit a entry.

https://github.com/Automattic/WP-Job-Manager/blob/master/includes/forms/class-wp-job-manager-form-submit-job.php#L92

jom:
Fixes #826

* Always treat these cookies as session cookies.
* Remove them on WordPress logout.

#### Testing instructions:

* Start a job listing but don't go past preview step.
* Logout and go back to job posting page. Verify you aren't resuming the previous listing.


spencerfinnell:
Nice catch.

PR #1067, opened 2017-06-29
Minor string fixes

garrett-eclipse:
Minor spelling and localization fixes


jom:
Nice catch! Looks like we had a mix of North American/UK versions of customize/customise in the readme.

garrett-eclipse:
No worries, the only 'customising' I left was in the URLs as localizing that string would break the link.

PR #1083, opened 2017-07-11
Fix/rename mixtape prefix

pgk:

No logic changes, only renames and include file renames.

* Rename mixtape `WPJM_REST` prefix to `WP_Job_Manager_REST` as discussed in #1036 

#### Testing instructions:

* pull this repo and run `./scripts/build_mixtape.sh`
* No fatals, wpjm rest api works

### DIFF
--- a/.mixtapefile
+++ b/.mixtapefile
@@ -1,3 +1,3 @@
 sha=da21513ec58babe6f8ada1b8ccf21e39003e04b0
-prefix=WPJM_REST
+prefix=WP_Job_Manager_REST
 destination=lib/wpjm_rest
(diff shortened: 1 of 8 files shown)


dbtlr (approved):
LGTM! 

One comment I have is completely unrelated, but in looking at the testing instructions, it seems like we have some manual build instructions. It might be useful to match some of the standards the Jetpack team are using and use NPM commands to make updating, building, and testing require as little knowledge as possible.

I know we're not really using npm right now for anything else, but it is a tool that comes preloaded on most systems, so it is pretty easy to use as a loader.

pgk:
👍

PR #1113, opened 2017-07-25
Release 1.28.0

jom:
This is a PR to track deploy tasks 1.28.0 release.

- [x] Commit replace /1.27.1/1.28.0/

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,18 @@
+= 1.28.0 =
+* Enhancement: Improves support for Google Job Search by adding `JobListing` structured data. (@jom; https://github.com/Automattic/WP-Job-Manager/pull/1115)
+* Enhancement: Adds ability for job types to be mapped to an employment type as defined for Google Job Search. (@jom; https://github.com/Automattic/WP-Job-Manager/pull/1112)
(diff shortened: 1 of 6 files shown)


thedecanus:
New to GitHub - Not sure on the correct way to create a suggestion for a new feature but anyway here goes:

**List Jobs Within Locations**

I'd like to suggest creating a way for locations that are entered to have their own page that lists any jobs within them. 

If a job listings location was entered into the post, then that location would then generate its own page. If the location did not exist beforehand then a new page would be created.

Alongside this feature, their could be a widget that displays the "Top Locations". A user could then select "x" amount of locations to show in a grid/list item format. Each location displayed would have the number of jobs posted within that location and would link through to the relevant locations pages.

**It could look something like this:**

London (120) - Southampton (240) - Portsmouth (30)

Manchester (25) - Liverpool (200) - Birmingham (55)

jom:
Hi, @thedecanus! Welcome to GH :)

If you wouldn't mind posting that enhancement request [as a new issue](https://help.github.com/articles/creating-an-issue/), we can track it there. This thread is just for the 1.28.0 release. Thanks!

PR #1112, opened 2017-07-25
Add employment type meta field to job type terms

jom:
Fixes #1052 

#### Changes proposed in this Pull Request:

* In preparation for #1046, this PR adds Google's preset definitions for employment types. Admins can map job type taxonomy terms to Google's defined employment types.
* Will attempt to map terms that were set up using WPJM's defaults on update.

See: `employmentType` in [Google's Job Posting definition
](https://developers.google.com/search/docs/data-types/job-postings#definitions)

#### Testing instructions:

* Create a new job type set to an employment type. Edit the new employment type and verify the new setting was saved.
* Edit an existing job type and verify the employment type was updated.

#### Proposed changelog entry for your changes:
* Allows job types to be mapped to an employment type as defined for Google Job Search.

screen shot 2017-07-25 at 7 13 21 pm
screen shot 2017-07-25 at 7 14 25 pm

PR #1115, opened 2017-07-26
Improve support for Google Job Search with JobPosting structured data

### LINKED ISSUE #1046: Native support for Google Job Search
This new Google search facet was only introduced today, but I think providing native support would be a great way to broaden the audience for our job openings.

The new markup is explained in detail here - https://support.google.com/webmasters/answer/7388807

jom:
Fixes #1046 

Base is `add/employment-type#1052` (requires #1112 to be merged first; for review, I would look at each commit individually skipping the first few)
 
* Improves support for Google Job Search by putting structured data (JSON-LD) in page footer for single job listings.
* Removes existing microdata from default templates
* Adds `wpjm_get_job_listing_structured_data` filter for overriding auto-generated structured data. Return false to disable the behavior. 
* Minor: Added `wpjm_[get_]the_job_description()` and implemented in templates. 

See: https://developers.google.com/search/docs/data-types/job-postings

#### Testing instructions:

* Google has a [snazzy testing tool](https://search.google.com/structured-data/testing-tool). Paste in published job listing URLs and verify the content is accurate.

**Note for theme devs:** I'm using `wp_footer` to add this to the job listing pages. Structured data can be placed anywhere, but I found a few more examples of folks including it in the footer. I'm flexible and open to changing this and just adding it to the standard job listing content (which would allow theme developers to do the same eventually), but this will at least give a valid `JobListing` object for all users by default. Using JSON-LD will require theme developers to remove the microdata from their overridden templates. Not doing so won't cause huge issues, but may cause Google to think two different job listings are appearing on a page.

* Improves support for Google Job Search by providing a complete `JobListing` record.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -44,6 +44,8 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'auto-draft_to_publish', array( $this, 'set_expiry' ) );
 		add_action( 'expired_to_publish', array( $this, 'set_expiry' ) );
 
+		add_action( 'wp_footer', array( $this, 'output_structured_data' ) );
+
 		add_filter( 'the_job_description', 'wptexturize'        );
 		add_filter( 'the_job_description', 'convert_smilies'    );
 		add_filter( 'the_job_description', 'convert_chars'      );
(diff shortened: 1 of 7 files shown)


spencerfinnell:
Sweeeeeet. Can we add some version numbers to template files? Won't do much now but will hopefully make changes to them more noticeable.

jom:
@spencerfinnell Like inline documentation (e.g. '@since 1.28.0 Microdata removed in favor of structured data placed in page footer`) in the template files or something more advanced?

spencerfinnell:
Was just thinking something like WooCommerce https://github.com/woocommerce/woocommerce/blob/master/templates/archive-product.php -- but keeping a sort of inline changelog with it as well would be a nice added bonus!

jom:
@spencerfinnell Definitely do-able. Looking at the `git blame` it seems like `@version` just matches the last version it was significantly modified?

spencerfinnell:
Yeah it's bumped to the version number it was changed in.

jom:
👍 I'll add them in a separate PR before 1.28.0.

dbtlr (approved):
LGTM!

onegreenspider:
You guys rock! This is exactly what I've been looking for.  Thank you for the new structured data improvements!

PR #1120, opened 2017-07-27
Add noindex to expired/filled job listings

### LINKED ISSUE #1117: Improve support for popular sitemap generators
For now, I propose trying to add support for:
- [x] [Jetpack](https://jetpack.com/)
- [x] [Yoast SEO](https://wordpress.org/plugins/wordpress-seo/)
- [x] [All in One SEO](https://wordpress.org/plugins/all-in-one-seo-pack/)
-  ~~Google XML Sitemaps~~ (No filters!)

In order to support each, we need to be able to:
- Make sure the custom post type (`job_listing`) is added.
- Make sure only published (not `preview` or `expired`) job listings are included.
- Make sure filled job listings are excluded. If we don't, [Google considers it spam](https://developers.google.com/search/docs/data-types/job-postings).

In this issue, I might also sneak in adding a `noindex` meta tag to expired / filled job listing, per Google's [recommendation](https://developers.google.com/search/docs/data-types/job-postings).

jom:
Further Fixes #1117 ;)

* Adds `noindex` meta tag to expired and filled job listings to make Google happy (might be flagged as spam if we don't).
* Adds `wpjm_allow_indexing_job_listing` filter to override this.

#### Testing instructions:

* Make sure expired and filled job postings have the `noindex` value for the `robots` meta field in their single page header.
* Make sure non-expired/filled job postings do _not_ have this tag.
* Also make sure non-job listings don't have this tag.

* Instructs search engines not to index expired and filled job listings.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -44,6 +44,7 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'auto-draft_to_publish', array( $this, 'set_expiry' ) );
 		add_action( 'expired_to_publish', array( $this, 'set_expiry' ) );
 
+		add_action( 'wp_head', array( $this, 'noindex_expired_filled_job_listings' ) );
 		add_action( 'wp_footer', array( $this, 'output_structured_data' ) );
 
 		add_filter( 'the_job_description', 'wptexturize'        );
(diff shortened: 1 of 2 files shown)


spencerfinnell:
Check out https://developer.wordpress.org/reference/functions/wp_no_robots/

jom:
@spencerfinnell I thought about that but wasn't sure about the inclusion of `follow`. Google doesn't put that in their recommended `robots` tag for removing expired/filled job listings. 

If I'm reading W3C correctly, `follow` is by default so it might just be implied. In that case I'm fine with just using `wp_no_robots()`

ethodev:
This is now integrated into WPJobManager. What's the chances Google would see filled posts as spam? I fear I will miss out on traffic from people who would still find my niche site useful as they can view similar jobs or more jobs from the employer who filled theirs, or of course post a resume etc.

jom:
You can override the behavior with this filter in your theme:
`add_filter( 'wpjm_allow_indexing_job_listing', '__return_true' );`

PR #1123, opened 2017-07-28
WPML: Add `lang` to query string for Ajax

### LINKED ISSUE #973: AJAX get_listings/ Returns Location and Date in Wrong Language
I am searching for a Job in English. Primary language is German.
The job is returned in English, but the location is "Überall" (no location is set), it should be "Anywhere".
The date is "Veröffentlich vor 4 Wochen" instead of "Released 4 Weeks ago".

So apparently WP-Job-Manager gets the correct language (Otherwise the Job title and Type wouldn't be in English) but outputs date and location in default language.

I'm using WPML 3.7.0 and WP-Job-Manager 1.25.3 

I guess it's a bug.

image

jom:
Just using `wpml_switch_language` wasn’t changing the site URL and is
just used in overriding `WP_Query`. Seems like if you have the query
var `lang` set WPML will also handle the locale switching for you.

Fixes #973 

#### Testing instructions:

* `[jobs]` shortcode: Make sure the location and date locale match the language for the listings for both the default and a non-default language.

* Improved support with WPML in the `[jobs]` shortcode.

### DIFF
--- a/includes/3rd-party/wpml.php
+++ b/includes/3rd-party/wpml.php
@@ -9,22 +9,11 @@
  * @since 1.26.0
  */
 function wpml_wpjm_init() {
-	add_action( 'get_job_listings_init', 'wpml_wpjm_set_language' );
 	add_filter( 'wpjm_lang', 'wpml_wpjm_get_job_listings_lang' );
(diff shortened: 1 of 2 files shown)


jom:
Reverted in #1129

Issue #1105, opened 2017-07-18
Add action for sending new user notifications


jom:
#### Proposed Added Functionality
In `wp_job_manager_create_account()`,  fire a new action (`wpjm_notify_new_user`) that is, as a default, hooked to `wp_job_manager_notify_new_user`. This will allow users to override this functionality more easily.

See: #1103

PR #1125, opened 2017-07-28
Make notification for new users an action

jom:
Adds new action `wpjm_notify_new_user` which defaults to `wp_job_manager_notify_new_user()`.

Fixes: #1105


jom:
FYI: Sneaking into 1.28.0 @joelworsham

PR #1139, opened 2017-08-09
Register and use original tinymce link and not use wplink.

### LINKED ISSUE #920: Disallow linking to other posts inside the WP editor
When I try to add a link inside the [submit_job_form] page in the Description field, I shouldn't be able to see links to other posts.
#### What Happened Instead
I can see links to all post types made on the site. This happens on all Editor fields including the [submit_resume_form] page (ref in gh-137). This only happens when you are **logged in**.

screen shot 2017-04-05 at 9 30 20 pm

screen shot 2017-04-05 at 9 30 43 pm

cc @annezazuu @bikedorkjon @kraftbj

turtlepod:
Fixes #920 

* Instead of modify the search query in wplink (tinymce), simply register and use "link" tinymce plugin.

#### Testing instructions:

* Check this video: https://drive.google.com/file/d/0B9TdkMXV_Q7GS1R2aHNtUVUydlE/view?usp=sharing

* Use "link" tinymce plugin.


jom:
One hesitation I have about using TinyMCE's core `link` plugin is making sure the bundled version works with both TinyMCE 4.1.3 (version in WPJM's minimum WordPress [4.3.1]) thru 4.6.2 (current in WordPress 4.8). 

If we can verify it works in all changed versions of WordPress since 4.3.1 (see [codex](https://codex.wordpress.org/TinyMCE)), I'd be more open to including it. If so, this PR should also revert the changes made in #922.

jom:
Closing this for now, @turtlepod. Happy to re-open if we can show compatibility with our oldest supported version.

PR #1168, opened 2017-09-08
Add license helper to WPJM core for premium add-ons

jom:
#### To do:
- [x] Add tests

* Moves and updates license/helper code from individual wpjobmanager.com plugins to WPJM core.
* Licenses are now stored in one option. Old license activations should be migrated although WordPress MU users should re-activate their licenses.
* WordPress MU users are now activated network-wide (which makes sense as this is for plugin updates). Before there was a hack-ish work-around to use the active site to test for plugin updates.
* Moves license management to WP Admin > Job Listings > Add-ons > Licenses.
* Requests JSON responses from wpjobmanager.com's License API. 
* Uses new docblock notation of `WPJM-Product` in our plugins to pass along the wpjobmanager.com's product slug (consistently also happens to be the plugin's slug, but I decoupled the two). 

Dev note: I kept the usage of the British `licence` in the code but used `license` in all strings. Now that I'm already migrating the old options to a single option, I'm happy with moving all over to `license` if people think it is necessary.

Internal note: I'm updating all of our plugins using a standard branch name `change/wpjm-core-updater`.

screen shot 2017-09-08 at 2 24 47 pm

### DIFF
--- a/includes/helper/class-wp-job-manager-helper-api.php
+++ b/includes/helper/class-wp-job-manager-helper-api.php
@@ -1,0 +1,173 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * WP_Job_Manager_Helper_API
+ */
+class WP_Job_Manager_Helper_API {
+
+	const API_BASE_URL = 'https://wpjobmanager.com/';
+
+	/**
+	 * The single instance of the class.
+	 *
+	 * @var self
+	 * @since  1.29.0
+	 */
+	private static $_instance = null;
+
+	/**
+	 * Allows for accessing single instance of class. Class should only be constructed once per call.
+	 *
+	 * @since  1.29.0
+	 * @static
+	 * @return self Main instance.
+	 */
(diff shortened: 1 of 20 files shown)


includes/helper/class-wp-job-manager-helper-api.php
```diff
@@ -0,0 +1,136 @@
...
+				}
+				return array(
+					'error_code' => wp_remote_retrieve_response_code( $request ),
+					'error' => 'Error code: ' . wp_remote_retrieve_response_code( $request ),
+				);
+			}
+			return false;
+		}
+
+		$response = json_decode( wp_remote_retrieve_body( $request ), true );
```
dbtlr: It's worthwhile to put a @ before the json_decode(), as a broken response from the server could cause a fatal error otherwise (stupid PHP)

spencerfinnell:
@jom Just a note this page won't be accessible if https://github.com/Automattic/WP-Job-Manager/blob/master/includes/admin/class-wp-job-manager-admin.php#L112 is filtered off.

PR #1182, opened 2017-09-21
REST API: Job Types

pgk:
Adds a job-type endpoint under `wp/v2`.

* Enables `show_in_rest` for the job type taxonomy
* Adds a custom nested field, via `register_rest_field`

#### Testing instructions:

* Update mixtape locally (`./node_modules/.bin/mixtape build` )
* Do CRUD requests on `wp/v2/job-types`

* REST API: Add Job Types Support


jom (approved):
Tested and works for me.

Minor note: Some public methods are missing explicit `public` keywords.

pgk:
Good catch! Done :)

PR #1200, opened 2017-10-10
Localization fix for license

garrett-eclipse:
In translations it's inconsistent with use of licence & license.
I tried to standardize by using en_US in strings and comments but didn't touch any variable/function references to avoid unforseen issues.

Fixes #

* Standardize license verbiage for en_US

### DIFF
--- a/includes/helper/class-wp-job-manager-helper.php
+++ b/includes/helper/class-wp-job-manager-helper.php
@@ -12,7 +12,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  */
 class WP_Job_Manager_Helper {
 	/**
-	 * @var array Messages when updating licences.
+	 * @var array Messages when updating licenses.
 	 */
 	protected $licence_messages = array();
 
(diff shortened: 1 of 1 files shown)


includes/helper/class-wp-job-manager-helper.php
```diff
@@ -468,7 +468,7 @@ private function activate_licence( $product_slug, $licence_key, $email ) {
 	private function deactivate_licence( $product_slug ) {
 		$licence = $this->get_plugin_licence( $product_slug );
 		if ( empty( $licence['licence_key'] ) || empty( $licence['email'] ) ) {
-			$this->add_error( $product_slug, __( 'licence is not active.', 'wp-job-manager' ) );
+			$this->add_error( $product_slug, __( 'license is not active.', 'wp-job-manager' ) );
```
jom: Good catch. Comments and option names use licence (WPJM's original dev was a Brit ;)) but all strings should use US spelling (license).

garrett-eclipse:
Thanks @jom I've reverted the comments so this should just be the translatable strings. Cheers

jom:
Thanks for the PR, @garrett-eclipse! Should be in the 1.29.1 release.

garrett-eclipse:
Thanks @jom

PR #1210, opened 2017-10-20
REST API: Enable/Disable via setting (and filter)

pgk:

* This adds a setting in the admin area that allows users to opt-out of
enabling the api (enabled by default)
* Add a filter to programmatically control this too.

screen shot 2017-10-20 at 11 33 39

N.B. This assumes there is a proper mixtape install in place.

#### Testing instructions:

* Navigate to Job Listings > Settings > REST API
* Test that toggling the checkbox enables/disables the wpjm/v1 api namespace

### DIFF
--- a/includes/rest-api/class-wp-job-manager-models-settings.php
+++ b/includes/rest-api/class-wp-job-manager-models-settings.php
@@ -38,7 +38,16 @@ class WP_Job_Manager_Models_Settings extends WP_Job_Manager_REST_Model_Settings
 			include_once $parent . '/admin/class-wp-job-manager-settings.php';
 		}
 
-		return WP_Job_Manager_Settings::instance()->get_settings();
+		$settings = WP_Job_Manager_Settings::instance()->get_settings();
+		$filtered_settings = array();
+		foreach ( $settings as $key => $value ) {
+			// We don't want the rest api setting block to be tweaked via the rest api.
+			if ( 'rest_api' === $key ) {
+				continue;
+			}
+			$filtered_settings[ $key ] = $value;
+		}
+		return $filtered_settings;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


jom:
I'm a bit torn about adding a setting for this. For a typical user, what would be the purpose of _disabling_ the REST API? For the users who know what this is for and still want to disable it, would they be able to just add a filter to disable it?

Since this will start enabling the REST API by default, I'd like us to do some pen testing for the increased attack surface before merging this.

pgk:
@jom Re: enabling the api setting, just followed what WooCommerce does:

screen shot 2017-10-20 at 12 38 25

But fine with having just a filter for it. I would say that this would be the final pr that would be merged with master, after everything else is there.

wpfan2099:
many thanks - that looks great - thankx for discussing this.

includes/rest-api/class-wp-job-manager-models-settings.php
```diff
@@ -38,7 +38,16 @@ public function get_settings() {
 			include_once $parent . '/admin/class-wp-job-manager-settings.php';
 		}
 
-		return WP_Job_Manager_Settings::instance()->get_settings();
+		$settings = WP_Job_Manager_Settings::instance()->get_settings();
+		$filtered_settings = array();
+		foreach ( $settings as $key => $value ) {
```
jom: Are you thinking this would be more explicit than `unset( $settings['rest_api'] );`?

jom (approved):
Once we're ready to enable-by-default, this LGTM. Made one low-priority comment.

donnapep:
Given that we'll soon be adding Gutenberg blocks to WPJM, and those blocks will need to use the REST API, it doesn't make sense to add a setting for this. Not sure about the filter either.

PR #1222, opened 2017-11-01
Sanitize field input using different strategies

### LINKED ISSUE #1205: Non-Latin URLs not encoded/decoded correctly?
To repro:

1. Get a URL with no-Latin chars in it, like this one:

https://www.colsadna.co.il/%D7%94%D7%95%D7%A1%D7%A4%D7%AA-%D7%A1%D7%93%D7%A0%D7%94/

2. Post a new job.
3. Use that URL as application method.
4. On the job page after you submit, it doesn't work:

screen shot 2017-10-16 at 10 38 25 am

722598-zen

jom:
Fixes #1205 

* Allows text fields to specify their sanitization strategy. Built-in options: `url`, `email`, `url_or_email`.  Callables with a single argument allowed.
* Changes the `application` field to use the correct sanitization strategy. This will allow for octets in URLs.
* Changes the `company_website` field to use the `url` sanitization. 

#### Testing instructions:

* Apply for jobs using standard URLs, URLs with non-Latin characters, emails, bad emails, dangerous strings, etc.

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -283,14 +283,39 @@ abstract class WP_Job_Manager_Form {
 	/**
 	 * Navigates through an array and sanitizes the field.
 	 *
-	 * @param array|string $value The array or string to be sanitized.
-	 * @return array|string $value The sanitized array (or string from the callback).
+	 * @since 1.22.0
+	 * @since 1.29.1 Added the $sanitizer argument
+	 *
+	 * @param array|string    $value      The array or string to be sanitized.
+	 * @param string|callable $sanitizer  The sanitization method to use. Built in: `url`, `email`, `url_or_email`, or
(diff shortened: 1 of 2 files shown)


tripflex:
The only thing I would bring up is in regards to the `preg_match` used to check if the value is a URL.  

First thing would be if users decide to enter the URL like this:
`//google.com/something`

I've also had a ton of instances where users have contacted me because a field they are using doesn't have a valid URL (meaning the user entered `google.com/something` instead of `//google.com/something`) and is the reason I added a setting and handling in my Field Editor to automatically prepend `//` to a URL that is missing it

My first thought was to use `parse_url` to handle this, here's a real quick sandbox to show the differences:
https://glot.io/snippets/ev374vwkbw

jom:
I've changed the URL discovery to `null !== parse_url( $value, PHP_URL_HOST )`. It speaks to a secondary issue with this field: validation. Might be worth another issue.

dbtlr (approved):
Looks good to me

PR #1226, opened 2017-11-08
Don't allow submit_job_board shortcode in admin

seanvfs:
Don't process shortcode in admin

Fixes #

*

#### Testing instructions:

*

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -75,6 +75,8 @@ class WP_Job_Manager_Shortcodes {
 	 * @return string|null
 	 */
 	public function submit_job_form( $atts = array() ) {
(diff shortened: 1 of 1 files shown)


jom:
What is this fixing/preventing? Please provide more information.

seanvfs:
This is fixing an issue where the submit_job_board shortcode executes code in the admin of a page with the shortcode present in the editor.

spencerfinnell:
What is it executing? 

It's standard shortcode implementation so you would be seeing this for everything if something was being executed.

seanvfs:
It was rendering the shortcode output to the admin.  I just tested with a fresh install and couldn't reproduce the problem.  My apologies.  It must have been related to that specific instance.

jom:
@seanvfs Thanks for the PR. If you're able to reproduce and it is caused by something strange that WPJM is doing, feel free to put up another or create an issue.

PR #1238, opened 2017-11-23
Consider featured flag when randomly sorting job listing results

### LINKED ISSUE #971: Sorting options break the featured listings
I expect that adding basic sorting options to jobs shortcode would keep featured listings at the top of the listings page.

#### What Happened Instead
If I add any sorting option to the jobs shortcode (besides default date added), the featured listings are not sticky anymore. e.g. if I sort listings randomly, the featured listings get mixed up in the list.  

Use [jobs orderby=rand] with some listings featured. The featured listings will not be sticky anymore.

jom:
Fixes #971 

* When using `get_job_listings()` and setting the `orderby` argument to `rand_featured`, featured listings are still considered and randomly listed at the top of all other listings. 

#### Testing instructions:

* Easiest to test in the `[jobs]` shortcode by setting the `orderby` argument:
`[jobs orderby=rand_featured]` and for the old, all shuffled behavior, using `[jobs orderby=rand]`.

* Enhancement: When retrieving job listings in `[jobs]`, setting `orderby` to `rand_featured` will still place featured listings at the top. 

cc: @bikedorkjon @aheckler

### DIFF
--- a/tests/php/includes/factories/class-wp-unittest-factory-for-job-listing.php
+++ b/tests/php/includes/factories/class-wp-unittest-factory-for-job-listing.php
@@ -36,6 +36,9 @@ class WP_UnitTest_Factory_For_Job_Listing extends WP_UnitTest_Factory_For_Post {
 			$args['meta_input'] = array();
 		}
 		$args['meta_input'] = $this->generate_args( $args['meta_input'], $this->default_job_listing_meta );
+		if ( ! empty( $args['meta_input']['_featured'] ) ) {
+			$args['menu_order'] = -1;
+		}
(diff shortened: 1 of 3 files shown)


aheckler:
> If we think this is generally not the intended behavior, we could reverse this (`rand` being random and [for example] `rand_featured` to have the new behavior).

This would be my inclination, since up to now, `rand` has not considered the featured flag anyway. Might as well stick with that behavior and only introduce `rand_featured` to get the new "random but with featured still on top" behavior.

jom:
@aheckler I feel comfortable with that. Makes sense to me. I'll change the PR.

PR #1249, opened 2017-12-01
Fix issue with "sanitize_posted_field" method when an array of values is provided.

albionselimaj:
Right now, the sanitizer in `includes/abstracts/abstract-wp-job-manager-form.php` checks if the given value is an array, and if so, does the following:
`array_map( array( $this, 'sanitize_posted_field' ), $value, $sanitizer );`

I think the author tried to pass the `$value` and `$sanitizer` as function parameters to the `sanitize_posted_field` method call, but `array_map` doesn't work this way, at it expects every other argument to be an array, so it throws a PHP Warning: `array_map(): Argument #3 should be an array`

This bug also prevents multilevel fields, like image galleries, and other custom fields, from being saved.

- BUGFIX: `array_map(): Argument #3 should be an array` error.

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -294,10 +294,13 @@ abstract class WP_Job_Manager_Form {
 	protected function sanitize_posted_field( $value, $sanitizer = null ) {
 		// Sanitize value
 		if ( is_array( $value ) ) {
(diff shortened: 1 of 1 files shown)


albionselimaj:
Thanks for the fast reply. I've updated it now for PHP 5.2 compatibility.
So, when could we expect an update that includes this fix? This issue prevents some (custom) fields from saving, specifically those that should be saved as an array.

jom:
I'll try to get it out early next week.

PR #1273, opened 2018-01-04
Test fix: WordPress REST API now returns 401 for unauthorized logged in users

jom:
cc @pgk

Feel free to change the way I fixed this test.

See: https://core.trac.wordpress.org/changeset/42421


pgk:
Looks good to me!

PR #1282, opened 2018-01-10
No longer deactivate/remove expired license records

jom:

* No longer deactivate expired license records.

#### Testing instructions:

* Have a license expire for a paid WPJM add-on plugin. Make sure the license information stays in the WPJM instance after a plugin update check.

### DIFF
--- a/includes/helper/class-wp-job-manager-helper.php
+++ b/includes/helper/class-wp-job-manager-helper.php
@@ -167,9 +167,7 @@ class WP_Job_Manager_Helper {
 			'email'          => $licence['email'],
 		) );
 
-		if ( isset( $response['errors'] ) ) {
-			$this->handle_api_errors( $product_slug, $response['errors'] );
-		}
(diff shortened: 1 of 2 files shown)


jom:
Good catch! I'll set the error with the error key to prevent multiple from showing up. 

The activation and deactivation language is a bit strange. To me, it says "this license is tied to this instance." Especially with a developer license with subscriptions, I'm not sure if we want them to have to reactivate all their licenses if their subscription lapses for a failed payment. Can you think of other language we can use or a way we could make it more clear?

alexsanford:
@jom This fixed the multiple error messages for me 🎉 

The other problem is that when I update the expiry of my license to be in the future (i.e. I "renew" my license), the error message doesn't go away, even after I do a check for updates on my site.

As for the language, I think I was mostly confused as to how to get that error message to go away. Since it wasn't happening on update, I figured maybe it had something to do with activating the plugin. But I think there will be less confusion if the error message disappears when the license is renewed.

PR #1299, opened 2018-01-25
Adds hook after user makes a job edit

### LINKED ISSUE #1298: Allow redirects after a successful job edit
We currently let people redirect after job submission but not after editing a job. Maybe we can do this here-ish:

https://github.com/Automattic/WP-Job-Manager/blob/1.29.2/includes/forms/class-wp-job-manager-form-edit-job.php#L133-L140

Inspired by:

https://wordpress.org/support/topic/redirect-after-edit-job/

jom:
Fixes #1298 

* Moves edit job form submission handling outside of `WP_Job_Manager_Form_Edit_Job::output()` so redirections can happen.
* Adds a new action `job_manager_user_edit_job_listing` that fires after a user saves a job listing from the dashboard.

Example code for handling the redirect:
```php
add_action( 'job_manager_user_edit_job_listing', function() {
	wp_redirect( 'https://example.com/good-job' );
	exit();
} );
```

### DIFF
--- a/includes/forms/class-wp-job-manager-form-edit-job.php
+++ b/includes/forms/class-wp-job-manager-form-edit-job.php
@@ -178,6 +178,17 @@ class WP_Job_Manager_Form_Edit_Job extends WP_Job_Manager_Form_Submit_Job {
 			}
 
 			/**
+			 * Fire action after the user edits a job listing.
+			 *
+			 * @since 1.30.0
+			 *
+			 * @param int    $job_id        Job ID.
+			 * @param string $save_message  Save message to filter.
+			 * @param array  $values        Submitted values for job listing.
+			 */
+			do_action( 'job_manager_user_edit_job_listing', $this->job_id, $save_message, $values );
+
+			/**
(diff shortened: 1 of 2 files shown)


westaussie:
Hey @jom ... thanks for taking a look at this :-)

Wondering if by default the redirection should be set to go to the same place as the redirect after save (ie Job Dashboard), and be able to be overridden by a filter as needed? In my case I'm overriding the redirect location depending on logged in Role, so Admins see all Jobs and Employers/Candidates see their own respective Jobs.

Cheers :-)

jom:
@westaussie For now, I haven't changed the default redirect. You should be able to override using the example code provided above. Let me know if I'm not understanding correctly.

rfzamith:
Hi guys, quick question, how can I use the code below for creating a new vacancy, and not just for editing an existing vacancy?

jom:
@rfzamith This filter was just for redirection after editing a post. I'd hook into  `job_manager_job_submitted` and have its priority be more than `10` (so email notifications are sent).

PR #1306, opened 2018-02-05
Use REST API to retrieve available add-ons

### LINKED ISSUE #1288: Update add-ons page to use REST API
Right now the add-ons page is currently being pulled from https://wpjobmanager.com/add-ons and is dependent on consistent structuring of the products on that page. We should change that over to a REST API call to wpjobmanager.com to decouple that from the design/structure of that page.

jom:
Fixes #1288 

* Move to use new wpjobmanager.com REST API to pull available add-ons. Uses Woo's implementation.

#### Testing instructions:

* Visit the Add-ons page in Admin.

screen shot 2018-02-07 at 10 50 11 am


includes/admin/class-wp-job-manager-addons.php
```diff
@@ -38,47 +39,75 @@ public static function instance() {
...
-
-			if ( ! is_wp_error( $raw_addons ) ) {
-
-				$raw_addons = wp_remote_retrieve_body( $raw_addons );
-
-				// Get Products
-				$dom = new DOMDocument();
-				libxml_use_internal_errors(true);
-				$dom->loadHTML( $raw_addons );
+	private function get_extensions( $category = null, $term = null ) {
```
alexsanford: I'm noticing that we're using the terms "add-ons" and "extensions" in the code, more-or-less interchangeably. Is there a distinction? Should we pick one term and stick with it?

includes/admin/class-wp-job-manager-addons.php
```diff
@@ -89,9 +118,12 @@ public function output() {
 			if ( isset( $_GET['section'] ) && 'helper' === $_GET['section'] ) {
 				do_action( 'job_manager_helper_output' );
 			} else {
-				echo '<h1 class="screen-reader-text">' . __( 'WP Job Manager Add-ons', 'wp-job-manager' ) . '</h1>';
-				echo '<div id="job-manager-addons-banner" class="notice updated below-h2"><strong>' . __( 'Do you need multiple add-ons?', 'wp-job-manager' ) . '</strong> <a href="https://wpjobmanager.com/add-ons/bundle/" class="button">' . __( 'Check out the core add-on bundle &rarr;', 'wp-job-manager' ) . '</a></div>';
-				echo $addons;
+				$term = isset( $_GET['term'] ) ? sanitize_text_field( $_GET['term'] ) : null;
```
alexsanford: We're adding in the ability to filter on `term` here, but we don't expose this functionality in the UI. Is this something we're planning to add later? Right now it just seems like unused code.

jom: 👍 The end-point handles it and was going to add a filter, but might just take it out completely for now.

includes/admin/class-wp-job-manager-addons.php
jom: I kept going back and forth on this one. I generally prefer extensions over add-ons, but not 100% sure themes fit with that. I guess add-ons is more generic. I'll just go back to pure add-ons.

Issue #1263, opened 2017-12-22
Add usage tracking


donnapep:
When a user opts in to usage tracking, or a user has already opted in to usage tracking and a certain amount of time has passed, I want WP Job Manager and system info to be logged per site, so that I can make better decisions about what features to focus on improving, and to enable targeted messaging.

# Tracked Usage Data

## WP Job Manager Info

_Please note that site URL and admin email address should always be sent with each event type (`wpjm_stats_log` and `wpjm_system_log`)._

The following data should be logged under the `wpjm_stats_log` event.

### Employers
- [x] Total number of users with the _Employer_ role

### Jobs
- [x] Total number of expired jobs
- [x] Total number of preview jobs
- [x] Total number of pending jobs
- [x] Total number of pending payment jobs
- [x] Total number of published jobs

Total number of published or expired jobs that have:
  - [x] A company logo
  - [x] One or more job types selected

#### Job Data
Total number of published or expired jobs that have a non-empty (i.e. not whitespace or `null` etc.) value for:
  - [x] Location
  - [x] Application Email or URL
  - [x] Company Name
  - [x] Company Website
  - [x] Company Tagline
  - [x] Company Twitter
  - [x] Company Video
  - [x] Listing Expiry Date

Total number of published or expired jobs that have the following settings checked:
  - [x] Position Filled
  - [x] Featured Listing

Total number of published or expired jobs that:
- [x] Have been posted by a guest user

### Job Types
- [x] Total number of job types
- [x] Total number of job types that have a non-empty description
- [x] Total number of job types that have _Employment Type_ set

Total number of published or expired jobs of each type (if they exist):
  - [x] Freelance
  - [x] Full Time
  - [x] Internship
  - [x] Part Time
  - [x] Temporary

### Job Categories
- [x] Total number of job categories
- [x] Total number of job categories that have a non-empty description

You'll need to check the _Categories_ box in _Settings_ > _Job Listings_ in order to see these.

## System Info
System info should be logged under the `wpjm_system_log` event.
- [x] List of activated plugins and their versions, including WPJM
- [x] Active theme and version
- [x] WordPress version
- [x] PHP version 
- [x] Locale / language
- [x] Multisite

# How is Data Collected?
Tracking data is sent automatically once a user opts in, and then once every 2 weeks for as long as the option is enabled.

## Opt-in Setting
A new checkbox setting (default unchecked) is added to the bottom of the _General_ settings page:
- Label: Enable usage tracking
[...]

PR #1312, opened 2018-02-08
Add Usage Tracking

alexsanford:
See https://github.com/Automattic/WP-Job-Manager/issues/1263

Adds Usage Tracking to WP Job Manager. The Usage Tracking functionality is copied from Sensei, and tweaked to support WPJM. Please see the individual commits for better visibility into the specific changes.

## Testing

- Ensure that the opt-in dialog works. To reset, run the following:

```php
delete_option( 'wpjm_usage_tracking_opt_in_hide' );
delete_option( 'job_manager_usage_tracking_enabled' );
```

- Ensure that the setting in the General tab works.

- Ensure that the cron job is scheduled (`wpjm_usage_tracking_send_usage_data` - the Crontrol plugin is a great way to check this).

- Ensure that the data is sent by manually running the cron job (the Crontrol plugin can also be used for this) and monitoring the HTTP request that is sent.

- Ensure, in the HTTP request, the `jobs` URL parameter value represents the number of job listings on your site.

- If usage tracking is not enabled, the cron job may still be scheduled, but the data should not be sent when the job runs.

- Ensure that Usage Tracking can only be managed by admins.

- Ensure the cron job is unscheduled when the plugin is deactivated.

- Ensure that the links in the opt-in dialog and Settings page go to the right place.

- ~[WIP] Ensure that the activation wizard displays the opt-in, and it works properly.~ This will need some discussion, so I'm going to put it in a separate PR.

### DIFF
--- a/includes/class-wp-job-manager-usage-tracking.php
+++ b/includes/class-wp-job-manager-usage-tracking.php
@@ -11,7 +11,16 @@ include dirname( __FILE__ ) . '/../lib/usage-tracking/class-usage-tracking-base.
  **/
 class WP_Job_Manager_Usage_Tracking extends WP_Job_Manager_Usage_Tracking_Base {
 
+	const WPJM_SETTING_NAME = 'job_manager_usage_tracking_enabled';
+
 	const WPJM_TRACKING_INFO_URL = 'https://wpjobmanager.com/document/what-data-does-wpjm-track';
+
+	protected function __construct() {
+		parent::__construct();
+
+		// Add filter for settings.
+		add_filter( 'job_manager_settings', array( $this, 'add_setting_field' ) );
+	}
 
 	/*
 	 * Implementation for abstract functions.
@@ -26,11 +35,11 @@ class WP_Job_Manager_Usage_Tracking extends WP_Job_Manager_Usage_Tracking_Base {
 	}
 
 	protected function get_tracking_enabled() {
-		// TODO
+		return get_option( self::WPJM_SETTING_NAME  ) || false;
 	}
 
 	protected function set_tracking_enabled( $enable ) {
-		// TODO
+		update_option( self::WPJM_SETTING_NAME, $enable );
 	}
 
 	protected function current_user_can_manage_tracking() {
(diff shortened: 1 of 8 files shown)


includes/class-wp-job-manager-usage-tracking.php
```diff
@@ -0,0 +1,46 @@
...
+		// TODO
+	}
+
+	protected function current_user_can_manage_tracking() {
+		return current_user_can( 'manage_options' );
+	}
+
+	protected function opt_in_dialog_text() {
+		return sprintf( __( "We'd love if you helped us make WP Job Manager better by allowing us to collect
+			<a href=\"%s\" target=\"_blank\">usage tracking data</a>.
```
donnapep: This is strange, but it's not opening in a new tab for me. Does it work for you?

alexsanford: Hmm... something is stripping off the `target` attribute. I'm seeing the same behaviour in Sensei as well. I'll look into it. Maybe it's something WP does with all notices?

donnapep: It was introduced in the latest round of changes because it works in 1.9.20.

donnapep:
When I deactivated the plugin it did not remove the cron job.

includes/class-wp-job-manager-usage-tracking.php
alexsanford: Bahaha! I tried to blame WordPress, but actually it was me 😆

Pushed a fix.

alexsanford:
> When I deactivated the plugin it did not remove the cron job.

Two problems:

1. I didn't write that code. That's fixed now 😄 

2. I added the code to `uninstall.php` (which is where the other scheduled hooks are unscheduled), but it appears that nothing actually runs `uninstall.php` on plugin deactivation. @jom do you have any insight here?

donnapep:
> I added the code to uninstall.php (which is where the other scheduled hooks are unscheduled), but it appears that nothing actually runs uninstall.php on plugin deactivation. @jom do you have any insight here?

I don't think we can put that code there since it will only be run when the user deletes the plugin, but not if they just deactivate it.

jom:
@donnapep is right, that is only executed when uninstalled ([#](https://developer.wordpress.org/plugins/the-basics/uninstall-methods/#method-2-uninstall-php)).

I wonder if we should move those schedule hook removals to `deactivate_{$plugin}` hook. I think usually the `uninstall.php` cleans up options, etc.

alexsanford:
That makes sense. I got mixed up between "deactivating" and "uninstalling" 🙂 

I've set up the deactivation hook, and it looks like it's working now!

alexsanford:
Thanks all!

> Only thing I might add is the new options to uninstall.php. We should probably re-confirm if they uninstall and reinstall.

Very good point. I'll go ahead and merge this as-is so it's not blocking further functionality, but I'll open a new PR with the changes to `uninstall.php`.

PR #1324, opened 2018-02-13
Add usage tracking opt-in to setup wizard

alexsanford:
In the setup wizard, display the opt-in dialog inline as a checkbox.

## Testing

- Disable usage tracking and unhide the opt-in dialog - `delete_option( 'wpjm_usage_tracking_opt_in_hide' );`

- Visit `/wp-admin/index.php?page=job-manager-setup`.

- You should not see the opt-in dialog. Instead, the opt-in text and checkbox should be inline in the wizard.

- When the `Start setup` button is pressed, usage tracking should be enabled or disabled, depending on the value of the checkbox. You can check this on the settings page.

- Regardless of the checkbox value, the opt-in dialog should be hidden afterward.

- If the user opt to skip the wizard, usage tracking should not be enabled and the opt-in dialog should not be hidden.

- If usage tracking has already been enabled, the opt-in text and checkbox should not be displayed in the wizard.

### DIFF
--- a/includes/admin/class-wp-job-manager-setup.php
+++ b/includes/admin/class-wp-job-manager-setup.php
@@ -145,18 +145,7 @@ class WP_Job_Manager_Setup {
 	 * Usage tracking opt in text for setup page.
 	 */
 	private function opt_in_text() {
-		return sprintf(
-
-			/*
-			 * translators: the href tag contains the URL for the page
-			 * telling users what data WPJM tracks.
-			 */
-			__(
-				'Check the box below to allow us to collect <a href="%s" target="_blank">usage tracking data</a>.  No sensitive information is collected.',
-				'wp-job-manager'
-			),
-			WP_Job_Manager_Usage_Tracking::WPJM_TRACKING_INFO_URL
-		);
+		return WP_Job_Manager_Usage_Tracking::get_instance()->opt_in_checkbox_text();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


jom:
I'm a bit concerned about the interruption to the flow of the setup wizard. With the tabs at the top I can tell I'm in a wizard but I'm not sure it is instantly obvious that I need to choose an option for tracking usage data in order to proceed. That UI definitely makes sense in the notice dialog, but I wonder if a checkbox would make more sense here.

alexsanford:
I like the checkbox idea. I feel like the implementation might be a bit less messy too. We could just submit it through a form and wouldn't have to worry about getting the ajax-y stuff to work.

@donnapep What do you think? Do you think a checkbox (checked by default) would work just as well, in terms of getting the opt-ins that we want?

spencerfinnell:
I agree with @jom -- they don't feel like "proceed" buttons. 

If buttons are kept they could just be no the "Done" step. But I don't think you'll get very many opt-ins. A checkbox on a normal step makes the most sense to me.

donnapep:
> What do you think? Do you think a checkbox (checked by default) would work just as well, in terms of getting the opt-ins that we want?

The only problem is that in order to be GDPR-compliant, we can't check that by default. They have to explicitly opt in. I'm thinking that's probably why I prototyped it with buttons, because we likely wouldn't get very many people checking the box before clicking the button. But when there are only 2 buttons, they are forced to make a decision. But if 2 buttons confuses the NUX, then that's also a problem. Where's a designer when you need one? 😆

spencerfinnell:
Relevant WooCommerce discussion on GDPR: https://github.com/woocommerce/woocommerce/issues/17837#issuecomment-346020774

donnapep:
For this one, since we don't have a lot of time, we might as well just go with the checkbox (default unchecked) for now until we're able to figure something better out.

alexsanford:
Changed to checkbox, and updated testing instructions.

donnapep:
Something a bit strange. I've deleted the `wpjm_usage_tracking_opt_in_hide` option. While I don't see the opt-in dialog in the wizard 🙌 , I no longer see it outside of the wizard either 🙁

alexsanford:
@donnapep do you have usage tracking enabled? You can check in your settings. If it is already enabled, it will not show the opt-in dialog, or the wizard checkbox.

donnapep:
Oh, whoops. 🤦‍♀️  I forgot to ensure the usage tracking setting was disabled. As you were...

donnapep:
Could we make the checkbox be identical to the one on the settings page, except perhaps leaving off the _Enable usage tracking_ label? That would save some translations and gives some consistency between pages.

screen shot 2018-02-16 at 9 40 22 am

donnapep:
> 99% of the time this will result in a bunch of 0 counts and default taxonomies, but the environment data might be valuable still (PHP version, WordPress version, etc)

Given that we won't be logging the settings or the system info in this first iteration, can we defer this? It could be valuable once those are in but as you've pointed out, would result in a bunch of 0 counts for our initial usage tracking data.

jom:
@donnapep Didn't realize that. Yup, that works.

PR #1331, opened 2018-02-14
Add jobs usage tracking

donnapep:

This PR logs the following usage tracking data:
- Total number of expired jobs
- Total number of preview jobs
- Total number of pending jobs
- Total number of pending payment jobs
- Total number of published jobs

Total number of published or expired jobs that have:
  - A company logo
  - One or more job types selected

#### Testing instructions:

1. Check that the tests run.
2. Set job listings with varying statuses including _Expired_, _Pending Approval_, _Pending Payment_, '_Preview_ and _Active_ (i.e. publish). (Note that in order to set a job to _Pending Payment_ status, the WooCommerce Paid Listings plugin must be installed.)
3. Assign a company logo and select a job type for some of the jobs.
4. For bonus points, in _Job Listings_ > _Settings_ > _Job Listings_, check the _Multi-select Listing Types_ setting, and ensure that some jobs have multiple job types selected, but that they are only included once in the count.

Check that the following properties are logged to Tracks correctly:
- `jobs_type`
- `jobs_logo`
- `jobs_status_expired`
- `jobs_status_pending`
- `jobs_status_pending_payment`
- `jobs_status_preview`
- `jobs_status_publish`

### DIFF
--- a/includes/class-wp-job-manager-usage-tracking-data.php
+++ b/includes/class-wp-job-manager-usage-tracking-data.php
@@ -27,14 +27,14 @@ class WP_Job_Manager_Usage_Tracking_Data {
 		$count_posts = wp_count_posts( 'job_listing' );
 
 		return array(
-			'employers'            => self::get_employer_count(),
-			'jobs_type'            => self::get_job_type_count(),
-			'jobs_logo'            => self::get_company_logo_count(),
-			'jobs_expired'         => isset( $count_posts->expired ) ? $count_posts->expired : 0,
-			'jobs_pending'         => $count_posts->pending,
-			'jobs_pending_payment' => isset( $count_posts->pending_payment ) ? $count_posts->pending_payment : 0,
-			'jobs_preview'         => isset( $count_posts->preview ) ? $count_posts->preview : 0,
-			'jobs_publish'         => $count_posts->publish,
+			'employers'                   => self::get_employer_count(),
+			'jobs_type'                   => self::get_job_type_count(),
+			'jobs_logo'                   => self::get_company_logo_count(),
+			'jobs_status_expired'         => isset( $count_posts->expired ) ? $count_posts->expired : 0,
+			'jobs_status_pending'         => $count_posts->pending,
+			'jobs_status_pending_payment' => isset( $count_posts->pending_payment ) ? $count_posts->pending_payment : 0,
+			'jobs_status_preview'         => isset( $count_posts->preview ) ? $count_posts->preview : 0,
+			'jobs_status_publish'         => $count_posts->publish,
 		);
 	}
 
--- a/includes/class-wp-job-manager-usage-tracking-data.php
+++ b/includes/class-wp-job-manager-usage-tracking-data.php
@@ -9,12 +9,12 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-/**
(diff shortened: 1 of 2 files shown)


includes/class-wp-job-manager-usage-tracking-data.php
```diff
@@ -1,17 +1,20 @@
...
 	exit;
 }
 
-/**
- * Supplies the usage tracking data for logging.
- *
- * @package Usage Tracking
- * @since 1.30.0
- */
+	/**
```
alexsanford: Is there a reason to have this indented?

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
```diff
@@ -44,4 +106,118 @@ public function test_employers() {
...
+		// Create some media attachments.
+		$media = $this->factory->attachment->create_many(
+			6, array(
+				'post_type'   => 'job_listing',
+				'post_status' => 'publish',
+			)
+		);
+
+		// Add logos to some listings with varying statuses.
+		add_post_meta( $this->listings[0], '_thumbnail_id', $media[0] );
```
alexsanford: It's not very clear here which jobs the attachments are being added to. It also seems brittle. Will this break if we change how many jobs of various statuses are being generated?

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
```diff
@@ -44,4 +106,118 @@ public function test_employers() {
...
+	 *
+	 * @since 1.30.0
+	 * @covers WP_Job_Manager_Usage_Tracking_Data::get_usage_data
+	 * @covers WP_Job_Manager_Usage_Tracking_Data::get_job_type_count
+	 */
+	public function test_get_job_type_count() {
+		$terms = $this->factory->term->create_many( 6, array( 'taxonomy' => 'job_listing_type' ) );
+
+		// Assign job types to some jobs.
+		wp_set_object_terms( $this->listings[0], $terms[0], 'job_listing_type', false );
```
alexsanford: Similar to above, it's not very clear here which jobs the types are being added to, and I think it will be brittle.

includes/class-wp-job-manager-usage-tracking-data.php
donnapep: Hmm...the linter was complaining about this earlier but seems to be fine when I outdent it again as per https://github.com/Automattic/WP-Job-Manager/pull/1331/commits/dfb0fec392ff30223cdd1467b11118c0378daeb3. 🤷‍♀️

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
donnapep: > It's not very clear here which jobs the attachments are being added to.

Fair point. I got lazy. 😛  Fixed in https://github.com/Automattic/WP-Job-Manager/pull/1331/commits/23b6b0407565f7d84671a7f61d7d17c7551e218c.

> Will this break if we change how many jobs of various statuses are being generated?

Yes, it's all isolated in the `create_job_listings` function though. This seems acceptable to me as the same thing would happen if you defined it inside the test case itself and then decided to change the number of jobs to create.

PR #1332, opened 2018-02-14
Remove separate text domain for usage tracking

alexsanford:
As per pxLjZ-4tU-p2 it will be simpler to use the text domain for the plugin rather than having a separate text domain for usage tracking.

## Testing

- Ensure tests pass.
- Ensure strings display correctly in the opt-in dialog.


jom:
I’ll check this in the morning, but I wonder if `makepot` can pick up on the dynamic domain.

PR #1335, opened 2018-02-15
Usage tracking for Jobs data

alexsanford:
Adds usage tracking to count jobs with non-empty:

- Location
- Application email or URL
- Company name
- Company site
- Company tagline
- Company Twitter handle
- Company video
- Job expiry date
- Filled position checkbox
- Featured listing checkbox

@donnapep I took a different approach to testing than you did in #1331. Instead of a helper that gets called in `setUp`, I have one that is called in individual tests with some params. ~There's going to be some work in getting the two PR's to work together, so I'll rebase after #1331 is merged.~ Rebased this, and refactored a bit.

## Testing

- Ensure the tests pass

- Create jobs with and without values for the above attributes. Ensure that when usage data is sent, the counts are correct.

### DIFF
--- a/tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
+++ b/tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
@@ -232,7 +232,7 @@ class WP_Test_WP_Job_Manager_Usage_Tracking_Data extends WPJM_BaseTest {
 		$published = 3;
 		$expired   = 2;
 
-		$this->create_job_listings( '_job_location', 'Toronto', $published, $expired );
+		$this->create_job_listings_with_meta( '_job_location', 'Toronto', $published, $expired );
 
 		$data = WP_Job_Manager_Usage_Tracking_Data::get_usage_data();
 		$this->assertEquals( $published + $expired, $data['jobs_location'] );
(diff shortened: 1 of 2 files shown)


donnapep:
These seem to be pulling back jobs with any status. Ideally, they should be limited to published or expired jobs. Otherwise we're going to have a mix of stats, some filtered by status and some not.

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
```diff
@@ -418,27 +331,69 @@ public function test_jobs_company_twitter() {
...
+			'meta_input' => array(
+				$meta_name => $meta_value,
+			),
+		) );
+
+		// Create expired listings.
+		$this->factory->job_listing->create_many( $expired, array(
+			'meta_input' => array(
+				$meta_name     => $meta_value,
+				'_job_expires' => "2018-01-01",
```
donnapep: I'm not sure this will actually set the status to expired. The post status may need to be explicitly set to Expired?

alexsanford: That's right. This is changed in a later commit 🙂

Issue #825, opened 2016-12-06
Update arguments in calls to get_terms()


aheckler:
We have a handful of calls to `get_terms()` using the deprecated style:

```php
get_terms( 'some_taxonomy' , $args );
```

We should update them to the new way:

```php
$args = array (
    'taxonomy' => 'some_taxonomy',
    'more' => 'args',
    ...
);

get_terms( $args );
```

https://developer.wordpress.org/reference/functions/get_terms/

jom:
Bumped this for now until it starts throwing actual deprecation warnings in WP core _or_ until we bump WPJM's minimum core requirement to 4.5. Right now core, although the naming shows deprecated, just treats it as an alternative calling for `wp_term()`. I imagine they will legitimately deprecate it one day and call `_deprecated_argument()`.

Right now it appears in two files. One is a template file (`templates/job-filters.php`) and the other is for the CPT filter (includes/admin/class-wp-job-manager-cpt.php). I started doing a `$wp_version` check in the CPT file, but doing it in the template file is a bit clunky. 

If we want to add support for this _before_ bumping minimum WP version to 4.5, I suggest creating another global wrapper function and do the WP version check in there.

spencerfinnell:
Mentioned in the PR but I'll vote to bumping minimum version vs adding extra code. Being on WP.org I think the minimum requirements are stated clear enough and 4.5+ seems reasonable.

jom:
Our bump to a minimum of WordPress 4.7 in the 1.31.0 release will allow us to incorporate this change now.

PR #1340, opened 2018-02-16
Add job categories usage tracking

donnapep:

This PR logs the following usage tracking data:
- Total number of job categories (`job_categories`)
- Total number of job categories that have a non-empty description (`job_categories_desc`)

#### Testing instructions:

1. Check that the tests run.
2. Ensure that the _Categories_ setting in _Settings_ > _Job Listings_ is checked.
3. Add some job categories and give a few of them a description.
4. Check that the properties are logged to Tracks correctly.

### DIFF
--- a/includes/class-wp-job-manager-usage-tracking-data.php
+++ b/includes/class-wp-job-manager-usage-tracking-data.php
@@ -29,6 +29,7 @@ class WP_Job_Manager_Usage_Tracking_Data {
 		return array(
 			'employers'                   => self::get_employer_count(),
 			'job_categories'              => wp_count_terms( 'job_listing_category', array( 'hide_empty' => false ) ),
+			'job_categories_desc'         => self::get_job_category_has_description_count(),
 			'jobs_type'                   => self::get_job_type_count(),
 			'jobs_logo'                   => self::get_company_logo_count(),
 			'jobs_status_expired'         => isset( $count_posts->expired ) ? $count_posts->expired : 0,
@@ -53,6 +54,33 @@ class WP_Job_Manager_Usage_Tracking_Data {
 		);
 
 		return $employer_query->total_users;
+	}
+
+	/**
+	 * Get the number of job categories that have a description.
+	 *
+	 * @since 1.30.0
+	 *
+	 * @return int Number of job categories with a description.
+	 **/
+	private static function get_job_category_has_description_count() {
+		$count = 0;
+		$terms = get_terms(
+			array(
+				'taxonomy'   => 'job_listing_category',
+				'hide_empty' => false,
+			)
+		);
+
+		foreach ( $terms as $term ) {
+			$description = isset( $term->description ) ? trim( $term->description ) : '';
+
+			if ( ! empty( $description ) ) {
+				$count++;
+			}
+		}
+
+		return $count;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/class-wp-job-manager-usage-tracking-data.php
```diff
@@ -54,6 +56,33 @@ private static function get_employer_count() {
...
+	/**
+	 * Get the number of job categories that have a description.
+	 *
+	 * @since 1.30.0
+	 *
+	 * @return int Number of job categories with a description.
+	 **/
+	private static function get_job_category_has_description_count() {
+		$count = 0;
+		$terms = get_terms(
```
jom: TIL, while investigating #825, that the new signature for `get_terms()` (in the way you used it here) is annoyingly not backward compatible for WordPress instances before version 4.5. I think it will fatal ([#](https://github.com/WordPress/WordPress/blob/4.4.14/wp-includes/taxonomy.php#L1070)).

I punted on #825 for now. It is only half-way deprecated (doesn't throw notice) in WP core to use `get_terms( $taxonomy, $args );`, but still feels weird to use deprecated behavior on new code. We could do a version check here or we could create a global wrapper function (that does the version check).

spencerfinnell: Bump minimum version? :D 4.1 is from 2014. WooCommerce minimum is 4.7.

jom: @spencerfinnell I think we should, but it might be nice to have some usage data for how many folks are pre-4.5 :).

spencerfinnell: I think it is. But waiting a bit after some data makes sense. I would imagine it matches WC pretty well.

donnapep: @jom Do you mean do a WP version check and based on that, determine which version of `get_terms` to call? Or something else?

jom: Yes, a WP version check.

donnapep: But if I'm going to be calling an older version of `get_terms` anyway based on the version check, why not just forego a version check and call it using the older syntax regardless of version?

jom: That's fine because it doesn't throw any deprecation notice right now. It is also what we're doing elsewhere in the plugin (from #825).

spencerfinnell: Its just a gamble if core with truly deprecate the first string argument or not. Unlikely they will but it does introduce the possibility of needing future updates.

donnapep: Fixed in https://github.com/Automattic/WP-Job-Manager/pull/1340/commits/1cc5ba30298a4ab7ebcd9d2909cee5f91d26d61f.

donnapep:
> When Job Categories is not not enabled in Settings > Job Listings, we get a WP_Error in the job_categories field of the usage data.

Fixed in https://github.com/Automattic/WP-Job-Manager/pull/1340/commits/b22ddba72882154edb34fc7f11af60126bd91092.

> But evidently that setting is off in my test environment, and so the tests do not pass.

I wasn't able to reproduce this, but it shouldn't matter if you have the setting checked on your site or not. Tests run in a clean environment 😕

alexsanford:
Tests are still failing on my local installation, but since they're passing for others, and passing on CI, I'm going to assume it's something wrong on my end 🙂

PR #1348, opened 2018-02-17
Remove test dependencies from usage data tests

alexsanford:
Specifically, ensure that job listing categories and types are enabled before running the usage data tests.

## Testing

- Ensure the full test suite passes (`phpunit`)

- Ensure that the usage data tests pass when run independently (`phpunit tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php`)

### DIFF
--- a/tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
+++ b/tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
@@ -54,6 +54,20 @@ class WP_Test_WP_Job_Manager_Usage_Tracking_Data extends WPJM_BaseTest {
 	private $publish;
 
 	/**
+	 * Set up for tests.
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		// Ensure job categories and types are enabled.
+		update_option( 'job_manager_enable_categories', 1 );
+		update_option( 'job_manager_enable_types', 1 );
+		unregister_post_type( 'job_listing' );
+		$post_type_instance = WP_Job_Manager_Post_Types::instance();
+		$post_type_instance->register_post_types();
+	}
+
+	/**
 	 * Create a number of job listings with different statuses.
 	 */
 	private function create_default_job_listings() {


tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
```diff
@@ -53,6 +53,20 @@ class WP_Test_WP_Job_Manager_Usage_Tracking_Data extends WPJM_BaseTest {
...
 
+	/**
+	 * Set up for tests.
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		// Ensure job categories and types are enabled.
+		update_option( 'job_manager_enable_categories', 1 );
+		update_option( 'job_manager_enable_types', 1 );
```
donnapep: I'd rather enable these options explicitly in the individual test cases, as I also have tests that run when they're not enabled. The majority of test cases shouldn't need this set.

donnapep:
Just curious what the motivation for this PR is. Does it resolve the issue with your tests failing locally?

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
```diff
@@ -53,6 +53,20 @@ class WP_Test_WP_Job_Manager_Usage_Tracking_Data extends WPJM_BaseTest {
...
+		unregister_post_type( 'job_listing' );
```
donnapep: A more appropriate place for the unregistration to happen would be inside a `tearDown` function, since it's essentially undoing something created earlier.

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
```diff
@@ -53,6 +53,20 @@ class WP_Test_WP_Job_Manager_Usage_Tracking_Data extends WPJM_BaseTest {
 	 */
 	private $publish;
 
+	/**
+	 * Set up for tests.
+	 */
+	public function setUp() {
```
donnapep: If this only needs to run once for this entire test file, instead of once per test case, then a better place for this code would be in the static `setUpBeforeClass` function. This will save us some build time.

alexsanford:
> Just curious what the motivation for this PR is. Does it resolve the issue with your tests failing locally?

Yeah exactly. Currently, the full test suite passes when run together, but the `test_class.wp-job-manager-usage-tracking-data.php` file does not pass when run on its own, because it has (implicit) dependencies on other tests being run first. This PR aims to remove those dependencies and make the tests pass independently.

This PR is not critical for the 1.30.0 release.

tests/php/tests/includes/test_class.wp-job-manager-usage-tracking-data.php
jom: It needs to happen before `$post_type_instance->register_post_types();` because that doesn't update the post types (based on the options above) if it has already been registered.

donnapep: It ran OK for me when I put it in a `teardown`. 🤷‍♀️

jom (approved):
Seems safe 😃

Issue #1202, opened 2017-10-11
Replace chosen.js?


dsaric-dev:
Hello,

The included select element enhancement library (chosen.js) doesn't support mobile browsers. Are there any plans on replacing it with one that does provide mobile support such as select2 or selectize?

Thanks,
Damir

jom:
What issue are you having on your mobile browser? Please also include your browser/OS information. I'm able to select [multiple] job categories from both Chrome and Safari on iOS.

dsaric-dev:
The issue is that chosen.js library is disabled for mobile browsers, so the dropdown list defaults to html select. This also means that placeholder isn't being used, but defaults to mobile browser default of "0/n items selected."

spencerfinnell:
https://github.com/harvesthq/chosen/pull/1388

tripflex:
iOS seems to be a common issue with mobile stuff ... had similar issue with Flatpickr and iOS when using native HTML5 elements, geezzz ... @dsaric-dev can you tell us what browser you're using, the version, and what version of iOS you're using?

dsaric-dev:
@tripflex, I'm not personally using iOS, and it's not important, as this is relevant for all mobile browsers. Chosen.js doesn't support them, and it doesn't seem as it will support them any time soon.

From https://harvesthq.github.io/chosen/#faqs:
_What browsers are supported?
All modern desktop browsers are supported (Firefox, Chrome, Safari and IE9). Legacy support for IE8 is also enabled. **Chosen is disabled on iPhone, iPod Touch, and Android mobile devices** [(more information)](https://github.com/harvesthq/chosen/pull/1388)._

dsaric-dev:
Is there any progress on or interest in this issue/enhancement? @jom? Anyone else?

tripflex:
I think switching to Select2 or even just using the select lib from Semantic UI would be a nice improvement 😁

spencerfinnell:
https://github.com/jshjohnson/Choices is nice as well.

richardmtl:
Reported in 1533329-zen

codestor4:
![img_1747](https://user-images.githubusercontent.com/20779676/49800202-eda35600-fd6c-11e8-837d-3a08704ee79d.PNG)


1637325-zen

PR #1354, opened 2018-02-22
Stagger release of new asset enqueuing strategy until 1.32.0

jom:

* This will stagger the new asset enqueueing strategy to 1.32.0. This will give devs more time to add this to their themes and plugins. Theme and plugin developers can try out the upcoming behavior by adding this to their `wp-config.php`:
`define( 'JOB_MANAGER_TEST_NEW_ASSET_BEHAVIOR', true );`

* It also introduces a new filter for the frontend CSS. I think this should encourage the use of better patterns instead of having plugin and theme developers use `is_wpjm()` functions and filters. If we don't choose to stagger the release, I think we should still merge ed03d50.

Easily copy-able code to test using the behaviors (with that constant defined):
```php
add_filter( 'job_manager_chosen_enabled', function( $chosen_used_on_page ) {
	global $post;
	if ( is_singular()
		&& is_a( $post, 'WP_Post' )
		&& has_shortcode( $post->post_content, 'resumes' )
	) {
		$chosen_used_on_page = true;
	}
	return $chosen_used_on_page;
} );
```
#### Testing instructions:

* Make sure the chosen library and `frontend.css` are enqueued on all frontend pages. Once you define the constant, make sure they are only enqueued on WPJM pages. 

#### Post-merge:
- [ ] Create issue for 1.32.0

### DIFF
--- a/wp-job-manager.php
+++ b/wp-job-manager.php
@@ -307,7 +307,16 @@ class WP_Job_Manager {
 			'i18n_confirm_delete' => __( 'Are you sure you want to delete this listing?', 'wp-job-manager' ),
 		) );
 
-		if ( is_wpjm() ) {
+
+		/**
+		 * Filter whether to enqueue WPJM core's frontend scripts. By default, they will only be enqueued on WPJM related
+		 * pages.
+		 *
+		 * @since 1.30.0
+		 *
+		 * @param bool $is_frontend_style_enabled
+		 */
+		if ( apply_filters( 'job_manager_enqueue_frontend_style', is_wpjm() ) ) {
 			wp_enqueue_style( 'wp-job-manager-frontend', JOB_MANAGER_PLUGIN_URL . '/assets/css/frontend.css', array(), JOB_MANAGER_VERSION );
 		}
 	}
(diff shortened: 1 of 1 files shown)


jom:
@alexsanford 

- Correct: Old behavior is enqueueing the style regardless of if the JS is loaded as well. New behavior (really post-1.32.0) is: enqueue the style if we think the JS is also going to be enqueued. In a different PR, I think we could change it so the chosen CSS is just _registered_ here and then enqueued in either the `term-multiselect` and `multiselect` fields and the `job-filters.php` template ([here](https://github.com/Automattic/WP-Job-Manager/blob/3f1a060a2bfc146a6e5feca5e5686a667e7f82a6/templates/job-filters.php#L18)). This could be breaking, which is why it'd be good to do it in a different PR and test with plugins, etc.
- Good reminder. I'll do that when I merge.

jom:
Considering chat in #1202, I also wonder if there is a way we could do this so it is easier to slowly deprecate `chosen.js` later on.

jom:
I've been thinking about #1202 and this ticket. I think this, by itself, will help because it will force folks to be explicit about when they are using `chosen`.

PR #1356, opened 2018-02-22
When only WPJM widget, use lightweight CSS

jom:
Fixes issue with widgets not styling correctly with new asset enqueuing strategy.

* Move styles that are used by widgets into a lightweight frontend LESS/CSS. Use that CSS on pages with _just_ WPJM. On pages that contain other frontend elements, use the larger stylesheet with those styles mixed in.

#### Testing instructions:

* Verify styling of widgets and shortcodes are unchanged.


donnapep (approved):
Works great! 👍

PR #1387, opened 2018-03-16
Added new filter for job dashboard actions which allows success message

jonasvogel:
When creating custom action handlers for the job dashboard actions, there is no possibility to add a success message. While error messages can be realized with `throw new Exception(…)`, there is no way to access `$this->job_dashboard_message` from within a hook to set the success message.

I therefore created a new filter in addition to the existing action hook `job_manager_my_job_do_action` which expects to return a string as success message. The message then will be added to `$this->job_dashboard_message` and displayed to the user.

* Add new filter for job dashboard actions which allows success message

#### Testing instructions:

* See gist: https://gist.github.com/jonasvogel/fb9a10b77c75df0da82645cf6ea20975


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -158,6 +158,12 @@ public function job_dashboard_handler() {
 
 				do_action( 'job_manager_my_job_do_action', $action, $job_id );
 
+
+				$success_message = apply_filters( 'job_manager_job_dashboard_do_action', '', $action, $job_id );
+				if ( $success_message ) {
+					$this->job_dashboard_message = '<div class="job-manager-message">' . $success_message . '</div>';
+                }
```
jom: Minor text formatting issue on this line (spaces instead of tabs).

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -158,6 +158,12 @@ public function job_dashboard_handler() {
 
 				do_action( 'job_manager_my_job_do_action', $action, $job_id );
 
+
+				$success_message = apply_filters( 'job_manager_job_dashboard_do_action', '', $action, $job_id );
```
jom: You have a double space above this line. I'd recommend a more specific filter name (`job_manager_job_dashboard_success_message`). Please add documentation for this new filter. 

https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters


We should probably also consolidate `job_manager_job_dashboard_do_action_{$action}` and `job_manager_my_job_do_action`, but that can be a different PR.

jom (changes requested):
Thanks for this PR! I have a few comments.

includes/class-wp-job-manager-shortcodes.php
alexsanford: +1 on the suggested filter name. `job_manager_job_dashboard_success_message ` is much clearer IMO.

jom (approved):

PR #1390, opened 2018-03-20
Enqueue the datepicker UI when using a date field in the front-end

jom:

* Enqueue the jQuery UI CSS when using the datepicker on the front-end.

#### Testing instructions:

* Add a datepicker field and verify it is styled correctly on the front-end.
```
	$fields['job']['fancy_date'] = array(
		'label'       => __( 'Fanciest of Dates', 'wp-job-manager' ),
		'type'        => 'date',
		'required'    => false,
		'priority'    => 5
	);
	return $fields;
} );
```

Builds off of #1353

### DIFF
--- a/includes/admin/class-wp-job-manager-admin.php
+++ b/includes/admin/class-wp-job-manager-admin.php
@@ -82,14 +82,10 @@ class WP_Job_Manager_Admin {
 	 * Enqueues CSS and JS assets.
 	 */
 	public function admin_enqueue_scripts() {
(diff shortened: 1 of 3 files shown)


alexsanford (approved):
Looks great! I guess Application Deadline must do this itself? There doesn't seem to be any style issues with that date field (thus why I missed this issue). Maybe we should consider removing that styling from Application Deadline in favour of this solution?

(Also I edited the description to link to the correct PR 😄 )

PR #1406, opened 2018-03-26
Add form values filter

jonasvogel:
To allow custom handling of job submit and edit form values, add a filter to `WP_Job_Manager_Form->get_posted_fields`

* Instead of just returning the processed values, apply a filter to them.

#### Testing instructions:

```
add_filter( 'job_manager_get_posted_fields', function($values, $fields) {
    // Do something with the values
    return $values;
}, 10, 2);
```


alexsanford:
@jonasvogel Could you give an example of a use case for this that isn't already covered?

We do [have a filter](https://github.com/jonasvogel/WP-Job-Manager/blob/8324bbc7014fcf77c7659bb887ea1ee51b9902cc/includes/abstracts/abstract-wp-job-manager-form.php#L346) that allows you to supply a handler for a field type, which should allow for more focused control here.

tripflex:
👍  LGTM -- there's numerous possible reasons for needing this, as yes there is a filter for the field "type" but there's no way to modify or set field values before they are processed ... so if you need to modify or set a value for a standard field type ... it's not going to be easy to do, and would require a lot of code just to hack around getting a simple value set.

Example: user wants to remove `job_title` field from showing on submit form and set it from another field's value .... there's no easy way to do this

jom (approved):
LGTM! Thanks for adding the documentation.

Issue #672, opened 2016-05-12
Email Notifications to Admin When New Jobs Are Submitted


jenhooks:
An email notification sent to the site administrator when a new job is posted is a popular request from users, and seems like an intuitive addition to WP Job Manager.

2651428-t

spencerfinnell:
Related #645

dkolski:
Have a look to "post status notifier" which is a great plugin to manage notifications whenever a status of  post type occurs. You can define rules and define for which post-type you want to apply the notification and for which status transition (status before and after).
It costs 20$.

quantumleap33:
Here is a helpful [link](https://wpjobmanager.com/document/tutorial-send-email-employer-job-listing-approved/).
Also [Better Notification for Wordpress](https://wordpress.org/plugins/bnfw/) is really good for that and it's free. Developer is also responsive in the forums.

tripflex:
This is possible in the emails addon as well BTW https://plugins.smyl.es/wp-job-manager-emails

jom:
Bumping this to 1.31.0, which will have a more comprehensive email notification system.

codestor4:
requested: 1052655-zen

Issue #446, opened 2015-04-20
Feature Request: Notification to be sent to the listing owner via email when a listing is about to expire.


adampickering:
Options along the lines of these:

![img](https://dl.dropboxusercontent.com/spa/nnc1royxnaf5onn/cl-ykor2.jpg)

adampickering:
Would be also great to email the site admin as well as the listing owner. Perhaps an option to enable this?

kochhase:
this feature would be nice to have! And is there a plan to give this jobowner the posibility to have a discount for the next jobpage?

Roy1980:
I need this badly!

adampickering:
:+1:

Designerplus:
nice one!

Roy1980:
Any news about this?

mikejolley:
https://wordpress.org/plugins/post-status-notifier-lite/ can be used to end an email once expired.  No short term plans to add this feature request to core.

kochhase:
sorry Mike,
I talk to Timo Reith, he is gernan guy. I use his post status notifier pro 1.8. But how should he do this job? He can do this only when the jobstatus changed. Or is he able to look in the time field to evaluate the periods? I don´t now how.

timoreith:
That's right, my plugin (Post Status Notifier, link above posted by Mike) only triggers on post status changes. I could imagine a cronjob that walks through all job listings and checks the expiration date. When it encounters a match, like four days before expiration from the screenshot, it updates the listing and adds a category "expiring". Then you could create a PSN notification rule that triggers for posts having this category.
That's just my first idea. I guess we could find a better solution.

kochhase:
Hi Timo, that sounds good. Do you think, we could than send a link to spezific Jobpackets to give the  jobowner the posibility to have a discount for the next jobpacket?

kochhase:
Hi Timo,
do you know, when we can aspect some trigger for this problem? I need that feature for my update of meerjobs. thank you for help!!

craigg5:
I would really like to see this feature added also!

tripflex:
This is a feature available in the emails addon FYI https://plugins.smyl.es/wp-job-manager-emails

adampickering:
Still I think it really should be a core feature

annezazu:
599924-zen

PR #1422, opened 2018-04-05
Adds core email notification handling

jom:
This is a large chunk of functionality that I'm adding. This is the largest/first PR that just adds an MVP and starts resolving #672 (as well as others later).

Additional PRs will be added to:
- Show job details in notification to admin when new jobs are created. 
- Finish the actual notification for #672.
- Add an administration interface for enabling/disabling each notification in settings.
- Add additional notifications for editing job listings and notice of expiring job listings (#446).
- I'm also going to create a PR for the Resume's plugin to test compatibility with a WPJM extension plugin.

This PR:
- Adds the core functionality for sending notifications in WPJM.
- Includes a very minimal version of a notification that will be sent to admins when a new job listing is created.
- Adds unit tests functionality included so-far.

## Testing Instructions
- Verify unit tests pass.
- On submitting a new job listing from the `[submit_job_form]` shortcode, verify a very basic email is sent to the WordPress admin when submitted.

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-email-template.php
+++ b/includes/abstracts/abstract-wp-job-manager-email-template.php
@@ -1,0 +1,88 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly
+}
+
+/**
+ * Abstract class for an email notification built using templates.
+ *
+ * @package wp-job-manager
+ *
+ * @since 1.31.0
+ */
+
+abstract class WP_Job_Manager_Email_Template extends WP_Job_Manager_Email {
+
+	/**
+	 * Get the rich text version of the email content.
+	 *
+	 * @return string
+	 */
+	public function get_rich_content() {
+		return $this->get_template( false );
+	}
+
+	/**
+	 * Get the plaintext version of the email content.
+	 *
+	 * @return string
+	 */
+	public function get_plain_content() {
+		if ( $this->has_template( true ) ) {
+			return $this->get_template( true );
+		}
+		return strip_tags( $this->get_rich_content() );
+	}
+
+	/**
+	 * Get the contents of a template.
+	 *
+	 * @param bool $plain_text
+	 * @return string|bool
+	 */
+	public function get_template( $plain_text = false ) {
+		$template = $this->locate_template( $plain_text );
+		if ( ! $template ) {
+			return false;
+		}
+		$args = $this->get_args();
+		$email = $this;
+
+		ob_start();
+		include $template;
+		return ob_get_clean();
+	}
+
+	/**
+	 * Check to see if a template exists for this email.
+	 *
+	 * @param bool $plain_text
+	 * @return bool
+	 */
+	public function has_template( $plain_text = false ) {
+		$template_file = $this->locate_template( $plain_text );
+		return $template_file && file_exists( $template_file );
+	}
+
+	/**
+	 * Locate template for this email.
+	 *
+	 * @param bool $plain_text
+	 * @return string
+	 */
+	protected function locate_template( $plain_text ) {
+		return locate_job_manager_template( $this->get_template_file_name( $plain_text ) );
+	}
+
+	/**
+	 * Generate the file name for the email template.
+	 *
+	 * @param bool $plain_text
+	 * @return string
+	 */
+	protected function get_template_file_name( $plain_text = false ) {
+		$class_name = get_class( $this );
+		return WP_Job_Manager_Email_Notifications::get_template_file_name( $class_name::get_key(), $plain_text );
+	}
+}
--- a/includes/abstracts/abstract-wp-job-manager-email-template.php
+++ b/includes/abstracts/abstract-wp-job-manager-email-template.php
@@ -83,6 +83,8 @@ abstract class WP_Job_Manager_Email_Template extends WP_Job_Manager_Email {
 	 */
 	protected function get_template_file_name( $plain_text = false ) {
 		$class_name = get_class( $this );
-		return WP_Job_Manager_Email_Notifications::get_template_file_name( $class_name::get_key(), $plain_text );
+		// PHP 5.2: Using `call_user_func()` but `$class_name::get_key()` preferred.
+		$email_notification_key = call_user_func( array( $class_name, 'get_key') );
+		return WP_Job_Manager_Email_Notifications::get_template_file_name( $email_notification_key, $plain_text );
 	}
 }
--- a/includes/abstracts/abstract-wp-job-manager-email-template.php
+++ b/includes/abstracts/abstract-wp-job-manager-email-template.php
@@ -39,12 +39,12 @@ abstract class WP_Job_Manager_Email_Template extends WP_Job_Manager_Email {
 	 * Get the contents of a template.
 	 *
 	 * @param bool $plain_text
-	 * @return string|bool
+	 * @return string
 	 */
 	public function get_template( $plain_text = false ) {
 		$template = $this->locate_template( $plain_text );
 		if ( ! $template ) {
-			return false;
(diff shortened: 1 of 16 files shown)


includes/abstracts/abstract-wp-job-manager-email-template.php
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * Get the plaintext version of the email content.
+	 *
+	 * @return string
+	 */
+	public function get_plain_content() {
+		if ( $this->has_template( true ) ) {
+			return $this->get_template( true );
+		}
+		return strip_tags( $this->get_rich_content() );
```
alexsanford: Would it work to call the parent function here instead?

includes/abstracts/abstract-wp-job-manager-email-template.php
```diff
@@ -0,0 +1,90 @@
...
+
+abstract class WP_Job_Manager_Email_Template extends WP_Job_Manager_Email {
+
+	/**
+	 * Get the rich text version of the email content.
+	 *
+	 * @return string
+	 */
+	public function get_rich_content() {
+		return $this->get_template( false );
```
alexsanford: `get_template` may return a bool, whereas this function claims to return a string. Is this ok?

includes/abstracts/abstract-wp-job-manager-email-template.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	/**
+	 * Get the plaintext version of the email content.
+	 *
+	 * @return string
+	 */
+	public function get_plain_content() {
+		if ( $this->has_template( true ) ) {
+			return $this->get_template( true );
```
alexsanford: `get_template` may return a bool, whereas this function claims to return a string. Is this ok?

includes/abstracts/abstract-wp-job-manager-email-template.php
```diff
@@ -0,0 +1,90 @@
...
+	 * Generate the file name for the email template.
+	 *
+	 * @param bool $plain_text
+	 * @return string
+	 */
+	protected function get_template_file_name( $plain_text = false ) {
+		$class_name = get_class( $this );
+		// PHP 5.2: Using `call_user_func()` but `$class_name::get_key()` preferred.
+		$email_notification_key = call_user_func( array( $class_name, 'get_key') );
+		return WP_Job_Manager_Email_Notifications::get_template_file_name( $email_notification_key, $plain_text );
```
alexsanford: This feels like a circular dependency to me. Since this file is responsible for rendering emails with a template, would it make sense for it to have knowledge of where to look for templates, rather than deferring to `WP_Job_Manager_Email_Notifications`? Not a show-stopper, but I'm uneasy with this extra dependency for some reason :)

includes/class-wp-job-manager-email-notifications.php
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	/**
+	 * Returns the total number of deferred notifications to be sent.
+	 *
+	 * @access private
+	 *
+	 * @return int
+	 */
+	public static function _get_deferred_notification_count() {
```
alexsanford: It looks like this function is only used by the unit tests. Should we mention that in the doc comment?

alexsanford (approved):
This looks really great! I tested it and it works well. I have a few nitpicks, but otherwise I think this looks good. Thanks for your hard work on this one! 🙂

tripflex:
Regarding the email templates, if you guys are interested, I wouldn't mind looking at open sourcing a slimmed down version of my WP Job Manager Emails addon to allow customization of core email templates, that way users could have a UI for building the emails out:

https://plugins.smyl.es/wp-job-manager-emails/

Of course it would be open source, and would not have a bunch of ads to purchase "premium" version (bc that annoys me so much when other plugins do that)  ... we can discuss this more in Slack if you guys are interested

includes/abstracts/abstract-wp-job-manager-email-template.php
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Abstract class for an email notification built using templates.
+ *
+ * @package wp-job-manager
+ *
+ * @since 1.31.0
+ */
+
+abstract class WP_Job_Manager_Email_Template extends WP_Job_Manager_Email {
```
tripflex: Just FYI, `WP_Job_Manager_Email` class is perfectly fine, but wanted to point out that my addon uses `WP_Job_Manager_Emails` ... just something to keep in mind

includes/class-wp-job-manager-email-notifications.php
```diff
@@ -0,0 +1,334 @@
...
+			$headers[] = 'From: ' . $args['from'];
+		}
+
+		if ( ! self::send_as_plain_text( $email_notification_key ) ) {
+			$headers[] = 'Content-Type: text/html';
+		}
+
+		$content = self::get_email_content( $email_notification_key, $args );
+
+		return wp_mail( $args['to'], $args['subject'], $content, $headers, $args['attachments'] );
```
tripflex: Would be good to have a filter here for the ability to short circuit the email if needed.  Right now if I ever need to do this through `wp_mail` I have to short circuit by setting `to` and `message` to null/false values (until easy way to short circuit is integrated) https://core.trac.wordpress.org/ticket/35069

includes/abstracts/abstract-wp-job-manager-email-template.php
jom: Good catch! Done.

includes/abstracts/abstract-wp-job-manager-email-template.php
jom: I had the logic for locating the template here, but actually moved it to `WP_Job_Manager_Email_Notifications` because of [upcoming work](https://github.com/Automattic/WP-Job-Manager/tree/add/job-details) where _segments_ of the emails will be rendered in that class. Otherwise, `WP_Job_Manager_Email_Notifications` would have to call a static on a incomplete abstract class (which I think is possible, but I feel uneasy with). 

The alternative was creating a global functions, but I try to keep those to a minimum.

includes/abstracts/abstract-wp-job-manager-email-template.php
jom: I'll change the signature of `get_template()` method 5e9d1a0f8297d9275c7c0703c613e7085fa8f817

includes/abstracts/abstract-wp-job-manager-email-template.php
jom: I'll change the signature of `get_template()` method 5e9d1a0f8297d9275c7c0703c613e7085fa8f817

includes/class-wp-job-manager-email-notifications.php
jom: Done 4082f95acda790eedb26fc9678548db8161e4721

includes/abstracts/abstract-wp-job-manager-email-template.php
jom: Noted 👍

includes/class-wp-job-manager-email-notifications.php
jom: eecfeee6d7fee77ac62c2f534d60201fdcfa24f6 work?

jom:
@tripflex Thanks for taking a look at this. I think adding a UI editor for these might be nice at some point. For now the admin UI will be pretty simple (mostly just enable/disable). I went the template route for now because it makes sense for a lot of the notifications, but I think a child class `WP_Job_Manager_Email` (e.g. `WP_Job_Manager_Email_Editable`) and accompanying UI could be a nice addition, possibly borrowing from your plugin. We can chat more on Slack 👍.

Issue #1362, opened 2018-03-01
Delete plugin data on uninstall


donnapep:
As part of the GDPR, we need to delete all data (especially user data) when a user uninstalls WPJM.

WordPress core shows a prompt when deleting a plugin to indicate that both the plugin and its data will be deleted. The user must accept this confirmation before anything gets removed:

![screen shot 2018-03-01 at 1 32 08 pm](https://user-images.githubusercontent.com/1190420/36862502-f977351c-1d54-11e8-9036-65aa4950af83.jpg)

We'll need to be sure to communicate that WPJM will now delete ALL of its data, as there may be site admins who have grown accustomed to deleting the plugin and then immediately re-installing it (perhaps installing a different version), and take for granted that most of their data will remain intact.

# What to Delete?

The following data should be deleted when WP Job Manager is deleted if that is not the case already. _Note that for a multi-site install, this means that the data deletion needs to happen across all sites._

## Custom Post Types

_Done: https://github.com/Automattic/WP-Job-Manager/pull/1418_

All posts for the following custom post type should be moved to the Trash:
- `job_listing`

## Terms & Taxonomies

_Done: https://github.com/Automattic/WP-Job-Manager/pull/1419_

Term and taxonomy data for the following taxonomies should be deleted:
- `job_listing_category`
- `job_listing_type`

## Pages

_Done: https://github.com/Automattic/WP-Job-Manager/pull/1421_

The following pages should be moved to the Trash:
- The page referenced in the _Submit Job Form Page_ field of the settings
- The page referenced in the _Job Dashboard Page_ field of the settings
- The page referenced in the _Job Listings Page_ field of the settings

## Options

_Done: https://github.com/Automattic/WP-Job-Manager/pull/1427_

The following options should be deleted:
- `wp_job_manager_version`
- `job_manager_installed_terms`
- `wpjm_permalinks`
- `job_manager_helper`
- `job_manager_date_format`
- `job_manager_google_maps_api_key`
- `job_manager_usage_tracking_enabled`
- `job_manager_usage_tracking_opt_in_hide`
- `job_manager_per_page`
- `job_manager_hide_filled_positions`
- `job_manager_hide_expired`
- `job_manager_hide_expired_content`
- `job_manager_enable_categories`
- `job_manager_enable_default_category_multiselect`
- `job_manager_category_filter_type`
- `job_manager_enable_types`
- `job_manager_multi_job_type`
- `job_manager_user_requires_account`
[...]

PR #1431, opened 2018-04-10
Add data deletion for WPJM Cron Jobs

alexsanford:
Contributes to #1362

This PR, on plugin deletion, deletes the cron jobs associated with WPJM. The following cron jobs are deleted:

- `job_manager_check_for_expired_jobs`
- `job_manager_delete_old_previews`
- `job_manager_clear_expired_transients`
- `job_manager_usage_tracking_send_usage_data`

## Testing

- First, you may want to back up your data, or copy it to a fresh WordPress installation.

- Ensure that the tests pass.

- With WPJM installed, inspect the cron jobs. The ones listed above should be scheduled. You can use a plugin such as WP Crontrol for this.

- Deactivate the WPJM plugin.

- Verify the WPJM cron jobs have been unscheduled.

- Please also test on a multisite installation. When the plugin is deleted from the Network Admin OR deactivated from an individual site, the WPJM cron jobs should be removed.

### DIFF
--- a/includes/class-wp-job-manager-data-cleaner.php
+++ b/includes/class-wp-job-manager-data-cleaner.php
@@ -342,7 +342,8 @@ class WP_Job_Manager_Data_Cleaner {
 	}
 
 	/**
-	 * Cleanup cron jobs.
+	 * Cleanup cron jobs. Note that this should be done on deactivation, but
+	 * doing it here as well for safety.
 	 *
 	 * @access private
 	 */
--- a/includes/class-wp-job-manager-data-cleaner.php
+++ b/includes/class-wp-job-manager-data-cleaner.php
@@ -38,6 +38,17 @@ class WP_Job_Manager_Data_Cleaner {
 		'job_listing_type',
 	);
 
+	/** Cron jobs to be unscheduled.
+	 *
+	 * @var $cron_jobs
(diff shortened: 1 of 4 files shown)


jom:
We should probably do this on deactivation instead of uninstall. It looks like we currently have this in `uninstall.php`, but might want to move it to a deactivation hook (like we did for usage tracking).

alexsanford:
@jom Good point!

I opted to add the unscheduling of the cron jobs in a deactivation hook, but also to leave the implementation as-is to unschedule them on plugin delete. This should be redundant, but I feel that the data cleaning code should be thorough, in case those cron jobs have stuck around for some reason (maybe the plugin somehow gets deactivated without the hook running properly?)

I don't think the redundancy is too messy, but would like to hear your thoughts 🙂

jom:
Works for me! We can also probably remove the old calls in `uninstall.php`.

alexsanford:
Good catch! I removed them, and the option deletion code too.

I don't know about the Usage Tracking code in `uninstall.php` though. It's redundant, but it also makes sense for the usage tracking class to clean up after itself. We don't need both the data cleaner AND the usage tracking to clean up those options, but I don't want to remove the functionality from either 😄

PR #1433, opened 2018-04-10
Add class to the job title and replace H1 tag with H2

### LINKED ISSUE #1424: Job Summary shortcode outputs H1 element; add class to the title
As raised in this thread: https://wordpress.org/support/topic/seo-issue-5/

If you use the Job Summary shortcode, it's usually within a different page that will already have an H1; we should probably use an H2. 

I haven't come across this shortcode being used "in the wild" much, so if we were to change this, I doubt it would have much impact on sites. If any themes are already styling this, they'll have to update their styles which could be a "breaking" change on some sites. To make it easier to target the element and fix the styles, we should add a class to it as well. So, instead of 

`<h1>job title here</h1>`

I suggest we use something like this instead

`<h2 class="job_summary_title">job title here</h2>`

I realize that we open things up with a `<div class="job_summary_content">` right before, but adding a class there makes targeting the h2 with (or excluding from) a particular style even easier.

abdullah1908:
Fixes #1424

**Changes proposed in this Pull Request:**

- Added class to the job title

- Replaced H1 tag with H2

### DIFF
--- a/templates/content-summary-job_listing.php
+++ b/templates/content-summary-job_listing.php
@@ -8,7 +8,7 @@
  * @author      Automattic
  * @package     WP Job Manager
  * @category    Template
- * @version     1.27.0
+ * @version     1.31.0
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
--- a/templates/content-summary-job_listing.php
+++ b/templates/content-summary-job_listing.php
@@ -34,7 +34,7 @@ global $job_manager;
 
 	<div class="job_summary_content">
 
-		<h1><?php wpjm_the_job_title(); ?></h1>
+		<h2 class="job_summary_title"><?php wpjm_the_job_title(); ?></h2>
 
 		<p class="meta"><?php the_job_location( false ); ?> &mdash; <?php the_job_publish_date(); ?></p>
 


jom:
Hi @abdullah1908! Thanks for sending this in. We'll ship it with 1.31.0. Would you mind changing the version for the template (at the top of the file) to `1.31.0`? Thank you!

spencerfinnell:
@jom is there an expected release date for 1.30.0? Since the only way to target this before was the h1 tag directly and this is now removed themes will need to adjust accordingly.

richardmtl:
@spencerfinnell Yeah, I did raise that concern in the original issue:

<blockquote> If any themes are already styling this, they'll have to update their styles which could be a "breaking" change on some sites.</blockquote>
 
I haven't seen this used much "in the wild" though, in the sites that I've done support for. I think this PR makes things better all-around for the plugin; do you think it'll be a big deal?

spencerfinnell:
@jom Thanks.

@richardmtl Honestly not sure how many people use this shortcode. But an `h2` tag will still likely have _some_ styling inside a page so it will at least still look like a title. Hopefully it will take a similar shape to what anyone previous needed to override it to. We don't have any specific styles for this in our themes.

abdullah1908:
@jom @rachelmcr ,

I have updated the template version as well 💯 

Let me know if anything for me relevant to this PR.

Thanks

PR #1443, opened 2018-04-12
Update readme.txt and readme.md

### LINKED ISSUE #1424: Job Summary shortcode outputs H1 element; add class to the title

abdullah1908:
Update readme.txt and readme.md on the behalf of Fix #1424


abdullah1908:
@jom ,

Can you please review this PR as well with reference of Fix #1424

abdullah1908:
Hello @jom ,

I have made significant contributions to the second most Job Board solution on [wordpress.org](https://wordpress.org/plugins/simple-job-board/) named as **[Simple Job Board](https://wordpress.org/plugins/simple-job-board/)** and I am the part of the [PressTigers](presstigers.com) team. I'll surely contribute consistently to the WP Job Manager and hopefully this PR will have some worth.

Thanks

PR #1448, opened 2018-04-13
Add settings UI for email notifications

jom:
Fixes #1439 

Example of future settings:
* For #446 we'll add a setting for how many days ahead we should warn employers before their listings expire.
* We could pretty easily make the subject customizable.
* For some notifications, the entire notification could be customizable. 

#### Changes proposed in this Pull Request:

* Restructures `WP_Job_Manager_Settings` (318c1405663e7de250588264296e616b0ccbc967) to support the new field type `multi` (compound settings stored in serialized arrays) and `mutli_enable_expand`.
* Adds basic fields for enabled and send in plain/rich text.
* Does some minor UI change to settings (adds line separator) to help a bit with the new email notifications screen.

**Round 2 Design**:
email-notifications-demo
screen shot 2018-04-14 at 5 19 49 pm

<a href="https://user-images.githubusercontent.com/68693/38757143-d89fff5a-3f63-11e8-98b5-a36ea6b5adda.png">Round 1 Design</a>

#### Testing instructions:

* Ensure all tests pass.
* Try enabling/disabling these settings and make sure the notifications are sent when appropriate in the right type (rich/plain).

cc: @tripflex 

To show what it looks like when we add another setting:
screen shot 2018-04-15 at 2 30 03 pm

PR #1449, opened 2018-04-16
Send notice to employers of expiring job listings

### LINKED ISSUE #446: Feature Request: Notification to be sent to the listing owner via email when a listing is about to expire.

jom:
Fixes #446 

Depends on #1448 (~rebase~ rebased)

* Send a notice to employers when job listing is about to expire.

#### Testing instructions:

* Create a job listing that expires tomorrow. 
* Enable the notification and have it notify 1 day away.
* Run `do_action( 'job_manager_email_daily_notices' );` and verify the email is sent. 

* Notify employers when their listings are expiring.

### DIFF
--- a/includes/class-wp-job-manager-email-notifications.php
+++ b/includes/class-wp-job-manager-email-notifications.php
@@ -432,6 +432,8 @@ final class WP_Job_Manager_Email_Notifications {
 	 * Sending notices to employers for expiring job listings.
 	 */
 	public static function send_employer_expiring_notice() {
+		self::maybe_init();
+
 		$email_key   = call_user_func( array( 'WP_Job_Manager_Email_Employer_Expiring_Job', 'get_key' ) );
 		if ( ! self::is_email_notification_enabled( $email_key ) ) {
 			return;
--- a/includes/class-wp-job-manager-email-notifications.php
+++ b/includes/class-wp-job-manager-email-notifications.php
@@ -274,6 +274,15 @@ final class WP_Job_Manager_Email_Notifications {
 			);
 		}
 
+		$job_expires = get_post_meta( $job->ID, '_job_expires', true );
+		if ( ! empty( $job_expires ) ) {
+			$job_expires_str = date_i18n( get_option( 'date_format' ), strtotime( $job_expires ) );
+			$fields['job_expires'] = array(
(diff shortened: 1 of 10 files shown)


includes/emails/class-wp-job-manager-email-employer-expiring-job.php
```diff
@@ -0,0 +1,163 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly
+}
+
+/**
+ * Email notification to employers when a job is updated.
```
alexsanford: This description needs to be updated

includes/emails/class-wp-job-manager-email-employer-expiring-job.php
```diff
@@ -0,0 +1,163 @@
...
+		return $args['author']->user_email;
+	}
+
+	/**
+	 * Expand arguments as necessary for the generation of the email.
+	 *
+	 * @param $args
+	 * @return mixed
+	 */
+	protected function prepare_args( $args ) {
```
alexsanford: Just a thought (not blocking for this PR):

All three email subclasses so far have the `job_id` logic in this method, and two of them have the `user_id` logic as well. I wonder if it would be worth writing a couple of helper methods so that this logic isn't duplicated? I know it's not a lot of code, but it seems like it could end up being copied in a lot of subclasses.

includes/class-wp-job-manager-email-notifications.php
```diff
@@ -416,6 +428,48 @@ public static function send_as_plain_text( $email_notification_key ) {
 		return apply_filters( 'job_manager_email_send_as_plain_text', $send_as_plain_text, $email_notification_key );
 	}
 
+	/**
+	 * Sending notices to employers for expiring job listings.
+	 */
+	public static function send_employer_expiring_notice() {
+		self::maybe_init();
+
+		$email_key   = call_user_func( array( 'WP_Job_Manager_Email_Employer_Expiring_Job', 'get_key' ) );
```
alexsanford: Why do we need to use `call_user_func` here?

alexsanford (approved):
Looks good! I have a few little comments, but no blockers.

🚀

PR #1450, opened 2018-04-16
[WIP] Allow users to relist a job listing before expiry

### LINKED ISSUE #848: Provide relist option before expiration

alexsanford:
Fixes #848

* Gives users an option on the Job Dashboard to Relist a job when it is within 5 days of expiry.
* The option is only available for Job listings whose status is `publish`.

#### Issue

When the job is relisted, and the Moderate New Listings setting is enabled, the listing's status will be changed to `pending` and will no longer be public until approved. We may want to discuss a better implementation for this.

#### Testing instructions:

* Go to the Job Dashboard page. Notice the option to Relist the job listing even though it is not expired.
* Relist the listing. The new expiry date should be calculated based on the old expiry date, not the current date.
* [WIP] Test with the WooCommerce Paid Listings add-on. The flow should be the same, and relisting the job listing should use up one listing in a purchased Job Package Product. _(Currently this does not seem to be working)_

Add ability to re-list a job listing before it is expired.

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -112,7 +112,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		// Load job details
 		if ( $this->job_id ) {
 			$job_status = get_post_status( $this->job_id );
-			if ( 'expired' === $job_status || 'publish' === $job_status && job_manager_job_expiring_soon( $this->job_id ) ) {
+			if ( 'expired' === $job_status || job_manager_job_can_be_relisted( $this->job_id ) ) {
 				if ( ! job_manager_user_can_edit_job( $this->job_id ) ) {
 					$this->job_id = 0;
 					$this->step   = 0;
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -803,11 +803,18 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 
 		// Continue = change job status then show next screen
 		if ( ! empty( $_POST['continue'] ) ) {
-			$job = get_post( $this->job_id );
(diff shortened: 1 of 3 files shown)


alexsanford:
Some notes on this:

- It doesn't currently work with WC Paid Listings (untested on Simple Paid Listings) and we have some considerations for that.
- The implementation here trades one problem for another. IMO the point of relisting early is to avoid the blip of downtime you get between expiry and relisting. With this implementation, the blip still happens if the listing requires approval.
- Some possible alternatives:
  - Do not go through the Edit flow when relisting. Instead, just extend the expiry date. This would likely mean more custom work in the Paid Listings plugins. It may, however, mean that we can skip approval, though not necessarily.
  - Have a way of copying the listing to a pending one requiring relisting, and then merge it back in when approved. This could work well but might be complex.
  - Have another way to submit a listing for approval without changing the status to pending (maybe using some special meta).

Some notes about the integration with WC Paid Listings:

- Users should use one of the slots in their package in order to relist a listing. This means that a single listing may take multiple slots in a package. This should be fine as long as we never explicitly recalculate the package count (I did a code investigation, and believe that this is fine, but it's worth the mention).
- We need to be a bit careful when dealing with subscription packages of type "Listing" as a special case:
  - If the relisting is done with the same package, the package count should **not be updated**. This is because the listing simply retains the slot that it is already in.
  - If the relisting is done using a new package/subscription, the package count of the original subscription should be decremented, because that slot should be freed up. The package count will be incremented on the new package that the listing is being added to.

benjiguth:
Any updates on this one? We need it!

benjiguth:
Was hoping to see this last year. Still begging for this feature. @htdat @iamgabrielma can you provide any insight here? @Automattic @jom @alexsanford @spencerfinnell @kraftbj @donnapep 
Come on everyone BE THE HEROS I KNOW YOU ARE!!!!!

Braehler:
Any updates on this one?

cena:
3297703-zen

peakmvpstl:
Are their any updates on this?

PR #1469, opened 2018-04-25
Emails should have contexts and use template paths for their context

jom:
As I started thinking about how we'd implement this in our extension plugins, I realized we want to give emails a context so they can appear in different setting screens _and_ they should follow a template path and default template path for overriding in plugins.

@alexsanford  I also closed that circular dependency for the template file name method 😉

### DIFF
--- a/includes/class-wp-job-manager-email-notifications.php
+++ b/includes/class-wp-job-manager-email-notifications.php
@@ -394,7 +394,7 @@ final class WP_Job_Manager_Email_Notifications {
 	 * Add email notification settings for a context.
 	 *
 	 * @param array  $settings
-	 * @param string $context (Defaults to Job Manager core)
+	 * @param string $context
 	 * @return array
 	 */
 	public static function add_email_settings( $settings, $context ) {
--- a/includes/class-wp-job-manager-email-notifications.php
+++ b/includes/class-wp-job-manager-email-notifications.php
@@ -28,12 +28,12 @@ final class WP_Job_Manager_Email_Notifications {
 	public static function init() {
 		add_action( 'job_manager_send_notification', array( __CLASS__, '_schedule_notification' ), 10, 2 );
 		add_action( 'job_manager_email_init', array( __CLASS__, '_lazy_init' ) );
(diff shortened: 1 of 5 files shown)


includes/class-wp-job-manager-email-notifications.php
```diff
@@ -325,13 +307,16 @@ private static function get_job_detail_fields( WP_Post $job, $sent_to_admin, $pl
 	/**
 	 * Output email header.
 	 *
-	 * @param WP_Job_Manager_Email $email          Email object for the notification.
-	 * @param bool                 $sent_to_admin  True if this is being sent to an administrator.
-	 * @param bool                 $plain_text     True if the email is being sent as plain text.
+	 * @param string $email_notification_key  Email object for the notification.
```
alexsanford: `s/object/key`

includes/class-wp-job-manager-email-notifications.php
```diff
@@ -340,29 +325,88 @@ public static function output_header( $email, $sent_to_admin, $plain_text = fals
...
 		}
 		include $template_segment;
 	}
 
 	/**
-	 * Add email notification settings.
+	 * Checks for an alternative email template segment in the template path specified by the current email.
+	 * Useful to provide alternative email headers and footers for a specific WPJM extension plugin.
+	 *
+	 * @param string  $email_notification_key  Email object for the notification.
```
alexsanford: `s/object/key`

alexsanford (commented):
Just some small inconsistencies in the docs that we might want to fix.

includes/class-wp-job-manager-email-notifications.php
jom: Fixed these in 6dd46dd2d7c431b867df4ba281e3e9a1f2bc1277

includes/class-wp-job-manager-email-notifications.php
jom: Fixed in 6dd46dd2d7c431b867df4ba281e3e9a1f2bc1277

PR #1472, opened 2018-04-26
When deleting users without merging delete their job listings

jom:
While deleting a user and `Delete all content` is selected, job manager will now allow job listings to be deleted as well.

## Testing Instructions
- In Incognito, post several job listings as a new user.
- Delete that user and choose `Delete all content`. 
- Verify job listings are deleted.

screen shot 2018-04-26 at 4 30 17 pm

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -244,7 +244,8 @@ class WP_Job_Manager_Post_Types {
 				'query_var' 			=> true,
 				'supports' 				=> array( 'title', 'editor', 'custom-fields', 'publicize', 'thumbnail' ),
 				'has_archive' 			=> $has_archive,
-				'show_in_nav_menus' 	=> false
+				'show_in_nav_menus' 	=> false,
+				'delete_with_user'		=> true,
 			) )
 		);
 


donnapep:
@jom Do you think we should consider including some facility to delete a user's data without having to delete the user? Re: [this comment](https://wpjobmanager.com/2018/04/26/whats-new-in-wp-job-manager-1-31-0/).

jom:
@donnapep Yes, but that seems like a different / much larger PR. Before adding full GDPR compliance, I'm curious what tools WP core adds.

donnapep:
> Before adding full GDPR compliance, I'm curious what tools WP core adds.

Yeah, I'm just wondering how long we can afford to wait. More and more people are asking about this. We may need to investigate the options sooner rather than later.

jom:
Looks like [many GDPR tickets](https://core.trac.wordpress.org/query?status=accepted&status=assigned&status=closed&status=new&status=reopened&status=reviewing&keywords=~gdpr&milestone=4.9.6&order=priority) are landing in [4.9.6 a beta next week](https://make.wordpress.org/core/2018/04/23/dev-chat-summary-april-18th-4-9-6-week-3/) slated for release on 2018-05-15.

PR #1481, opened 2018-05-03
Use spaces to separate image extensions.

### LINKED ISSUE #1479: Cosmetic change to "Uploaded files need to be..." string.
A user in 1103206-zen reports that the [mention of JPG files](https://github.com/Automattic/WP-Job-Manager/blob/fa9347765e51a9b7cb95554dad841ac7b1256504/wp-job-manager-functions.php#L1291) in the "Uploaded files need to be..." string looks messy:

uploaded_files_need_to_be

 Can we make this look a bit prettier?

oscarssanchez:
Hello,

Could you use this PR which fixes #1479 ?

* Add spaces before and after | to separate the output display when an image type is not valid. 

Thanks!

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -1288,18 +1288,18 @@ function job_manager_upload_file( $file, $args = array() ) {
 function job_manager_get_allowed_mime_types( $field = '' ){
 	if ( 'company_logo' === $field ) {
 		$allowed_mime_types = array(
-			'jpg|jpeg|jpe' => 'image/jpeg',
-			'gif'          => 'image/gif',
-			'png'          => 'image/png',
+			'jpg | jpeg | jpe' => 'image/jpeg',
+			'gif'              => 'image/gif',
+			'png'              => 'image/png',
 		);
 	} else {
 		$allowed_mime_types = array(
-			'jpg|jpeg|jpe' => 'image/jpeg',
-			'gif'          => 'image/gif',
-			'png'          => 'image/png',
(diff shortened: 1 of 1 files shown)


tripflex:
I don't think this is the right way to fix this.  We should keep the key names without spaces as that is exactly how they are in core WordPress ... really all I think that needs to be done for this is to just do a string replacement on `|` with `, `

tripflex:
Something like this should be used instead:

```php
$allowed_mime_types = array(
    'jpg|jpeg|jpe' => 'image/jpeg',
    'gif'          => 'image/gif',
    'png'          => 'image/png',
);

$allowed_mime_types_csv = implode( ', ', str_replace( '|', ', ', array_keys( $allowed_mime_types ) ) );
```

Produces:

```php
string(24) "jpg, jpeg, jpe, gif, png"
```

oscarssanchez:
Hi @tripflex,

Thanks for your feedback, i see your point, it makes sense that every jpeg extension is separated by a comma as other image formats are separated likewise, hence the cosmetic change looks more ordered.

If it is fine with Auttomatic, i can implement the changes right away.

Thanks again!

jom:
Hi @oscarssanchez! Thanks for this PR. I agree with @tripflex. Even though the `job_manager_get_allowed_mime_types()` function isn't used for that yet, I'd rather update the parsing of the output in the error for `job_manager_upload_file()` to handle `|`.

jom:
Closing for now, but feel free to ask for reopen if you get a chance to refactor.

PR #1491, opened 2018-05-15
Use strict sanitization for usernames

### LINKED ISSUE #1490: BUG: Sorry, that username already exists
In some cases `wp_job_manager_create_account()` attempts to create a user with a username that has already been taken.

#### Steps to Reproduce
1. Enable the setting for account creation during job submission
2. Create a job using the front end form with the email `mike+test@domain.com`
3. Create a second job using the front end form with the email `mike+test@anotherdomain.com`
4. You will see an error `Sorry, that username already exists!`

WP 4.9.5
WPJM 1.31.0

timothyjensen:
In order to accurately check if the username exists we have to use the same level of strictness as the WP core function `wp_insert_user()`, which is to say `$strict = true`.

Fixes #1490

* Set strict sanitization for usernames

#### Testing instructions:

1. Enable the setting for account creation during job submission
2. Create a job using the front end form with the email mike+test2@domain.com
3. Create a second job using the front end form with the email mike+test2@anotherdomain.com
4. There should no longer be an error message upon submission

Fix bug where new user could not be created due to conflicting user name

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -532,7 +532,7 @@ function wp_job_manager_create_account( $args, $deprecated = '' ) {
 		extract( $args );
 	}
 
-	$username = sanitize_user( $username );
+	$username = sanitize_user( $username, true );
 	$email    = apply_filters( 'user_registration_email', sanitize_email( $email ) );
 
 	if ( empty( $email ) ) {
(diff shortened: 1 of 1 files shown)


jom (approved):
Thanks for this PR! LGTM.

PR #1503, opened 2018-05-21
optimises a call to fetch all post meta in the duplication code

tomjn:

* This swaps out an SQL query for `get_post_meta` with a single argument

This should simplify the code, and improve performance via `WP_Cache`. The biggest improvements are likely to be seen when a persistent object cache is available

#### Testing instructions:

* duplicate listings

* Optimisations to listing duplication

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -1388,10 +1388,9 @@ function job_manager_duplicate_listing( $post_id ) {
 	/*
 	 * Duplicate post meta, aside from some reserved fields.
 	 */
-	$post_meta = $wpdb->get_results( $wpdb->prepare( "SELECT meta_key, meta_value FROM {$wpdb->postmeta} WHERE post_id=%d", $post_id ) );
+	$post_meta = get_post_meta( $post_id );
 
 	if ( ! empty( $post_meta ) ) {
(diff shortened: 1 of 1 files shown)


jom:
As is, this breaks duplication functionality. I think this is because `get_post_meta()` returns values in an array and we treat these as `unique` (so just want the first one, if it exists).

In this example, `post_ID = 1277` is the duplicate and has its value set as a serialized array.
screen shot 2018-09-07 at 1 00 29 pm

My other concern is about plugin interactions with filters in `get_metadata()`, such as `get_{$meta_type}_metadata`. I'm not sure if WPML or other plugins that do things with metadata in view would prevent us from getting the raw value.

jom (changes requested):
As is, this breaks duplication functionality. I think this is because `get_post_meta()` returns values in an array and we treat these as `unique` (so just want the first one, if it exists).

In this example, `post_id = 1277` is the duplicate and has its value set as a serialized array.
screen shot 2018-09-07 at 1 00 29 pm

My other concern is about plugin interactions with filters in `get_metadata()`, such as `get_{$meta_type}_metadata`. I'm not sure if WPML or other plugins that do things with metadata in view would prevent us from getting the raw value.

PR #1510, opened 2018-05-28
Add Salary Field

### LINKED ISSUE #1320: Add salary field to satisfy Google's Job Search schema
Assuming a user fills out all the job form fields correctly, Google will still throw a warning (_not_ an error) because listings do not have a salary by default. We should consider adding this:

https://wpjobmanager.com/document/tutorial-adding-a-salary-field-for-jobs/

... to the core plugin, IMHO.

We should also consider renaming the `Location` field to `Address`, since if you don't enter an exact street address, Google will throw other warnings (or actual errors, depending on how the user enters the location). `Address` is less ambiguous in this regard.

Related to #1313

rajeebbanstola:
Fixes #1320 

* Add Salary Field

#### Testing instructions:

* Copy the output HTML and paste here - https://search.google.com/structured-data/testing-tool/u/0/

* Add Salary Field


RajeebTheGreat:
- Added Salary field to both backend and frontend form - https://github.com/Automattic/WP-Job-Manager/pull/1510/files#diff-c117e2e96cdacd51451ce7c82afd7bafR60
https://github.com/Automattic/WP-Job-Manager/pull/1510/files#diff-3f735c69f0317a05c6e6d037ec955995R194
- Added job salary to the feed - https://github.com/Automattic/WP-Job-Manager/pull/1510/files#diff-45bedf8b655c4c72a2f18be3dc8f459aR456
- Added salary to the template
- Added schema markup for salary - https://github.com/Automattic/WP-Job-Manager/pull/1510/files#diff-50b2a3d20d5e9f9fc1f6059bb05182b1R403
- Added functions to retrieve job salary

**Issues:**
1. Right now, I'm adding filters for job salary currency and unit text. Maybe we could add that in options? - https://github.com/Automattic/WP-Job-Manager/pull/1510/files#diff-50b2a3d20d5e9f9fc1f6059bb05182b1R413
2. I used a briefcase icon for the salary. That is not appropriate?
3. Just like location, which prints "Anywhere" if the location is not set, we could add some text if user doesn't enter "salary".
4. Writing tests
5. Making code compatible with WP Rest API

wp-job-manager-template.php
```diff
@@ -1098,6 +1121,58 @@ function get_the_company_twitter( $post = null ) {
...
+	 * @since 1.0.0
+	 *
+	 * @param string  $job_salary
+	 * @param WP_Post $post
+	 */
+	return apply_filters( 'the_job_salary', $job_salary, $post );
+}
+
+/**
+ * Displays or retrieves the current company name with optional content.
```
sboisvert: This seems to be inaccurate :)

wp-job-manager-template.php
```diff
@@ -1098,6 +1121,58 @@ function get_the_company_twitter( $post = null ) {
...
+	if ( ! $post || 'job_listing' !== $post->post_type ) {
+		return;
+	}
+
+	$job_salary = $post->_job_salary;
+
+	/**
+	 * Filter the returned job salary.
+	 *
+	 * @since 1.0.0
```
sboisvert: I suspect this is inaccurate :)

wp-job-manager-template.php
```diff
@@ -399,6 +399,29 @@ function wpjm_get_job_listing_structured_data( $post = null ) {
...
 
+
+	$salary = get_the_job_salary( $post );
+	if ( ! empty( $salary ) ) {
+		$data['baseSalary'] = array();
+		$data['baseSalary']['@type'] = 'MonetaryAmount';
+		/**
+		 * Filter for the salary currency.
+		 *
+		 * @since 1.0.0
```
sboisvert: I suspect this is inaccurate :)

wp-job-manager-template.php
```diff
@@ -1098,6 +1121,58 @@ function get_the_company_twitter( $post = null ) {
...
+	 * @param string  $job_salary
+	 * @param WP_Post $post
+	 */
+	return apply_filters( 'the_job_salary', $job_salary, $post );
+}
+
+/**
+ * Displays or retrieves the current company name with optional content.
+ *
+ * @since 1.0.0
```
sboisvert: This seems to be inaccurate :)

sboisvert (dismissed):
This looks great!

A few questions:
1) in  includes/admin/class-wp-job-manager-writepanels.php why do we add it as priority 2? 

2) It seems we're specifying MONTH as the default "unit" but not showing it inside the user interface for when the user adds the salary. I'm not sure about other places in the world but in Canada, most salaries are expressed either as a Yearly sum for full time salaried position or hourly for part time or full time not-salaried positions. I don't have a strong opinion for what the default should be, but it should be clear to the user inside the UI when adding a salary.

3) There seems to be a few instances of copy pasting and not going over the copy pasted information to ensure it was correct :) I've flagged them inline. But it's a good habit to take to be thorough and review everything yourself for things like that :)

RajeebTheGreat:
@sboisvert 

Thank you for the review.

1. Since there is no milestone set, I added `TBD` in the `@since` comments. We can change that once the milestone is set?  Is that optimal? 

2. I wanted to show the job salary field close to the job location field. Other fields below that are related to the company, so it makes sense to show related field like the location close to each other.

3. Regarding the unit, I was thinking we should add a select option near the salary field. Maybe @jom has any idea on this? If we are not setting any unit, I will show the default in form fields label.

4. @jom What do you think about showing some text like "Negotiable" when the user doesn't enter the salary?

spencerfinnell:
Just chiming in to please consider this a major change and to milestone/release accordingly.

sboisvert:
@RajeebTheGreat it looks good from my perspective, I think on @jom or someone else with closer insight into the plugin might know the answer to the remaining questions

thedigitalcircle:
Hi
Can think also about min & max salary(Salary Range) for the Salary field also. If I insert range in salary field, https://search.google.com/structured-data/testing-tool/u/0/ tools only pull the min salary.
An example https://www.jobsfrontier.com/job/web-developer-trigent-software-4168/
If I test through Structured Data Testing tools,
http://prntscr.com/mp5grz
Shuvashish

onubrooks:
Closing this in favor of https://github.com/Automattic/WP-Job-Manager/pull/2263

PR #1512, opened 2018-05-29
Fix: Undefined index in class-wp-job-manager-cpt.php

albionselimaj:
The following error happens on admin screens when `post_type` isn't set in `$wp->query_vars`:
```
Notice: Undefined index: post_type in \plugins\wp-job-manager\includes\admin\class-wp-job-manager-cpt.php on line 689
```
I've added an additional check to avoid this.


jom (approved):
LGTM! Thanks for the PR.

PR #1518, opened 2018-06-07
Data Privacy: Add user meta to user extract

### LINKED ISSUE #1506: Data Privacy: Add user meta to user extract
The meta keys that we add to user profiles can be included in WP Core's new data export feature.

shantanu2704:
Fixes #1506

Following are the meta keys stored by the plugin -

1.` _company_logo`
2.` _company_name`
3. `_company_website`
4. `_company_tagline`
5. `_company_twitter`
6. `_company_video`

The data contained within these keys can now be a part of the data export requested by the user.

* Add a new class to handle the user data export, modelled after the data exporter in Core.
* Add a new test class to test the exported data

#### Testing instructions:
* Run unit tests.
* Test cases covered :
    1. All the aforementioned keys contain data.
    2. Only some of the keys contain data.
    3. An invalid email address is passed to the exporter.

 The user meta stored by the plugin can now be exported


includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,87 @@
...
+
+			foreach ( $user_meta_keys as $user_meta_key => $name ) {
+				$user_meta = get_user_meta( $user->ID, $user_meta_key, true );
+
+				if ( empty( $user_meta ) ) {
+					continue;
+				}
+
+				if ( '_company_logo' === $user_meta_key ) {
+					$user_meta = wp_get_attachment_url( $user_meta );
```
mdbitz: wp_get_attachment_url can return false. We should check for this condition and return as it isn't valid data to export.

tests/php/tests/includes/test_class.wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,177 @@
+<?php
+/**
+ * Defines a class to test the data exporter
+ *
+ * @package wp-job-manager-test
+ * @since 1.31.1
+ */
+
+require 'includes/class-wp-job-manager-data-exporter.php';
```
mdbitz: We shouldn't need this require in the text class. Instead we should have the Test Class extends WPJM_BaseTest which does some setup of the plugin.

tests/php/tests/includes/test_class.wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,177 @@
...
+							'Company Video'   => array(
+								'name'  => 'Label',
+								'value' => 'https://example.com/company/video',
+							),
+						),
+					),
+					'done' => true,
+				),
+			], // End of first set of parameters.
+			[
```
mdbitz: Can we add a third case to handle if none of the Job Manager meta data is set?

shantanu2704: Sure!

shantanu2704:
When can we merge this. @mdbitz?

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,89 @@
...
+
+			foreach ( $user_meta_keys as $user_meta_key => $name ) {
+				$user_meta = get_user_meta( $user->ID, $user_meta_key, true );
+
+				if ( empty( $user_meta ) ) {
+					continue;
+				}
+
+				if ( '_company_logo' === $user_meta_key ) {
+					if ( !$user_meta = wp_get_attachment_url( $user_meta ) ) {
```
emrikol: Could this be broken out into two separate lines for better readability?

Something like:

```
$user_meta = wp_get_attachment_url( $user_meta );
if ( false !== $user_meta ) {
```

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,89 @@
...
+						continue;
+					}
+				}
+
+				$user_data_to_export[ "$name" ] = array(
```
emrikol: Does `$name` need to be in double quotes?

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,89 @@
...
+				'data'        => $user_data_to_export,
+			);
+
+			return array(
+				'data' => $export_items,
+				'done' => true,
+			);
+		}
+	}
+}
```
emrikol: From PHPCS:
> File must end with a newline character

tests/php/tests/includes/test_class.wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,194 @@
...
+	 *
+	 * @dataProvider data_provider
+	 * @param array $args The metadata to be added.
+	 * @param array $expected The expected output.
+	 */
+	public function test_user_data_exporter( $args, $expected ) {
+		// ARRANGE.
+		$this->setupUserMeta( $args, $expected );
+		$exporter = new WP_Job_Manager_Data_Exporter();
+		if ( $id = email_exists( 'johndoe@example.com' ) ) {
```
emrikol: Could this be broken out into two separate lines for better readability?

Something like:

```
$id = email_exists( 'johndoe@example.com' );
if ( false !== $id ) {
```

tests/php/tests/includes/test_class.wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,194 @@
...
+						'group_label' => __( 'WP Job Manager User Data' ),
+						'item_id'     => '', // the item_id depends on the ID of the user.
+						'data'        => array(),
+					),
+					'done' => true,
+				),
+			), // End of third set of parameters.
+		);
+	}
+}
```
emrikol: From PHPCS:
> File must end with a newline character

shantanu2704: I don't know why GH isn't tracking it but both the files have an empty line at the end when I open them locally

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @since
+	 */
+	class WP_Job_Manager_Data_Exporter {
+		/**
+		 * Register the user data exporter method
+		 *
+		 * @param array $exporters The exporter array.
+		 * @return array $exporters The exporter array.
+		 */
+		public function register_wpjm_user_data_exporter( $exporters ) {
```
jom: This needs to be a static function as it is being hooked with a static reference.

wp-job-manager.php
```diff
@@ -112,6 +113,9 @@ public function __construct() {
 		add_action( 'wp_logout', array( $this, 'cleanup_job_posting_cookies' ) );
 		add_action( 'init', array( 'WP_Job_Manager_Email_Notifications', 'init' ) );
 
+		// Filters
+		add_filter( 'wp_privacy_personal_data_exporters', array( 'WP_Job_Manager_User_Data_Exporter', 'register_wpjm_user_data_exporter' ) );
```
jom: This class is called `WP_Job_Manager_Data_Exporter`. As mentioned earlier, this method also needs to be a static function when using this type of callable reference.

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,90 @@
...
+					}
+				}
+
+				$user_data_to_export[ $name ] = array(
+					'name'  => __( 'Label', 'wp-job-manager' ),
+					'value' => $user_meta,
+				);
+			}
+
+			$export_items = array(
```
jom: This should be an array inside an array (with one item). See core's implementation for an example:
https://github.com/WordPress/WordPress/blob/6fd8080e7ee7599b36d4528f72a8ced612130b8c/wp-includes/user.php#L2914-L2919

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -0,0 +1,90 @@
...
+
+				if ( '_company_logo' === $user_meta_key ) {
+					$user_meta = wp_get_attachment_url( $user_meta );
+					if ( false === $user_meta ) {
+						continue;
+					}
+				}
+
+				$user_data_to_export[ $name ] = array(
+					'name'  => __( 'Label', 'wp-job-manager' ),
```
jom: I'm guessing this should just be `$name` and not `__( 'Label', 'wp-job-manager' )`

includes/class-wp-job-manager-data-exporter.php
```diff
@@ -68,13 +68,13 @@ public static function user_data_exporter( $email_address ) {
...
-				$user_data_to_export[ $name ] = array(
-					'name'  => __( 'Label', 'wp-job-manager' ),
+				$user_data_to_export[] = array(
+					'name'  => $name,
 					'value' => $user_meta,
 				);
 			}
 
-			$export_items = array(
+			$export_items[] = array(
```
jom: Looks good! I think the last thing is to set up this array (`$export_items = array();`) before adding an item to it.

shantanu2704: So always declare variables before using them?

jom:
Looks great! Thanks, @shantanu2704!

shantanu2704:
Thank you, sir!

PR #1533, opened 2018-06-27
Add Webpack/Eslinting for Gutenberg blocks

dbtlr:
Pulled from #1521 

This PR adds in all of the webpack building and eslinting required to build a Gutenberg block.

## The Build

- The build is pretty standard webpack/babel, with some things borrowed from the Gutenberg build.

- I've included it in our `grunt`-based existing build system.

- When we start building the other blocks, there are a few things we may need to consider:
  - Sharing components between blocks.
  - Splitting bundles (for both synchronous and asynchronous loading).


.eslintrc.js
```diff
@@ -0,0 +1,176 @@
...
+	parserOptions: {
+		sourceType: 'module',
+		ecmaFeatures: {
+			jsx: true,
+		},
+	},
+	globals: {
+		wp: true,
+		window: true,
+        document: true,
```
alexsanford: Funky indentation here

includes/class-wp-job-manager-blocks.php
```diff
@@ -0,0 +1,65 @@
...
+			return;
+		}
+
+		add_action( 'init', array( $this, 'register_blocks' ) );
+	}
+
+	/**
+	 * Register all Gutenblocks
+	 */
+	public function register_blocks() {
```
alexsanford: This is likely to fail since the jobs block is gone. Should we just stub this function out?

dbtlr: removed!

spencerfinnell:
WordPress' Babel preset package will handle Babel config: https://github.com/WordPress/packages/tree/master/packages/babel-preset-default

PR #1536, opened 2018-07-02
Cleans up build scripts and updates path of built blocks

jom:
This PR cleans up the files that are packaged in the released plugin and also moves the built block directory to `build/blocks`. Ideally, we'll move all built assets to `build/[css|js|blocks|...]` and it'd be nice to have it be ready for that.

To test build, run `grunt build` and see the `tmp/build` directory.


dbtlr (approved):
LGTM

PR #1537, opened 2018-07-02
Setup Jest as Javascript test runner

donnapep:


jest.config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "preset": "@wordpress/jest-preset-default",
+  "testMatch": [
+    "**/tests/**/*.js?(x)"
```
jom: Would it speed it up at all to limit this to the assets directory?

donnapep: It didn't seem to make a difference, but if we eventually want to write tests for the other JS files, it would seem to make sense to have those tests live in `assets/` as well. Updated in https://github.com/Automattic/WP-Job-Manager/pull/1537/commits/7cc172a628a5ddc65eb68519d1d39406c3f867af.

PR #1561, opened 2018-07-27
Allow users to relist expired jobs in bulk

### LINKED ISSUE #1555: Add bulk relist option to job listings in wp-admin.
There's an "Expire Jobs" option on Job Listings > All Jobs in wp-admin, but no option to relist in the bulk actions. Would this be possible to add?

1289546-zen

AmandaJBell:
Fixes #1555

*Allow users to relist expired entries in bulk from wp-admin. If a user relist's an entry it is set to published, and if the expiry date of that entry has passed it has no new expiry date (however, I can alter these as needed).

#### Testing instructions:

* Open the Job Listings in wp-admin
* Add two jobs, giving one an 'expired' status, and one a 'preview' status (or any status that is not 'Active' or 'Expired')
* Select both jobs in the 'All Jobs' subsection
* Select 'Relist Expired Jobs' from the 'Bulk Actions' dropdown, and hit 'Apply'
* The job that was set to expired should now be 'Active' and the job that was set to 'Preview' should still be inactive.

Add bulk relist option to job listings from wp-admin

### DIFF
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -208,6 +208,13 @@ class WP_Job_Manager_CPT {
 		}
 		return false;
 	}
+	
+	/**
+	 * Performs bulk action to relist a single expired job.
+	 *
+	 * @param int $post_id Post ID.
+	 * @return bool
+	 */
 	public function bulk_action_handle_relist_job( $post_id ) {
 		$job_data = array(
 			'ID'          => $post_id,
@@ -221,6 +228,7 @@ class WP_Job_Manager_CPT {
 		}
 		return false;
 	}
+
 	/**
 	 * Performs bulk action to mark a single job listing as filled.
 	 *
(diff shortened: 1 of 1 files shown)


AmandaJBell:
Please let me know if I missed anything or if I should make any changes. This is my first contribution so any feedback is appreciated!

Confido2018:
Thank you verry much for your hard work. 
Only, I have one issue left. When the jobs are relisted, the publish date stays the same as before. Problem is, jobseekers think it's an old job and not vacant anymore. Is it possible to change the publish date to the refresh date (when refresh in the bulk relist option? Thank you verry much in advance.  
Kind regards, Fabian

alexsanford (commented):
When I relist an expired job from the edit page, it is assigned a new expiry date. I'm thinking we should do the same for the bulk relist?

jom:
Closing for now, but let us know when it is ready for review again.

PR #1569, opened 2018-08-05
Preselect job types by slug in query string on `[jobs]` page

### LINKED ISSUE #1474: Preselect job types by slug in query string on `[jobs]` page.
Similar to categories (`http://example.com/jobs/?search_category=19`), we should allow for pre-selection of job types by query string (`http://example.com/jobs/?search_job_type=internship`).

https://github.com/Automattic/wp-job-manager/blob/72935f2ef628a1157781cf310cba04d5535b2fe8/includes/class-wp-job-manager-shortcodes.php#L304-L306

Temporary work around:
```php
add_filter( 'job_manager_output_jobs_defaults', function( $defaults ) {
	if ( ! empty( $_GET['search_job_type'] ) ) {
		$defaults['selected_job_types'] = sanitize_text_field( $_GET['search_job_type'] );
	}
	return $defaults;
} );
```

felipeelia:
Fixes #1474 

* Allow job types pre-selection by query string, similar to what is already possible with categories.

#### Testing instructions:

* Access the page with `[jobs]` shortcode
* Append `?search_job_type=<types-slugs-separated-by-comma>`

Allow job types pre-selection by query string

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -330,13 +330,7 @@ class WP_Job_Manager_Shortcodes {
 			$atts['filled'] = ( is_bool( $atts['filled'] ) && $atts['filled'] ) || in_array( $atts['filled'], array( 1, '1', 'true', 'yes' ), true );
 		}
 
-		// Array handling.
-		$atts['categories']         = is_array( $atts['categories'] ) ? $atts['categories'] : array_filter( array_map( 'trim', explode( ',', $atts['categories'] ) ) );
-		$atts['job_types']          = is_array( $atts['job_types'] ) ? $atts['job_types'] : array_filter( array_map( 'trim', explode( ',', $atts['job_types'] ) ) );
-		$atts['post_status']        = is_array( $atts['post_status'] ) ? $atts['post_status'] : array_filter( array_map( 'trim', explode( ',', $atts['post_status'] ) ) );
-		$atts['selected_job_types'] = is_array( $atts['selected_job_types'] ) ? $atts['selected_job_types'] : array_filter( array_map( 'trim', explode( ',', $atts['selected_job_types'] ) ) );
-
-		// Get keywords and location from querystring if set.
+		// Get keywords, location, category and type from querystring if set.
 		if ( ! empty( $_GET['search_keywords'] ) ) {
 			$atts['keywords'] = sanitize_text_field( $_GET['search_keywords'] );
 		}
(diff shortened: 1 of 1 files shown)


DanielNetzer:
Trying to implement this seems to fail and not filter the jobs according to the job type passed.

felipeelia:
Hi @DanielNetzer!

Both approaches work in a WordPress fresh install, with 2017 theme, but in different ways. Using `selected_job_types` I'm able to see the filters. Using `job_types` I can see the same results, but the filters aren't visible.

With `selected_job_types`:

image

With `job_types`:

image

I'm not sure, but this can be a config difference or a theme specificity. What theme are you using?

DanielNetzer:
I'm using Divi, but I'm not using the regular filters. I've implemented my own search search form.

felipeelia:
> I've implemented my own search search form.
In that case, you can use the `job_manager_output_jobs_defaults` filter, as suggested by @jom in the [original issue](https://github.com/Automattic/WP-Job-Manager/issues/1474). Anyway, let's see how the plugin folks prefer to handle cases like yours.

About _WP Jobs Manager Predefined Regions_, I'm assuming that you're talking about the Astoundify's one, am I right? If so, it seems that they based it on taxonomies, so you can use core [get_terms()](https://developer.wordpress.org/reference/functions/get_terms/) to retrieve the list you want. Something like 

```
$regions = get_terms( array(
    'taxonomy' => 'job_listing_region',
    'hide_empty' => false,
) );
```

should do it. ([Here](https://github.com/Astoundify/wp-job-manager-regions/blob/master/includes/class-taxonomy.php) is where the taxonomies are declared.)

DanielNetzer:
looks good, and you are right its the WP Jobs Manager Predefined Regions. thank you for the help. i've did something similar to what jon did.

I've created a shortcode for the entire form,

`/**
	 * Search job form.
	 */
	public function search_job() {

		$job_types_options = '';
		$job_category_options = '';
		$regions = get_terms( array(
			'taxonomy' => 'job_listing_region',
			'hide_empty' => false,
		) );
		
		echo `<script>console.log('` . json_encode( $regions ) . `');</script>`;

		foreach (get_job_listing_types() as $type) {
			$job_types_options .= '<option value="' . esc_attr( $type->slug ) . '">' . esc_html( $type->name ) . '</option>';
		}

		foreach (get_job_listing_categories() as $cat) {
			$job_category_options .= '<option value="' . esc_attr( $cat->term_id ) . '">' . esc_html( $cat->name ) . '</option>';
		}

		return '<form class="wp_job_manager_search_job" method="GET" action="https://bee-adv.co.il/nelian/home">
						<select class="input" id="search_category" name="search_category">
						<option disabled selected value> תחום </option>'
						. $job_category_options  .
						'</select>
						<select class="input" id="search_job_type" name="search_job_type">
						<option disabled selected value> סוג משרה </option>'
						. $job_types_options  .
						'</select>
						<input class="input" placeholder="איזור" type="text" id="search_location" name="search_location" />
						<input class="et_pb_searchsubmit" type="submit" value="חפש עבודה" />
					</form>
				</div>';
	}`

jom (approved):
Code looks great! Thank you.

PR #1570, opened 2018-08-05
Hide filled and expired positions from native WP search

### LINKED ISSUE #833: Filled and hidden positions displayed in WordPress native search
Hi,

I found a bug in the plugin. When you setup "WP Job Manager" and set "Hide filled positions" and "Hide content within expired listings" the positions are still displayed in the WordPress native search.

I expected filled and expired position to be hidden from normal WordPress search. When the setup says so. 

#### What Happened Instead
When searching I found that all of the expired and filled positions are displayed int the native search of WordPress.

1. Install Wordpress
2. Install WP Job Manager
3. Setup WP Job Manager (enable: "Hide filled positions" and "Hide content within expired listings")
4. Create 2 Jobs
5. Expire that job first one and mark the second one as filled
6. Do search with WordPress own search function using the name of the expired job or the filled job
7. Now the expired/filled position should be displayed in the search results

felipeelia:
Fixes #833

Configurations for Hiding filled or expired positions aren't respected in core search.

This PR introduces a method (`WP_Job_Manager_Post_Types::public_search_handler()`) to insert the necessary `meta_query` parameter to deal with filled positions.

To solve expired positions, currently, a bit hacky solution is needed. Although the `expired` post status is created with the `exclude_from_search` set to true, WordPress will ignore it if the WP_Query `post_status` parameter is empty, using only the post status `public` value. That way, this PR suggests setting `'public' => ! isset( $_GET['s'] )`.

I've already reported the problem with a [ticket in core](https://core.trac.wordpress.org/ticket/44737).

#### Testing instructions:

1. Install Wordpress
1. Install WP Job Manager
1. Setup WP Job Manager (enable: "Hide filled positions" and "Hide content within expired listings")
1. Create 2 Jobs
1. Expire that job first one and mark the second one as filled
1. Do search with WordPress own search function using the name of the expired job or the filled job

Hide filled and expired positions from native WP search

p.s.: Please ignore 3ed0288 and 1bd7465 commits. I've messed up my files with another branch :grimacing:

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -66,6 +66,7 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'update_post_meta', array( $this, 'update_post_meta' ), 10, 4 );
 		add_action( 'wp_insert_post', array( $this, 'maybe_add_default_meta_data' ), 10, 2 );
 
+		add_action( 'parse_query', array( $this, 'public_search_handler' ) );
 		add_action( 'parse_query', array( $this, 'add_feed_query_args' ) );
 
 		// Single job content.
@@ -301,7 +302,7 @@ class WP_Job_Manager_Post_Types {
 			'expired',
 			array(
 				'label'                     => _x( 'Expired', 'post status', 'wp-job-manager' ),
-				'public'                    => true,
+				'public'                    => ! isset( $_GET['s'] ),
 				'protected'                 => true,
 				'exclude_from_search'       => true,
 				'show_in_admin_all_list'    => true,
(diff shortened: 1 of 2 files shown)


includes/class-wp-job-manager-post-types.php
```diff
@@ -461,6 +462,35 @@ public function job_feed() {
...
+		}
+
+		// Remove filled positions, if necessary.
+		if ( 1 === absint( get_option( 'job_manager_hide_filled_positions' ) ) ) {
+			$meta_query = $query->get( 'meta_query' );
+			if ( ! is_array( $meta_query ) ) {
+				$meta_query = array();
+			}
+
+			$meta_query[] = [
```
jom: WPJM currently requires compatibility with PHP 5.2. Short-array syntax is currently not available.

felipeelia: Oh, you're right, sorry @jom. I've made a commit with the necessary change.

includes/class-wp-job-manager-post-types.php
```diff
@@ -301,7 +302,7 @@ public function register_post_types() {
 			'expired',
 			array(
 				'label'                     => _x( 'Expired', 'post status', 'wp-job-manager' ),
-				'public'                    => true,
+				'public'                    => ! isset( $_GET['s'] ),
```
jom: Would you mind adding a comment to your core trac issue above this line?
Something like:
```
// For background and to track issue, see https://core.trac.wordpress.org/ticket/44737.
```

jom (approved):
Code looks good! Thanks for this PR and submitting that ticket to core WP. 

I was wondering if we should check the `job_manager_hide_expired` before doing the `public` attribute hack or setting `exclude_from_search = true`, but maybe I'll do that in a separate PR after this is merged.

PR #1616, opened 2018-11-18
Add/display job category

### LINKED ISSUE #799: Display job category on default single job template
Outside of the jobs shortcode, we don't use the category anywhere. Should we add it to the single job template?

0xDELS:
Fixes #799 

Added a setting in the Job Listings to be able to turn on or off showing of the job category
Made a few changes on the content-single-job_listing-company.php and wp-job-manager-template.php to allow the categories to show on the frontend.

#### Testing instructions:
On the Job Listing Settings turn on the Categories 
On the Job Listing Settings turn on the Display Job Category
Add a job with a category or multiple, it will appear on the Individual Job Page

#### Proposed changelog entry for your changes: Added ability to display job categories

NOTE: I want to improve this by allowing the categories to link to other jobs from the same category.

screenshot 2018-11-18 at 17 25 13
screenshot 2018-11-18 at 17 25 28
screenshot 2018-11-18 at 17 27 04


0xDELS:
Hi,

I was getting some ERRORS with indentation and blank lines on the version 7.2.

Ran PHPCBF and fixed the lines on the comments that didn't have a space in between. In the version 7.2 is ok now, however, it failed on the 5.2. I think something went wrong with Travis.

alexsanford (commented):
Thanks for submitting this! I left some code comments below.

In the interest of preventing clutter, I'm not sure we should add a setting for this. I think just having the `job_manager_show_job_category` filter default to `false` and allowing users to enable the functionality through that filter.

If we do keep the setting, it should be moved up to be with the other Category settings ("Enable listing categories", etc).

We should also update the template version number since we're making a change to it.

wp-job-manager-functions.php
```diff
@@ -950,6 +951,13 @@ function job_manager_multi_job_type() {
 	return apply_filters( 'job_manager_multi_job_type', 1 === intval( get_option( 'job_manager_multi_job_type' ) ) );
 }
 
+/**
+ *  Checks if the job listing should display the category
```
pgk: Maybe add a `@return bool` here?

wp-job-manager-functions.php
```diff
@@ -950,6 +951,13 @@ function job_manager_multi_job_type() {
...
+ */
+function job_manager_show_job_category() {
+	return apply_filters( 'job_manager_show_job_category', 1 === intval( get_option( 'job_manager_show_job_category' ) ) );
```
pgk: cast to `bool`?

pgk (commented):
Couple more comments

gikaragia:
Resolving stale PR

PR #1621, opened 2018-11-21
Add support for remote positions

### LINKED ISSUE #918: Add support for remote job listings
Listings are currently tied to a single location. If no location is specified then the listings don't show up in a search.

Would be good to add support for 'location independent' job listings.

3135358-t

0xDELS:
Fixes #918 

When Adding a new Job Listing, now exists the ability to consider that as a Remote Position
If its Tagged as a Remote Position the Google Job Search Schema will be updated accordingly
Search for Remote Positions is available via the Location field.

#### Testing instructions:
Install WordPress and WP Job Manager
To add a Remote Position just check the Remote Position checkbox when adding a new listing
To search for remote positions, type "remote" or "Remote" in the Location Field

#### Proposed changelog entry for your changes: Add Support for Remote Positions

A CSS update will be needed when is a Remote Position to get it with the same look as a "normal" position, but i want to improve this at the end.

Any feedback to improve this PR is very much appreciated. 

Screenshots here
screenshot 2018-11-21 at 14 37 59
screenshot 2018-11-21 at 14 38 36
screenshot 2018-11-21 at 14 38 50


wp-job-manager-functions.php
```diff
@@ -66,7 +66,10 @@ function get_job_listings( $args = array() ) {
 		}
 
 		if ( ! empty( $args['search_location'] ) ) {
-			$location_meta_keys = array( 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' );
+			if ( 'remote' === $args['search_location'] || 'Remote' === $args['search_location'] ) {
```
jom, line 69: If we were to use this natural language search, we'd probably want to let these be translatable.

jom (commented):
Sorry for the late review. This needs a rebase. I wonder if Location and Remote/Telecommute need to be split a bit. For example, some businesses have an office location but have a work from home option (especially now). Others might be distributed but only allow employees in their country.

wp-job-manager-functions.php
masteradhoc: @0xDELS or @jom can u fix this and send it to the next update? This is still needed highly due to covid.

Hastibe:
A lot of the remote jobs that get posted on our job listings are "Remote in X" (X usually being a geographic region, such as the state or country), usually they are posted as "Anywhere in X" (as "Remote in X" sounds strange). Is there (or could there be) a way to accommodate these types of remote jobs with this code?

JEGardner:
For an example of how this could behave to support @Hastibe's  request, Job Board Fire do this really well, supporting multiple locations and multiple remote locations per job. The user is then able to search by location and remote location in one field:

girlboss location search

(screenshot taken from [jobs.girlboss.com](https://jobs.girlboss.com))

MikeSEONuts:
We are looking for this feature to be added to the plugin as it has become really important for SEO - has there been an update on this or what the conflicts were/are?

wp-job-manager-functions.php
onubrooks: > We are looking for this feature to be added to the plugin as it has become really important for SEO - has there been an update on this or what the conflicts were/are?

An update for this is in the works.

wp-job-manager-template.php
```diff
@@ -410,6 +410,11 @@ function wpjm_get_job_listing_structured_data( $post = null ) {
 		$data['jobLocation']            = [];
 		$data['jobLocation']['@type']   = 'Place';
 		$data['jobLocation']['address'] = wpjm_get_job_listing_location_structured_data( $post );
+		if ( $post->_remote_position ) {
+			$data['additionalProperty']          = array();
+			$data['additionalProperty']['@type'] = 'PropertyValue';
+			$data['additionalProperty']['value'] = 'TELECOMMUTE';
```
dadish, line 416: I think `jobLocationProperty` suits better here. Can we update this part to use that instead?
```suggestion
			$data['jobLocationType'] = 'TELECOMMUTE';
```

wp-job-manager-template.php
```diff
@@ -814,7 +823,11 @@ function get_the_job_location( $post = null ) {
 		return null;
 	}
 
-	return apply_filters( 'the_job_location', $post->_job_location, $post );
+	if ( $post->_remote_position ) {
+		return apply_filters( 'the_job_location', $post->_remote_position, $post );
+	} else {
+		return apply_filters( 'the_job_location', $post->_job_location, $post );
+	}
```
dadish, line 830: I am not sure we need this part. I think `job_location` and `remote_position` should be separated.

Is this covering some edge case? Why this was added?

wp-job-manager-functions.php
```diff
@@ -74,8 +74,12 @@ function get_job_listings( $args = [] ) {
 		}
 
 		if ( ! empty( $args['search_location'] ) ) {
-			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
+			if ( __( 'remote', 'wp-job-manager' ) === $args['search_location'] || __( 'Remote', 'wp-job-manager' ) === $args['search_location'] ) {
```
dadish, line 77: We could probably convert to lowercase before comparing. Less things to translate and handles all the case combinations.

```suggestion
			if ( strtolower( __( 'remote', 'wp-job-manager' ) ) === strtolower( $args['search_location'] ) ) {
```

Screen Shot 2022-06-13 at 14 18 51

wp-job-manager-template.php
```diff
@@ -782,13 +787,17 @@ function get_the_job_publish_date( $post = null ) {
...
+		return;
+	}
 	if ( $location ) {
 		if ( $map_link ) {
 			// If linking to google maps, we don't want anything but text here.
 			echo wp_kses_post(
 				apply_filters(
 					'the_job_location_map_link',
-					'<a class="google_map_link" href="' . esc_url( 'https://maps.google.com/maps?q=' . rawurlencode( wp_strip_all_tags( $location ) ) . '&zoom=14&size=512x512&maptype=roadmap&sensor=false' ) . '">' . esc_html( wp_strip_all_tags( $location ) ) . '</a>',
+					'<a class="google_map_link" href="' . esc_url( 'https://maps.google.com/maps?q=' . rawurlencode( wp_strip_all_tags( $location ) ) . '&zoom=14&size=512x512&maptype=roadmap&sensor=false' ) . '" target="_blank">' . esc_html( wp_strip_all_tags( $location ) ) . '</a>',
```
dadish, line 800: Nice! 👍

dadish (changes requested):

wp-job-manager-template.php
onubrooks: Nice

wp-job-manager-template.php
onubrooks: True we probably don't. I'll revert the change since @0xDELS hasn't responded in a while. What do you think about [this](https://github.com/Automattic/WP-Job-Manager/pull/1621#pullrequestreview-564261384) comment with respect to search which right now is either 'remote' or 'location' but not both?

dadish: @onubrooks Yes! Now that we have changes from the #2261 we can add support for multiple search locations and remote/non-remote listings

I'll push the implementation soon. 👍

dadish: @onubrooks I pushed some updates. Now it should allow you to search multiple locations at once and also search for remote locations as well. It is a text based UI. So when you search for job located in London, you search `London`. If you search for jobs that are remote in London, then you search `london remote`.


https://user-images.githubusercontent.com/2578542/173572033-3ba12bdc-c680-41f9-a1d4-2c89acdd382a.mp4

Let me know what you think.

onubrooks: This works for me!

PR #1623, opened 2018-11-23
"Load previous listings" bug fix #1604

### LINKED ISSUE #1604: "Load previous listings" appears when it shouldn't
#### Steps to Reproduce
1. Make sure you have more jobs that you are showing on the first page of job listings, so that a "Load More Listings" button or link appears
2. Click it to load more listings
3. Refresh the page

(Note: this was with 2017 theme and only WPJM active; it also happens on the Demo site with Indeed jobs listings)

#### What Happened Instead

A non-clickable line of text, "Load previous listings", shows up before the job listings but after the filters. I see no JS errors in console

4.9.8, 1.31.3

#### Browser / OS Version

Chrome, OSX

#### Screenshot / Video

screen shot 2018-10-23 at 14 40 00

#### Source

Originally reported here: https://wordpress.org/support/topic/load-previous-listings-not-working/

tonytettinger:
Fixes #1604 

As stated in the issue description the text "Load previous listings" incorrectly (unnecessarily) appears on the page when we click on the "Load More Listings" then refresh the Job Listings Page.

This is due to a few lines which are not necessary in the code, although I would suggest to open another ticket, which would be an enhancement to create a "Load Less Listings" link. Removing these lines should not affect any functionality of the page, except resulting in the expected page output.

With the current suggested changes the page loads as expected. (tested with 0 jobs, and jobs>jobs list per page). 
The load spinner shows, jobs are displayed correctly and stay displayed the same way on refresh.

Solution suggestion: Remove lines 30-44

Explanation of line removals (also added as code comments):

- Line 31: It's not necessary to add the 'load' class because the onClick handler adds this class (line 207 in the ajax-filter.js file)
- Line 35-40: These lines are responsible for outputting and handling the text that was requested to be removed from the page (also the set of matched elements is not well defined in the .wrap() jQuery element, so it is not functional)
- Line 42: This line modifies the output data unnecessarily.

#### Testing instructions:

As in the issue description:

1. Make sure you have more jobs that you are showing on the first page of job listings, so that a "Load More Listings" button or link appears
2. Click it to load more listings
3. Refresh the page

Page is expected to keep displaying the expanded job list without any incorrect text appearing.


assets/js/ajax-filters.js
```diff
@@ -27,21 +27,6 @@ jQuery( document ).ready( function ( $ ) {
 			xhr[index].abort();
 		}
 
-		if ( ! append ) {
```
tonytettinger: Line 31: It's not necessary to add the 'load' class because the onClick handler adds and handles this class on line 207.

assets/js/ajax-filters.js
```diff
@@ -27,21 +27,6 @@ jQuery( document ).ready( function ( $ ) {
...
-			$( results ).addClass( 'loading' );
-			$( 'li.job_listing, li.no_job_listings_found', results ).css( 'visibility', 'hidden' );
-
-			// Not appending. If page > 1, we should show a load previous button so the user can get to earlier-page listings if needed
-			if ( page > 1 && true !== target.data( 'show_pagination' ) ) {
```
tonytettinger: These lines are responsible for outputting and handling the text that was requested to be removed from the page (also the set of matched elements is not well defined in the .wrap() jQuery element, so it is not functional)

assets/js/ajax-filters.js
```diff
@@ -27,21 +27,6 @@ jQuery( document ).ready( function ( $ ) {
 			xhr[index].abort();
 		}
 
-		if ( ! append ) {
-			$( results ).addClass( 'loading' );
-			$( 'li.job_listing, li.no_job_listings_found', results ).css( 'visibility', 'hidden' );
-
-			// Not appending. If page > 1, we should show a load previous button so the user can get to earlier-page listings if needed
```
tonytettinger: These lines are responsible for outputting and handling the text that was requested to be removed from the page (also the set of matched elements is not well defined in the .wrap() jQuery element, so it is not functional).

assets/js/ajax-filters.js
```diff
@@ -27,21 +27,6 @@ jQuery( document ).ready( function ( $ ) {
...
-			if ( page > 1 && true !== target.data( 'show_pagination' ) ) {
-				var previous = jQuery('<strong>').text(job_manager_ajax_filters.i18n_load_prev_listings).wrap('<a class="load_more_jobs load_previous" href="#"></a>');
-				$( results ).before( previous );
-			} else {
-				target.find( '.load_previous' ).remove();
-			}
-
-			target.find( '.load_more_jobs' ).data( 'page', page );
```
tonytettinger: This changes the data to display on the page is not necessary.

tonytettinger: This changes the data to display on the page is not necessary.

assets/js/ajax-filters.js
```diff
@@ -27,21 +27,6 @@ jQuery( document ).ready( function ( $ ) {
 			xhr[index].abort();
 		}
 
-		if ( ! append ) {
-			$( results ).addClass( 'loading' );
```
tonytettinger: It's not necessary to add the 'load' class because the onClick handler adds and handles this class on line 207.

assets/js/ajax-filters.js
tonytettinger: These lines are responsible for outputting and handling the text that was requested to be removed from the page (also the set of matched elements is not well defined in the .wrap() jQuery element, so it is not functional).

tonytettinger:
You are welcome, thank you for the review, glad I could help!

PR #1632, opened 2018-12-05
Force classic block on job listings

### LINKED ISSUE #1626: Resolve incompatibility with block editor and job description rich text
When we implement #1625 the block editor will be enabled (by default) for `job_listing` CPT. Generally, the meta boxes seem compatible with WordPress 5.0 and the new block editor. There is a  problem with job listings that are edited with a mix of the frontend (job dashboard shortcode) and WP admin. For job listings created or edited in WP Admin from the block editor, blocks could get messed up when edited from the standard `wp_editor()` from the job dashboard. In A Dream World, we'd have a mini frontenberg block editor for the job dashboard editor. What complicates this is already complicated functionality/hack is the guest posting functionality of WPJM.

For now, we need something that helps us immediately mitigates this issue. Possible solutions:
- Prevent block editor from being used for job listings CPT.
- Prevent jobs with non-classic blocks from being edited on the job dashboard.
- See if it is possible to just allow the classic block in the block editor.

jom:
Fixes #1626

* Forces the classic block on job listings. This is due to potential conflicts between frontend rich text editor and block editor.

#### Testing instructions:

* Make sure job listings edited in classic block appear the same as when they are edited with the classic editor.

#### Known Issues
* Due to a bug in the block editor (TODO: Find issue on Gutenberg 😉), the classic block context menu still allows for converting classic to blocks. This shouldn't happen because no other blocks are allowed. When they go back into the job listing editor things explode because it doesn't follow the template.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -34,6 +34,7 @@ class WP_Job_Manager_Post_Types {
 	 */
 	public function __construct() {
 		add_action( 'init', array( $this, 'register_post_types' ), 0 );
+		add_action( 'init', array( $this, 'prepare_block_editor' ) );
 		add_filter( 'admin_head', array( $this, 'admin_head' ) );
 		add_action( 'job_manager_check_for_expired_jobs', array( $this, 'check_for_expired_jobs' ) );
 		add_action( 'job_manager_delete_old_previews', array( $this, 'delete_old_previews' ) );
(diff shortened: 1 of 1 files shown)


spencerfinnell:
This properly captures changes made via `wp_editor()` on the frontend?

jom:
@spencerfinnell It appears when there is just a classic block on a page, no block editor artifacts are put into `post_content`. Definitely worth additional testing.

Also, just added a known issue for an annoying exception.

spencerfinnell:
Awesome. There is definitely a Gutenberg issue open somewhere about converting to non-existing blocks breaking everything; but I can't find it either.

includes/class-wp-job-manager-post-types.php
```diff
@@ -73,6 +74,28 @@ public function __construct() {
...
+	/**
+	 * Forces job listings to just have the classic block. This is necessary with the use of the classic editor on
+	 * the frontend.
+	 *
+	 * @param array   $allowed_block_types
+	 * @param WP_Post $post
+	 * @return array
+	 */
+	public function force_classic_block( $allowed_block_types, $post ) {
+		if ( 'job_listing' === $post->post_type ) {
```
donnapep: I don't quite get why this is needed when the template is locked on the CPT. It seems to work just fine without the filter. Is there a "before" use case I can test to understand the impact of the filter?

jom: This is a bit of a future proof depending on how they fix the issue with converting classic block to blocks.

donnapep (approved):
LG!

includes/class-wp-job-manager-post-types.php
spencerfinnell: I like this.

PR #1636, opened 2018-12-07
Introduce WP admin notices and refactor for setup wizard

### LINKED ISSUE #1181: Setup Wizard redirect should not be triggered while DOING_AJAX
I'm working on a WordPress plugin bundler which activates a list of plugins with a few AJAX requests and I have a problem with the fact that WP Job Manager tries to redirect at the setup wizard even if it is an AJAX request.

If I queue 2 ajax requests (one activates WP Job Manager and the second one WooCommerce) they should both work without any redirect since the user won't interact with the actual page.

This [redirect](https://github.com/Automattic/WP-Job-Manager/blob/2e9bfa47bdcb1ab439bc41b38d65e74b726b9d14/includes/admin/class-wp-job-manager-setup.php#L86) makes sense only if the user will interact with the Setup Wizard, but in this case, the HTML will be just a response into an ajax success callback like this:
screen shot 2017-09-20 at 11 01 05

Maybe avoid that redirect when DOING_AJAX is true?

jom:
Fixes #1181 (in a very long winded way 😃)

* Primarily: Introduces admin notices helper for introducing notices in admin dashboard. 
* Removes the old way of redirecting to setup page on first activation. I never liked this as it made it sometimes made it difficult to access for bulk activations or if something interrupts activation. 
* Refactors setup class by:
  * Moving HTML to views.
  * Moving processing logic out of `output()` method.

#### Testing instructions:

* In an existing WPJM instance, switch to this branch. Deactivate and then reactivate the plugin. Nothing should happen and you shouldn't be prompted to set up WPJM.
* I a new instance, activate the plugin for the first time. You shouldn't be redirected to the setup wizard but on the dashboard, plugin, and job listing pages you should have a notice to set up WPJM.

#### To do
- [x] Add tests for `WP_Job_Manager_Admin_Notices` class.


includes/admin/views/html-admin-notice-core-setup.php
```diff
@@ -0,0 +1,23 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+?>
+<div class="updated wpjm-message">
+	<p>
+		<?php
+		echo wp_kses_post( __( 'You are nearly ready to start listing jobs with <strong>WP Job Manager</strong>.', 'wp-job-manager' ) );
```
jom: Definitely open to feedback on this thrown-together prompt 😉

spencerfinnell:
All I'll ask for is an easy way to disable this. :)

spencerfinnell:
Remove this action would work, but a filter would be nice.

jom:
@spencerfinnell Which part are you hoping to disable? Did you prefer the old redirect-on-activation or were you entirely disabling the setup wizard?

jom:
@spencerfinnell Where in the flow do you think would work best?

If it is queued up for a user to see, you should be able to prevent it with:
```php
add_action( 'job_manager_init_admin_notices', function() {
   remove_action( 'job_manager_admin_notice_core_setup', array( 'WP_Job_Manager_Admin_Notices', 'display_core_setup' );
}, 11 );
```

I suppose you can also remove the action that calls `init_core_notices` to prevent any future core notices, too. 

We could also add a filter to short-circuit adding the notice to begin with in the `add_notice` method, but then if someone activated their theme after the plugin, it would still show.

spencerfinnell:
@jom yeah it’s basically between that code and a filter. Either is and and depends on the order of activation. 

job_manager_show_core_setup filter (on by default) to kill any hooks added would be my vote.

jom:
@spencerfinnell I added `job_manager_show_admin_notice_{$notice}` 😄 

This should suppress it:
```php
add_filter( 'job_manager_show_admin_notice_core_setup', '__return_false' ); 
```

alexsanford (approved):
This works well! The setup wizard seems to work like it did before, and I like the notice.

Do you think that just the dashboard, plugin, and job listings pages is enough for the setup notice? I wonder if it should show up on more WPJM pages...have you looked at what WooCommerce does with their notice?

Not a blocker for me though. Good job on this! 🙂

jom:
@alexsanford I added it (95913136fd7a2c3565d3e360f2104049f6c23950) to the taxonomies and add-ons page. We currently suppress notices on Settings. I could enable it, but it looks like WP would place them just below the tabs (looks a bit funny).

Were there other places you had in mind?

alexsanford:
> Were there other places you had in mind?

Not really. I had thought of settings, but it definitely looks weird to have it below the tabs. That said, WooCommerce seems to do it this way - http://cld.wthms.co/QrnoJ6

I don't think it's a big deal, but I wonder if it makes sense to follow the WooCommerce approach. I just tested it, and they seem to enable the notice on every WooCommerce-related page (including settings) plus the Dashboard and Plugins pages. Not saying we need to do it that way, but it does make it a bit more persistent, and I'm not sure it makes a lot of sense to have it displayed on some WPJM pages but not others.

Thoughts?

alexsanford:
LGTM! 🎉

PR #1642, opened 2018-12-11
Allow job listing archive slug to be customized

### LINKED ISSUE #1486: When themes support archive page, allow for custom archive page slug
Right now, when a theme has support for `job-manager-templates` the archive page slug defaults to `jobs`, which often conflicts with the automatically created `[jobs]` page. 

We should allow people to customize this slug in Permalink settings (when theme support is enabled) and have it default to something that won't conflict (like `job-listings`).

This will require an upgrade rule. All old instances that have set something for the post type's `has_archive` when declaring the post type (either by theme support or filter), we'll want to have it default to whatever they've set. Otherwise, when people enable theme support, we should default it to the non-conflicting slug.

jom:
Fixes #1486

* Rename the `wpjm_permalinks` option to `job_manager_permalinks` for consistency.
* Along with the rename, I also shifted it to store the settings in JSON. JSON encoding/decoding is a bit faster and more secure.
* Add option for people to change Job Listings archive page from Permalink Settings _if_ their theme declares support for `job-manager-templates`. 
* Before this customization was allowed, the default would be `/jobs`. This often conflicted with the standard permalink for the `[jobs]` shortcode. For _new_ installations, the default will be `/job-listings`.

#### Testing instructions:

Three scenarios should be tested:
* New install with theme support: 
  * On a fresh instance with the active theme declaring support (`add_theme_support( 'job-manager-templates' );`) , activate job manager (this branch). 
  * Go to WP Admin > Settings > Permalinks.
  * Ensure the `Job listing archive page` option shows up with a default for `job-listings`.
  * Change setting to another valid slug (e.g. `fancy-job-listings`). 
  * Save settings. 
  * Visit `http://example.com/fancy-job-listings` and verify it works and is an archive of published job listings.
* New install without theme support: 
  * On a fresh instance without the active theme declaring support, activate job manager (this branch). 
  * Go to WP Admin > Settings > Permalinks.
  * Ensure the `Job listing archive page` option does not show up.
  * In MySQL, verify `job_manager_permalinks` option in `{prefix_}options` contains `"jobs_archive":""`.
  * Both `http://example.com/job-listings` and `http://example.com/jobs` should not be the jobs archive page.
* Old install with theme support: 
  * On an instance that already has job manager with the active theme declaring support (`add_theme_support( 'job-manager-templates' );`), make sure the plugin is on `master` or the last release.
  * Go to WP Admin > Settings > Permalinks; save permalinks as-is at the bottom.
  * In MySQL `{prefix_}options`, verify `job_manager_permalinks` option doesn't exist.
  * Switch to this branch.
  * Refresh the permalink settings page.
  * In MySQL `{prefix_}options`, verify `job_manager_permalinks` option exists that is the current representation of the old `wpjm_permalinks` _and_ `wpjm_permalinks` has been removed.
  * Ensure the `Job listing archive page` option shows up with a default for `jobs`.
  * Change setting to another valid slug (e.g. `fancy-job-listings`). 
  * Save settings. 
  * Visit `http://example.com/fancy-job-listings` and verify it works and is an archive of published job listings.

* Enhancement: Allow jobs archive page slug to be customized in Permalink Settings.

### DIFF
--- a/includes/admin/class-wp-job-manager-permalink-settings.php
+++ b/includes/admin/class-wp-job-manager-permalink-settings.php
@@ -76,6 +76,24 @@ class WP_Job_Manager_Permalink_Settings {
 			'permalink',
 			'optional'
 		);
+		if ( current_theme_supports( 'job-manager-templates' ) ) {
+			add_settings_field(
+				'wpjm_job_listings_archive_slug',
+				__( 'Job listing archive page', 'wp-job-manager' ),
+				array( $this, 'job_listings_archive_slug_input' ),
+				'permalink',
+				'optional'
+			);
+		}
+	}
+
+	/**
+	 * Show a slug input box for job listing archive slug.
+	 */
+	public function job_listings_archive_slug_input() {
+		?>
+		<input name="wpjm_job_listings_archive_slug" type="text" class="regular-text code" value="<?php echo esc_attr( $this->permalinks['jobs_archive'] ); ?>" placeholder="<?php echo esc_attr( $this->permalinks['jobs_archive_rewrite_slug'] ); ?>" />
+		<?php
 	}
 
 	/**
@@ -129,6 +147,10 @@ class WP_Job_Manager_Permalink_Settings {
 			$permalink_settings['job_base']      = sanitize_title_with_dashes( $_POST['wpjm_job_base_slug'] );
 			$permalink_settings['category_base'] = sanitize_title_with_dashes( $_POST['wpjm_job_category_slug'] );
 			$permalink_settings['type_base']     = sanitize_title_with_dashes( $_POST['wpjm_job_type_slug'] );
+
+			if ( isset( $_POST['wpjm_job_listings_archive_slug'] ) ) {
+				$permalinks['jobs_archive'] = sanitize_title_with_dashes( $_POST['wpjm_job_listings_archive_slug'] );
+			}
 
 			update_option( WP_Job_Manager_Post_Types::PERMALINK_OPTION_NAME, wp_json_encode( $permalink_settings ) );
 
--- a/includes/admin/class-wp-job-manager-permalink-settings.php
+++ b/includes/admin/class-wp-job-manager-permalink-settings.php
@@ -118,12 +118,17 @@ class WP_Job_Manager_Permalink_Settings {
 				switch_to_locale( get_locale() );
 			}
 
-			$permalinks                  = (array) get_option( 'wpjm_permalinks', array() );
+			/**
+			 * Option `wpjm_permalink` was renamed to match other options in 1.32.0.
+			 *
+			 * Reference to the old option will be removed in 1.34.0.
+			 */
+			$permalinks                  = (array) get_option( WP_Job_Manager_Post_Types::PERMALINK_OPTION_NAME, get_option( 'wpjm_permalink', array() ) );
 			$permalinks['job_base']      = sanitize_title_with_dashes( $_POST['wpjm_job_base_slug'] );
 			$permalinks['category_base'] = sanitize_title_with_dashes( $_POST['wpjm_job_category_slug'] );
 			$permalinks['type_base']     = sanitize_title_with_dashes( $_POST['wpjm_job_type_slug'] );
 
-			update_option( 'wpjm_permalinks', $permalinks );
+			update_option( WP_Job_Manager_Post_Types::PERMALINK_OPTION_NAME, $permalinks );
 
 			if ( function_exists( 'restore_current_locale' ) ) {
 				restore_current_locale();
--- a/includes/admin/class-wp-job-manager-permalink-settings.php
+++ b/includes/admin/class-wp-job-manager-permalink-settings.php
@@ -137,11 +137,11 @@ class WP_Job_Manager_Permalink_Settings {
 			}
 
 			/**
-			 * Option `wpjm_permalink` was renamed to match other options in 1.32.0.
+			 * Option `wpjm_permalinks` was renamed to match other options in 1.32.0.
 			 *
 			 * Reference to the old option will be removed in 1.34.0.
 			 */
-			$legacy_permalink_settings = wp_json_encode( get_option( 'wpjm_permalink', array() ) );
+			$legacy_permalink_settings = wp_json_encode( get_option( 'wpjm_permalinks', array() ) );
 			$permalink_settings = (array) json_decode( get_option( WP_Job_Manager_Post_Types::PERMALINK_OPTION_NAME, $legacy_permalink_settings ), true );
 
 			$permalink_settings['job_base']      = sanitize_title_with_dashes( $_POST['wpjm_job_base_slug'] );
--- a/includes/admin/class-wp-job-manager-permalink-settings.php
+++ b/includes/admin/class-wp-job-manager-permalink-settings.php
@@ -149,7 +149,7 @@ class WP_Job_Manager_Permalink_Settings {
 			$permalink_settings['type_base']     = sanitize_title_with_dashes( $_POST['wpjm_job_type_slug'] );
 
 			if ( isset( $_POST['wpjm_job_listings_archive_slug'] ) ) {
-				$permalinks['jobs_archive'] = sanitize_title_with_dashes( $_POST['wpjm_job_listings_archive_slug'] );
+				$permalink_settings['jobs_archive'] = sanitize_title_with_dashes( $_POST['wpjm_job_listings_archive_slug'] );
 			}
 
 			update_option( WP_Job_Manager_Post_Types::PERMALINK_OPTION_NAME, wp_json_encode( $permalink_settings ) );
--- a/includes/admin/class-wp-job-manager-permalink-settings.php
+++ b/includes/admin/class-wp-job-manager-permalink-settings.php
@@ -123,12 +123,14 @@ class WP_Job_Manager_Permalink_Settings {
 			 *
 			 * Reference to the old option will be removed in 1.34.0.
 			 */
-			$permalinks                  = (array) get_option( WP_Job_Manager_Post_Types::PERMALINK_OPTION_NAME, get_option( 'wpjm_permalink', array() ) );
-			$permalinks['job_base']      = sanitize_title_with_dashes( $_POST['wpjm_job_base_slug'] );
-			$permalinks['category_base'] = sanitize_title_with_dashes( $_POST['wpjm_job_category_slug'] );
-			$permalinks['type_base']     = sanitize_title_with_dashes( $_POST['wpjm_job_type_slug'] );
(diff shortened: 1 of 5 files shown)


spencerfinnell:
Code looks good. Will try to test soon.

includes/class-wp-job-manager-post-types.php
```diff
@@ -736,19 +738,40 @@ public static function get_permalink_structure() {
...
-		$permalinks['type_rewrite_slug']     = untrailingslashit( empty( $permalinks['type_base'] ) ? _x( 'job-type', 'Job type slug - resave permalinks after changing this', 'wp-job-manager' ) : $permalinks['type_base'] );
+		// Ensure rewrite slugs are set. Use legacy translation options if not.
+		$permalinks['job_rewrite_slug']          = untrailingslashit( empty( $permalinks['job_base'] ) ? _x( 'job', 'Job permalink - resave permalinks after changing this', 'wp-job-manager' ) : $permalinks['job_base'] );
+		$permalinks['category_rewrite_slug']     = untrailingslashit( empty( $permalinks['category_base'] ) ? _x( 'job-category', 'Job category slug - resave permalinks after changing this', 'wp-job-manager' ) : $permalinks['category_base'] );
+		$permalinks['type_rewrite_slug']         = untrailingslashit( empty( $permalinks['type_base'] ) ? _x( 'job-type', 'Job type slug - resave permalinks after changing this', 'wp-job-manager' ) : $permalinks['type_base'] );
+		$permalinks['jobs_archive_rewrite_slug'] = untrailingslashit( empty( $permalinks['jobs_archive'] ) ? 'job-listings' : $permalinks['jobs_archive'] );
```
jom: Note: I didn't translate the new default (`job-listings`) but would be open to it. I didn't want people to use translations to _change_ the slug (like old with the legacy support we're keeping with the other ones), but I can see the case for having the _default_ be language specific.

includes/admin/class-wp-job-manager-permalink-settings.php
```diff
@@ -118,12 +136,23 @@ public function settings_save() {
...
-			$permalinks                  = (array) get_option( 'wpjm_permalinks', array() );
-			$permalinks['job_base']      = sanitize_title_with_dashes( $_POST['wpjm_job_base_slug'] );
-			$permalinks['category_base'] = sanitize_title_with_dashes( $_POST['wpjm_job_category_slug'] );
-			$permalinks['type_base']     = sanitize_title_with_dashes( $_POST['wpjm_job_type_slug'] );
+			/**
+			 * Option `wpjm_permalink` was renamed to match other options in 1.32.0.
+			 *
+			 * Reference to the old option will be removed in 1.34.0.
+			 */
+			$legacy_permalink_settings = wp_json_encode( get_option( 'wpjm_permalink', array() ) );
```
donnapep: `wpjm_permalink` is missing the 's'.

includes/class-wp-job-manager-post-types.php
```diff
@@ -736,19 +738,40 @@ public static function get_permalink_structure() {
...
 		}
 
-		$permalinks = wp_parse_args(
-			(array) get_option( 'wpjm_permalinks', array() ),
+		/**
+		 * Option `wpjm_permalink` was renamed to match other options in 1.32.0.
+		 *
+		 * Reference to the old option and support for non-standard plugin updates will be removed in 1.34.0.
+		 */
+		$legacy_permalink_settings = wp_json_encode( get_option( 'wpjm_permalink', array() ) );
```
donnapep: `wpjm_permalink` is missing the 's'.

includes/admin/class-wp-job-manager-permalink-settings.php
```diff
@@ -118,12 +136,23 @@ public function settings_save() {
 				switch_to_locale( get_locale() );
 			}
 
-			$permalinks                  = (array) get_option( 'wpjm_permalinks', array() );
-			$permalinks['job_base']      = sanitize_title_with_dashes( $_POST['wpjm_job_base_slug'] );
-			$permalinks['category_base'] = sanitize_title_with_dashes( $_POST['wpjm_job_category_slug'] );
-			$permalinks['type_base']     = sanitize_title_with_dashes( $_POST['wpjm_job_type_slug'] );
+			/**
+			 * Option `wpjm_permalink` was renamed to match other options in 1.32.0.
```
donnapep: `wpjm_permalink` is missing the 's'.

includes/class-wp-job-manager-post-types.php
```diff
@@ -736,19 +738,40 @@ public static function get_permalink_structure() {
 			switch_to_locale( get_locale() );
 		}
 
-		$permalinks = wp_parse_args(
-			(array) get_option( 'wpjm_permalinks', array() ),
+		/**
+		 * Option `wpjm_permalink` was renamed to match other options in 1.32.0.
```
donnapep: `wpjm_permalink` is missing the 's'.

donnapep (commented):
I haven't tested this yet, as I'd like to have those few typos fixed first in case they affect something.

donnapep:
> Four scenarios should be tested:

Wondering what the test instructions are for the fourth scenario. 🙂

includes/class-wp-job-manager-post-types.php
jom: Fixed in a5980321f66801dfc5d6e0805cd31f4edf15f2d1

includes/admin/class-wp-job-manager-permalink-settings.php
jom: Fixed in a5980321f66801dfc5d6e0805cd31f4edf15f2d1

jom:
@donnapep Consolidated two :)

donnapep:
I tested _Old install with theme support_ first and here's what I found:

> * In MySQL `{prefix_}options`, verify `job_manager_permalinks` option exists that is the current representation of `wpjm_permalinks`.
Mostly, although `job_manager_permalinks` has an additional entry for `"jobs_archive":"jobs"`.

> * In MySQL `{prefix_}options`, verify `job_manager_permalinks` option exists that is the current representation of the old `wpjm_permalinks` _and_ `wpjm_permalinks` has been removed.
Mostly, although `job_manager_permalinks` has an additional entry for `"jobs_archive":"jobs"` and the `wpjm_permalinks` entry was _not_ deleted.

> * Change setting to another valid slug (e.g. `fancy-job-listings`).
> * Save settings.

After changing the setting and saving, it goes back to `jobs`.

includes/class-wp-job-manager-post-types.php
tripflex: Another option would be to do something like I do in most of my plugins, which is to pull the post type single label, and use that (since a lot of themes modify this to be something like `listings` or whatever custom one they use)

```php
	public static function get_job_post_label(){

		$job_obj      = get_post_type_object( 'job_listing' );
		$job_singular = is_object( $job_obj ) ? $job_obj->labels->singular_name : __( 'Job' );

		if( ! $job_singular ) $job_singular = __( 'Job' );

		return $job_singular;

	}
```

jom:
@donnapep 
> Mostly, although job_manager_permalinks has an additional entry for "jobs_archive":"jobs".
That's fine.

> wpjm_permalinks entry was not deleted.

I'm not able to reproduce. You're sure your test database didn't have `job_manager_permalinks` (from previous test) when you deactivated/activated?  This might have happened if the instance was on this branch before switching to `master`. Also, it might have come back from an ajax call before you deactivated. I probably should have "deactivate the plugin and then delete the option."

@donnapep and @alexsanford : Fixed the persisting issue in 100f4008aa7c161e620fc1fb9325c5c74c917c18. That's what you get from renaming a variable last minute.

donnapep:
> I'm not able to reproduce. You're sure your test database didn't have `job_manager_permalinks` (from previous test) when you deactivated/activated?

Yup. I deactivate, delete `job_manager_permalinks`, run `select * from wp_options
where option_name in ('wpjm_permalinks', 'job_manager_permalinks')` and get this:

screen shot 2018-12-12 at 2 58 04 pm

After reactivating and running the same query again, I get this:

screen shot 2018-12-12 at 3 00 58 pm

I can see that `delete_option` is not being called because `false === get_option( 'job_manager_permalinks', false )` does not return `false`. It's already been set elsewhere by this point.

donnapep:
It's being set in the `get_permalink_structure` function, which is running before the `install` function.

On the plus side, the field is saving its value now. 🙂

jom:
Should be fixed now. Moved migration from legacy completely into `WP_Job_Manager_Post_Types`.

alexsanford:
Just re-tested and it's working very nicely!

The only thing I'm seeing is in the third case:

> Ensure the `Job listing archive page` option shows up with a default for `jobs`.

When I look at the option, it is blank, with the placeholder for `job-listings`. It does not have the value `jobs`. I think I saw that on a previous test, but it didn't have it when I tested this time.

jom:
@alexsanford I just pushed a small update, but I don't think it should affect the result of your test. I updated the testing instructions though because the activation/deactivation dance is no longer necessary.

donnapep:
_New install with theme support:_
> * Ensure the `Job listing archive page` option shows up with a default for `job-listings`.

I am not seeing `job-listings`. The default value is `jobs`.

_New install without theme support_
No issues.

_Old install with theme support_
No issues. I'm not seeing the behaviour that Alex is.

includes/class-wp-job-manager-post-types.php
```diff
@@ -736,19 +738,45 @@ public static function get_permalink_structure() {
...
 		}
 
-		$permalinks = wp_parse_args(
-			(array) get_option( 'wpjm_permalinks', array() ),
+		/**
+		 * Option `wpjm_permalinks` was renamed to match other options in 1.32.0.
+		 *
+		 * Reference to the old option and support for non-standard plugin updates will be removed in 1.34.0.
+		 */
+		$legacy_permalink_settings = '[]';
```
alexsanford: I wonder if it would make sense to consolidate this code with the similar code in `includes/admin/class-wp-job-manager-permalink-settings.php`? Maybe we call a function that checks for `wpjm_permalinks`, converts it to the `job_manager_permalinks` option if needed, and deletes the old option. Then we just call it from here and from `includes/admin/class-wp-job-manager-permalink-settings.php` to ensure that we're dealing with the up-to-date option.

Not a blocker for me, but it would reduce a bit of code duplication.

alexsanford (approved):
Tested again, and it all seems to work! 🎉 

Just one little comment, but not a blocker for me.

donnapep (approved):
This is working for me now. 🙂

includes/class-wp-job-manager-post-types.php
jom: done in 0df7616477c624018edfb32bf189c1feb0481c15

PR #1646, opened 2018-12-13
Add dropdown to select author on job listing editor

### LINKED ISSUE #846: Replace listing author ID number meta with a dropdown
WooCommerce has a nice selector to choose an user for an order. It would be nice to add this instead of requiring a site item to know the user ID of the intended author.

jom:
Fixes #846

~~Based on `change/move-to-select2`. I'll rebase when #1645 is merged.~~ 
dec-13-2018 17-38-48

* Add dropdown that is sourced from an AJAX call to select an author on a job listing page.
* Based on Woo's approach with some modifications. I added pagination at the expense of searching `first_name` and `last_name` meta fields.
* Bumped up the priority for the author field so it appears first in the list to fix an issue with `select2` and the block editor. Having it at the bottom of the screen caused an issue with the search input jumping around on activation.
* I tried to make it so we can easily implement this on resumes and applications.
* Fixed a tiny bug that threw a PHP notice when saving a listing that got bumped to a guest post.

#### Testing instructions:

* In both the block and classic editor, change the author of a job listing. Try different searches. 
* Try clearing (`x` on right of input) and saving. 
* Bonus points: go to edit a job listing and then in another tab log out and log in as a low-privileged user (subscriber). Go back to the job listing editor tab (without refreshing) and attempt to change user/search for user. Should fail.

* Enhancement: Change job listing authors more easily with dropdown in job listing editor.


alexsanford (approved):
This looks great! There are two tiny UX things to mention. If either of them is a simple fix, great. If not, I don't think either of them are critical.

- I have to click twice in order to change the author. I have to click the "Change" button, and then click the drop-down. It might not be immediately intuitive to our users that they have to do the extra click. I'm not sure if there's a great alternative, however. Given that they also need to be able to use the `x` button, I don't think we could just automatically open the drop-down. We could consider removing the "Change" button entirely, so we only have the drop-down UI. But then we would lose the link to the user. Probably best to leave it as-is, but I thought I would bring it up.

- In the block editor, after I change the author and click Update, the UI does not revert back to the link with the Change button. This is because the block editor saves over AJAX, and the page does not get reloaded. This is probably fine, but it's a small inconsistency with the UX of the Classic Editor.

alexsanford:
@jom Yeah, I'm fine to just leave this as-is, since the issues aren't that big, and there isn't a straightforward fix (except maybe removing the Change button). We can always revisit this later if we feel like it's a pain point. I'm guessing it won't be 🙂

jom:
cc: @tripflex I don't think this affects Field Editor, but just in case.

tripflex:
@jom nope it shouldn't, but thank you for the heads up either way ... I do have a setting to allow users to enable Chosen in the admin area, but even with that I don't think this is any kind of issue (it's disabled by default)

PR #1649, opened 2018-12-17
Release 1.32.0

jom:
Release for 1.32.0

[Closed Issues](https://github.com/Automattic/WP-Job-Manager/issues?q=is%3Aissue+milestone%3A1.32.0+is%3Aclosed) | [PRs](https://github.com/Automattic/WP-Job-Manager/pulls?q=is%3Apr+milestone%3A1.32.0+is%3Aclosed) | [Diff](https://github.com/Automattic/WP-Job-Manager/compare/1.31.3...release/1.32.0)

### Timeline
- Target Release: ~2019-01-15~ 2019-01-23
- Beta 2: 2019-01-18 [Beta 2 Release](https://github.com/Automattic/WP-Job-Manager/releases/tag/1.32.0-beta2) | [Download](https://github.com/Automattic/WP-Job-Manager/releases/download/1.32.0-beta2/wp-job-manager-1.32.0-beta2.zip)
- Beta 1: 2018-12-17 [Beta 1 Release](https://github.com/Automattic/WP-Job-Manager/releases/tag/1.32.0-beta1) | [Download](https://github.com/Automattic/WP-Job-Manager/releases/download/1.32.0-beta1/wp-job-manager-1.32.0-beta1.zip)

### Things to Test
- Theme compatibility on `[jobs]` shortcode with category select fields. Try with multiple category select option enabled and disabled.
- Theme compatibility with frontend styles. As deprecated in an earlier release, frontend styles must be told when to be enqueued using `job_manager_enqueue_frontend_style` filter. See [this code](https://github.com/Automattic/wp-job-manager/blob/c0938ad4a3224f793d6663d23b7a430906babb84/wp-job-manager.php#L448-L476).
- Permalink settings persist when moving an existing instance to 1.32.0.
- Author selection works when editing a job listing in WP admin.
- See changelog for additional things to look at.

Comment on this PR with any issues that are found.

### Changelog:
```
= 1.32.0 =
* Enhancement: Switched from Chosen to Select2 for enhanced dropdown handling and better mobile support. May require theme update.
* Enhancement: Draft and unsubmitted job listings now appear in `[job_dashboard]`, allowing users to complete their submission.
* Enhancement: Filled and expired positions are now hidden from WordPress search. (@felipeelia)
* Enhancement: Adds additional support for the new block editor. Restricted to classic block for compatibility with frontend editor.
* Enhancement: Job types can be preselected in `[jobs]` shortcode with `?search_job_type=term-slug`. (@felipeelia)
[...]

PR #1657, opened 2019-01-10
Only init select2 on visible select fields

jom:
Prevents some collisions with chosen on some themes.

From: https://github.com/Automattic/WP-Job-Manager/pull/1649#issuecomment-453259400

See the comment for testing instructions.


alexsanford:
Minor correction - I think this is the comment you meant to link to: https://github.com/Automattic/WP-Job-Manager/pull/1649#issuecomment-453259400

alexsanford:
Also, you may want to remove the "Fixes" word from the description, otherwise merging this will close the other PR 🙂

alexsanford (commented):
This works for me. The only reason I'm not approving now is because of the "Fixes" keyword (accidentally merging now would close the other PR, which would be annoying). But once that's changed, I'm good with this being merged 🙂

PR #1659, opened 2019-01-14
Set up backwards compatibility on taxonomy and archive pages

jom:
This fixes some issues with themes that load the job filters on taxonomy pages. Until themes tell Job Manager explicitly that they want to use `select2` on certain pages, this will load those assets on these extra pages.

### Testing Instructions

On Jobify, view job listing archive page. Make sure `select2` is initialized on the category select.


alexsanford (approved):
Looks good to me! 🚀

PR #1660, opened 2019-01-15
Use global for select2 args and set width to 100%

jom:
This fixes an issue with window resizing and renames the `job_manager_select2_multiselect_args` variable to more generic `job_manager_select2_args`.

To fix the issue with window resizing, it passes a more opinionated 100% width. Seems to look okay in the themes I tested with.

### Testing Instructions
- On multiple themes, check:
   - `[jobs]` filter with multiselect and normal select for the job categories.
   - `[submit_job_listing]` page with multiselect fields.

These should all be `select2` unless `chosen` has already loaded from the theme or the theme is explicitly blocking it.


assets/js/ajax-filters.js
```diff
@@ -239,14 +239,11 @@ jQuery( document ).ready( function ( $ ) {
 		return false;
 	} );
 
-	if ( $.isFunction( $.fn.select2 ) ) {
-		var select2_args = {
-			allowClear: true,
-			minimumResultsForSearch: 10
-		};
-		if ( 1 === parseInt( job_manager_ajax_filters.is_rtl, 10 ) ) {
```
donnapep: Just wondering is this was intentionally removed?

assets/js/multiselect.js
```diff
@@ -1,4 +1,6 @@
 /* global job_manager_select2_args */
-jQuery(function(){
-	jQuery( '.job-manager-multiselect:visible' ).select2( job_manager_select2_args );
+jQuery(function($){
+	if ( $.isFunction( $.fn.select2 ) && typeof job_manager_select2_args !== 'undefined' ) {
+		$('.job-manager-multiselect:visible').select2(job_manager_select2_args);
```
donnapep: We've lost the lovely whitespacing before and after the parentheses. 🙂

assets/js/term-multiselect.js
```diff
@@ -1,4 +1,6 @@
 /* global job_manager_select2_args */
-jQuery(function(){
-	jQuery( '.job-manager-category-dropdown:visible' ).select2( job_manager_select2_args );
+jQuery(function($){
+	if ( $.isFunction( $.fn.select2 ) && typeof job_manager_select2_args !== 'undefined' ) {
+		$('.job-manager-category-dropdown:visible').select2(job_manager_select2_args);
```
donnapep: Same here.

assets/js/ajax-filters.js
jom: That is now coming in from `job_manager_select2_args`.

wp-job-manager.php
```diff
@@ -410,9 +410,11 @@ public function frontend_scripts() {
 				$select2_args['dir'] = 'rtl';
 			}
 
+			$select2_args['width'] = '100%';
+
 			wp_localize_script(
-				'select2', 'job_manager_select2_multiselect_args',
-				apply_filters( 'job_manager_select2_multiselect_args', $select2_args )
```
alexsanford: Do we need to deprecate this? Or has it not been released yet?

jom: It hasn't been released yet. All `select2` stuff is in this release.

PR #1668, opened 2019-01-24
Revert: Hide filled and expired positions from native WP search 

### LINKED ISSUE #1667: 1.32.0 breaks search for anything *except* jobs throughout WordPress (frontend + backend)
#### Steps to Reproduce
1. In WP Admin go to Posts
2. Ensure there is at least one post
3. Search for a word found in any one of the posts
4. You will get no results unless that word happens to exist within a posted job as well

The same is true of front-end searches. All search queries are broken. When I disable this plugin, all works as expected again.

#### What I Expected

For WordPress to handle searches as normal.

#### What Happened Instead

WordPress searches return no results unless your search query happens to contain a job in the results.

1.32.0 -- It only began with this version.

#### Browser / OS Version

- This should be irrelevant as it's not a front-end issue. 
- Tried it with Firefox 65b12 and Chrome Version 71.0.3578.98
- macOS 10.14.2

### Related WP Forum Post

Found this one and commented on it. It describes exactly the same issue I'm experiencing: https://wordpress.org/support/topic/site-search-problem/#post-11127903

### Other

I have no search plugins installed and to my knowledge there is no custom code that interacts with the WordPress search system. The singular plugin that even comes close to interacting with search functionality (WP-Json-Feed), I've disabled and it makes no difference.

jom:
Reopens #833
Fixes #1667
Reverts #1570

* Restores old functionality of just hiding expired and filled positions on the archive pages.

#### Testing instructions:

* Search for standard posts and pages in WP search and make sure they appear.

* Fix: Non-jobs were not appearing in WP search results.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -69,7 +69,6 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'update_post_meta', array( $this, 'update_post_meta' ), 10, 4 );
 		add_action( 'wp_insert_post', array( $this, 'maybe_add_default_meta_data' ), 10, 2 );
 
-		add_action( 'parse_query', array( $this, 'public_search_handler' ) );
 		add_action( 'parse_query', array( $this, 'add_feed_query_args' ) );
 
 		// Single job content.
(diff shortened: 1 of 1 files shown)


alexsanford (approved):
I cannot reproduce the original issue. The search works fine for me on `master`.

That said, everything seems to work fine on this PR as well, so I'm fine with merging this.

PR #1708, opened 2019-03-29
Add job search submission button

### LINKED ISSUE #1409: Search form: Include actual "Search" or "Submit" button for accessibility
Raised in https://wordpress.org/support/topic/accessibility-issues-with-plugin/ 

<blockquote>This form does not contain a submit button, which creates issues for those who cannot submit the form using the keyboard. Submit buttons are INPUT elements with type attribute “submit” or “image”, or BUTTON elements with type “submit” or omitted/invalid.”</blockquote>

alexsanford:
Fixes #1409 

Adds a "Search Jobs" button to the job filters UI for accessibility purposes, and a filter to hide the button if desired.

## Approach

I decided to add the button directly to the template. This is consistent with the rest of the fields in the template. An alternative would be to add the button using the `job_manager_job_filters_search_jobs_end` action instead.

Some implications of adding directly to the template:

- Existing themes/sites with custom templates will not have this button unless they add it to their template.
- Themes or plugins that adds a similar button using the aforementioned action will need to disable the new button using the new filter (currently Jobify adds a button in this way).

## Testing instructions

- On the "jobs" page, ensure that the filters work when pressing Enter, and when clicking the button.
- Test with several themes, to ensure that the button looks and works ok on them all.
- Ensure that this code snippet hides the button:

```php
add_filter( 'job_manager_job_filters_show_submit_button', '__return_false' );
```


templates/job-filters.php
```diff
@@ -51,6 +51,12 @@
 			</div>
 		<?php endif; ?>
 
+		<?php if ( apply_filters( 'job_manager_job_filters_show_submit_button', true ) ) : ?>
```
jom: Let's add docs for new filter.

richardmtl:
<blockquote>Might be worth digging a bit more into the intent.</blockquote>
 
The intent, from users who have requested this, is that they are confused by the lack of a submit button, and they worry that folks won't know how to get results for their searches.

jom:
I think we can either have a button _or_ the auto-submit on blur/change. Otherwise, we will create patterns where results will be loaded and then re-loaded with the same content. We can possibly do this by having `ajax-filters.js` check for the button's existence, but we should talk to some theme folks.

alexsanford:
I took a look at how Jobify handles this. It seems that they add a button, but then prevent the button from actually triggering a form submission. When the button is clicked, that causes a blur on whichever field is in focus, and so even though the button technically does nothing, clicking on it still (indirectly) triggers an update.

This is probably the simplest option for us as well. It gives users a button, which should reduce confusion, but otherwise keeps the functionality as-is. Feels a little funny to have a button that technically does nothing, but it might actually be a good solution in this case.

I've added this to the PR. Any thoughts?

templates/job-filters.php
alexsanford: Done - 36edea5

jom:
Two things:
1) On Twenty Seventeen, the padding doesn't seem to have an effect.
2) Job types hook onto `job_manager_job_filters_end`, causing it to show up below the submit button. We might need to move where the button is located or change the hooks. We could hook it into `job_manager_job_filters_end` priority `19` (right before results). 

Screen Shot 2019-04-29 at 2 52 13 PM

alexsanford:
Thanks for having a look, @jom !

> On Twenty Seventeen, the padding doesn't seem to have an effect.

Good catch! It was a `float` issue that only showed up when the category dropdown was disabled. Fixed in 4fa9d99

> Job types hook onto job_manager_job_filters_end, causing it to show up below the submit button. 

Personally, I think this is fine. From a UX perspective, the tags are sort of their own thing anyway. I think the idea of clicking the checkboxes outside of the "form" and having the job list update automatically is pretty intuitive. Much more intuitive than the jobs updating on blur of a text field (as per the above screenshot).

Here's one possible option for going below the tags. I don't like it as much this way. I think it makes more sense for the button to be with the text fields and category dropdown.

Screen Shot on 2019-05-01 at 19:21:33

Steve98796:
I  use the Jobify theme and this has caused the following issue. Since the update I get this mess rather than the one line I had before with no Search Jobs:
image

alexsanford:
@Steve98796 Could you try adding the following code snippet, and see if it fixes the problem?

```php
add_filter( 'job_manager_job_filters_show_submit_button', '__return_false' );
```

Braehler:
@Steve98796 here is a css solution that is working for us
`@media only screen and (min-width:992px){
.home .search_categories {
    width: 26% !important;
}
.home .search_submit {
    float: right !important;
    width: 20% !important;
}
}
@media only screen and (min-width:992px){
.search_categories {
    width: 26% !important;
}
.search_submit {
    float: right !important;
    width: 20% !important;
}
}`

Steve98796:
Thanks for the replies. Looks like the latest Jobify update has resolved it.

PR #1709, opened 2019-04-05
Adds setting to show logo on Featured Jobs widget

### LINKED ISSUE #1700: Featured Jobs widget: Company logo is not displayed
Originally reported in 1862198-zen ; I was able to reproduce on a site with Twenty Seventeen.

The Featured Jobs widget does not display the company logo:

Screen Shot 2019-03-12 at 16 18 53

Even though it should, as far as I can tell:

https://github.com/Automattic/WP-Job-Manager/blob/master/includes/widgets/class-wp-job-manager-widget-featured-jobs.php#L113
https://github.com/Automattic/WP-Job-Manager/blob/master/templates/content-job_listing.php#L23

So I don't know what's blocking it from showing. Note: it's not just hidden because of CSS; the HTML output isn't there for it either.

jom:
Fixes #1700

* Adds setting for Featured Jobs widget to show logo. Matches behavior with Recent Jobs widget.

#### Testing instructions:

* Add Featured Job widget. Try enabling and disabling the **Show Company Logo** setting for the widget.

- New: Adds option to show company logo on Featured Jobs widget.

### DIFF
--- a/includes/class-wp-job-manager-widget.php
+++ b/includes/class-wp-job-manager-widget.php
@@ -181,8 +181,8 @@ class WP_Job_Manager_Widget extends WP_Widget {
 				case 'checkbox':
 					?>
 					<p>
-						<label for="<?php echo esc_attr( $this->get_field_id( $key ) ); ?>"><?php echo esc_html( $setting['label'] ); ?></label>
 						<input class="checkbox" id="<?php echo esc_attr( $this->get_field_id( $key ) ); ?>" name="<?php echo esc_attr( $this->get_field_name( $key ) ); ?>" type="checkbox" value="1" <?php checked( $value, 1 ); ?> />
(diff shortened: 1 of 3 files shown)


alexsanford (approved):
Code looks good, works well.

This is a nitpick, but I think the checkbox should be on the left of the label instead of the right - https://cld.wthms.co/jroaqG

That would match the positioning of the checkboxes in the default WP widgets.

I see that the checkbox is also on the right in the Recent Jobs widget...and maybe there are others? If it's going to be a big deal to change them all, then let's not worry about it. But otherwise, I'd say we should move it to the left.

Approving this. The checkbox position isn't a dealbreaker 🙂

jom:
@alexsanford Moved the checkbox 466755d29f48965cfc21cfff9c5ebb2839073ded :)

PR #1711, opened 2019-04-08
add fieldset-type-FIELDTYPE class to fieldset field wrapper element

### LINKED ISSUE #1710: Adding field type to default fieldset HTML element?
So right now in the default submit template, the `fieldset` HTML elements only have a class that is related to the field meta key.

This makes it difficult to target specific field types for customization, without having to create a `job-submit.php` template override .. which would be kinda overkill for just adding a class.

This is what is currently in the template:
https://github.com/Automattic/WP-Job-Manager/blob/6a1e5f044e32dae3602bedab5b06ada386441919/templates/job-submit.php#L41-L48

What do you guys think about maybe adding a field type specific class to it?  Maybe something like this:

```PHP
fieldset-type-<?php echo esc_attr( $field['type'] ); ?>
```

That way themes, plugins, or end users can target specific field types without having to use jQuery or any other javascript to dynamically add classes just to customize a specific field type.

tripflex:
Fixes #1710

* Add `fieldset-type-XXXX` class to default `fieldset` HTML elements (`XXXX` is the field type slug )

#### Testing instructions:

* Browser to submit page, inspect fieldset elements

### DIFF
--- a/templates/job-submit.php
+++ b/templates/job-submit.php
@@ -39,7 +39,7 @@ global $job_manager;
 		<?php do_action( 'submit_job_form_job_fields_start' ); ?>
 
 		<?php foreach ( $job_fields as $key => $field ) : ?>
-			<fieldset class="fieldset-<?php echo esc_attr( $key ); ?>">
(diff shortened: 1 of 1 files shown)


jom:
@tripflex Actually, using this in a PR I'm working on. Going to merge it. I'll update version.

PR #1722, opened 2019-04-24
Allow embed `<iframe>` in job description content

### LINKED ISSUE #1584: Allow additional HTML entities in job listing post content
In our recent HTML escaping PRs we started doing a pass on `the_content()` through `wp_kses_post()`. This is filtering out safe transformations (such as YouTube embeds) in `the_content` hook.

https://github.com/Automattic/WP-Job-Manager/blob/7ea7454fd98b9bf655b452f1193118ae5525e0fb/wp-job-manager-template.php#L532

To fix this, I think we should add a new context to `wp_kses_allowed_html` (allowing for everything in `posts` + ifames), pass it through `wp_kses` in `wpjm_get_the_job_description()`, and `echo` it with a `// WPCS: XSS ok.` exception in `wpjm_the_job_description()`. 

Alternatively, we could run `wp_kses_post` on the raw `post_content` value in `wpjm_get_the_job_description()` and then apply `the_content` filter on the final value.

#### Steps to Reproduce
1.  Create a job post with a YouTube link on one line.
2. View the job.

#### What I Expected
YouTube video to be displayed.

#### What Happened Instead
No YouTube embed in job description.

4.9.x / 1.31.3

#### Browser / OS Version
Chrome

#### Context / Source
https://wordpress.org/support/topic/why-embedded-youtube-wont-show-up-in-single-job-page/

### LINKED ISSUE #1718: Youtube video embeds not working
#### Steps to Reproduce
- Add a Youtube video to a job description from WP Admin.
- Video preview shows in the Editor.
- Save or Preview the post.
- Video embed is missing when viewing the Job Details page on the Frontend.
- Video embed is working with JobifyVersion: 3.8.6 & ListifyVersion: 2.5.0.
- Video embed not working with `Twenty Seventeen`, `Twenty Nineteen`.
- Video embed not working with `Avada 5.7.2` (reported by a user in support).

jom:
Fixes issue of embeds getting removed from job description content. Introduced by #1495 in 1.31.1 (July 2018).

Fixes #1584
Fixes #1718 

* Allows iframe in job description content.

#### Testing instructions:

* Add YouTube link in job description.
* Verify embed works correctly on frontend.

### To  Do
- [x] Add tests

### DIFF
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -548,13 +548,13 @@ function wpjm_get_the_job_description( $post = null ) {
 		return null;
 	}
 
-	$description = apply_filters( 'the_job_description', $post->post_content );
+	$description = apply_filters( 'the_job_description', wp_kses_post( $post->post_content ) );
 
 	/**
 	 * Filter for the job description.
 	 *
 	 * @since 1.28.0
-	 * @param string      $title Title to be filtered.
+	 * @param string      $job_description Job description to be filtered.
 	 * @param int|WP_Post $post
 	 */
 	return apply_filters( 'wpjm_the_job_description', $description, $post );
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -548,7 +548,7 @@ function wpjm_get_the_job_description( $post = null ) {
 		return null;
 	}
 
-	$description = apply_filters( 'the_job_description', get_the_content( $post ) );
+	$description = apply_filters( 'the_job_description', $post->post_content );
 
 	/**
 	 * Filter for the job description.
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -531,7 +531,7 @@ function wpjm_get_the_job_title( $post = null ) {
 function wpjm_the_job_description( $post = null ) {
 	$job_description = wpjm_get_the_job_description( $post );
 	if ( $job_description ) {
-		echo wp_kses_post( $job_description );
+		WP_Job_Manager_Post_Types::output_kses_post( $job_description );
 	}
 }
 
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -421,6 +421,52 @@ class WP_Job_Manager_Post_Types {
 	}
 
 	/**
+	 * Filter the post content of job listings.
+	 *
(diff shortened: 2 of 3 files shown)


tripflex:
While I understand the need for something like this --- doesn't allowing iframe in job content open up sites for potential exploitation by injecting malicious iframes?

jom:
@tripflex We could add `wp_kses_post` before it passes through `the_job_description` (which adds the embeds; like `the_content` filters). I’d be fine with that and should still fix this issue.

wp-job-manager-template.php
```diff
@@ -548,13 +548,13 @@ function wpjm_get_the_job_description( $post = null ) {
 		return null;
 	}
 
-	$description = apply_filters( 'the_job_description', get_the_content( $post ) );
+	$description = apply_filters( 'the_job_description', wp_kses_post( $post->post_content ) );
```
alexsanford: Seems like when we're calling `wpjm_the_job_description`, we'll be running `kses` twice. Is this necessary? Could we just call `WP_Job_Manager_Post_Types::output_kses_post` here, and not worry about it in `wpjm_the_job_description`?

Could be that there's a good reason to call it twice, and I'm just not seeing it 😄

alexsanford (approved):
Seems to work well. One question about how the `kses` is happening, but otherwise I think this one looks good!

wp-job-manager-template.php
jom: This comes from above comment. The issue is we don’t want to allow `iframes` from user input _but_ we do want to allow them when they come from embeds (from within `the_job_description` hooks). We really probably don’t need to filter it again (post embed), but seems to be standard to it on output as well (to catch other things from hooks?).

One other note about this: there is a tiny bit of functionality change happening here. `get_the_content` was being called incorrectly and I would have had to override the global `$post` variable to do it correctly. With that, pagination inside of a single post is removed. I don’t know if it would have really worked before and doubt it was in-use, but just thought it best to note it.

alexsanford: Works for me! I don't think the overhead of calling it twice will be significant, just wanted to be clear.

Yeah, I see what you mean about `get_the_content`. Definitely was wrong before. Hopefully the change doesn't cause weird side-effects, but it seems like we should be ok.

PR #1723, opened 2019-04-26
Allow limiting the number of file uploads per field

### LINKED ISSUE #516: Limit number of Files Uploaded
Is there any option to limit the number of files that are allowed to be uploaded depends on the Job Package?

jom:
Fixes #516

* Allows custom multiple file fields to be added with enforced limits on the number of files that can be uploaded.
* Updates `assets/js/jquery-fileupload/jquery.fileupload.js` and `assets/js/jquery-fileupload/jquery.iframe-transport.js` to their newest versions.

#### To Do
- [x] Enforce limit in backend.
- [x] Check in other browsers (Not tested: IE/Edge)

#### Testing instructions:
Snippet to use/adjust:
```
  $fields['job']['job_files'] = [
	'label'              => __( 'Job Files', 'wp-job-manager' ),
	'type'               => 'file',
	'required'           => false,
	'placeholder'        => '',
	'priority'           => 6,
	'ajax'               => false,
	'multiple'           => true,
	'file_limit'	     => 2,
	// 'file_limit_message' => 'You can only upload %d FILES, please and thank you',
  ];
  
  return $fields;
} );
```

* With this snippet, try validation with ajax enabled and disabled. Already uploaded files count toward limit and removing them should allow for more files to be uploaded.

Note: `'multiple' => false` will ignore file limit.


jom:
@tripflex Want to loop you in early on this one. The issue has a lot of Field Editor chatter, but haven't taken a look at what you're doing.

tripflex:
@jom thank you, yes my plugin already has native integration for this to handle max allowed files after it wasn't ever added in core (per that issue).

Just for reference, here's how i'm doing it.  The template override file adds some JS to define the max upload here:
https://github.com/tripflex/wp-job-manager-field-editor/blob/master/templates/form-fields/file-field.php#L16-L32

Here's the JS to handle that:
https://github.com/tripflex/wp-job-manager-field-editor/blob/master/assets/js/build/fileupload.js

Here's the method used to check:
https://github.com/tripflex/wp-job-manager-field-editor/blob/1a2466b30ea460bec73f4227d6fc1e71bccae691/classes/integration.php#L282-L338

Job check:
https://github.com/tripflex/wp-job-manager-field-editor/blob/fe97ca6694f4044942a338dd0daf25cc193ae2de/classes/job/fields.php#L451

Resume check:
https://github.com/tripflex/wp-job-manager-field-editor/blob/74d25521c0e062ee8af93ec7325c847a366d6776/classes/resume/fields.php#L314

tripflex:
I don't see anything that would be a major conflict right off the top of reviewing it, but i will go through and test myself to make sure, and then i can make sure to integrate with the native handling for an upcoming release (and maybe deprecate the old handling for older versions of WPJM that people may still be using).

Targeting this for 1.33.0?

jom:
@tripflex Yup, targeting it for 1.33.0 (probably a month out). Let me know if you find anything we should add/change to core's implementation. There is some stuff I don't like about Ajax uploader's validation (`window.alert`), but I think we'll save that for another PR.

assets/js/ajax-file-upload.js
```diff
@@ -5,15 +5,39 @@ jQuery(function($) {
...
+					fileLimitLeft = fileLimit - currentFiles;
+					$file_field.data( 'file_limit_left', fileLimitLeft );
+				}
+
+				if ( false !== fileLimitLeft && fileLimitLeft <= 0 ) {
+					var message = 'Exceeded upload limit';
+					if( $file_field.data( 'file_limit_message' ) ) {
+						message = $file_field.data( 'file_limit_message' );
+					} else if ( typeof job_manager_job_submission !== 'undefined' ) {
+						message = job_manager_job_submission.i18n_over_upload_limit;
```
alexsanford: Should we use JS i18n here? https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/

I guess it was introduced in 5.0, so maybe we should hold off until we drop 4.x support?

assets/js/job-submission.js
```diff
@@ -1,8 +1,55 @@
...
+	$( document.body ).on( 'change', '.fieldset-type-file input[type=file][multiple][data-file_limit]', function(){
+		var fileLimit     = parseInt( $(this).data( 'file_limit' ), 10 );
+		var currentFiles  = parseInt( $(this).siblings( '.job-manager-uploaded-files' ).first().children( '.job-manager-uploaded-file' ).length, 10);
+		var fileLimitLeft = fileLimit - currentFiles;
+		var rawElement    = $(this).get(0);
+
+		if ( typeof rawElement.files !== 'undefined' ) {
+			var filesUploaded = parseInt( rawElement.files.length, 10 );
+			if ( fileLimit && filesUploaded > fileLimitLeft ) {
+				var message = job_manager_job_submission.i18n_over_upload_limit;
```
alexsanford: See above message about JS i18n

assets/js/ajax-file-upload.js
jom: Yeah, I think let's hold off on that now. I'm a bit hesitant to bump minimum version to 5.0 with so many on 4.9.

PR #1731, opened 2019-04-30
Register post meta and show default fields in REST API

### LINKED ISSUE #1591: Application field auto populating new listings with user email (even if app method set to URL only)
https://github.com/Automattic/WP-Job-Manager/blob/68d32d068d1e095036c5ed2c1f8562c1bb00dbdf/includes/admin/class-wp-job-manager-writepanels.php#L66

Looks like by default as you can see in the code above, if meta doesn't exist on a listing when it's created from admin area, it auto populates that field value with the current user's email address.

My main concern with this is that even if the application method is configured in the settings for "Website URLs only" ... this field is still populated with a user's email address ...

My opinion on this is that we should either remove this entirely (only populating from existing value), check the application method and only populate if email is enabled, or add something in settings to enable/disable this feature

I think there are a few options we have to resolve this:

1. Completely remove the auto populate from user's email
2. Only populate if application method has email allowed (URL & Email, or Email only)
3. Add a setting to enable/disable this feature, or even just a text box to save as option for "application populate value" or something

Thoughts? @jom

### LINKED ISSUE #1697: REST API: Register custom field meta
Hello:-), I would like to be able to update by API the information included in a meta box (example of a meta box on the theme demo mycar "mileage")

1 - In the first case I can see the meta box field created from ACF. With ACF it is possible to do this by making a POST request:

curl -X POST \
https://test.com/wp-json/acf/v3/annonce/183\\
-H'Authorization: Basic xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' \
-H'Postman-Token: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' \
-H'cache-control: no-cache' \
-H' content-type: multipart/form-data; boundary=----WebKitFormBoundary7MA4YWxkTrZu0gW' \
-F'fields[mileage]=15019''

2 - Whit WP JOB Manager, In this second case I can't see the field of the meta box "mileage". It exists well but is not visible by the API.
I can't discover the meta box "mileage" fields via a GET request. This is the request I built, but it is not complete.

curl -X POST \
https://test2.com/wp-json/wp/v2/job-listings/183 \
-H'Authorization: Basic xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' \
-H'Postman-Token: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' \
-H'cache-control: no-cache' \
-H' content-type: multipart/form-data; boundary=----WebKitFormBoundary7MA4YWxkTrZu0gW' \
-F' fields[?????????????]=15019''.

Can you help me to complete understanding how to build the POST request that allows me to do this. Or I am willing to pay a plugin for the functionality to be available. 

Thank you in advance:-)

jom:
Fixes #1697
Fixes #1591 

### Changes proposed in this Pull Request:

- Refactors the way job listing fields are registered in WP admin. *This must/should be backwards compatible*
- Registers `job_listing` post meta and enables `show_in_rest`. 
- Moves sanitization of post meta higher up into `add_post_meta`/`update_post_meta`. There is a new `sanitize_callback` attribute on field settings. There is a default sanitization for core fields that mimics previous behavior. Meta values coming from the frontend input are getting double-sanitized in some cases, but because they could be passed off to other areas (email notifications, plugins, etc), probably best to keep it that way.
- Moves auth for editing to callbacks as well in `auth_edit_callback` attribute on field settings. Right now these don't do much because the CPT has tighter permissions in REST API.
- Introduces `auth_view_callback` which defaults to previous behavior for WP admin. The only exception is `job_expires`. That field is hidden unless you're an admin or an employer viewing their own jobs.

### Fields Now Public by REST API
- `_job_location`
- `_application`
- `_company_name`
- `_company_website`
- `_company_tagline`
- `_company_twitter`
- `_company_video`
- `_filled`
- `_featured`

### What's not in this PR
- This is Phase 1 of working on WPJM's REST API. Right now, non-read events are very limited. Meta fields have looser `auth` rules, but aren't accessible because `job_listings` are only editable in REST API by admins. We'll eventually open this up a bit more so that employers can post/edit their own listings.
- I kept the logic for syncing job expiry/post status in `writepanels`. It is actually probably a bit broken in the block editor right now. I started moving that logic to higher level filters that could be shared among the REST API and front/backends, but it is going to require a bit of an over hall. 
- Google's location meta fields (`geolocation_*`). These should probably be added eventually, but in a different PR (possibly read-only). This would be the first example of `'show_in_admin' => false`.

### Known Backwards Breaking Changes
- `job_manager_job_listing_data_fields` no longer gets passed `$post_id`. It may not always be set, but if needed you could check the global `$post_id` (which is what was being passed before). For setting values manually, I'd use the new `job_manager_job_listing_wp_admin_fields`, which does get passed the `$post_id`.

### Testing Instructions
- First, make sure no behavior changes when creating/modifying job listings in WP admin in both the block and classic editors. Some fields might have a bit stricter sanitization, but shouldn't break standard usage.
- Test with themes / plugins that add custom fields to the backend editor. These should all work with this PR.
- `GET http://{$HOST}/wp-json/wp/v2/job-listings/` should return listings and their public meta fields under `meta`.
- `GET http://{$HOST}/wp-json/wp/v2/job-listings/{$ID}` should return a specific listing and its public meta fields under `meta`.
- Authenticate as admin (I use [JWT Authentication for WP REST API](https://wordpress.org/plugins/jwt-authentication-for-wp-rest-api/)), but lots of browser+cookie based options to. Verify `_job_expires` is passed along in the above requests.
- `PUT http://{$HOST}/wp-json/wp/v2/job-listings/{$ID}` should allow editing of fields in `meta` (authenticated as admin).
- `POST http://{$HOST}/wp-json/wp/v2/job-listings/` should allow creating job listings as admin with meta fields.
- Try various bad strings to make sure things that should be sanitized are.
- Test the front end job submission form as a guest (make sure settings allow for that), new employer (start as logged out), existing employer (log in first), and admin.

### New Filters
- `job_manager_job_listing_wp_admin_fields` should be used to modify field values/placeholders in WP admin. Do not use to add or remove fields.


jom:
@tripflex This is another one for Field Editor compatibility. There is a lot going on, but most of it is trying to preserve existing behavior without duplicating it for the REST API. I'm trying to engineer this to be as backwards compatible as possible (if not 100%). By default, all custom fields won't have `show_in_rest` so there shouldn't be anything urgent for the Field Editor.

tests/php/tests/includes/rest-api/test_class.wp-job-manager-job-listings.php
```diff
@@ -156,6 +162,268 @@ public function test_get_job_listings_success_guest() {
 		$this->assertResponseStatus( $response, 200 );
 	}
 
+	/**
+	 * Tests to make sure public meta fields are exposed to guest users and private meta fields are hidden.
+	 */
+	public function test_guest_read_access_to_private_meta_fields() {
```
pgk: Yay tests!! Another method name could be `test_guest_can_read_only_public_fields`

tests/php/includes/class-wpjm-base-test.php
```diff
@@ -166,19 +166,21 @@ protected function assertNotPostStatus( $expected_post_type, $post ) {
 		$this->assertNotEquals( $expected_post_type, $post->post_status );
 	}
 
-	protected function get_user_by_role( $role ) {
+	protected function get_user_by_role( $role, $variant = '' ) {
```
pgk: do we need the `$variant` part of it? could we get away with e.g. 

```
return $this->login_as( $this->get_user_by_role( 'employer_b' ) );
```

includes/class-wp-job-manager-rest-api.php
```diff
@@ -0,0 +1,43 @@
...
+	 * @param WP_Post          $post     Post object.
+	 * @return WP_REST_Response
+	 */
+	public static function prepare_job_listing( $response, $post ) {
+		$current_user = wp_get_current_user();
+		$fields       = WP_Job_Manager_Post_Types::get_job_listing_fields();
+		$data         = $response->get_data();
+
+		foreach ( $data['meta'] as $meta_key => $meta_value ) {
+			if ( isset( $fields[ $meta_key ] ) && is_callable( $fields[ $meta_key ]['auth_view_callback'] ) ) {
```
pgk: Maybe also check for `isset( $fields[ $meta_key ]['auth_view_callback'] )`

includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -659,6 +608,15 @@ public function save_job_listing_data( $post_id, $post ) {
 				continue;
 			}
 
+			// Checkboxes that aren't sent are unchecked.
+			if ( 'checkbox' === $field['type'] ) {
```
pgk: +100 for this!

tests/php/includes/class-wpjm-base-test.php
jom: We could, but we'd need to parse the string to determine the actual role to set with `set_role`. This seems a bit cleaner.

includes/class-wp-job-manager-rest-api.php
jom: I've set defaults so it should always be set. Sometimes it could be `null`, but that shouldn't cause a problem with this.

tests/php/tests/includes/rest-api/test_class.wp-job-manager-job-listings.php
jom: Sounds good! 0a6a578df5a7772c80221c3268d173f9148897fc

tripflex:
@jom thanks this def could have potential breaking changes, but as long as the `job_manager_job_listing_data_fields` filter is still called (which it appears to be), I don't see any backwards compatibility issues that stand out to me (right now).

Couple questions:

1.) Will the Resumes addon be updated to follow this same style of handling?
2.) Any target date right now for the minor release (so i can make sure to test heavily before hand)

* as a side note -- very good job with the PHP documentation, i'm very big on that myself and love to see devs in the habit of documentation (always better to have more -- than not enough)

tripflex:
@jom couple things to note while i was searching for anything in my plugins that uses `job_manager_job_listing_data_fields`:

1.) WC Paid Listings relies on `$job_id` being passed, would need to be updated to support this change

I checked in my plugin, and I do not rely on the `$job_id` being passed at all, but maybe for backwards compatibility it may be best to still pass that value using the global `$post` as mentioned?  Or maybe have a method to pull that value from the writepanels class?

Reason being I don't see it having any issues with my plugins -- but my concern is with other addon plugins, or maybe even custom user code that may rely on that value being set

jom:
@tripflex 
> as long as the job_manager_job_listing_data_fields filter is still called (which it appears to be)

Yup, it will still be called, just a bit higher up the call chain. If you transition your usage so that fields can be called in the REST API, it needs to be hooked into on `init` and not `admin_init` (not sure if you were before). Additionally, if you transition to take advantage of the REST API registration, fields should always be included in the returned response and the auth callbacks should be used for visibility/access. For backwards compatibility, though, nothing should need to be changed.

> maybe for backwards compatibility it may be best to still pass that value using the global $post as mentioned? Or maybe have a method to pull that value from the writepanels class?

We could. I had it do that for a while, but I don't want to give the impression that this method will be called in the loop. If there is a way for me to sniff for hooks being added expecting two arguments (so I can throw a deprecation notice), I'd consider it so we can do a deprecation cycle with it. It is relatively new. I added it back when I made the change to WCPC. 

> Will the Resumes addon be updated to follow this same style of handling?

Eventually, but we don't have a timeline for it.

> WC Paid Listings relies on $job_id being passed, would need to be updated to support this change

Yup, this is a good example of a usage that I'll probably just transition to use `job_manager_job_listing_wp_admin_fields` (with version check).

FawadNL:
Just tested this locally and it works perfectly, great job!. When will this be merged?

includes/class-wp-job-manager-rest-api.php
```diff
@@ -0,0 +1,52 @@
...
+ * @since 1.33.0
+ */
+class WP_Job_Manager_REST_API {
+	/**
+	 * Sets up initial hooks.
+	 *
+	 * @static
+	 */
+	public static function init() {
+		add_action( 'rest_prepare_job_listing', array( __CLASS__, 'prepare_job_listing' ), 10, 2 );
```
alexsanford: This should be a filter

includes/class-wp-job-manager-rest-api.php
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	/**
+	 * Filters the job listing data for a REST API response.
+	 *
+	 * @param WP_REST_Response $response The response object.
+	 * @param WP_Post          $post     Post object.
+	 * @return WP_REST_Response
+	 */
+	public static function prepare_job_listing( $response, $post ) {
```
alexsanford: Is this function just hiding the field from being returned? I'm surprised that there isn't a way to add this as a callback to `register_meta`. I wonder if it would make more sense to call the `auth_view_callback` function when we call `register_meta` and use its value to determine `show_in_rest`? So it might look like:

```php
'show_in_rest' => $field['show_in_rest'] && call_user_func( $field['auth_view_callback'], ... ),
```

Maybe I'm bikeshedding here, and if so don't worry about it. It just seems like there should be a better way to do this than to add the field to the API response and then use a filter to remove it?

jom: Sorry, I was going to add a comment and forgot. The reason I did this filtering here instead of when deciding whether or not to `show_in_rest` is because for some auth checks we need `$post` in order to match existing auth check logic (`edit_others_posts` or `job_manager_user_can_edit_job( $post_id )`), which might not be available by the time we register our meta fields. I also expected there to be something for this in `register_meta`, but this was the best I could find.

includes/class-wp-job-manager-rest-api.php
jom: 406ca3ed47dd98ba0283945acf6590cd2a389b9b

PR #1732, opened 2019-04-30
Job Submission Form: Validate Description and Category Fields Client Side

### LINKED ISSUE #1163: Improve client-side UI validation
### Description of the problem
We should consider improving the client-side validation UI to prevent issues like https://github.com/Automattic/WP-Job-Manager/issues/1158. See also https://github.com/Automattic/WP-Job-Manager/issues/842.

Perhaps we could borrow from what WooCommerce does:

account-validation

roundhill:
Add js validation checks for the Description and Category fields on the job submission form.

Fixes #1163

* We add checks in js to validate the fields that don't work `out of the box` using HTML 5 form validation.
* If the validation fails, we use `setCustomValidity` to show a message. Note: This currently uses hard-coded english strings in the js. Any options there for localization?
* I had to do a hacky workaround to get the validation message to show for the `Description` field. Since it uses `tinymce`, the `<textarea>` is hidden, so I had to unhide it and make it 1 pixel tall... seems ok from my testing but could be a bit fragile.

#### Testing instructions:

* Create a job posting, but leave the Description field blank. You should see this message:
Screen Shot 2019-04-30 at 3 24 24 PM

* Enter something in the Description field, and save the posting. It should pass validation and show the preview.
* Enable categories in `/wp-admin/edit.php?post_type=job_listing&page=job-manager-settings#settings-job_listings`.
* Make another posting but don't select a category, you should see a message:
Screen Shot 2019-04-30 at 3 26 31 PM

* Choose a category, and save the submission. It should pass validation and show the preview.


assets/js/job-submission.js
```diff
@@ -3,8 +3,63 @@ jQuery(document).ready(function($) {
 		$(this).closest( '.job-manager-uploaded-file' ).remove();
 		return false;
 	});
-	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function() {
+	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function( event ) {
+		// Validate the category field if present
+		var jobCategory = $( '#job_category' );
```
jom: It is possible that the field has been filtered to not be required. 

Template file [form-fields/term-multiselect-field.php](https://github.com/Automattic/wp-job-manager/blob/332601223930d0a4e26b327700f754c705421e50/templates/form-fields/term-multiselect-field.php) would need to be updated to pass `required` from field settings to `job_manager_dropdown_categories()`. `job_manager_dropdown_categories()` would need to be updated to _probably_ add a class to the field. Adding the HTML5 `required` prop actually seems to do the HTML5 validation to the hidden field (on Chrome, at least), but it is positioned in a strange place (where the hidden field sits above the `select2` field). I'm not sure if this is cross browser (it seems like you had to make the `<textarea>` to be visible to get it to show for you?).

assets/js/job-submission.js
```diff
@@ -3,8 +3,63 @@ jQuery(document).ready(function($) {
...
+			jobCategoryInput.setCustomValidity( 'Category is required' );
+			jobCategoryInput.reportValidity();
+			
+			return;
+		}
+
+		// Validate the description field if present
+		if ( typeof tinymce !== undefined || tinymce.get( 'job_description' ) != null ) {
+			var jobDescription = tinymce.get('job_description').getContent();
+			var editorTextArea = $( '#job_description' );
```
jom: Technically this could also be made optional, but I'm not as concerned about this field. Seems pretty critical to the CPT. 

If we wanted to, it would be passing a `required` class in `editor_class` in [this template's](https://github.com/Automattic/wp-job-manager/blob/332601223930d0a4e26b327700f754c705421e50/templates/form-fields/wp-editor-field.php) editor arguments.

assets/js/job-submission.js
```diff
@@ -3,8 +3,63 @@ jQuery(document).ready(function($) {
...
-	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function() {
+	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function( event ) {
+		// Validate the category field if present
+		var jobCategory = $( '#job_category' );
+		if ( jobCategory.length && !jobCategory.val() ) {
+			event.preventDefault();
+			$(this).find( 'input[type=submit]' ).blur();
+
+			var jobCategoryInput = $( '.select2-search__field' )[0];
+			jobCategoryInput.setCustomValidity( 'Category is required' );
```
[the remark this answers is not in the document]
roundhill: Thanks, I went the `wp_localize_script` route.

assets/js/job-submission.js
```diff
@@ -3,8 +3,63 @@ jQuery(document).ready(function($) {
 		$(this).closest( '.job-manager-uploaded-file' ).remove();
 		return false;
 	});
-	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function() {
+	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function( event ) {
```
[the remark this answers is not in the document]
roundhill: I still left this as is, but abstracted some of the logic to functions to make it easier to grok (hopefully).

assets/js/job-submission.js
roundhill: I've added some checks that the `required-field` class is present. If not there it will skip the validation entirely. 

I modified the test code you sent me to also allow the description field to be optional if you want to test with it:
`add_filter( 'submit_job_form_fields', function( $fields ) {
		  $fields['job']['job_category']['required'] = false;
		  $fields['job']['job_description']['required'] = false;
		  if ( isset( $fields['job']['job_type'] ) ) {
		     $fields['job']['job_type']['required'] = false;
		     $fields['job']['job_description']['required'] = false;
		  }
		  return $fields;
} );`

assets/js/job-submission.js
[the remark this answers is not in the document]
jom: The new abstracted functions are 💯. Now that this event handler is also dealing with validation I still wonder about removing `:not(.prevent-spinner-behavior)` and putting it in a check around the spinner logic below. If a theme for some reason was blocking the spinner behavior, they might still want the fancy validation.

Something like:
```js
if ( ! $(this).hasClass( 'prevent-spinner-behavior' ) ) {
	$(this).find( '.spinner' ).addClass( 'is-active' );
	$(this).find( 'input[type=submit]' ).addClass( 'disabled' ).on( 'click', function() { return false; } );
}
```

jom (approved):
This looks really good! Tested on a few of the major themes and it seems to work well. Left one comment on old thread, but otherwise this looks ready to 🚢  😄 .

It needs a rebase. I was The Cause of these and am happy to do the rebase if you prefer.

assets/js/job-submission.js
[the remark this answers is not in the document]
roundhill: Made that change and rebased too!

assets/js/job-submission.js
```diff
@@ -67,8 +67,96 @@ jQuery(document).ready(function($) {
 		return is_valid;
 	});
 
-	$( document.body ).on( 'submit', '.job-manager-form:not(.prevent-spinner-behavior)', function() {
-		$(this).find( '.spinner' ).addClass( 'is-active' );
-		$(this).find( 'input[type=submit]' ).addClass( 'disabled' ).on( 'click', function() { return false; } );
+	$( document.body ).on( 'submit', function() {
```
jom: I think the `.job-manager-form` selector still makes sense here. I'm not sure we want this to fire for all forms on a page.

roundhill: Right! Updated in 59cfd51

jom: Nice! Looks good and still works well for me. 🚢 🚢

jom:
Looks great! Ready to merge 🚢 🚀

Issue #1728, opened 2019-04-30
Soft Deprecate PHP 5.2 to 5.6.19 Support


jom:
Following slightly behind core, I propose we add a message to 1.33.0 saying that in 1.34.0, WP Job Manager will no longer support PHP versions less than 5.6.20 (core's new minimum as of WordPress 5.2).

PR #1736, opened 2019-05-07
Maintenance: Move main class to its own file and perform PHPCBF

jom:
As a precursor for #1728, I wanted to finally move the main WPJM class to its own file and prepare the `wp-job-manager.php` file for upcoming changes. This is a large PR, but most of the changes are for file headers.

* Moves main `WP_Job_Manager` class to its own file.
* Adds/standardizes file headers in all files in project. Enables the rule to check for this.
* Makes some non-controversial PHPCS changes across project.
* Potentially breaking: There was a random global function and action in the main `wp-job-manager.php` file. In 6111497c23cc35fe7d3f8181df7f5ac649361f13, I moved that to `WP_Job_Manager_Post_Types`. Users removing that filter will now need to use this snippet. They could also remove the post type later in the `post_types_to_delete_with_user`, which is probably what they should have been doing. 
```
add_action( 'init', function() {
    remove_filter( 'post_types_to_delete_with_user', array( 'WP_Job_Manager_Post_Types', 'delete_user_add_job_listings_post_type' ) );
} );
```


jom:
Nice! Merging for now, but definitely welcome post-merge feedback on this one.

roundhill:
This PR was a real `Class act`!

PR #1743, opened 2019-05-09
Change how `[jobs]` maintains state when navigating away and refreshing

### LINKED ISSUE #1680: Odd behaviour with "Load More"
Originally reported here: https://wordpress.org/support/topic/going-back-on-job-listing-doenst-show-all-jobs/

I was able to reproduce.

To reproduce:

1. Visit the page with the [jobs] shortcode
2. Click on Load More if you have more jobs that what shows on the first page of results
3. Click through to one of those jobs
4. Use your browser's Back arrow

**What I expected:**

To be back at the page with the first page of results for the [jobs] shortcode

**What we get:**

A URL with /#s=1 , and only the jobs from the 2nd page of results, not the 1st (and no way to see the "previous" results)

The original reporter stated that this happens in Chrome and Edge, but not Firefox.

### LINKED ISSUE #1712: Add loading indicator back to `[jobs]` shortcode
Similar to #1680 
Caused by #1623

We should add a loading indicator back to the job listings list when listings are being initially loaded. Currently, there is no indicator that jobs are being pulled in the background.

Source: https://wordpress.org/support/topic/job_listings-loading-does-not-work-in-last-version-wpjm/

jom:
Fixes #1712 
Fixes #1680 

This fixes a lot of strange behavior with the `[jobs]` shortcode. Since it is a pretty major refactor, I also ran `prettier` on the file. Eventually I'd like to run this across all JS files in WPJM.

* Adds back in the loading indicator.
* Shows number of results and allows for reset/rss if the only filter change was deselecting a job type.
* Properly sets up the job type checkboxes after refresh or going back in browser history.
* Moves from HTML5's History API to SessionStorage. No more weird hash (#s=1) in URL and allows for multiple forms on same page.

Key changes to behavior:
* Within a browser session, always persist the form values on both refresh and going back to the `[jobs]` page after navigating away.
* Within a browser session, use cached/previously retrieved results when going back to `[jobs]` page _after navigating to a job listing_. If a user clicks elsewhere in the page and then goes back or just refreshes, the form will persist but the results will be reloaded _starting on page 1_. @alexsanford and I talked about this for a while and it seemed to make the most sense. 

#### Caveats/Known Issues

- If you have the Job Tags plugin activated, the tags filter still won't persist. That change will need to be made in the Job Tags plugin. There  is already an open issue for this.
- If you open a job listing in a new tab, the first time you refresh the `[jobs]` page it will persist the results. Refresh again and it will clear those cached results and start back on page 1.

#### Testing instructions:

* Create multiple job listings and make sure they are published. I'd recommend at least 10.
* Create a pages with:
  - `[jobs]`
  - `[jobs per_page="3"]` (change arg) 
  - `[jobs show_pagination="true"]`
  - `[jobs per_page="3" show_filters="false"]` (form not there to persist but results/page should per above)
  - `[jobs per_page="3" show_pagination="true" show_filters="false"]` (form not there to persist but results/page should per above)
* Test to make sure new behavior works across browsers and doesn't break major themes.


assets/js/ajax-filters.js
```diff
@@ -1,288 +1,456 @@
 /* global job_manager_ajax_filters */
-jQuery( document ).ready( function ( $ ) {
+jQuery( document ).ready( function( $ ) {
+	var supports_html5_sessionStorage = false;
+	if ( window.localStorage && typeof window.localStorage.setItem === 'function' ) {
+		supports_html5_sessionStorage = true;
+	}
```
roundhill: What if this were a function instead?

```
function supports_html5_session_storage() {
    return window.localStorage && typeof window.localStorage.setItem === 'function';
}
```

roundhill (changes requested):
I'm having problems with the first three shortcodes, it seems to only work if I have `show_filters=false` in the shortcode.

For the first three shortcodes you wanted tested (like the simple `[jobs]` shortcode), I see the filters/search UI but no job listings show up no matter what I search for. Shouldn't all of the listings show by default? Let me know if you can repro that. I didn't see any errors in the console.

Little UI nit, it's a little strange when two spinners display:
May-13-2019 15-48-44

assets/js/ajax-filters.js
jom: Changed 943aea9530671045dcf6eec6b8fa4102034a251f

jom:
@roundhill 476cc8422f2b41499b4ed8347235ce2fe4865e05 should fix the double indicator. My test environment has featured listings, which have a background color so the indicator doesn't show up.

roundhill (approved):
Working well for me now, thanks @jom!

assets/js/ajax-filters.js
```diff
@@ -1,288 +1,458 @@
...
-			var $keywords  = form.find( ':input[name="search_keywords"]' );
-			var $location  = form.find( ':input[name="search_location"]' );
-
-			// Workaround placeholder scripts
-			if ( $keywords.val() !== $keywords.attr( 'placeholder' ) ) {
-				keywords = $keywords.val();
-			}
+		try {
+			return window.sessionStorage.setItem( session_storage_key, JSON.stringify( state ) );
+		} catch ( e ) {}
```
alexsanford: Could we add a comment saying why it's ok to squelch the exception here?

assets/js/ajax-filters.js
```diff
@@ -1,288 +1,458 @@
...
+
+		var index = $( 'div.job_listings' ).index( $target );
+		var session_storage_key = session_storage_prefix + index;
+
+		try {
+			var state = window.sessionStorage.getItem( session_storage_key );
+			if ( state ) {
+				return JSON.parse( state );
 			}
+		} catch ( e ) {}
```
alexsanford: Could we add a comment saying why it's ok to squelch the exception here?

assets/js/ajax-filters.js
```diff
@@ -1,288 +1,458 @@
...
-		var featured     = target.data( 'featured' );
-		var filled       = target.data( 'filled' );
-		var job_types    = target.data( 'job_types' );
-		var post_status  = target.data( 'post_status' );
-		var index        = $( 'div.job_listings' ).index(this);
-		var categories, keywords, location;
-
-		if ( index < 0 ) {
-			return;
+	function job_manager_store_state( $target, state ) {
```
alexsanford: Could we add some comments for these functions? Even brief ones would be fine. It's not immediately clear (to me, anyway) what the purpose of each function is.

assets/js/ajax-filters.js
```diff
@@ -1,288 +1,458 @@
...
+					scrollTop: $target.offset().top,
+				},
+				600
+			);
+
+			return false;
+		} )
+		.on( 'update_results', function( event, page, append ) {
+			var data = '';
+			var $target = $( this );
```
alexsanford: Are we using the same prettier configuration as Calypso? Seems like it should still align assignments for us, but maybe not. It's been a while since I've looked at Calypso code 🙂

assets/js/ajax-filters.js
```diff
@@ -1,288 +1,458 @@
...
+		$target.triggerHandler( 'update_results', [ page, true ] );
 		return false;
 	} );
 
 	if ( $.isFunction( $.fn.select2 ) && typeof job_manager_select2_args !== 'undefined' ) {
 		var select2_args = job_manager_select2_args;
-		select2_args['allowClear']              = true;
-		select2_args['minimumResultsForSearch'] = 10;
+		select2_args[ 'allowClear' ] = true;
+		select2_args[ 'minimum$resultsForSearch' ] = 10;
```
alexsanford: Is the change to `minimum$resultsForSearch` correct?

assets/js/ajax-filters.js
```diff
@@ -1,288 +1,458 @@
...
+	$( window ).on( 'load', function() {
+		$( 'div.job_listings' ).each( function() {
+			var $target = $( this );
+			var $form = $target.find( '.job_filters' );
+			var $results_loaded = false;
+			var state = job_manager_get_state( $target );
+
+			if ( state ) {
+				// Restore the results from cache.
+				if (state.results) {
```
alexsanford: Check spacing

alexsanford (approved):
I have a few comments regarding formatting and comments, but this seems to work really well! 🚀

assets/js/ajax-filters.js
jom: Yes, I stole the config file from Calypso [[#](https://github.com/Automattic/wp-calypso/blob/master/client/blocks/site-preview/index.jsx#L25-L32)]. We could use our own, though, if we'd prefer.

assets/js/ajax-filters.js
jom: Fixed in e82e1543946ba67dfa5bab01ad1cc9f7ebc7e57b 😄

assets/js/ajax-filters.js
jom: Done in 38b06765d730da12f993c82916a9ba82c9d3eb3f

assets/js/ajax-filters.js
jom: Done in 38b06765d730da12f993c82916a9ba82c9d3eb3f

assets/js/ajax-filters.js
jom: Done in 38b06765d730da12f993c82916a9ba82c9d3eb3f

assets/js/ajax-filters.js
jom: Done in 38b06765d730da12f993c82916a9ba82c9d3eb3f

assets/js/ajax-filters.js
alexsanford: Ok, I was just surprised at the lack of alignment. But if it's consistent with Calypso then it's fine by me 👍

alexsanford:
@jom All of those changes look good! ✅

nonasun:
I use Robolist Lite theme, which uses job search on home page banner. Looks like there's problem with [jobs] shortcode, which caches old search value. Don't know if it's because of this plugin, or it's the theme. Here's the repo:
After setting keyword "tree", and click search button, it goes to a page with [jobs] shortcode, where url is https://2centsrpa.azurewebsites.net/vendors/?search_keywords=tree&search_location=&search_category=
indicating only "tree" keyword is used to search.
Click link back to home page, select a category to search, without setting "tree" in keyword, it goes again to the page with  [jobs] shortcode. Now the url is https://2centsrpa.azurewebsites.net/vendors/?search_keywords=&search_location=&search_category=45, indicating it's using category 45 to search only, no "tree" keyword.
But in the search area, tree also shows and the search result is with tree as a keyword.
Any idea if this issue is known and/or related to this?

jom:
Hi @nonasun! By default, WPJM stores the filter form values on back/refresh and only uses the query string for the first load. The query string isn’t used by default for new searches (after the first load of `[jobs]`), but could be by your theme somehow. It might be worth reaching out to your theme developers. We’d be happy to chat with them to see how we can either get their theme compatible with this PR or see if there is a change we can make in the plugin to make it easier for themes to work with.

nonasun:
@jom Thank you very much for the explanation. But for following searches after the first one, the new search value is always added. E.g. after searching using category 1, go back to home page and search category 2, then both category 1 and category 2 are in the filter. Go back to home page and search category 3, you will see category 1, 2 and 3. So the new value is added. Does WPJM provide function to set a search field, instead of adding? If so, how to do it? Then what the theme should do is to set value for each search criteria using user input each time it searches?
Thanks again.

PR #1647, opened 2018-12-14
List unsubmitted job listings in job dashboard and allow completion

jom:
Fixes #1644

#### Changes proposed in this Pull Request:
I thought this was going to be more difficult. Maybe I'm overlooking something, but this seems to work for me. Have just done some testing but will do more later.

* For `draft` (started in WP Admin) and `preview` (started in frontend) job listings, list them in the `[job_dashboard]` and allow users to continue their submission.
* This does not allow for continuation of listings stuck in `pending_payment` (WC Paid Listings). We'll need to build off this PR in that plugin.
* It shows the post status of `preview` as `Preview`, but that might be confusing. We could map the human readable label for `preview` to `draft`'s label in `get_the_job_status()`. 
* Folks who want to disable this can filter `job_manager_get_dashboard_jobs_args` and remove the post statuses. 

#### Testing instructions:

* Using both WP admin and then again through `[submit_job_listing]` form, start some listings. In WP admin, just save draft. In frontend, stop at Preview stage.
* Visit `[job_dashboard]` page. 
* Click `Continue Submission` under either listing. 
* Should allow you to move to Preview stage and post (assuming no WCPL or similar plugin is active).

Issue #1377, opened 2018-03-09
Add link to Submit New Job after submission


richardmtl:
Requested here: https://wordpress.org/support/topic/go-back-to-create-a-job-offer/

We could add a link to https://github.com/Automattic/WP-Job-Manager/blob/master/templates/job-submitted.php which says something like "Submit another job here" or something.

linoman-zz:
hello dear Richard - many many thanks for bringing this up to our awareness- GREAT!! Thanks!!

donnapep:
Instead of surfacing a link to submit another job as part of the confirmation notice, would it make sense to display the confirmation notice and show the form to submit another job below it? It would save the user an extra click at least and streamline the process for those who have multiple jobs to post.

PR #1746, opened 2019-05-12
Introduce Save Draft button on job submission form

### LINKED ISSUE #505: Add ability to save job drafts to edit or submit them later
Requested in Help Scout ticket 6624

jom:
Fixes #505 

* This adds a `Save Draft` button on the job submission form. It builds off of the functionality in #1647. 
* New filter: `submit_job_form_can_resume_later` (prefix matches other filters on submission form).
* Set button after Preview step so that visitors who pressed return/enter while in a text field would move through submission process.
* Added `secondary` class so that themes could make the button different/less prominent that main button.
* ~Note: HTML5 + server side validation must pass in order to save as a draft. We might be able to get around this for some fields, but we need it for fields related to registration. I think if we tackle this, it would be in a second phase.~ 4f7e426c0bd5d8ceefd863e70fc50c94b098e46c

To disable: `add_filter( 'submit_job_form_can_resume_later', '__return_false' );`

#### Testing instructions:

* Ensure jobs can be saved as drafts for logged in users or when user registration is enabled/forced. It shouldn't show up for guests or when users are unable to create accounts from the job submission page (job submission settings).
* Once saved, drafts should be accessible and resume-able from the `[job_dashboard]` page.

* Feature: Allow job listings to be saved as drafts during the submission process.


jom:
@roundhill 
> It would be nice to create a new listing from the job dashboard. Maybe this is #1377?

Seems like a separate issue, but I agree. Seems like a new enhancement request.

> You have to fill out all of the required fields before you can save the draft, which may defeat the purpose of saving a draft for some users?

Yup, caught that as well (see PR description). I think in order to fix it, we'll need to introduce a "draft required" field (essentially the registration fields). I can work on that in this round.

jom:
Rebased with changes from `master`.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -548,6 +548,7 @@ public function submit() {
 				'job_fields'         => $this->get_fields( 'job' ),
 				'company_fields'     => $this->get_fields( 'company' ),
 				'step'               => $this->get_step(),
+				'can_resume_later'   => $this->can_resume_later(),
```
alexsanford: Since the word "resume" (meaning "continue") in code is spelled the same as "resumé" (as in, CV), I wonder if it would be better to use a different word to avoid confusion?

Maybe `can_continue_later`?

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -566,16 +567,20 @@ public function submit_handler() {
 			// Get posted values.
 			$values = $this->get_posted_fields();
 
-			if ( empty( $_POST['submit_job'] ) ) {
+			$is_saving_draft = $this->can_resume_later() && ! empty( $_POST['save_draft'] );
```
alexsanford: See comment above regarding use of the word `resume`.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -983,6 +996,34 @@ public function done_before() {
...
+			// If these are enabled, we know an account will be created on save.
+			$can_resume_later = true;
+		}
+
+		/**
+		 * Override if visitor can resume job submission later.
+		 *
+		 * @param bool $can_resume_later True if they can resume job later.
+		 */
+		return apply_filters( 'submit_job_form_can_resume_later', $can_resume_later );
```
alexsanford: See comment above regarding use of the word `resume`.

assets/js/job-submission.js
```diff
@@ -50,6 +50,21 @@ jQuery(document).ready(function($) {
 
 	$( '.fieldset-type-file input[type=file][multiple][data-file_limit]' ).trigger( 'update_status' );
 
+
+	$( document.body ).on( 'click', '#submit-job-form .button.save_draft', function() {
+		var $form    = $(this).closest( '#submit-job-form' );
+		var is_valid = true;
+
+		$form.find( 'div.draft-required input[required]').each( function() {
```
alexsanford: Do you have a way to test this? When I inspect the HTML I don't see any fields with a class of `draft-required`.

jom: It is on the registration fields. You have to be signed out.

includes/forms/class-wp-job-manager-form-submit-job.php
jom: Yup, I was wondering the same but I thought it was fine as a verb. I'll change it though. 71e145dc93ed39f42a3ca2f6b4480448ffe325a7

assets/js/job-submission.js
alexsanford: Cool, thanks.

When I leave the required fields blank and try to save the draft, I get the validation notice for the _last_ required field in the form. This is backwards from the default browser behaviour, where we get the validation for the _first_ required field. Is this something we can fix?

jom: Yup f3ab44aeed6871022a0a08cb73c548444071adba

PR #1759, opened 2019-05-30
Create a more unique cache storage key per `[jobs]` instance

jom:
Fixes an issue in 1.33.0-beta.1 where multiple `[jobs]` instances across pages would collide and have their state shared. I considered just using either `window.location.href` (or variation of that and/or hash), but this seemed a bit cleaner.

* Include the `post_id` when generating the key used in `sessionStorage`.

#### Testing instructions:

* Verify tests from #1743 still pass (shouldn't need to make multiple variations with arguments).
* Create multiple pages with a `[jobs]` shortcode. 
* Within the same tab/session, visit one, enter a filter and/or load more pages of results, and then visit the other page with a `[jobs]` shortcode. Verify the filter form values from the first page don't show up on the second page.


alexsanford (approved):
Looks good, works well! 🚀

PR #1774, opened 2019-06-05
Sanitize and unslash additional input

### LINKED ISSUE #1769: Maintenance: Fix input not sanitized notices
Building off of #1760 

Remove the following from `phpcs.xml.dist` and then fix the new issues when running `phpcs`.

Replace:
```xml
	<rule ref="WordPress.Security.ValidatedSanitizedInput.InputNotSanitized">
		<severity>4</severity>
	</rule>
```

and
```xml
	<rule ref="WordPress.Security.ValidatedSanitizedInput.MissingUnslash">
		<severity>4</severity>
	</rule>
```

with this in the official rules section:
```
	<rule ref="WordPress.Security.ValidatedSanitizedInput"/>
```

jom:
Fixes #1769

* Adds further sanitization and unslashing to input (`$_GET`, `$_REQUEST`, `$_POST`) and `$_SERVER`.

#### Review instructions:

This should get in early in 1.34.0's cycle. The main risks here/things to double check 
- I've sanitized an array using `sanitize_text_input` directly instead of using `array_map`.
- I've sanitized a string too strictly. I don't recall seeing controversial things... most actual stored user input was sanitized on storage earlier.

I've done a self-review to double check, but could definitely use another look.

For nonce checks, I've opted to `wp_unslash` the input and ignore the sanitization warning. There is an interesting discussion about this in a [WPCS issue](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/869). I don't think we should modify the nonce at all. Unslashing is just undoing something that WP does earlier in the call.


includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -603,7 +605,7 @@ protected function upload_file( $field_key, $field ) {
 			}
 
 			$file_urls       = array();
-			$files_to_upload = job_manager_prepare_uploaded_files( $_FILES[ $field_key ] );
+			$files_to_upload = job_manager_prepare_uploaded_files( $_FILES[ $field_key ] ); //phpcs:ignore WordPress.Security.ValidatedSanitizedInput
```
jom: See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1720 for context

includes/class-wp-job-manager-ajax.php
```diff
@@ -118,27 +117,24 @@ public static function do_jm_ajax() {
...
 		$per_page           = isset( $_REQUEST['per_page'] ) ? absint( $_REQUEST['per_page'] ) : absint( get_option( 'job_manager_per_page' ) );
 
 		if ( is_array( $search_categories ) ) {
 			$search_categories = array_filter( array_map( 'sanitize_text_field', array_map( 'stripslashes', $search_categories ) ) );
 		} else {
-			$search_categories = array_filter( array( sanitize_text_field( stripslashes( $search_categories ) ) ) );
+			$search_categories = array_filter( array( sanitize_text_field( wp_unslash( $search_categories ) ) ) );
 		}
 
+		$types              = get_job_listing_types();
```
jom: I moved this from above to make that block just have input

includes/abstracts/abstract-wp-job-manager-form.php
GaryJones: When ignoring, use the full `Standard.Category.SniffName.ViolationCode`, not just `Standard.Category.SniffName`. Otherwise you might be ignoring other checks that are in / added to the same sniff.

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -492,7 +492,9 @@ protected function get_posted_field( $key, $field ) {
 		if ( ! isset( $field['sanitizer'] ) ) {
 			$field['sanitizer'] = null;
 		}
-		return isset( $_POST[ $key ] ) ? $this->sanitize_posted_field( $_POST[ $key ], $field['sanitizer'] ) : '';
+
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
GaryJones: I guess the violation here is because it's being run through a custom sanitization method, and not one of the usual recognised functions? If so, add a reason to the ignore to explain as much, to make it more obvious:

```
// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- $this->sanitize_posted_field() checks for foo and then calls sanitize_text_field()...
```

(for example).

includes/abstracts/abstract-wp-job-manager-form.php
GaryJones: Add this PR comment as a reason on the ignore:

```
//phpcs:ignore WordPress.Security.ValidatedSanitizedInput - See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1720.
```

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -105,7 +105,7 @@ public static function init_core_notices() {
 	 */
 	public static function dismiss_notices() {
 		if ( isset( $_GET['wpjm_hide_notice'] ) && isset( $_GET['_wpjm_notice_nonce'] ) ) {
-			if ( ! wp_verify_nonce( sanitize_key( wp_unslash( $_GET['_wpjm_notice_nonce'] ) ), 'job_manager_hide_notices_nonce' ) ) {
+			if ( ! wp_verify_nonce( wp_unslash( $_GET['_wpjm_notice_nonce'] ), 'job_manager_hide_notices_nonce' ) ) { // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
GaryJones: Add a comment, to briefly explain the reason for the ignore - i.e. not sanitizing the nonce.

includes/admin/class-wp-job-manager-writepanels.php
```diff
@@ -631,15 +634,15 @@ public function save_job_listing_data( $post_id, $post ) {
...
+					update_post_meta( $post_id, $key, date( 'Y-m-d', strtotime( sanitize_text_field( wp_unslash( $_POST[ $key ] ) ) ) ) );
 				}
 			} elseif ( '_job_author' === $key ) {
 				if ( empty( $_POST[ $key ] ) ) {
 					$_POST[ $key ] = 0;
 				}
 				$wpdb->update( $wpdb->posts, array( 'post_author' => $_POST[ $key ] > 0 ? absint( $_POST[ $key ] ) : 0 ), array( 'ID' => $post_id ) );
 			} elseif ( isset( $_POST[ $key ] ) ) {
-				update_post_meta( $post_id, $key, $_POST[ $key ] );
+				update_post_meta( $post_id, $key, wp_unslash( $_POST[ $key ] ) ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Input sanitized in registered post meta config
```
GaryJones: Maybe include a file reference, so someone can more easily go and find out what the sanitization is exactly?

```
// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Input sanitized in registered post meta config - see /includes/foo/post-meta.php
```

includes/abstracts/abstract-wp-job-manager-form.php
jom: We need to ignore both the Unslash and Sanitization here. I couldn't find where to ignore multiple rules on the same line. Do you know how?

GaryJones: Just use comma separation:

```
// phpcs:ignore A.B.C.D, A.B.C.E
```

(or maybe remove the space after the comma if that doesn't work)

jom: That works! Very long line.

PR #1783, opened 2019-06-06
Prevent loading of listings when first page of results come preloaded

jom:

* In version 1.32.3, `ajax-filters.js` didn't load listings on page load if there were no job filters (`[jobs show_filters=false]` pre-populates the first page of results. This restores that functionality.

#### Testing instructions:

* On Jobify, view a taxonomy archive page. They generate a listing of results inside of a `<div class="job_listings">...</div>` and mock the look of `[jobs]`. This will keep `ajax-filters.js` from loading with an empty query.
* Create a page with `[jobs show_filters="false"]` and verify it loads jobs correctly on multiple pages and maintains state (if you have two pages loaded, click on a listing in result, and click browser's back button, 2 pages should be loaded right away).


jom:
You can see the original functionality here:
https://github.com/Automattic/WP-Job-Manager/blob/1.32.3/assets/js/ajax-filters.js#L267-L285

PR #1785, opened 2019-06-06
Allow multiple file uploads if limit is not set

jom:

* Fix the logic that limits to 1 file on non-multiple files.

#### Testing instructions:

* Test with testing instructions in #1783.
* Also test with this snippet and verify there isn't a limit on number of files that can be uploaded
```
  $fields['job']['job_files'] = [
	'label'              => __( 'Job Files', 'wp-job-manager' ),
	'type'               => 'file',
	'required'           => false,
	'placeholder'        => '',
	'priority'           => 6,
	'ajax'               => false,
	'multiple'           => true,
  ];
  
  return $fields;
} );
```


jom:
This seems like an obvious typo that I didn't catch. I'm going to merge for now, but feel free to post merge review.

PR #1793, opened 2019-06-11
Revert $form_data to form_data in submitted Ajax filter data

jom:
Fixes issue where submitted data for Ajax filters was incorrectly changed in 1.33.0 by my commit from `form_data` to `$form_data`. This isn't used in core but is by themes/plugins.

This should fix the breakage from https://wordpress.org/support/topic/search-by-regions-and-keywords-not-working-since-1-33-0/


andreilupu:
This PR would help a lot so we don't have to change the form_data key in quite a few themes and plugins 🙏

PR #1801, opened 2019-06-19
Fix PHPCS errors around lack of nonce checking

### LINKED ISSUE #1768: Maintenance: Fix nonce check validation notices
Building off of #1760 

Remove the following from `phpcs.xml.dist` and then fix the new issues when running `phpcs`.
```xml
	<rule ref="WordPress.Security.NonceVerification">
		<severity>4</severity>
	</rule>
```

I imagine most if not all will be ignored, but it is worth checking to see if we can add additional nonce checks.

jom:
Fixes #1768

* Adds nonce checking when hiding license notifications in WP admin. It isn't serious, but does technically change data. 
* Everything else either has a nonce check when possible or the data is use for flow/filtering and can be considered safe.
* Note: Some methods required a slight refactor in order to prevent unreadable code covered in comments. When necessary, I chunked together some reading of `$_GET` and `$_POST` variables and wrapped in `phpcs:disable` and `phpcs:enable`. I used this sparingly and tried to keep it in tight blocks of code.

Style notes: I'm trying to put ignore statements above the line they are ignoring. Some exceptions (put at the end of the line):
- When it is embedded in a [multiline] `if()` statement.
- When it hinders readability (sometimes within blocks of variable sets).

### DIFF
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -270,9 +270,11 @@ class WP_Job_Manager_CPT {
 	public function action_notices() {
 		global $post_type, $pagenow;
 
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended -- No changes made from input.
 		$handled_jobs    = isset( $_REQUEST['handled_jobs'] ) && is_array( $_REQUEST['handled_jobs'] ) ? array_map( 'absint', $_REQUEST['handled_jobs'] ) : false;
 		$action          = isset( $_REQUEST['action_performed'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['action_performed'] ) ) : false;
 		$actions_handled = $this->get_bulk_actions();
+		// phpcs:enable WordPress.Security.NonceVerification.Recommended
 
 		if (
 			'edit.php' === $pagenow
@@ -329,6 +331,7 @@ class WP_Job_Manager_CPT {
 			),
 		);
 
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- No changes or data exposed based on input.
 		$selected_category = isset( $_GET['job_listing_category'] ) ? sanitize_text_field( wp_unslash( $_GET['job_listing_category'] ) ) : '';
 		echo "<select name='job_listing_category' id='dropdown_job_listing_category'>";
 		echo '<option value="" ' . selected( $selected_category, '', false ) . '>' . esc_html__( 'Select category', 'wp-job-manager' ) . '</option>';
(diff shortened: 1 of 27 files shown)


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -521,7 +525,7 @@ public function output_job( $atts ) {
 		);
 
 		if ( ! $atts['id'] ) {
-			return;
+			return null;
```
jom: This was to get rid of a sniff issue. It wasn't a bug, but it doesn't like `void` being used for a string.

includes/admin/class-wp-job-manager-cpt.php
```diff
@@ -329,6 +331,7 @@ public function jobs_by_category() {
 			),
 		);
 
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- No changes or data exposed based on input.
 		$selected_category = isset( $_GET['job_listing_category'] ) ? sanitize_text_field( wp_unslash( $_GET['job_listing_category'] ) ) : '';
```
pgk: This along with other similar checks could be potentially refactored to a single method.

something like `<somename>::get_sanitized_var( $var, $default='' );`

includes/admin/class-wp-job-manager-permalink-settings.php
```diff
@@ -135,26 +135,32 @@ public function settings_save() {
...
-			$permalink_settings['job_base']      = isset( $_POST['wpjm_job_base_slug'] ) ? sanitize_title_with_dashes( wp_unslash( $_POST['wpjm_job_base_slug'] ) ) : '';
-			$permalink_settings['category_base'] = isset( $_POST['wpjm_job_category_slug'] ) ? sanitize_title_with_dashes( wp_unslash( $_POST['wpjm_job_category_slug'] ) ) : '';
-			$permalink_settings['type_base']     = isset( $_POST['wpjm_job_type_slug'] ) ? sanitize_title_with_dashes( wp_unslash( $_POST['wpjm_job_type_slug'] ) ) : '';
+		$permalink_settings = WP_Job_Manager_Post_Types::get_raw_permalink_settings();
 
-			if ( isset( $_POST['wpjm_job_listings_archive_slug'] ) ) {
-				$permalink_settings['jobs_archive'] = sanitize_title_with_dashes( wp_unslash( $_POST['wpjm_job_listings_archive_slug'] ) );
-			}
+		// phpcs:disable WordPress.Security.NonceVerification.Missing -- WP core handles nonce check for settings save.
+		$permalink_settings['job_base']      = isset( $_POST['wpjm_job_base_slug'] ) ? sanitize_title_with_dashes( wp_unslash( $_POST['wpjm_job_base_slug'] ) ) : '';
```
pgk: Kinda similarly, these three (and maybe more in other places) could be refactored to a method

includes/admin/class-wp-job-manager-cpt.php
jom: WPCS sniffs for those (or similar) functions. We could create a custom _global function_ (not class method) and set the `customUnslashingSanitizingFunctions` in PHPCS, but I've been trying to keep new global functions to a minimum.

includes/admin/class-wp-job-manager-permalink-settings.php
jom: Ditto for previous comment.

PR #1805, opened 2019-06-24
Try not to preserve form on refresh

### LINKED ISSUE #1804: ajax-filters.js conflicting with search query parameters passed in URL via a custom search form.
When passing parameters from a search form (as instructed by [Automattic](https://wpjobmanager.com/document/tutorial-creating-custom-job-search-form/) the results of any previous searches are what's retained in job-filters.php input(s) and the URL parameters are ignored.

The values are correctly passed. but ajax-filters.js is changing these to whatever was previously searched for on the page and doesn't return anything based on the input that was entered via the custom search form.

Using the latest version of WordPress 5.2.2 and WP Job Manager 1.33.2
Capture

jom:
Fixes #1804

* This would change how the form is preserved. We could use `performance.navigation.type` to detect reload/refresh, but it might not be consistent in certain browsers.
* This persists the form value on two common non-refresh events: form submission and link press (any on page). 
* It is a bit of a hack. Welcome to hear of alternatives.

_Note_: This is built off of `version/1.33` and could easily be moved to a point release if necessary.


richardmtl:
Note that https://github.com/Automattic/WP-Job-Manager/issues/1804#issuecomment-516562303 says that the patch didn't work for them.

PR #1825, opened 2019-07-22
Fix plugin activation issue

alexsanford:
The plugin activation hook for WPJM was not running properly. This PR ensures that the hook is run, and fixes some related issues with deactivation and i18n loading.

* Move activation and deactivation hook setup back to the root WPJM file.
* Access language files using the `JOB_MANAGER_PLUGIN_DIR` constant, rather than `__FILE__` (which is no longer correct).

#### Package

[wp-job-manager-255537b.zip](https://github.com/Automattic/WP-Job-Manager/files/3418673/wp-job-manager-255537b.zip)

#### Testing instructions:

To test one of the symptomatic issues ([#](https://wordpress.org/support/topic/something-critically-wrong-with-the-latest-update/)):

* Create a clean WordPress instance.
* Install and activate WP Job Manager on this branch (or using the package linked above).
* **Immediately after activation, before refreshing or going to another page,** hover over the Job Listings sidebar menu item.
* Ensure that the links are all there (i.e. "All Jobs", "Add New", etc).

### DIFF
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -96,10 +96,6 @@ class WP_Job_Manager {
 		add_filter( 'wp_privacy_personal_data_exporters', array( 'WP_Job_Manager_Data_Exporter', 'register_wpjm_user_data_exporter' ) );
 
 		add_action( 'init', array( $this, 'usage_tracking_init' ) );
-		register_deactivation_hook( __FILE__, array( $this, 'usage_tracking_cleanup' ) );
-
-		// Other cleanup.
-		register_deactivation_hook( __FILE__, array( $this, 'unschedule_cron_jobs' ) );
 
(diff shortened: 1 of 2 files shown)


adampickering:
Looks good apart from the suggestion of enabling “listing categories” default upon activation of the plugin, this often trips up a lot of people trying to find the setting.

adampickering:
Good to go! Test on local installs.

pgk (commented):
Code looks good! tested on a fresh local site and got this on hover:

Screenshot 2019-07-23 at 02 04 05

And then on refresh, this:

Screenshot 2019-07-23 at 02 04 51

Seems like the job types link was not there on first hover, but came in after the refresh.

pgk (approved):
Lgtm and works as advertised! The job type link seems to be unrelated to this and a different non-priority thing.

PR #1829, opened 2019-07-24
#1814 submitting existing jobs prefill fix

### LINKED ISSUE #1814: Submitting existing job doesn't pre-fill with posted values
WP Job Manager will use posted values to pre-fill form data in case when listing is new, but it doesn't seem to do that when editing existing listing, and instead pre-fills the form data with existing listing data (if a server side validation error occurs). 

#### Steps to Reproduce
1. Visit job dashboard, edit existing listing
2. Make a change to any field
3. Bypass client side validation (html5 required attribute) for some required field and make that field empty to trigger server side validation error
4. Any changes made in 2. are lost

Reproducible on https://demo.wpjobmanager.com

#### Screenshot / Video
https://www.dropbox.com/s/ccvdsilgaqvjzvp/2019-07-01_04-57-22.mp4?dl=0

ovidiul:
Fixes #1814 

*when submitting the form data of an existing job post, upon submit, the form tries to always fetch the saved job data and not the pending object, so we can test that if a form error occured, we keep the proposed object for further editing in frontend.

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -514,7 +514,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		$this->init_fields();
 
 		// Load data if necessary.
-		if ( $this->job_id && !$this->errors ) {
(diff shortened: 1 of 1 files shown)


gikaragia:
Done as part of #2103

PR #1831, opened 2019-07-24
#1815 selecting all job types filter fix search display

### LINKED ISSUE #1815: Selecting all job types removes the search results count
#### Steps to Reproduce
1. On the jobs page, select all but one job types:

image

2. Note that it shows "Search completed. Found X matching records"
3. Now select all job types...

#### What I Expected

It would still say "Search completed. Found X matching records"

#### What Happened Instead

That section disappears:

image

5.2.2 / 1.33.2

#### Context / Source

Reported in 2152212-zen

ovidiul:
Fixes #1815

*this fixes the "Search completed. Found %d matching record." display by adjusting the $job_types_filtered parameter to correctly output true value when all job categories are searched

fixes the search ajax response when all filtered ategories are being checked

### DIFF
--- a/includes/class-wp-job-manager-ajax.php
+++ b/includes/class-wp-job-manager-ajax.php
@@ -142,7 +142,7 @@ class WP_Job_Manager_Ajax {
 		}
 
 		$types              = get_job_listing_types();
-		$job_types_filtered = ! is_null( $filter_job_types ) && count( $types ) !== count( $filter_job_types );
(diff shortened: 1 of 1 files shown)


alexsanford:
As per @jom's comment (https://github.com/Automattic/WP-Job-Manager/issues/1815#issuecomment-529416565) I'm not sure we want to do this. When all the job types are checked, this is the default state, and I don't think it makes sense to display the "Search Completed" bar with the "Reset" link (unless there is another field being searched as well).

jom:
Closing for now. I think we don't want to show this when we have the default state.

PR #1848, opened 2019-08-06
Added page path into the result.

scholz-timo:
Fixes #

Added page path to the get_job_listings cache key.
*

#### Testing instructions:
- Install qtranslate-xt (https://github.com/qTranslate/qtranslate-xt)
- Add 2 Languages.
- Run setup for wp-job-manager.
- Provide translations for all the wp-job-manager pages.
- Create jobs and provide translations.
- Go to the job listing page and switch the language.
*


jom:
Closing this. I think it will cause some unnecessary cache duplication for many users. I suggest you inject this into the query args using the `get_job_listings_query_args` filter for now.

PR #1851, opened 2019-08-07
Enhancement: Added setting to bypass trash for job listings.

JuanchoPestana:

On the Settings screen, the "Delete Data On Uninstall" says: "Once removed, this data cannot be restored.". Now, strictly speaking (and I'm well aware this is not a critical issue) the job listings CAN be restored, since the uninstall procedure uses the 'wp_trash_post()' which sets the job listings posts with a "trash" status. If a user, after deleting the plugin (with the 'Delete Data on Uninstall' option active) re-installs the plugin, the job listings appear on the trash for 30 days and they can be restored. This means that, after deleting the plugin, the WordPress Database will store information (wp_posts table and wp_postmeta table) for 30 days.

In my opinion (I might be wrong!) this is a bit confusing for a user, since the plugin specifically says that "data cannot be restored". I think this may lead the user to the false conclusion (it did to me) that the plugin is deleting EVERYTHING from the database when it really isn't.

I propose the following change:
I'm attaching a screenshot that shows it better, but basically it's to modify the text (description) of the current 'Delete Data On Uninstall' Setting and to add a new Setting that asks the user if they want to bypass the trash for job listings or not.
This could also be applied for the 'jobs', 'job-dashboard', 'post-a-job' (or whatever name chosen by the user) pages created, which, when uninstalling the plugin, are sent to the trash. But this is out of the scope of this PR.

The implementation of my suggestion is in this PR, just in case it's considered a good idea.

Please let me know your thoughts on this, if I'm missing something, or if I'm complicating things instead of helping!

#### Testing instructions:

* Install the plugin (with this PR code)
* Go to the Settings page
* Activate the "Bypass Trash For Job Listings" setting
* Create a new job listing
* Delete the plugin
* Check the wp_posts and wp_postmeta tables. No data from the created job listing should exist

Added setting to bypass trash for job listings.

#### Screenshot:
enhancement


onubrooks:
Hey there @JuanchoPestana 
Thanks for your contribution. This is interesting work. Do you mind rebasing?

JuanchoPestana:
@onubrooks - apologies for not getting to this in time. It's been a couple of fast-paced weeks :). Thanks for helping out with this!

Issue #1856, opened 2019-08-12
job-submission.min.js throws js error expects job_description to be WP Editor


tripflex:
Resolved in #1808 -- but it seems that never made it's way into the minified version.

When i submitted the PR should I have included the updated `.min.js` version as well, or should minification be ran before release (for all files)?

I can submit a PR for the updated `job-submission.min.js` but want to make sure we get the process of handling minified files moving forward to some kind of standard so we don't come  across this issue again 👍

roundhill:
I think we should minify it as part of the PR. Any thoughts @pgk or @jom?

I created a PR for the minified file: #1859

tripflex:
Closing this as resolved in #1859

Issue #1808, opened 2019-06-25
job-submission.js throws js error expects job_description to be WP Editor


tripflex:
https://github.com/Automattic/WP-Job-Manager/blob/7b9df224e2bfc6107e8a0f3069eaaec4ef94a01b/assets/js/job-submission.js#L154-L165

If a user modifies (via plugin or PHP) the job description field to change it to say, a textarea, this will result in JS error on the page, as TinyMCE might not even be set.

Probably should add a conditional if statement around it to check if tinymce even exists on window object before assuming it does

tripflex:
Disregard this .. not sure how I didn't see the check right above it ..durp!

jom:
Actually, I’m wondering if this should be `&&` instead of `||`.
https://github.com/Automattic/WP-Job-Manager/blob/7b9df224e2bfc6107e8a0f3069eaaec4ef94a01b/assets/js/job-submission.js#L142

The check in the `setTimeout()` looks good, though.

tripflex:
@jom yeah i agree it should be `&&` because tinymce should exist AND job_description not equal to null

tripflex:
So i'm going to re-open this issue, as this is still a problem, but it's due to checking against `undefined` without quotes, when it should be in quotes

tripflex:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#Typeof_operator_and_undefined

Example:
https://glot.io/snippets/fdt65d486p

`undefined` will only be true when the var has been set, but not defined, ie:

```javascript
var x;
```

But in the case where `job_description` is not using WP Editor (or any other field on the page), `tinymce` is never actually loaded on the page, so it must be checked against `"undefined"`

PR #1859, opened 2019-08-13
Fixed check for existing tinymce description field in job-submission.js

roundhill:
Looks like the operator was incorrect in the `descriptionFieldIsPresent()` function, it should use an `&&` to only pass if `tinymce` var exists AND we have a `job_description` field.

I also ran the `gulp` command to minify the job-submission.js file. See #1856.

**To Test**
* Verify that an empty job description field that is using tinymce shows a validation error.
* Change the field type from `wp-editor` to `text` [at this line](https://github.com/Automattic/WP-Job-Manager/blob/master/includes/forms/class-wp-job-manager-form-submit-job.php#L240), and verify that it also shows a verification message with no errors in the js console.


tripflex:
@roundhill yeah we even talked about it being `&&` but I must have forgot to include that in the PR https://github.com/Automattic/WP-Job-Manager/issues/1808#issuecomment-505571813
https://github.com/Automattic/WP-Job-Manager/issues/1808#issuecomment-505572780

Whoops!  Looks good to me though, I will make sure from now on to also include the minified file in any PR submitted as well.

Going to test it on my local dev site just to make sure and will report back here shortly.

assets/js/job-submission.js
```diff
@@ -153,7 +153,7 @@ jQuery(document).ready(function($) {
 
 	// Listen for changes to the description field to clear validity
 	setTimeout( function() {
-		if ( typeof tinymce === "undefined" || tinymce.get( 'job_description' ) == null ) {
+		if ( !descriptionFieldIsPresent() ) {
```
alexsanford: Should add a space after the `!` here.

roundhill: Thanks, fixed it in two places in f687665.

PR #1869, opened 2019-08-28
Add a filter for determining if job_dashboard_handler() should be ran.

### LINKED ISSUE #1858: Add a filter to shortcode_action_handler() method to run job_dashboard_handler() method.
Hi, 

I am trying to use the `[job_dashboard]` in a template using `do_shortcode()` method, but this way, the `job_dashboard_handler()` method of the `WP_Job_Manager_Shortcodes` class doesn't run since you have` is_page()` and `has_shortcode()` flags in the `shortcode_action_handler()` method where the earlier method was called. 

Is it possible to add a filter in `shortcode_action_handler()` method so that we can conditionally run the `job_dashboard_handler()` method???

Maybe something like this: 
```
public function shortcode_action_handler() {
		global $post;
		
		if( is_page() && has_shortcode( $post->post_content, 'job_dashboard' ) ) {
			$run_handler = true;
		} else {
			$run_handler = false;
		}
		
		$run_handler = apply_filters( 'wannna_run_this', $run_handler );
		
		if( $run_handler ) {
			$this->job_dashboard_handler();
		}
	}
```

Thanks

pgk:
Fixes #1858 

Adds a filter to the `shortcode_action_handler()` method. 
Not sure if it is really necessary or a good idea to tamper with this anyway.

* Add a filter to the `shortcode_action_handler()` method.


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -69,7 +69,16 @@ public function __construct() {
 	public function shortcode_action_handler() {
 		global $post;
 
-		if ( is_page() && has_shortcode( $post->post_content, 'job_dashboard' ) ) {
+		/**
+		 * Determine if the shortcode action handler should run.
+		 *
+		 * @param bool $should_run_handler Should the handler run.
+		 *
+		 * @return bool
```
alexsanford: As per the [guide](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters), the `@return` statement should not be used.

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -69,7 +69,16 @@ public function __construct() {
...
+		 */
+		$should_run_handler = apply_filters( 'job_manager_should_run_shortcode_action_handler', is_page() && has_shortcode( $post->post_content, 'job_dashboard' ) );
```
alexsanford: I wonder if we should include more context here? I guess there isn't much we can supply that isn't already accessible from globals. Any thoughts?

PR #1798, opened 2019-06-17
Fix PHPCS errors around direct database access and lack of caching

jom:
Fixes #1770

## Changes proposed in this Pull Request:

This PR makes a few more potentially breaking changes than previous PHPCS PRs.

### Ignored Warnings

I didn't see any elegant alternatives for any of these.

- Cache version deletion: The prefix is dynamic and should work with WPJM plugins. I didn't see a workaround for this. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/58e34723976d9cadc580481d1fc8099d316aa242
- Cache manager count handling: Similar to above the prefix is dynamic. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/6dc0367a0e517b0df0e39853171307b9c4175b73
- Post search in WP admin: Previous behavior searches all meta fields here. `WP_Query` connects the meta query with `AND` so it would have to match one of the core post fields (title, content, etc) _and_ a meta field. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/4db57fd557d8eec6b70f4d830fa269244216f015
- Installer data update for menu order: This seems the easiest way to update all job listings without loading them all. We could use `wpdb::update()`, but this seems fine for legacy data-update code. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/12703dcdd67c756f6a643d640f2b367eb508effd
- Retrieving raw meta values when duplicating post: Easiest way to receive raw values of meta data when duplicating a job listing. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/2def4869f067b0557b1069c4508690b8d1125612

### Fixed

These had better alternatives that we could use. I either made existing tests more robust or created new tests to check for breaks in behavior.

- `uninstall.php`: Switched to using `get_sites()` to retrieve all sites we should remove data on MS. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/39473df4c4569b3950efecc4d885aa378788ae1e
- Update author field: Switched to using `wp_update_post()` and removing actions to prevent calling self. https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/ac6c9f42d46982d6205880dbd1f8932ff143fb4f
- Use `WP_Query`/`get_posts` to retrieve expired job listings https://github.com/Automattic/WP-Job-Manager/pull/1798/commits/69c51d4df4fd36b54fea35eed4706dc86153fcf6 
[...]

PR #1876, opened 2019-09-03
Revert removal of direct database access in two places

### LINKED ISSUE #1843: WPJM 1.33.4 breaks functionality of "Download Attachments" plugin
Reported in 2240943-zen

Steps to reproduce:

1) On a site with WPJM 1.33.3, install and activate https://wordpress.org/plugins/download-attachments/
2) In Settings -> Attachments, make sure "All users" and "job_listing" are selected, and Save
3) Create new job listing from Job listings -> Add New
4) In the Download Attachments metabox, add a PDF file, and click Save in that same meta box. 
5) Publish the job listing; refresh the Edit Listing page in WP-Admin; note that the PDF attachment is still there
6) Update to WPJM 1.33.4
7) Follow steps 4 and 5 again
8) After publishing and refreshing, note that the PDF attachment is no longer there in WP-Admin for the job listing.

### LINKED ISSUE #1877: Polylang - Featured Jobs (fields) not translating
There appears to be an issue using Polylang with the latest update of WP Job Manager.

This specifically affects custom fields e.g. – hours, salary, closing date

Whatever is entered into the field on the English translation e.g. 27 hours per week
Will then appear on both English and the Alternative language job advert, rather than the correct language. 

It appears that it will only allow you to save one version of the language for the field type. This is only affecting feature jobs.

If you change the Alternative language to e.g. 27 awr yr wythnos
This will then reflect on both Alternative and English language job advert.

Any ideas on a fix?

jom:
Fixes #1843
Fixes #1877

* Revert change made in #1798 that avoided direct database access to update the post author. 
* Revert an additional change made in same PR around updating the menu order for the featured flag. This was causing issues around alternative versions in Polylang (#1877).
* This seems to fix #1843 and potentially saves us from other third-party plugin conflicts as well as a bug that would probably be introduced when/if we enable revisions.

#### Testing instructions:

* Update post author using WPJM meta box in WP Admin when editing job listing. Verify it updates on save.
* Follow steps in #1843 to make sure bug is fixed.
* Follow steps in https://github.com/Automattic/WP-Job-Manager/issues/1877#issuecomment-528397340 to make sure that bug is also fixed.

### DIFF
--- a/includes/admin/class-wp-job-manager-writepanels.php
+++ b/includes/admin/class-wp-job-manager-writepanels.php
@@ -604,6 +604,8 @@ class WP_Job_Manager_Writepanels {
 	 * @param WP_Post $post (Unused).
 	 */
 	public function save_job_listing_data( $post_id, $post ) {
+		global $wpdb;
+
 		// These need to exist.
 		add_post_meta( $post_id, '_filled', 0, true );
 		add_post_meta( $post_id, '_featured', 0, true );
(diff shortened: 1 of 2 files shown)


alexsanford (approved):
This seems to fix the noted issues for me.

PR #1883, opened 2019-09-13
Allow using term slugs for preselecting category filter

### LINKED ISSUE #1300: Allow preselecting categories on job's page by slug
I used to be able to append a query string to the jobs page URL to automatically narrow down the jobs list to a category, like so:

`http://wpjm.local/?search_categories=test`

... which would only show jobs in the `test` category. This no longer appears to be working.

jom:
Fixes #1300

* This allows for pre-selecting categories in `[jobs]` filter by term slug instead of just term ID.
* Note: Existing functionality was to use `/jobs/?search_category=7` to pre-select category by term ID. I used the same field (instead of the one in the issue) and allowed it to be a comma separated list of IDs and/or slugs. For example: `/jobs/?search_category=7,nice-category,okay` should select the category with term ID 7 and slugs `nice-category` and `okay` (if they have jobs in them).

#### Testing instructions:

* Enable categories with multiple select in WPJM settings (should work without multiple select, but will just take the first one listed probably).
* Add categories and published/accepted jobs in those categories.
* Using the `search_category` query variable, use a combination of the term ID or slug to pre-select the categories on the filter form.
* Verify the terms are selected.

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -360,9 +360,23 @@ class WP_Job_Manager_Shortcodes {
 
 		// Array handling.
 		$atts['categories']         = is_array( $atts['categories'] ) ? $atts['categories'] : array_filter( array_map( 'trim', explode( ',', $atts['categories'] ) ) );
+		$atts['selected_category']  = is_array( $atts['selected_category'] ) ? $atts['selected_category'] : array_filter( array_map( 'trim', explode( ',', $atts['selected_category'] ) ) );
 		$atts['job_types']          = is_array( $atts['job_types'] ) ? $atts['job_types'] : array_filter( array_map( 'trim', explode( ',', $atts['job_types'] ) ) );
 		$atts['post_status']        = is_array( $atts['post_status'] ) ? $atts['post_status'] : array_filter( array_map( 'trim', explode( ',', $atts['post_status'] ) ) );
 		$atts['selected_job_types'] = is_array( $atts['selected_job_types'] ) ? $atts['selected_job_types'] : array_filter( array_map( 'trim', explode( ',', $atts['selected_job_types'] ) ) );
+
+		// Normalize field for categories.
+		if ( ! empty( $atts['selected_category'] ) ) {
+			foreach ( $atts['selected_category'] as $cat_index => $category ) {
+				if ( ! is_numeric( $category ) ) {
+					$term = get_term_by( 'slug', $category, 'job_listing_category' );
+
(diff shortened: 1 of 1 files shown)


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -360,10 +360,24 @@ public function output_jobs( $atts ) {
 
 		// Array handling.
 		$atts['categories']         = is_array( $atts['categories'] ) ? $atts['categories'] : array_filter( array_map( 'trim', explode( ',', $atts['categories'] ) ) );
+		$atts['selected_category']  = is_array( $atts['selected_category'] ) ? $atts['selected_category'] : array_filter( array_map( 'trim', explode( ',', $atts['selected_category'] ) ) );
 		$atts['job_types']          = is_array( $atts['job_types'] ) ? $atts['job_types'] : array_filter( array_map( 'trim', explode( ',', $atts['job_types'] ) ) );
 		$atts['post_status']        = is_array( $atts['post_status'] ) ? $atts['post_status'] : array_filter( array_map( 'trim', explode( ',', $atts['post_status'] ) ) );
 		$atts['selected_job_types'] = is_array( $atts['selected_job_types'] ) ? $atts['selected_job_types'] : array_filter( array_map( 'trim', explode( ',', $atts['selected_job_types'] ) ) );
 
+		// Normalize field for categories.
```
alexsanford: I wonder if it would make sense to move this functionality into `get_job_listings`, which this ultimately calls, to make this functionality more widely available? For example, I could see it being useful in the REST API in the future?

alexsanford (approved):
Looks good, works well for me!

I have one little comment about where to put the functionality, but I'm happy with it as-is if moving it would be complicated.

includes/class-wp-job-manager-shortcodes.php
jom: This doesn't actually get passed directly to `get_job_listings`, but just pre-selects the categories on the search form. Do you think we should change the behavior to override the search form and use the query string when performing the actual search?

alexsanford: Ah I think I see my confusion. It gets passed directly in the non-ajax case (see line 425). So I was thinking that we could just do this handling inside get_job_listings for the attributes handed in.

But I forgot about the ajax case. For that, we would need to do it this way in order to pre-populate the field, as you said. So probably this solution is best for now. If we want to add this handling more generically later, we can deal with it then 🙂

richardmtl:
This will require documentation updates. cc @danjjohnson

PR #1887, opened 2019-09-20
Just return cached results when the fields arg is `id=>parent`

### LINKED ISSUE #1784: Wrong posts array from cached query in get_job_listings()
https://github.com/Automattic/WP-Job-Manager/blob/585e5cb5ca8332797909c3b370e0cfa394012967/wp-job-manager-functions.php#L182

The caching logic  of `get_job_listings()` completely disregards the other cases of the `'fields'` parameter (`ids`, `id=>parent`), and always populates the WP_Query::posts array with WP_Post objects unconditionally, breaking any existing code that assumes that e.g. an array of integers is returned.

The  side effects are only observable on the second and subsequent runs of e.g. `get_job_listings( [ 'fields' => 'ids' ] )`.

jom:
Builds off of #1835 to fix #1784

* When getting listings via `get_job_listings()` and setting `fields` to either `ids` or `id=>parent`, just return the results as cached.

#### Testing instructions:

* I used this super exciting snippet to test:
```php
add_action( 'wp', function() {
  if ( ! isset( $_GET['debug'] ) ) {
	return;
  }
  
  echo '<pre>';
  var_dump( get_job_listings( [ 'fields' => $_GET['debug'] ] ) );
  exit;
} );
```

I then visited `?debug=ids` and `?debug=id%3D>parent` and verified the `posts` var was set to either an array of ids or an array of `stdClass` with `ID` and `post_parent` vars set.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -179,7 +179,8 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 					&& isset( $cached_query_posts->posts )
 					&& is_array( $cached_query_posts->posts )
 				) {
-					if ( 'ids' === $query_args['fields'] ) {
+					if ( in_array( $query_args['fields'], array( 'ids', 'id=>parent' ), true ) ) {
+						// For these special requests, just return the array of results as set.
 						$posts = $cached_query_posts->posts;
 					} else {
 						$posts = array_map( 'get_post', $cached_query_posts->posts );


roundhill (approved):
LGTM :shipit:

manzoorwanijk:
There is a potential problem with this PR! When it's `id=>parent`, It has to be an array of `WP_Post` objects where the keys are the Post IDs. But the cashed result will return an array of `stdClass` objects from `json_decode`.

jom:
@manzoorwanijk According to [WP_Query's documentation](https://developer.wordpress.org/reference/classes/wp_query/#return-fields-parameter), `id=>parent` is supposed to return an array of `stdClass` objects with `ID` and `post_parent` attributes.

PR #1893, opened 2019-09-27
Attempt to fix #1891

### LINKED ISSUE #1891: A "Keywords" field still does not work properly (also with custom search form) [1.34.0]
Hi,

I just updated WP Job Manager to 1.34.0 Beta Version to check is there everything already good with the "keywords" field. Unfortunately the error still occurs.

What it does mean?

If I write on my homepage in "keywords" word "driver" and press Enter, then I get job listings, but the field "keywords" is not filled in with my kewyord. It is happening after refresh page.

If I will write e.g. "cook" in keywords and go to "Job listings", the field keywords after a few seconds is filled by "cook".

Can you check it before an official relase?

andreilupu:
This PR handle the following things:
* A better event handling for the filters form, separated from the `.job_listings` wrapper.
* Refactor the localStorage key so it can be identified outside of the listings page.

More reasoning in https://github.com/Automattic/WP-Job-Manager/issues/1891#issuecomment-535883328


assets/js/ajax-filters.js
```diff
@@ -34,14 +34,17 @@ jQuery( document ).ready( function( $ ) {
 	 * Get the session storage key for the job listings instance.
 	 */
 	function get_session_storage_key( $target ) {
-		var index          = $( 'div.job_listings' ).index( $target );
-		var unique_page_id = $target.data( 'post_id' );
+		var index   = $( 'div.job_listings' ).index( $target );
+		var pageUrl = window.location.href;
```
jom: I'm not sure what this is solving as we were already falling back to the URL if the page ID wasn't being set. I'd prefer to use page ID where possible and fall back to URL. I believe we added that fallback for themes like yours which  have `[jobs]` elements on non-`[jobs]` pages.

andreilupu: Yep, I understand the reasoning behind it and the history of the commits, but as I said in #1891 the behavior is not consistent when the form is submitting filter values from one page to another, especially when on one page is with the fallback URL in the key name and on the redirected page is with page ID.

assets/js/ajax-filters.js
```diff
@@ -490,8 +486,16 @@ jQuery( document ).ready( function( $ ) {
 
 			return false;
 		} )
-		.on( 'submit', function() {
-			return false;
+		.on( 'submit', function(event) {
```
jom: We just changed it to stop persisting here and instead only when moving away from page on a job listing result.

assets/js/ajax-filters.js
jom: Shouldn't it only persist on the same page? I'm not sure I understand the steps to reproduce still.

assets/js/ajax-filters.js
andreilupu: That behavior is still happening, when the filters form is on a page that contains `. job_listings`, probably in the majority of cases, the form submission should be prevented.

This PR is about cases when the filtering template is out of the [jobs]. For example when you create a widget that contains only `get_job_manager_template( 'job-filters.php' );` and use the browser's form submission process, and redirect the search from a homepage without listings to the selected Listings page. At this point, this case is impossible, because the filtering logic is bound to the listings logic.

Also, I just realized this part of code has an issue, `persist_form` should be called after the `if` on line 494, not before.

Braehler:
Any updates here on this?

jom:
I think this has been fixed in other PRs.

PR #1899, opened 2019-10-09
Fix Job Dashboard table spacing and layout.

### LINKED ISSUE #1655: Dashboard tables do not look good on Twenty Nineteen theme
#### Steps to Reproduce
1. Install and activate Twenty Nineteen theme.
2. Load the Job Dashboard page.

#### What I Expected

The table to look ok.

#### What Happened Instead

The table was quite squished, and many cells have to break their text over multiple lines. This was the case even when the window took up my full laptop screen (15" MBP on default resolution).

- WP 5.0.2
- WPJM 1.31.3

#### Browser / OS Version

- OSX 10.14.x
- Chrome 71

#### Screenshot / Video

screen shot on 2019-01-10 at 12 26 48

davidparkercodes:
Fixes #1655 

* Fix the original issue in #1655 of the Twenty Nineteen theme media query shrinking the table when it did not need to. 
  * _Note: in the original issue there is an "Applications" column._
* Change the inner cell to have min-width and the padding to be the same on all sides.
* Center the .date column text. 

#### Testing instructions:

* Install and activate the Twenty Nineteen theme.
* Load the Job Dashboard page.
* See if the table's spacing is consistent.
* Try other themes and repeat.

Change spacing and padding inside #job-manager-job-dashboard div.


jom:
Sorry about this, but this PR will need a rebase. These rules have recently moved to a `scss` file.

PR #1915, opened 2019-10-18
Do not trailingslash ajax endpoint when using Plain permalink structure

jom:
This was breaking WPML when permalink is on parameter mode. Props to @vukvukovich for tracking this down and finding the fix.

* Remove `trailingslashit` around WPJM's ajax endpoint method.

#### Testing instructions:

* Switch permalink structure to `Plain`.
* Activate WPML with multiple languages.
* Create job listings in with versions in both the primary language and at least one other language.
* Visit the `[jobs]` page.
* From default language, make sure that language's listings are displayed. 
* From the other language, make sure that language's listings are displayed.

* Bonus (I've tested this scenario): Test above with a subdirectory WP install.


alexsanford (approved):
Looks good, works well!

Issue #1558, opened 2018-07-26
Use $query_args When Determining Random Ordering


spencerfinnell:
#### Steps to Reproduce
1. Use the Listify theme.
2. Sort listings by Random.

#### What I Expected

Listings to `shuffle` on each load.

#### What Happened Instead

They appear in the same order.

#### Context / Source

When random ordering was optimized/fixed it switched from looking at `$query_args` to `$args` for the status of `orderby`.

`$args` has no filter so there no way to intercept a previously non-random query and make it random.

https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L198

I think https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L198 should check for `$query_args['orderby']` and in https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L125 set a flag that it is `rand_featured`. 

```
if ( $rand_featured ) {
	usort( $result->posts, '_wpjm_shuffle_featured_post_results_helper' );
} elseif ( 'rand' === $query_args['orderby'] ) {
	shuffle( $result->posts );
}
```

spencerfinnell:
@jom Sorry I missed you yesterday. Not sure my initial proposal is the best way, but I'm also at a loss for any other elegant solution.

Introducing another filter for `$args` isn't great either.

PR #1916, opened 2019-10-21
[WIP] Refactor `get_job_listings()` to a class

jom:

- Moves majority of `get_job_listings()` into two classes:
  - Base class: `WP_Job_Manager_Listing_Query_Generator`
  - WPJM class: `WP_Job_Manager_Query_Generator`
- Functionality should be maintained as much as possible. There are a few minor changes. `get_job_listings()` tests seem to pass.
- Ideally, we'll use `WP_Job_Manager_Listing_Query_Generator` in Resume Manager to benefit from all the shared code that was in `get_job_listings()` and `get_resumes()`.
- I was initially thinking this for 1.34.1, but it seems like too big of a change.

- To do: 
  - [ ] Tests
  - [ ] Explore #1558 options.
  - [ ] Deprecate old global functions that are no longer used. Check for overrides and warn developers if they exist.


tripflex:
Adding myself to this as well just to stay in the loop please

PR #1919, opened 2019-10-24
Prevent some jobs from being edited on the frontend

jom:
Not to be confused with `job_manager_user_can_edit_job()`, which checks if a particular _user_ can edit a job listing, this allows for blocking certain listings from being edited on the frontend.

To test, you can play with this filter:
```
add_filter( 'job_manager_job_is_editable', function( $job_is_editable, $job_id ) {
    return false;
}, 10, 2 );
```

Make sure the `Edit` links goes away from the `[job_dashboard]` listing. If you temporarily disable snippet and then, load `[job_dashboard]` page, and then re-enable the snippet, the `Edit` links shouldn't work when you click on them.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -868,6 +868,37 @@ class WP_Job_Manager_Post_Types {
 		return $data;
 	}
 
+
+	/**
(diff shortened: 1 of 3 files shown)


alexsanford (approved):
:shipit:

PR #1925, opened 2019-10-29
Add ability to force email notifications to be enabled

jom:

* Some emails must stay enabled if they are critical to a plugin behavior or disabled if a pre-condition of their use is not met. This allows for us force certain email notifications to stay either enabled or disabled. 

#### Testing instructions:

* Test _Apply With Resume_ notification settings in Resumes plugin (on `master`).

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -516,9 +516,12 @@ class WP_Job_Manager_Settings {
 	 * @param string $ignored_placeholder
 	 */
 	protected function input_checkbox( $option, $attributes, $value, $ignored_placeholder ) {
+		if ( ! isset( $option['hidden_value'] ) ) {
+			$option['hidden_value'] = '0';
+		}
 		?>
 		<label>
-		<input type="hidden" name="<?php echo esc_attr( $option['name'] ); ?>" value="0" />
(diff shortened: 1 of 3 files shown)


includes/admin/class-wp-job-manager-settings.php
```diff
@@ -823,6 +826,18 @@ protected function input_multi_enable_expand( $option, $attributes, $values, $pl
 		$enable_option['name']       = $option['name'] . '[' . $enable_option['name'] . ']';
 		$enable_option['type']       = 'checkbox';
 		$enable_option['attributes'] = [ 'class="sub-settings-expander"' ];
+
+		if ( isset( $enable_option['force_value'] ) && is_bool( $enable_option['force_value'] ) ) {
+			if ( true === $enable_option['force_value'] ) {
```
alexsanford: Could this just be `if ( $enable_option['force_value'] ) { ... } else { ... }`? We've already checked for `is_bool` so I'm not sure we need to be so specific with our if/else here.

alexsanford (approved):
🎉

includes/admin/class-wp-job-manager-settings.php
jom: Good shout 44b93fd038bddada0d5ba2416b0d8bd3e92bb654

PR #1926, opened 2019-10-29
Allow notifications to be sent immediatly and get response back

jom:

* In Resumes we needed to be able to send notifications right away and be told if we were successful. This adds a public method to the email notifications class to allow us to do that. 

NOTE: Failed test is the geolocation test (unrelated to this PR). I think we went over our quota today.

#### To Do
* [x] Unit  tests

### DIFF
--- a/includes/class-wp-job-manager-email-notifications.php
+++ b/includes/class-wp-job-manager-email-notifications.php
@@ -100,6 +100,39 @@ final class WP_Job_Manager_Email_Notifications {
 	}
 
 	/**
+	 * Send email notifications for a specific email now and return the status. Not usually necessary but useful if you
+	 * need to check to make sure a notification was sent.
+	 *
+	 * @param string $email_key Email key to send notifications for.
+	 * @return bool True if an email notification was sent.
+	 */
+	public static function send_notifications( $email_key ) {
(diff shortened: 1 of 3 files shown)


includes/class-wp-job-manager-email-notifications.php
```diff
@@ -99,6 +99,39 @@ public static function send_deferred_notifications() {
...
+			if (
+				! is_string( $email[0] )
+				|| ! isset( $email_notifications[ $email[0] ] )
+				|| $email[0] !== $email_key
+			) {
+				continue;
+			}
+
+			$email_class            = $email_notifications[ $email[0] ];
+			$email_notification_key = $email[0];
```
alexsanford: Isn't this just `$email_key`?

alexsanford (commented):
Code and tests look good. Do you have any manual testing instructions to see whether this is working?

includes/class-wp-job-manager-email-notifications.php
jom: Yup! 7e0172da42a92c2aa3d5bc60adbc8634124b94fd

alexsanford (approved):
Awesome! :shipit:

PR #1940, opened 2019-11-25
define this as var for use out of context in validation errors for ajax file upload

### LINKED ISSUE #1939: AJAX File Upload errors are pushed to undefined (due to this not being the right context)
In the ajax upload js file, when `$.each(data.result.files` is called, it essentially changes the context of `this`, resulting in `this.validation_errors` to be undefined.  This results in a javascript error, ultimately never alerting the user of the error and just failing silently.

https://github.com/Automattic/WP-Job-Manager/blob/10898d081080f31d7f569c278495c0a05a60c5f2/assets/js/ajax-file-upload.js#L97

Basically we need to define a var like `var self = this;` before `$.each` is called, that way you can instead do `self.validation_errors.push` without errors.  I will open a PR to resolve this.

This issue was found while attempting to debug another problem, where for instance the user uploads a `.doc` file, which was originally `.docx` (so the mime type does not match), and in `wp_check_filetype_and_ext` it returns false since the mime type does not match the extension, it then fails with a `Sorry, this file type is not permitted for security reasons.` error from core under `_wp_handle_upload` due to `$type` being false.

tripflex:
Fixes #1939 

* Define `this` for use out of context to prevent undefined javascript error


tripflex:
@jom can we get this merged before the next release? kinda a big issue with errors not showing for file uploads

PR #1941, opened 2019-11-25
Add $email_notification_key to job_manager_email_do_send_notification filter

tripflex:
Currently in the `job_manager_email_do_send_notification` there's no way to determine what email "hook" or "key" is actually being sent -- making it very difficult or impossible to short circuit since we don't know what email is being sent.

This PR adds `$email_notification_key` as another arg in the filter

If we could get this merged ASAP i would greatly appreciate it, it's blocking me from being able to short circuit some emails


jom:
This is accessible from the `$email` argument. You can get its class and call `$email_class::get_key()`.

tripflex:
You're right, must have missed that thank you -- closing this PR

PR #1944, opened 2019-12-06
New filter to set email / url application as default field

### LINKED ISSUE #1872: Allow short circuit application field forced validation
Based on this support thread:
https://wordpress.org/support/topic/way-to-change-field-hidden-by-apply-for-job-button/

Basically the `application` field is hard coded for validation of either URL or Email ... which is fine as that should be the standard, but in the instance that a user wants to allow something other than just URL or Email, would be good to have a filter added around here:
https://github.com/Automattic/WP-Job-Manager/blob/6c35acabbbc7eacb03bf72ea21461f293f0a0ec0/includes/forms/class-wp-job-manager-form-submit-job.php#L437-L446

To allow short circuiting the required validation as email or URL (without having to hack apart validation through core WordPress `is_email` field)

Maybe changing `if ( ! is_email( $values['job']['application'] ) ) {` to something like:

```php
if ( ! is_email( $values['job']['application'] ) && apply_filters( 'job_manager_validate_fields_application_force_validate_url', true ) ) {
```

renatho:
Fixes #1872 

* Add a new filter to force the application field to skip the email / url validation. The filter name is `job_manager_application_field_skip_email_url_validation`

This filter should be combined with `submit_job_form_fields` and `job_manager_job_listing_data_fields`, else the sanitizer will clear the field before the validation. Also, the user can hide the application method options from settings, using the `job_manager_settings` filter, once it doesn't make more sense if we override these options with the filters.

A full customization example with comments:

```php
/**
 * Skip application email / url validation.
 */
function application_field_skip_email_url_validation() {
	return true;
}
add_filter( 'job_manager_application_field_skip_email_url_validation', 'application_field_skip_email_url_validation' );

/**
 * Override submit job form fields.
 */
function custom_submit_job_form_fields( $fields ) {
	// Skip application sanitizer from screen
	$fields[ 'job' ][ 'application' ][ 'sanitizer' ] = null;
	// Also changes the label from frontend, if desired
	$fields[ 'job' ][ 'application' ][ 'label' ] = 'Lorem ipsum';
	return $fields;
}
add_filter( 'submit_job_form_fields', 'custom_submit_job_form_fields' );

/**
 * Override job listing data fields.
 */
function custom_job_manager_job_listing_data_fields( $fields ) {
	// Skip application sanitizer from database
	$fields[ '_application' ][ 'sanitize_callback' ] = null;
	// Also changes the label from backend, if desired
	$fields[ '_application' ][ 'label' ] = 'Lorem ipsum';
	return $fields;
}
add_filter( 'job_manager_job_listing_data_fields', 'custom_job_manager_job_listing_data_fields' );

/**
 * Hide application method options from settings.
 */
function hide_application_method_options_from_settings( $settings ) {
	foreach ( $settings[ 'job_submission' ][ 1 ] as $key => $field ) {
		if ( 'job_manager_allowed_application_method' === $settings[ 'job_submission' ][ 1 ][ $key ][ 'name' ] ) {
			unset( $settings[ 'job_submission' ][ 1 ][ $key ] );
		}
	}
	return $settings;
}
add_filter( 'job_manager_settings', 'hide_application_method_options_from_settings' );

/**
 * Set job application method to "custom".
 */
function set_job_application_method_custom() {
	$method = new stdClass();
	$method->type = 'custom';
	return $method;
}
add_filter( 'the_job_application_method', 'set_job_application_method_custom' );

/**
 * Handle display for "custom" application method.
 */
function display_custom_application_method() {
	echo '<p>';
	esc_html_e( 'To apply for this job, please visit us in the office.', 'my-text-domain' );
	echo '</p>';
}
add_action( 'job_manager_application_details_custom', 'display_custom_application_method' );
```

#### Testing instructions:

* Add the filters in the theme `functions.php` (explanations in the code comments)
* Create a page with the shortcode `[submit_job_form]`
* Go to the page created and submit a new job with the field `Application email/URL` with "Lorem ipsum" value (no email or URL format)

* Added the new filter `job_manager_application_field_skip_email_url_validation` that force the `Application email/URL` field on post job to skip the email / url validations

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -167,22 +167,6 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	}
 
 	/**
-	 * Checks if it is forced application default field
-	 *
-	 * @return bool
-	 */
-	protected function is_forced_application_default_field() {
-		/**
-		 * Force application field to skip email / URL validations and sanitizer
-		 *
-		 * @since 1.x.x
-		 *
-		 * @param bool  $is_forced Whether the application field is forced to default.
-		 */
-		return apply_filters( 'job_manager_application_force_default_field', false );
-	}
-
-	/**
 	 * Initializes the fields used in the form.
 	 */
 	public function init_fields() {
@@ -207,10 +191,6 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 				$application_method_placeholder = __( 'Enter an email address or website URL', 'wp-job-manager' );
 				$application_method_sanitizer   = 'url_or_email';
 				break;
-		}
-
-		if ( $this->is_forced_application_default_field() ) {
-			$application_method_sanitizer = null;
 		}
 
 		if ( job_manager_multi_job_type() ) {
@@ -349,6 +329,22 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	}
 
 	/**
+	 * Checks if application field should use standard validation
+	 *
+	 * @return bool
+	 */
+	protected function should_application_field_use_standard_validation() {
+		/**
+		 * Force application field to use standard validation (skip email and URL validations)
+		 *
+		 * @since 1.x.x
+		 *
+		 * @param bool  $is_forced Whether the application field is forced to use standard validation.
+		 */
+		return apply_filters( 'job_manager_application_field_use_standard_validation', false );
+	}
+
+	/**
 	 * Validates the posted fields.
 	 *
 	 * @param array $values
@@ -435,7 +431,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		}
 
 		// Application method.
-		if ( ! $this->is_forced_application_default_field() && isset( $values['job']['application'] ) && ! empty( $values['job']['application'] ) ) {
+		if ( ! $this->should_application_field_use_standard_validation() && isset( $values['job']['application'] ) && ! empty( $values['job']['application'] ) ) {
 			$allowed_application_method   = get_option( 'job_manager_allowed_application_method', '' );
 			$values['job']['application'] = str_replace( ' ', '+', $values['job']['application'] );
 			switch ( $allowed_application_method ) {
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -329,19 +329,19 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	}
 
 	/**
-	 * Checks if application field should use standard validation.
+	 * Checks if application field should use skip email / URL validation.
 	 *
 	 * @return bool
 	 */
-	protected function should_application_field_use_standard_validation() {
+	protected function should_application_field_skip_email_url_validation() {
 		/**
-		 * Force application field to use standard validation (skip email and URL validations).
+		 * Force application field to skip email / URL validation.
 		 *
 		 * @since 1.x.x
 		 *
-		 * @param bool  $is_forced Whether the application field is forced to use standard validation.
+		 * @param bool  $is_forced Whether the application field is forced to skip email / URL validation.
 		 */
-		return apply_filters( 'job_manager_application_field_use_standard_validation', false );
+		return apply_filters( 'job_manager_application_field_skip_email_url_validation', false );
 	}
 
 	/**
@@ -431,7 +431,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		}
 
 		// Application method.
-		if ( ! $this->should_application_field_use_standard_validation() && isset( $values['job']['application'] ) && ! empty( $values['job']['application'] ) ) {
+		if ( ! $this->should_application_field_skip_email_url_validation() && isset( $values['job']['application'] ) && ! empty( $values['job']['application'] ) ) {
 			$allowed_application_method   = get_option( 'job_manager_allowed_application_method', '' );
 			$values['job']['application'] = str_replace( ' ', '+', $values['job']['application'] );
 			switch ( $allowed_application_method ) {
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -172,6 +172,13 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	 * @return bool
 	 */
 	protected function is_forced_application_default_field() {
+		/**
+		 * Force application field to skip email / URL validations and sanitizer
+		 *
+		 * @since 1.34.1
+		 *
+		 * @param bool  $is_forced Whether the application field is forced to default.
+		 */
 		return apply_filters( 'job_manager_application_force_default_field', false );
 	}
 
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -329,13 +329,13 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	}
 
 	/**
-	 * Checks if application field should use standard validation
+	 * Checks if application field should use standard validation.
 	 *
 	 * @return bool
 	 */
 	protected function should_application_field_use_standard_validation() {
 		/**
-		 * Force application field to use standard validation (skip email and URL validations)
+		 * Force application field to use standard validation (skip email and URL validations).
 		 *
 		 * @since 1.x.x
 		 *
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -175,7 +175,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		/**
 		 * Force application field to skip email / URL validations and sanitizer
 		 *
-		 * @since 1.34.1
+		 * @since 1.x.x
 		 *
 		 * @param bool  $is_forced Whether the application field is forced to default.
 		 */
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -435,7 +435,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		}
 
 		// Application method.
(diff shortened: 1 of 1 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -166,6 +166,22 @@ public function get_job_id() {
...
+	/**
+	 * Checks if it is forced application default field
+	 *
+	 * @return bool
+	 */
+	protected function is_forced_application_default_field() {
+		/**
+		 * Force application field to skip email / URL validations and sanitizer
+		 *
+		 * @since 1.34.1
```
jom: For now let's set `@since`'s to 1.x.x so we can flag to change it when we release. I moved this to 1.34.2, but the next release might be 1.35.0.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -166,6 +166,22 @@ public function get_job_id() {
 		return absint( $this->job_id );
 	}
 
+	/**
+	 * Checks if it is forced application default field
+	 *
+	 * @return bool
+	 */
+	protected function is_forced_application_default_field() {
```
jom: This is a tough one to name, but I'm not sure this gets at what it does. I'm slightly leaning towards making this _just_ about validation and not sanitization. For sanitization they can use the `submit_job_form_fields` filter to change the `sanitizer`. In that case, we could move the filter down and name it something like `job_manager_application_field_use_standard_validation` (feel free to iterate on that one).

jom (commented):
Nice work tracking down the inner workings of these form classes! Couple of comments about naming/possible alternative approach.

includes/forms/class-wp-job-manager-form-submit-job.php
renatho: Okay! Should I always do like this, using only the major version? Or is it only for this case?

includes/forms/class-wp-job-manager-form-submit-job.php
renatho: Yes! It was very tough to name and really it's not so good.
I like your approach! It reduces the complexity \o/

alexsanford (commented):
A few notes below, and a few issues:

### Setting

I wonder if this filter should also cause the setting to be hidden?

Screen Shot 2019-12-19 at 15 34 23

The setting doesn't make a lot of sense if we are not validating the email or URL. Then again, will it be confusing if we hide it?

### Adding http prefix

When I add the filters as per the PR description, and submit a job with "Lorem Ipsum" in the "Application email/URL" field, it adds `http://` to the beginning of the string, and URL encodes it:

Screen Shot 2019-12-19 at 15 40 43

### Multiple filters

I wonder if we should find a way to only use the one filter in order to disable the validation. So maybe this filter should also disable the sanitizer, so we don't have to use multiple filters. Do you think this makes sense?

Additionally, we still have the field being labelled "Application email/URL" on both the frontend and the backend. To change this, the user would have to make yet another change using the `submit_job_form_fields` filter. I wonder if we should also set that to a "sensible default" when this filter is used?

renatho:
> I wonder if this filter should also cause the setting to be hidden?
> 
> Screen Shot 2019-12-19 at 15 34 23
> 
> The setting doesn't make a lot of sense if we are not validating the email or URL. Then again, will it be confusing if we hide it?

I think it's a great idea because with the filter I think it does not make more sense.

renatho:
> When I add the filters as per the PR description, and submit a job with "Lorem Ipsum" in the "Application email/URL" field, it adds `http://` to the beginning of the string, and URL encodes it:
> 
> Screen Shot 2019-12-19 at 15 40 43

I investigated and identified the problem that happens here. The following code is creating a filter to sanitize the value on save of the post meta.
https://github.com/Automattic/WP-Job-Manager/blob/6363b97778d02f428b7ded22c98a0c88464a85fe/includes/class-wp-job-manager-post-types.php#L1270

Actually, I don't know if it's a bug on the `submit_job_form_fields` filter (which overrides the following object and not the object above) or if it's expected.

https://github.com/Automattic/WP-Job-Manager/blob/6363b97778d02f428b7ded22c98a0c88464a85fe/includes/forms/class-wp-job-manager-form-submit-job.php#L203-L308

And unless we'd do a refactory in this part, we could not merge the sanitizer from the `submit_job_form_fields` filter here too because it's a different approach.

renatho:
> I wonder if we should find a way to only use the one filter in order to disable the validation. So maybe this filter should also disable the sanitizer, so we don't have to use multiple filters. Do you think this makes sense?

Actually it was the first approach of this PR. But after some reviews, we decided to remove the sanitizer part to make the filter less complex.

But based on the following comment, I think it would better to adopt this approach (clear both the validation and the sanitizer). So we could solve it for this case now overriding the sanitizer for both objects. What do you think?

https://github.com/Automattic/WP-Job-Manager/pull/1944#issuecomment-567952538

renatho:
> Additionally, we still have the field being labelled "Application email/URL" on both the frontend and the backend. To change this, the user would have to make yet another change using the `submit_job_form_fields` filter. I wonder if we should also set that to a "sensible default" when this filter is used?

For this specific issue, based on the original support thread, I think it would not be a problem because the user would use the field to this specific meaning.

https://wordpress.org/support/topic/way-to-change-field-hidden-by-apply-for-job-button/

> I would like to change this to a blank field where employers can enter any text that is desired in this field (e.g., so, something like this could be entered: “Please apply by September 25th by submitting both a cover letter and resume to Bob at bob@mycompany.com or applying on our website at http://www.mycompany.com/jobs&#8221;).

But I agree with you. It could be helpful for other cases. Do you think something like to send another param (the new label) to the filter?

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -328,6 +328,22 @@ public function use_recaptcha_field() {
...
+	 */
+	protected function should_application_field_use_standard_validation() {
+		/**
+		 * Force application field to use standard validation (skip email and URL validations)
+		 *
+		 * @since 1.x.x
+		 *
+		 * @param bool  $is_forced Whether the application field is forced to use standard validation.
+		 */
+		return apply_filters( 'job_manager_application_field_use_standard_validation', false );
```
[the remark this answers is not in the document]
renatho: I'll wait the result of our discussions because we can change the filter meaning :)

alexsanford:
Thinking about this some more, I think the concept of making the filter do one thing is probably better. It would not be that useful on its own, but we can provide a code snippet to show people all the different filters to use to fully customize this. If we try to make the filter handle everything, I think it will get too messy.

So I guess we can have the filter only remove the validation, as long as we have filters to do everything else (remove the setting, change the field name, etc)? Maybe we could update the PR description with a full snippet that would show how to change everything?

renatho:
> Thinking about this some more, I think the concept of making the filter do one thing is probably better. It would not be that useful on its own, but we can provide a code snippet to show people all the different filters to use to fully customize this. If we try to make the filter handle everything, I think it will get too messy.
> 
> So I guess we can have the filter only remove the validation, as long as we have filters to do everything else (remove the setting, change the field name, etc)? Maybe we could update the PR description with a full snippet that would show how to change everything?

Sounds good! I just don't know yet how we could get around this problem: https://github.com/Automattic/WP-Job-Manager/pull/1944#issuecomment-567952538

I'll try to think about a solution on Monday.

includes/forms/class-wp-job-manager-form-submit-job.php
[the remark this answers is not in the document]
renatho: I renamed to `job_manager_application_field_skip_email_url_validation`. While I was writing it looked a little better. Does it make sense to you?

renatho:
@alexsanford I found filters for everything. So the complexity added to the code was only the new filter to remove the validation \o/
Thank you for your review! With the new filters, it's really better!
If you see anything else, please let me know! ;)

alexsanford (approved):
This works well for me! 🎉 

Approving, although I caught one more thing that we might want to add to our example code. When a job listing is displayed on the frontend, it still assumes that the value is a URL:

Screen Shot 2020-01-07 at 13 53 55

Here is the code for the `a` tag:

Screen Shot 2020-01-07 at 13 54 08

This can be fixed with yet another filter and action. Here is an example:

```php
/**
 * Set job application method to "custom".
 */
function set_job_application_method_custom() {
	$method = new stdClass();
	$method->type = 'custom';
	return $method;
}
add_filter( 'the_job_application_method', 'set_job_application_method_custom' );

/**
 * Handle display for "custom" application method.
 */
function display_custom_application_method() {
	echo '<p>';
	esc_html_e( 'To apply for this job, please visit us in the office.', 'my-text-domain' );
	echo '</p>';
}
add_action( 'job_manager_application_details_custom', 'display_custom_application_method' );
```

Might be good to add this to the example code, if we don't think it's going to overly complicate things.

renatho:
@alexsanford Good catch! Tested here too and added to the example.
Thank you for your review =)

Hastibe:
@renatho -- this looks terrific, but I am encountering an issue (maybe due to my lack of understanding about how to use this modification?). 

I've pasted in the customization example to my child theme's functions.php, and on the submit job form, it now allows me to enter any text (not just an email address or URL), but when the job post is published, it just displays the "To apply for this job, please visit us in the office." 

What do I need to modify so that it actually displays the text entered in the "Lorem ipsum" (_application) field on the submit job form, instead of just the "To apply for this job, please visit us in the office." text?

renatho:
> @renatho -- this looks terrific (I was the one who opened [the original support thread](https://wordpress.org/support/topic/way-to-change-field-hidden-by-apply-for-job-button/) way back when, and am just discovering all this amazing work!), but I am encountering an issue (maybe due to my lack of understanding about how to use this modification?).
> 
> I've pasted in the customization example to my child theme's functions.php, and on the submit job form, it now allows me to enter any text in the "Lorem ipsum" field (not just an email address or URL), but when the job post is published, it just displays the `"To apply for this job, please visit us in the office."` text.
> 
> What do I need to modify so that it actually displays the text entered in the "Lorem ipsum" (_application) field on the submit job form, instead of just the `"To apply for this job, please visit us in the office."` text?

Hey @Hastibe!

I think you can achieve your goals tweaking 2 parts of the snippet.

```php
/**
 * Set job application method to "custom".
 */
function set_job_application_method_custom( $method ) {
	$method->type = 'custom';
	return $method;
}
add_filter( 'the_job_application_method', 'set_job_application_method_custom' );

/**
 * Handle display for "custom" application method.
 */
function display_custom_application_method( $apply ) {
	echo '<p>';
	echo( 'To apply for this job, please visit us in the office. ' . str_replace( 'http://', '', $apply->url ) );
	echo '</p>';
}
add_action( 'job_manager_application_details_custom', 'display_custom_application_method' );
```

Basically, the first one will continue with the current `$method`, so we can use the `url` value (the customized input) in the other one. In the second one we'll tweak the value, removing the `http://`, and concat the message with what you want. ;)

I hope it works for you!

Hastibe:
@renatho -- thank you so much for your helpful response, and this looks like it's working--hurray! 🥳

PR #1952, opened 2019-12-20
Fix/issue 1912

### LINKED ISSUE #1912: Mark as Filled action generates incorrect messages with paginated jobs on job dashboard
From https://wordpress.org/support/topic/is-anyone-facing-this-problem-with-filled-or-non-filled-positions/

Screencast: https://streamable.com/i8n7y

Steps to reproduce:

1) Make sure you have enough jobs on your jobs dashboard that they are paginated at least to a page 2.
2) On the jobs dashboard, mark a job as filled; you will get the correct "X has been filled" message.
3) Click on the "2" in the pagination at the bottom of the jobs list.
4) You'll see a random "This position has already been filled." error message at the top.
5) Click again back to page 1, you'll see the same "This position has already been filled." message.
6) Click the "mark as unfilled"; you'll see the correct "X has been marked as not filled." message.
7) Click once again on the "2"; you'll see a random "This position is not filled." message at the top.
8) Click again back to page 1, the same "This position is not filled." message is still there.

These erroneous messages are confusing. :)

JuanchoPestana:
Fixes #1912 

* The paginate_links() function adds the url args (from the current url) to each pagination link. The actions (mark as filled, not filled, etc) in the job dashboard were triggered with the url (For example: url/?action=mark_as_filled). The consequences of this were that if you changed the page, the pagination link would carry the "action, job_id, and _wpnonce" parameters to the new page, causing the action to be triggered again; thus, getting the message that the job was already filled/not filled. 
* This PR removes the args from the url for the pagination links so that the actions are not triggered again when changing pages.

#### Testing instructions:

1. Make sure you have enough jobs on your jobs dashboard that they are paginated at least to a page 2.
2. On the jobs dashboard, mark a job as filled; you will get the correct "X has been filled" message.
3. Click on the "2" or the "next arrow" in the pagination at the bottom of the jobs list.
4. You should not see any messages

* Remove args from pagination_link urls


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -272,6 +274,16 @@ public function job_dashboard( $atts ) {
...
 	}
 
+	/**
+	 * Filters the url from paginate_links to avoid multiple calls for same action in job dashboard
+	 *
+	 * @param string $link
+	 * @return string
+	 */
+	public function filter_paginate_links( $link ) {
+		return esc_url( remove_query_arg( [ 'action', 'job_id', '_wpnonce' ], $link ) );
```
alexsanford, line 284: I don't think we need to `esc_url` for this, as it will be done after the filter completes.

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -61,6 +61,8 @@ public function __construct() {
 		add_shortcode( 'job', [ $this, 'output_job' ] );
 		add_shortcode( 'job_summary', [ $this, 'output_job_summary' ] );
 		add_shortcode( 'job_apply', [ $this, 'output_job_apply' ] );
+
+		add_filter( 'paginate_links', [ $this, 'filter_paginate_links' ], 10, 1 );
```
alexsanford, line 65: I wonder if this solution is too imprecise? For example, if there are other pagination links on this post, this will potentially change those links as well.

I wonder if it would be better to use the [`job_manager_pagination_args`](https://github.com/Automattic/WP-Job-Manager/blob/master/templates/pagination.php#L33) filter instead, and modify the `base` argument?

JuanchoPestana: Hey Alex! Thanks a lot for taking the time to review this PR!

Thanks for the `job_manager_pagination_args` filter suggestion. I tried to implement this approach, but the problem here is that the `paginate_links()` function adds the current URL args to the links; so even if we send in the 'base' argument a URL without the conflicting args, the `paginate_links()` adds them. The only filter I found we could hook to AFTER the conflicting args being re-added is the `paginate_links` one. 

Any other possible approaches? Let me know your thoughts!

includes/class-wp-job-manager-shortcodes.php
JuanchoPestana: Done!

JuanchoPestana:
Hey Alex! The Travis build is failing because of a PHPUnit version problem... Is this something I can fix on my side?

includes/class-wp-job-manager-shortcodes.php
alexsanford: Ah thanks for the explanation. Makes sense.

I guess another approach would be to try to check the URL to see if it is the one we expect, but that seems like overkill. This is probably fine as-is, and we can iterate if needed.

alexsanford:
> The Travis build is failing because of a PHPUnit version problem... Is this something I can fix on my side?

Looks like this was fixed in https://github.com/Automattic/WP-Job-Manager/pull/2031. Could you rebase this onto `master` and see if that fixes the tests?

JuanchoPestana:
> Looks like this was fixed in #2031. Could you rebase this onto `master` and see if that fixes the tests?

Hey @alexsanford! Just did this. No issues. Thanks again, and let me know if there's anything else I could do to help!

alexsanford:
Hmm this fixed the tests but it looks like something weird happened in the rebase:

---

Screen Shot 2020-06-19 at 15 56 51

---

Screen Shot 2020-06-19 at 15 56 47

---

Are you able to fix this so that we only have the commits for this PR rebased onto `master`?

JuanchoPestana:
Hey @alexsanford! I'm really really sorry for my delay... It's been a couple busy weeks!

I tried fixing this rebase issue but I'm having trouble... I know it's probably not the best approach, but I decided instead (to avoid extending the wait) to create a new PR with the fix. Does this work for you? If it doesn't, please let me know and I'll try to make some time to fix this rebase issue...

Here's the new PR: #2037 

Again, sorry for this!

alexsanford:
No worries! Closing in favour of #2037

Issue #1860, opened 2019-08-14
Job drafts empty after saving


Braehler:
#### Steps to Reproduce
1. Customer had an active listing or is creating a listing
2. Saves the draft, success message is shown
3. Continue to edit the draft later, and the draft is empty except the title

#### What I Expected
When continuing to edit the draft, all previously filled fiels should show up, including the images

#### What Happened Instead
Drafts are empty except the title and the company logo.
But in the admin area you can see content of the listing
This can be caused by 2 things:
Either there is problem with the "draft status" on its own. Because when you manualy change the status in the admin area from "draft" to "outdated" the user gets the prefilled data back
or
There is a conflict with the WC PAid Listings add on. If you choose a different duration of the listing when you continue to edit the listing, all prefilled fields are gone. When you choose the same duration the listing is prefilled.

#### WordPress / WP Job Manager Version
Word Press 5.2.2
WPJM: 1.33.3

#### Browser / OS Version
FF 68-0.1 64bit
Chrome: 76.0.3809.100 64bit

richardmtl:
Hi!

I am unable to reproduce this; however, we do have an open issue where something similar happens, with the WC Paid Listings addon. I will close this issue in favour of that one, then. Since that issue is in a private repository, you won't be able to see it, but *we* will be able to see the link between the two, and get back to you when we resolve it. 

Also, this issue is very familiar to me; do you have a forum thread open for it? If so, can you link to it, please? Thanks!

Braehler:
Hey,
here is the link to the open thread in the forum:
https://wordpress.org/support/topic/job-drafts-empty/

PR #1953, opened 2019-12-26
Fix preview and editing conflicts

renatho:
Fixes https://github.com/Automattic/wp-job-manager-wc-paid-listings/issues/138 (possibly)
**It fixes an issue from the WCPL repo.**

I reproduced the issue only twice. After that, the bug didn't happen more. I also tried to recreate the environment some times, but nothing changed.

After many tests, I reproduced a similar bug that can be related (I hope) - steps bellow.

#### What I understood as expected behavior based on code and commits:

When a user is creating a new job and goes to preview, if they leave the page and return, they can continue editing. But if they go to edit another job, they should be able to edit this one.

* It fixes a bug that prevents the user to edit other jobs while the preview cookie is set. It can solve some related issues: https://github.com/Automattic/wp-job-manager-wc-paid-listings/issues/138 and https://github.com/Automattic/WP-Job-Manager/issues/1860. But I'm not sure because I wasn't more able to reproduce it.

#### What more we can do:

* On the WCPL we could remove the `job_id` input hidden because we use the `$_GET` or the preview cookie. I'm just not sure if it could break some other thing or it'll be dead code - If someone could confirm on code review, it would be great =)

https://github.com/Automattic/wp-job-manager-wc-paid-listings/blob/6f5f1e608fba24ab57caed8a03c7bfa8fd82c5d6/includes/class-wc-paid-listings-submit-job-form.php#L191

#### Testing instructions:

1. Active the plugins "WooCommerce" and "WP Job Manager - WooCommerce Paid Listings".
2. Create a WooCommerce product with the product type "Job Package".
3. Go to WP-admin > Job Listings > Settings > Job Submission, and set "Paid Listings Flow" to 'Choose a package before entering job details'.
4. Go to submit job page and select any package.
5. Enter job details and click Save Draft.
6. Go to submit job page again and select any package.
7. Enter details for a second job and click Preview
8. Go to job dashboard and click Continue Submission on the draft you saved on step 5.
9. Select any package. The form should appear with the draft information - the saved in the step 5.

Also can be followed the mentioned issues steps to try simulate the other bugs:
- https://github.com/Automattic/wp-job-manager-wc-paid-listings/issues/138
- https://github.com/Automattic/WP-Job-Manager/issues/1860

#### Another bug identified:

A bug on choose package while creating new job after a preview: https://github.com/Automattic/wp-job-manager-wc-paid-listings/issues/147

* Fixes the job editing when a new job is pending to edit as a preview

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -110,6 +110,9 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		}
 
 		$this->job_id = ! empty( $_GET['job_id'] ) ? absint( $_GET['job_id'] ) : 0;
+		if ( 0 === $this->job_id ) {
+			$this->job_id = ! empty( $_POST['job_id'] ) ? absint( $_POST['job_id'] ) : 0;
+		}
 		// phpcs:enable WordPress.Security.NonceVerification.Missing,  WordPress.Security.NonceVerification.Recommended
 
 		if ( ! job_manager_user_can_edit_job( $this->job_id ) ) {
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -109,7 +109,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 			$this->step = is_numeric( $_GET['step'] ) ? max( absint( $_GET['step'] ), 0 ) : array_search( sanitize_text_field( $_GET['step'] ), array_keys( $this->steps ), true );
 		}
 
-		$this->job_id = ! empty( $_REQUEST['job_id'] ) ? absint( $_REQUEST['job_id'] ) : 0;
+		$this->job_id = ! empty( $_GET['job_id'] ) ? absint( $_GET['job_id'] ) : 0;
 		// phpcs:enable WordPress.Security.NonceVerification.Missing,  WordPress.Security.NonceVerification.Recommended
 
 		if ( ! job_manager_user_can_edit_job( $this->job_id ) ) {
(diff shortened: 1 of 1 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -109,7 +109,7 @@ public function __construct() {
 			$this->step = is_numeric( $_GET['step'] ) ? max( absint( $_GET['step'] ), 0 ) : array_search( sanitize_text_field( $_GET['step'] ), array_keys( $this->steps ), true );
 		}
 
-		$this->job_id = ! empty( $_REQUEST['job_id'] ) ? absint( $_REQUEST['job_id'] ) : 0;
+		$this->job_id = ! empty( $_GET['job_id'] ) ? absint( $_GET['job_id'] ) : 0;
```
jom: I'll need to do more testing, but I'm worried this might break other things. I think I remember when doing the last round of sanitization I tried this and it broke something. I wonder if this is hiding a scenario where we're passing `job_id` in `POST` and in the query string but the `POST` variable is incorrect?

renatho: The case that I identified where we are passing the `job_id` the both and the `POST` is incorrect, is when we have a preview to edit on the cookie and we try to edit another job.

What I'm thinking that we could do to keep working also with the `POST`, would be to get in the order `$_GET['job_id']` / `$_POST['job_id']` instead to trust on the EGPCS order. I don't know if we can have other problems related to this, but if you think it can be safer I can test.

What do you think?

renatho: Hey @jom !

I added this workaround: https://github.com/Automattic/WP-Job-Manager/pull/1953/commits/f2e7c8e6b04bfa727dfa82c21c4f1f5eb13e1ea5. So if there is any case that we aren't catching or some 3rd plugin that makes something with the `POST`, we'll get. We'll just get first the `GET`. ;)

jom (approved):
Looks good!

PR #1987, opened 2020-04-03
Fix search by query string with empty categories

### LINKED ISSUE #1977: Searching by category slug returns all jobs if category has no jobs
First reported here: https://wordpress.org/support/topic/job-category-page/

I think I was able to reproduce, but not 100% certain.

Try running a /?search_category=slug where slug = a category on the site that has no jobs associated with it. It returns all jobs instead of nothing.

renatho:
Fixes #1977

The goal of this PR was to fix the search by query string, when passing slugs of empty categories.
There were 2 possible approaches.

1) One approach is a cleaner solution, but I think the UX is worse. Basically, the first option is to fetch the results based on the query string if the categories filter in the form is empty.  See: https://github.com/Automattic/WP-Job-Manager/commit/6bde71f04eba276795e4a70808dc56cccb02d452
2) The other approach (the current code of this branch), is to add the categories from the query string to the form, and as a consequence, the result is filtered correctly considering also the query string. It's better in terms of UX because the user has the reference of what is being filtered in the form and can go forth and back any moment.

#### Testing instructions:

* Create some jobs and add them to categories.
* Create one category to be empty.
* In the jobs page, add the query string: `?search_category=my-empty-category` with the slug of the empty category. The result should be empty.
* Try to filter by categories with results and make sure that the results are correct.
* Go to the `wp-admin`
* Go to  `Job Listings` → `Settings`
* Active the option `Multi-select Categories`
* Try to filter per multiple categories using the query string: `?search_category=my-empty-category,my-other-category` (separate per comma)

#### Note:

The `search_category` for multiple categories is only supported if the `Multi-select Categories` is activated.


wp-job-manager-functions.php
```diff
@@ -1094,6 +1137,8 @@ function job_manager_dropdown_categories( $args = '' ) {
 		$r['pad_counts'] = true;
 	}
 
+	$r['search_category_slugs'] = wpjm_get_category_slugs_from_search_query_string();
```
alexsanford, line 1140: For consistency with how the other args are handled, I wonder if this should be wrapped with `if ( ! isset( $r['search_category_slugs'] ) ) { ... }`?

alexsanford (approved):
LGTM, works well!

Overall, I like the approach. I'm slightly concerned that it could cause confusion if users notice that there is a category that exists in the dropdown in some cases, and not in other cases. Not sure if that will be a big problem though 🤷‍♂

wp-job-manager-functions.php
renatho: I think what you mean is not have the `search_category_slugs` in the object when it's not necessary, right?

I really don't know if has this prop would be a problem, because if we don't have a search query string, it would be only a `[]`. But what we could do is change this line to something like:

```php
$search_category_slugs = wpjm_get_category_slugs_from_search_query_string();

if ( ! empty( $search_category_slugs ) ) {
  $r['search_category_slugs'] = $search_category_slugs
}
```

WDYT?

renatho:
> I'm slightly concerned that it could cause confusion if users notice that there is a category that exists in the dropdown in some cases, and not in other cases. Not sure if that will be a big problem though 🤷‍♂

I agree with you. The other alternative that I thought (https://github.com/Automattic/WP-Job-Manager/commit/6bde71f04eba276795e4a70808dc56cccb02d452) would be to filter without to add the categories in the select, but the user would not have a reference about what is being filtered.

To have a better UX in this case, I think we should rethink this search UI. At least, I don't have a simple idea how to fix it without to change the UI. 🤔

gikaragia (approved):
Looks good, works well!

PR #1988, opened 2020-04-05
Add feature to prevent assets cache while debugging

renatho:

* The assets URL only changes when changing the version. So while we are changing the assets, the assets can keep cached. This PR introduces a feature to add a different debug query string to the asset files while debugging in each reload.

I prefer to add it to the code, so it solves for all developers that use the project in the same manner. But another alternative (if we don't want to introduce it to the plugin code), would be to add the filters in our development environment.

#### Testing instructions:

* Define the `SCRIPT_DEBUG` as `true` in your env (it can be done in the `wp-config.php`).
* Change any JS or CSS file and access the page that loads it.
* Check if there is a unique id in the `debug` query string in the script import.
* Define the `SCRIPT_DEBUG` as `false` in your env.
* Access the page again.
* Check if the `debug` query string was not included.

### DIFF
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -100,6 +100,10 @@ class WP_Job_Manager {
 
 		// Defaults for WPJM core actions.
 		add_action( 'wpjm_notify_new_user', 'wp_job_manager_notify_new_user', 10, 2 );
+
+		// Prevent assets cache.
+		add_filter( 'script_loader_src', [ $this, 'prevent_debug_assets_cache' ] );
+		add_filter( 'style_loader_src', [ $this, 'prevent_debug_assets_cache' ] );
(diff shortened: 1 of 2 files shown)


renatho:
Hum. Thinking more, I don't know if it's a good practice to add it in plugins, because it will impact all assets of the project where the plugin is installed.

Wait for opinions. 😜

alexsanford:
Hmm yeah this would be potentially pretty far-reaching, which may not be what we want. Another option is for developers to use a plugin such as [this one](https://wordpress.org/plugins/prevent-browser-caching/) in their dev environment.

renatho:
Thank you for your opinion and the plugin suggestion @alexsanford !!
I'm closing this PR because we can use some solutions in our own dev environments.

PR #1989, opened 2020-04-06
Add validation in the post job to prevent add external images

### LINKED ISSUE #1982: Set a current company image (even from the other domain etc.)
**Describe the bug**
I found a problem with replacing the images (if you have already updaten one image) on the submit form.

What it means?

If you already have added job listing and you want to edit it, then you can easly put any photo (even which is from outsite of website) which you want and update the listing.

**To Reproduce**
1. Go to 'Post a Job'
2. Add just sample listing (remember to add an image)
3. Then after publish it go to "Job Dashboard" and click "Edit" next to the listing.
4. Go to the field "Company Detalis and find a field named "Logo" (company_logo) - now just turn on "Developer Mode" in your browser and then find the detalis of the updated image.
Find code similar to this:

`<input type="hidden" class="input-text" name="current_company_logo" value="{here can be number or link}">`

and then replace the content for any link to image which you want (it can be any image which you can find at internet or just an image which already exist on your website e.g "quotes" https://wpjobmanager.com/wp-content/uploads/2014/02/2014-02-28-at-20.02-1024x908.png" [logo etc.) bettwen th quotes and hit a button "Save changes".

What we can see? That the image has been changed in the listing. It contains just the full URL which we added. Also in meda libary you can see the empty files (which does not contain the thumbnail.

P. S. You can also put there ID of attachmend and it will be used from the libary of Wordpress. So even you can use somone pictures without re-upload them.

**Screenshots**
https://imgur.com/a/juIcW2i

**Expected behavior**
I think that it should take this information form different sources (e.g. get_post_meta or something). Or it should be validated: if somone wants to change it. Because then somone can use anything there.

**WordPress Environment**
- WordPress Version:  Newest.
- WP Job Manager Version: Newest.
- PHP Version: 7.3
- Other important details:

renatho:
Fixes #1982

* This PR introduces a new validation when posting a job to prevent add external images.
* The issue mentions the possibility to introduce the media ID in the field. Considering the code, using the ID instead upload a new image is expected to work, so I didn't change it for now.

#### Testing instructions:

* Post a new job and approve it.
* Go to "Job Dashboard" and edit the new job.
* Inspect the code and find `<input type="hidden" class="input-text" name="current_featured_image" value="...">` and `<input type="hidden" class="input-text" name="current_company_logo" value="...">`
* Edit the value adding an external image. Trying to simulate different situations. Eg.: https://www.copaair.com/documents/20182/2614143/boeing-737max9-mobile.jpg
* Save changes and make sure that the validation doesn't allow to complete the changes.

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -379,46 +379,30 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 					}
 					if ( ! empty( $check_value ) ) {
 						foreach ( $check_value as $file_url ) {
+							// Check image path.
+							$baseurl = wp_upload_dir()['baseurl'];
+							if ( ! is_numeric( $file_url ) && 0 !== strpos( $file_url, $baseurl ) ) {
+								throw new Exception( __( 'Invalid image path.', 'wp-job-manager' ) );
+							}
+
+							// Check mime types.
+							if ( ! empty( $field['allowed_mime_types'] ) ) {
+								$file_url  = current( explode( '?', $file_url ) );
+								$file_info = wp_check_filetype( $file_url );
+
+								if ( ! is_numeric( $file_url ) && $file_info && ! in_array( $file_info['type'], $field['allowed_mime_types'], true ) ) {
+									// translators: Placeholder %1$s is field label; %2$s is the file mime type; %3$s is the allowed mime-types.
+									throw new Exception( sprintf( __( '"%1$s" (filetype %2$s) needs to be one of the following file types: %3$s', 'wp-job-manager' ), $field['label'], $file_info['ext'], implode( ', ', array_keys( $field['allowed_mime_types'] ) ) ) );
+								}
+							}
+
+							// Check if attachment is valid.
 							if ( is_numeric( $file_url ) ) {
 								continue;
 							}
 							$file_url = esc_url( $file_url, [ 'http', 'https' ] );
 							if ( empty( $file_url ) ) {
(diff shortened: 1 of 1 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -407,6 +407,22 @@ protected function validate_fields( $values ) {
...
 					}
 				}
+				if ( 'file' === $field['type'] ) {
+					if ( is_array( $values[ $group_key ][ $key ] ) ) {
+						$check_value = array_filter( $values[ $group_key ][ $key ] );
+					} else {
+						$check_value = array_filter( [ $values[ $group_key ][ $key ] ] );
+					}
+					if ( ! empty( $check_value ) ) {
+						foreach ( $check_value as $file_url ) {
```
gkaragia, line 417: We could probably merge this check with the one above it to avoid code duplication. For example:
```
foreach ( $check_value as $file_url ) {
	if ( ! empty( $field['allowed_mime_types'] ) ) {
		// do mime checks here
	}
	// do the check that you added here.
}
```

renatho: Good!
I took the opportunity to refactor also the other validation block related to the file.

https://github.com/Automattic/WP-Job-Manager/pull/1989/commits/422f88882a24dcc064aeb534ccf2551c163337a8

jom (approved):
Looks good, works well!

tripflex:
This has broke CDN hosted images and other issues, and there is no filter or any other way to get around this besides downgrading to older version of WPJM @jom @gkaragia @renatho see #2035

PR #2001, opened 2020-05-03
add filter for user role on submit job user creation

### LINKED ISSUE #1997: Filter for user role in class-wp-job-manager-form-submit-job.php
**Is your feature request related to a problem? Please describe.**
In the job submit_handler an account is created if the user is not logged in. The role is set to the option get_option( 'job_manager_registration_role' ). Could you please add a filter to allow this option to be customised?

**Describe the solution you'd like**
Could you please add the following filter in includes/forms/class-wp-job-manager-form-submit-job.php, to replace Line 670 with:

'role'     => apply_filters('change_user_role_on_job_submission', get_option('job_manager_registration_role' ), $_POST),

**Describe alternatives you've considered**
I know there is a setting in the job manager settings but if there are more than one user roles, or the role needs changing dynamically depending on job submission values, then this filter makes that possible.

**Additional context**

tripflex:
Fixes #1997

* Adds `submit_job_form_create_account_role` filter

### Notes

So while this can be done via `job_manager_create_account_data` filter, I still went ahead and added it for the submit job account creation, since there are numerous addon plugins and themes that will call `wp_job_manager_create_account` so this one will be specific to user account creation from job submit area.

Added `$this` reference to be able to access fields or anything else from the filter if needed

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -667,7 +667,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 								'username' => ( job_manager_generate_username_from_email() || empty( $input_create_account_username ) ) ? '' : $input_create_account_username,
 								'password' => ( wpjm_use_standard_password_setup_email() || empty( $input_create_account_password ) ) ? '' : $input_create_account_password,
 								'email'    => sanitize_text_field( wp_unslash( $input_create_account_email ) ),
-								'role'     => get_option( 'job_manager_registration_role' ),
+								'role'     => apply_filters( 'submit_job_form_create_account_role', get_option( 'job_manager_registration_role' ), $values, $this ),
 							]
 						);
 					}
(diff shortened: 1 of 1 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -667,7 +667,7 @@ public function submit_handler() {
 								'username' => ( job_manager_generate_username_from_email() || empty( $input_create_account_username ) ) ? '' : $input_create_account_username,
 								'password' => ( wpjm_use_standard_password_setup_email() || empty( $input_create_account_password ) ) ? '' : $input_create_account_password,
 								'email'    => sanitize_text_field( wp_unslash( $input_create_account_email ) ),
-								'role'     => get_option( 'job_manager_registration_role' ),
+								'role'     => apply_filters( 'submit_job_form_create_account_role', get_option( 'job_manager_registration_role' ), $values, $this ),
```
jom, line 670: Mind adding documentation for this? I've put it on the next milestone.

tripflex: ahh sorry didn't see this, yeah i'll get an update pushed tomorrow

tripflex:
@jom Updated with documentation for the filter

jom (approved):
👍 Thanks for adding the docs. Looks good.

Issue #1976, opened 2020-02-27
Company Logo image sizes


richardmtl:
This is something that comes up on a semi-regular basis. See here, for example:

https://wordpress.org/support/topic/company-logo-9/

It'd be good to allow finer control over the logo size, so that people don't need to hack around with CSS or by blanket-deactivating cropping under Image Sizes. 

Related: https://github.com/Automattic/WP-Job-Manager/issues/1361 and https://github.com/Automattic/WP-Job-Manager/issues/1308

Marking as a bug as per @alexsanford , and Low (but I feel that it falls halfway between Normal and Low, ha!)

danjjohnson:
2752526-zen

Hastibe:
Also, just want to note that I couldn't get @richardmtl's suggestion (to deactivate cropping) in the support thread that he referenced to have any effect, despite trying it on a test site with just WP Job Manager active on Twenty Sixteen--I do appreciate all @richardmtl's efforts (and his opening this issue), though!

...Related to hoping this issue could be resolved, I'm a novice with GitHub, and is there a way to see what issues are being actively worked on, or anything like a roadmap for further development?

alexsanford:
We are currently in a phase of focusing on maintenance and bugfixes for WP Job Manager and its extensions. Once we've had a chance to do some triage of the backlog, this will go into our queue of bugs to be fixed.

And of course, PR's are always welcome! 🙂

Hastibe:
@alexsanford -- thanks for the clarification. Is there a roadmap of sorts for what maintenance and bugfixes are being worked on for WP Job Manager and its extensions, then?

I just see a lot of needs for maintenance and bugfixes, but hardly any activity (in terms of actual updates being made)--some of the extensions (with reported bugs) haven't been updated in over a year, which is really frustrating to see!

lamdayap:
Hey team and Richard!👋
I just added a small PR to allow the logos to display as rectangular, I added it as an option in Admin. Let me know whether this solves the original issue :) 

@Hastibe I'll share a small snippet of CSS as a hotfix that you can add. I've posted it on your thread in the WP forums.

jinnypark:
Another request in the forum thread [here](https://wordpress.org/support/topic/cropped-logos/).

geominy:
Hi team, do you perhaps have an update on this issue? Using WP Job Manager and would love a fix for the company logos.

Thank you!

vaccinemedia:
[...]

PR #2017, opened 2020-05-27
Add option to allow for rectangular logos - fixes issue 1976

lamdayap:
Fixes #
Issues with rectangular logos being resized into a square causing a squished look: https://github.com/Automattic/WP-Job-Manager/issues/1976
I added an option to in the job_listings admin page will dynamically adjust the css height property for company logos to auto so that users can display a rectangular logo. I made this an option in admin so that users have a choice of enforcing square logos still as it does create a neater look

### Testing instructions
Go into the job listing's settings page on Admin and toggle on the `Rectangular logos` option

### Screenshot / Video
Screen Shot 2020-05-27 at 9 09 44 pm
Screen Shot 2020-05-27 at 9 10 51 pm

### DIFF
--- a/assets/css/job-listings.less
+++ b/assets/css/job-listings.less
@@ -49,6 +49,17 @@ ul.job_listings {
 				vertical-align: middle;
 				box-shadow: none;
 			}
+			img.rectangular_company_logo {
+				width: 42px;
+				height: auto;
+				max-height: 80%;
+				position: absolute;
+				left: 1em;
+				float: left;
+				margin-right: 1em;
+				vertical-align: middle;
+				box-shadow: none;
+			}
 			div.position, div.location, ul.meta {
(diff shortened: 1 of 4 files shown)


tripflex:
Thanks for submitting a PR! I agree with @alexsanford

The check should probably be moved into `the_company_logo` function, using a filter to check whether to enable it or not, and just add the appropriate class for the output.  I say that because over the past 6+ years working for clients with WP Job Manager everyday, i've only been asked about this a handful of times, so I would assume it's a pretty edge case feature which would be best as a filter.

Many users and themes also override the template files so by doing this we move the functionality and support into `the_company_logo` function, so it will be compatible with customized template files and then we don't have to modify the default template file either.

Since `get_the_company_logo` accepts `$size` as second parameter should be fairly easy to do.

PR #2031, opened 2020-06-18
Fix test issues on Travis

renatho:
It's also being fixed here: https://github.com/Automattic/wp-job-manager-resumes/pull/321

* Fix PHPUnit issues on Travis.
  * Recently our builds stopped working because a WP compatibility issue with PHPUnit. So we updated to use the `5.7.*` for all PHP versions.
  * It caused another issue with the `runInSeparateProcess`, that also was fixed with: https://phpunit.readthedocs.io/en/latest/annotations.html#preserveglobalstate


renatho:
I'll wait to test it again tomorrow because it looks like we exceeded the daily quota from Google API with the key that we have configured in Travis. I added a `print_r` to figure out the issue here: https://travis-ci.org/github/Automattic/WP-Job-Manager/jobs/699879179

PR #2032, opened 2020-06-27
Fix jobs preview

### LINKED ISSUE #2028: Previewing new job created from editing existing submission fails
Original issue reported [here](https://github.com/Automattic/wp-job-manager-wc-paid-listings/issues/154).

---

reported in  2983442-zen

#### Steps to Reproduce
1. Complete job submission form ([submit_job_form])
2. Click on "Preview" page
3. Return to [job_dashboard]
4. Return to the page that the job submission form ([submit_job_form]) is on
5. Click on "You are editing an existing job. Create A New Job"
6. Enter details of new job into job submission form ([submit_job_form])
7. Click "Preview"

#### What I Expected

- the preview of the new job post is loaded, as normal.

#### What Happened Instead

-  the form refreshes and all entries into the job submission form ([submit_job_form]) are lost.

#### PHP / WordPress / WPJM version

[my test site]
-  PHP 7.3.5
- WP 5.4.1
- WPJM  1.34.1

#### Browser / OS version
- FF/Mac
- Chrome/Mac

renatho:
Fixes #2028

* Ensure that we clean the preview job clicking on "Create A New Job"
  * It wasn't working, because of the code flow/hooks order. We updated to run the form code previously when sending the `job_manager_form` also from `GET`. Previously it only happened from the `POST`.

### Testing instructions

1. Complete job submission form ([submit_job_form])
2. Click on "Preview" page
3. Return to [job_dashboard]
4. Return to the page that the job submission form ([submit_job_form]) is on
5. Click on "You are editing an existing job. Create A New Job"
6. Enter details of new job into job submission form ([submit_job_form])
7. Click "Preview"

### DIFF
--- a/includes/class-wp-job-manager-forms.php
+++ b/includes/class-wp-job-manager-forms.php
@@ -50,7 +50,7 @@ class WP_Job_Manager_Forms {
 	 */
 	public function load_posted_form() {
 		// phpcs:ignore WordPress.Security.NonceVerification.Missing -- Input is used safely.
-		$job_manager_form = ! empty( $_POST['job_manager_form'] ) ? sanitize_title( wp_unslash( $_POST['job_manager_form'] ) ) : false;
+		$job_manager_form = ! empty( $_REQUEST['job_manager_form'] ) ? sanitize_title( wp_unslash( $_REQUEST['job_manager_form'] ) ) : false;
 
 		if ( ! empty( $job_manager_form ) ) {
 			$this->load_form_class( $job_manager_form );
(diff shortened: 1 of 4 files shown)


includes/class-wp-job-manager-forms.php
```diff
@@ -50,7 +50,7 @@ public function __construct() {
 	 */
 	public function load_posted_form() {
```
renatho, line 51: I was thinking to rename this method, but I think it's not too weird to justify a deprecation only to rename the method. WDYT?

gkaragia: Let's leave it like this.

gikaragia (approved):
Looks good, works well!

PR #2036, opened 2020-07-09
Add a filter to enable external file blocking

### LINKED ISSUE #2035: Edit listing invalid image path error for image replaced by CDN URL
**Describe the bug**
_I'm not sure if this should be called a bug or enhancement so please change if necessary._
We use the[ Listify theme](https://themeforest.net/item/listify-wordpress-directory-theme/9602611) on our website, and recently installed an image optimization plugin ([Optimole](https://optimole.com/)) to process the images. This plugin adds their CDN url as a prefix to the original in order to refer to the processed image. While all working fine we now noticed that this causes an "invalid path error" when a user wants to edit their listing. This since in the `/includes/forms/class-wp-job-manager-form-submit-job.php` the code checks whether the image path start with <domain>/wp-content/uploads whereas in our case the URL looks like:  `https://mlpnsv5uiutt.i.optimole.com/........../https://salonstoelhuren.flywheelstaging.com/wp-content/uploads/job-manager-uploads/featured_image/2020/07/test.jpeg`. 

**To Reproduce**
Install Optimole and edit a listing. 

**Expected behavior**
Also allow images to be served from an external CDN when the images is the same (not sure if this would cause a security risk?)

In my case I now could add a code like this (P.s. I'm not a PhP programmer so code probably sucks)
```
if (strpos( $file_url, 'optimole') !== false ) {
   $file_url = substr($file_url, strpos($file_url, 'https://', strpos($file_url, 'https://')+1), strlen($file_url)); 
}
```
but it would be tedious to change this every time after an update, so if someone with more PHP skills now how I could add this to functions.php or something that would be great!

- I have deactivated the Optimole plugin and that solved the issue

**WordPress Environment**
- WordPress Version: 5.4.2
- WP Job Manager Version: 1.34.2
- PHP Version: 7.3

gkaragia:
Fixes #2035

* Hides blocking of external files behind a filter.
* The default behaviour is to not block external files as this can break sites which use CDNs.

### Testing instructions

* Follow the instructions in #1989 and check that external files are allowed.
* Add the filter:
```
add_filter(
	'job_manager_submit_job_reject_external_files',
	function( $val, $key, $group_key, $field ) {
	return true;
	},
	10,
	4
);
```
* Check that external files are now not allowed.

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -381,8 +381,20 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 						foreach ( $check_value as $file_url ) {
 							// Check image path.
 							$baseurl = wp_upload_dir()['baseurl'];
-							if ( ! is_numeric( $file_url ) && 0 !== strpos( $file_url, $baseurl ) ) {
-								throw new Exception( __( 'Invalid image path.', 'wp-job-manager' ) );
+
+							if ( ! is_numeric( $file_url ) ) {
+								/**
+								 * Set this flag to true to reject files from external URLs during job submission.
+								 *
+								 * @since 1.34.3
+								 *
+								 * @param bool  $reject_external_files  The flag.
+								 */
+								$reject_external_files = apply_filters( 'job_manager_submit_job_reject_external_files', false );
+
+								if ( $reject_external_files && 0 !== strpos( $file_url, $baseurl ) ) {
+									throw new Exception( __( 'Invalid image path.', 'wp-job-manager' ) );
+								}
 							}
 
 							// Check mime types.
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -379,8 +379,6 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 					}
 					if ( ! empty( $check_value ) ) {
 						foreach ( $check_value as $file_url ) {
-							// Check image path.
-							$baseurl = wp_upload_dir()['baseurl'];
 
 							if ( ! is_numeric( $file_url ) ) {
 								/**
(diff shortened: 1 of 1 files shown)


renatho:
Just mentioning the original issue to notify that this feature will be behind a filter: #1982

tripflex:
👍

I think it would also be good to add some other params to the filter, so if users want to setup handling to only reject external images, on say, a specific meta key or something, it would be possible by adding more to the filter.

Maybe something like:
```
$reject_external_files = apply_filters( 'job_manager_submit_job_reject_external_files', false, $key, $field, $values, $this );
```

At a minimum I would say at least `$key` and `$field`

tripflex:
Or another idea could be to use the filter to define the allowed remote URLs that can be used, maybe something like this:

```php
if ( ! is_numeric( $file_url ) ) {

	$baseurl = wp_upload_dir()['baseurl'];

	/**
	 * Set this flag to true to reject files from external URLs during job submission.
	 *
	 * @param bool $allowed_file_urls The flag.
	 *
	 * @since 1.34.3
	 *
	 */
	$allowed_file_urls = apply_filters( 'job_manager_submit_job_allowed_file_urls', array( $baseurl ), $key, $field, $values );

	if ( ! empty( $allowed_file_urls ) && str_replace( $allowed_file_urls, '', $file_url ) === $file_url ) {
		throw new Exception( __( 'Invalid image path.', 'wp-job-manager' ) );
	}
}
```

Basically by checking using `str_replace` and passing an array of valid URLs, as long as one of them matches the value returned from `str_replace` should be different than `$file_url` .. so if it's not, return the error meaning it's not one of the valid supported URLs

Just throwing out ideas here ...

renatho:
I like the idea to make it more flexible. I think we could make a very simple filter, just passing the `$file_url`, `$key`, and `$group_key`. So we have the path and the field to customize any validation that we want. WDYT @gkaragia?

gkaragia:
Hey @tripflex @renatho I think that since defining a list of allowed URLs can be implemented in the added filter we should better keep it simple. I ended up adding $key, $group_key and $field as arguments.

tripflex:
👍

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -381,8 +381,23 @@ protected function validate_fields( $values ) {
 						foreach ( $check_value as $file_url ) {
 							// Check image path.
 							$baseurl = wp_upload_dir()['baseurl'];
```
tripflex, line 383: This can be moved inside the `if( ! is_numeric( $file_url ) ) {` statement as it's only used specifically for that code (not a big deal but wanted to mention it)

gkaragia:
Ah sorry I forgot adding this change. Fixed in a6efd19ec4f31d14899c755ea08953796814f837

includes/forms/class-wp-job-manager-form-submit-job.php
gkaragia: Ah sorry I forgot adding this change. Fixed in a6efd19ec4f31d14899c755ea08953796814f837

PR #2037, opened 2020-07-10
Remove query args from paginate_links urls

### LINKED ISSUE #1912: Mark as Filled action generates incorrect messages with paginated jobs on job dashboard

JuanchoPestana:
Fixes #1912 

* The paginate_links() function adds the url args (from the current url) to each pagination link. The actions (mark as filled, not filled, etc) in the job dashboard were triggered with the url (For example: url/?action=mark_as_filled). The consequences of this were that if you changed the page, the pagination link would carry the "action, job_id, and _wpnonce" parameters to the new page, causing the action to be triggered again; thus, getting the message that the job was already filled/not filled. 
* This PR removes the args from the url for the pagination links so that the actions are not triggered again when changing pages.

#### Testing instructions:

1. Make sure you have enough jobs on your jobs dashboard that they are paginated at least to a page 2.
2. On the jobs dashboard, mark a job as filled; you will get the correct "X has been filled" message.
3. Click on the "2" or the "next arrow" in the pagination at the bottom of the jobs list.
4. You should not see any messages

* Remove args from pagination_link urls

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -61,6 +61,8 @@ class WP_Job_Manager_Shortcodes {
 		add_shortcode( 'job', [ $this, 'output_job' ] );
 		add_shortcode( 'job_summary', [ $this, 'output_job_summary' ] );
 		add_shortcode( 'job_apply', [ $this, 'output_job_apply' ] );
+
+		add_filter( 'paginate_links', [ $this, 'filter_paginate_links' ], 10, 1 );
 	}
 
 	/**
@@ -270,6 +272,16 @@ class WP_Job_Manager_Shortcodes {
 		);
 
 		return ob_get_clean();
+	}
+
+	/**
+	 * Filters the url from paginate_links to avoid multiple calls for same action in job dashboard
+	 *
+	 * @param string $link
+	 * @return string
+	 */
+	public function filter_paginate_links( $link ) {
+		return remove_query_arg( [ 'action', 'job_id', '_wpnonce' ], $link );
 	}
 
 	/**
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -87,9 +87,7 @@ class WP_Job_Manager_Shortcodes {
 	 * Handles actions which need to be run before the shortcode e.g. post actions.
 	 */
 	public function shortcode_action_handler() {
-		global $post;
-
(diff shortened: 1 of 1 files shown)


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -272,6 +274,16 @@ public function job_dashboard( $atts ) {
...
 
+	/**
+	 * Filters the url from paginate_links to avoid multiple calls for same action in job dashboard
+	 *
+	 * @param string $link
+	 * @return string
+	 */
+	public function filter_paginate_links( $link ) {
+		return remove_query_arg( [ 'action', 'job_id', '_wpnonce' ], $link );
+	}
```
alexsanford, line 285: I find myself concerned with this again. I had imagined before that this would only be called if a WPJM-related shortcode existed on the page, but looking again I think this would be called on _any_ page that calls the `paginate_links` function. This seems dangerous to me.

Do you think it would make sense to add a condition here to ensure that it only happens on the page(s) we want? E.g. see the `shortcode_action_handler` method. Maybe something like:

```suggestion
	public function filter_paginate_links( $link ) {
		global $post;
		
		if ( is_page() && has_shortcode( $post->post_content, 'job_dashboard' ) ) {
			return remove_query_arg( [ 'action', 'job_id', '_wpnonce' ], $link );
		}
		
		return $link;
	}
```

If so, maybe we could also extract that conditional logic to a private method (assuming we use the same logic as in `shortcode_action_handler`). Maybe `private function is_job_dashboard_page() { ... }`

JuanchoPestana: Hey Alex! 
You're totally right! It's a bit too aggressive... Your suggestion works perfectly! I just implemented it...

PR #2053, opened 2020-09-04
Remove extends in SCSS

jom:

* From @renatho's review in the Resume build PR, I learned that SASS `@extends` behaves differently between the LESS equivalent and adds new inherited classes instead of just putting the styles from that class. To get it to generate CSS more similar to the previous LESS version, I've moved them over to mixins. For backwards compatibility, I kept the `.jm-icon`, `.clearfix`, and `.display-icon` classes.

### Testing instructions

* Confirm no visual regressions.

### DIFF
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -78,8 +78,8 @@ a.wpjm-activate-licence-link:active {
 
 			&::before {
 
-				@include jm-icon()
-					content: "\e803";
+				@include jm-icon();
+				content: "\e803";
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


assets/css/admin.scss
```diff
@@ -78,8 +78,8 @@ a.wpjm-activate-licence-link:active {
 
 			&::before {
 
-				@extend .jm-icon;
-				content: "\e803";
+				@include jm-icon()
+					content: "\e803";
```
renatho, line 82: There are some unexpected tabs.

renatho (approved):
Looks good and seems to be working well.
There are just some unexpected tabs.

assets/css/admin.scss
jom: Ahh, that was `lint:css` after I forgot the `;` above. Fixed in 07732a9188b3e0684ac6d1be47db268804a76624

renatho (approved):

PR #2060, opened 2020-09-18
Add job_manager_select2_filters_args for search_categories filtering field, and fix job_manager_select2_args pollution

### LINKED ISSUE #2058: ajax-filters.js modifies job_manager_select2_args causing select2 errors
https://github.com/Automattic/WP-Job-Manager/blob/796975e4023c30eb03c8d8fb4c445a8bbad45775/assets/js/ajax-filters.js#L510-L516

Because in JavaScript when you set a variable equal to one that is an object/array, it assigns it by reference (think `&$variable` in PHP), so the code above essentially modifies `job_manager_select2_args` setting `allowClear` to `true`.

The issue here is two fold, if `job_manager_select2_args` is being used elsewhere on the page, it essentially modifies that object for every field, and second issue is that when using `allowClear` in Select2, you must also have a placeholder value set.

So the correct way to fix this would be to use `jQuery.extend()`, like this:

```javascript
if ( $.isFunction( $.fn.select2 ) && typeof job_manager_select2_args !== 'undefined' ) {
	var select2_args = $.extend( {}, job_manager_select2_args, { allowClear: true, minimumResultsForSearch: 10 } );
	$( 'select[name^="search_categories"]:visible' ).select2( select2_args );
}
```

The only problem is this doesn't solve the issue with the missing placeholder, which could be set like this, to only set `placeholder` value if one does not already exist in `job_manager_select2_args`:

```javascript
var select2_args = $.extend( { placeholder: 'Select Category' }, job_manager_select2_args, { allowClear: true, minimumResultsForSearch: 10 } );
```

Problem with this, is that it's not a translated value, and we wouldn't want to set the `placeholder` on the main `job_manager_select2_args` object as that would then apply to every single select2 field that uses that object for initialization.

So I think the best solution would instead be to create a new object with `wp_localize_script` specifically for the filters area.  I'll go ahead and submit a PR to get you guys take on this.

tripflex:
Fixes #2058

* Add `job_manager_select2_filters_args` for select2 initialization on search categories field

### Testing instructions

* Load up job lists page, and check `job_manager_select2_filters_args` in console

### DIFF
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -432,10 +432,14 @@ class WP_Job_Manager {
 		wp_localize_script( 'wp-job-manager-ajax-filters', 'job_manager_ajax_filters', $ajax_data );
 
 		if ( isset( $select2_args ) ) {
-			$select2_filters_args = array_merge( $select2_args, [ 'allowClear'              => true,
-			                                                      'minimumResultsForSearch' => 10,
-			                                                      'placeholder'             => __( 'Any Category', 'wp-job-manager' )
-			] );
+			$select2_filters_args = array_merge(
+				$select2_args,
+				[
+					'allowClear'              => true,
+					'minimumResultsForSearch' => 10,
+					'placeholder'             => __( 'Any Category', 'wp-job-manager' ),
+				]
+			);
 
 			wp_localize_script(
 				'select2',
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -437,12 +437,10 @@ class WP_Job_Manager {
 			                                                      'placeholder'             => __( 'Any Category', 'wp-job-manager' )
 			] );
 
-			$select2_filters_args = apply_filters( 'job_manager_select2_filters_args', $select2_filters_args );
-
 			wp_localize_script(
 				'select2',
 				'job_manager_select2_filters_args',
-				$select2_filters_args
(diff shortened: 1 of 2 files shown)


includes/class-wp-job-manager.php
```diff
@@ -431,6 +431,19 @@ public function frontend_scripts() {
 		wp_register_script( 'wp-job-manager-job-submission', JOB_MANAGER_PLUGIN_URL . '/assets/js/job-submission.min.js', [ 'jquery' ], JOB_MANAGER_VERSION, true );
 		wp_localize_script( 'wp-job-manager-ajax-filters', 'job_manager_ajax_filters', $ajax_data );
 
+		if ( isset( $select2_args ) ) {
+			$select2_filters_args = array_merge( $select2_args, [ 'allowClear'              => true,
```
jom, line 435: Not sure why this isn't failing in CI, but I think this needs to be expanded to different lines (see phpcs result; phpcbf on this file should fix most of the issues). Might be worth rebasing with master at the same time as we've updated some of these rules.

jom (commented):
Minor PHPCS issues but otherwise this is good! There is a weird issue with the "clear" button on select2, but not related to this PR.

tripflex:
@jom sorry didn't have phpcs enabled, enabled it and fixed the issues

jom (approved):
Looks good!

PR #2069, opened 2020-10-26
Fixes #2068

### LINKED ISSUE #2068: is_wpjm_page returns true if no pages are assigned
**Describe the bug**
In a fresh WP Networks installation, if I do not create any WP Job Manager pages then the function `is_wpjm_page` returns true because `is_page` returns true if an empty value is passed.

**To Reproduce**
1. Install WP Networks
2. Install WP Job Manager
3. Skip the steps to setup Pages.
4. Check `is_wpjm_page` on any page via `functions.php`. It will always return `true`.

**Expected behavior**

ibndawood:
Added a check to `$wpjm_page_ids` before calling `is_page`. This is because according to the notes here: https://developer.wordpress.org/reference/functions/is_page/#notes `is_page` will return `true` if an empty value is passed.

Fixes #2068

Added a check to `$wpjm_page_ids` before calling `is_page`. This is because according to the notes here: https://developer.wordpress.org/reference/functions/is_page/#notes `is_page` will return `true` if an empty value is passed.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -771,8 +771,10 @@ function is_wpjm_page() {
 		 * @param int[] $wpjm_page_ids
 		 */
 		$wpjm_page_ids = array_unique( apply_filters( 'job_manager_page_ids', $wpjm_page_ids ) );
-
-		$is_wpjm_page = is_page( $wpjm_page_ids );
+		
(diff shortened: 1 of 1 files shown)


jom (commented):
Looks good! Thanks for the PR.

jom (approved):

PR #2075, opened 2020-11-26
Fix missing variable with widget caching

### LINKED ISSUE #2009: Variable $cache is undefined
I found this bug while finding untranslatable strings for a previous issue.

Please check this line, `$cache` is undefined: https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-widget.php#L105

I guess `$cache` is get like this:

```
$cache = wp_cache_get( $this->widget_id, 'widget' );

if ( ! is_array( $cache ) ) {
	$cache = [];
}
```

jom:
Fixes #2009

* Before updating the cached instances of a widget, it retrieves the current value and builds off of that. It is a bit confusing, but `$this->widget_id` is the general widget ID (`widget_featured_jobs`) where `$args['widget_id']` is the unique instance ID of the widget (`widget_featured_jobs-5`). 

### Testing instructions

* Level 1: Just add a WPJM widget to a widget area and make sure it renders as expected.
* Level 2: With object caching set up, make sure a widget's rendering is cached between page loads. You could test this by changing a displayed post title in the database (not with WordPress).

### DIFF
--- a/includes/class-wp-job-manager-widget.php
+++ b/includes/class-wp-job-manager-widget.php
@@ -102,6 +102,12 @@ class WP_Job_Manager_Widget extends WP_Widget {
 	 * @param string $content
 	 */
 	public function cache_widget( $args, $content ) {
+		$cache = wp_cache_get( $this->widget_id, 'widget' );
+
+		if ( ! is_array( $cache ) ) {
+			$cache = [];
+		}
+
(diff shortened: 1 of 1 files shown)


renatho:
Looks good and works well! 🎉
I tested debugging and make sure it was appending to the array instead of to create a new.

PR #2090, opened 2020-12-17
Add job_manager_get_form_action filter fixes #2089

### LINKED ISSUE #2089: Form ACTION always includes all query parameters
https://github.com/Automattic/WP-Job-Manager/blob/700dc8e28018fb7bd8e576e67d527a6477b455a9/includes/abstracts/abstract-wp-job-manager-form.php#L185-L195

This method is called when outputting templates to define `action` in the forms, but an issue i'm running into is that this ends up setting the `action` to the full URL of current page, including all query variables.

Normally this would be fine, but the problem i'm finding is that when a URL parameter is used to set the value on a field, WordPress throws a `A variable mismatch has been detected` error, which I tracked down to this specific line of code:
https://core.trac.wordpress.org/browser/trunk/src/wp-includes/class-wp.php#L299

The issue comes up with taxonomies, because they are submitted with the term ID whereas I allow users to define the slug in the query parameter (for better UX by having the "slug" in the URL and not just a number).

Unfortunately there isn't any way to filter on this to remove the taxonomy query parameters, so it would be great it we could add a filter for when this method is called, so i can modify the query param to remove the taxonomy ones.

i'll submit a PR for this

tripflex:
Currently without overriding the entire template, there's no way to customize or change the form action URL.  This filter allows customization of the form action URL.

Fixes #2089

* Adds `job_manager_get_form_action` filter

### Testing instructions

*

### New/Updated Hooks

```
/**
 * Alter form submit action URL
 *
 * Before submitting or editing a job, alter the posted values before they get stored into the database.
 *
 * @param string $action Form action value
 * @param WP_Job_Manager_Form $this
 *
 * @since @@version
 *
 */
return apply_filters( 'job_manager_get_form_action', $action, $this );
 ```

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -190,8 +190,20 @@ abstract class WP_Job_Manager_Form {
 	 */
 	public function get_action() {
 		$default_action = isset( $_SERVER['REQUEST_URI'] ) ? sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ) : '';
+		$action = esc_url_raw( $this->action ? $this->action : $default_action );
 
-		return esc_url_raw( $this->action ? $this->action : $default_action );
+		/**
+		 * Alter form submit action URL
+		 *
+		 * Before submitting or editing a job, alter the posted values before they get stored into the database.
+		 *
+		 * @param string $action Form action value
+		 * @param WP_Job_Manager_Form $this
+		 *
+		 * @since @@version
+		 *
+		 */
+		return apply_filters( 'job_manager_get_form_action', $action, $this );
 	}
 
 	/**
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -195,13 +195,12 @@ abstract class WP_Job_Manager_Form {
 		/**
 		 * Alter form submit action URL
 		 *
-		 * Before submitting or editing a job, alter the posted values before they get stored into the database.
+		 * @since 1.35.0
 		 *
-		 * @param string $action Form action value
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: includes/abstracts/abstract-wp-job-manager-form.php [hunk 1]
```diff
@@ -190,8 +190,20 @@ public function show_messages() {
...
 		$default_action = isset( $_SERVER['REQUEST_URI'] ) ? sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ) : '';
+		$action = esc_url_raw( $this->action ? $this->action : $default_action );
 
-		return esc_url_raw( $this->action ? $this->action : $default_action );
+		/**
+		 * Alter form submit action URL
+		 *
+		 * Before submitting or editing a job, alter the posted values before they get stored into the database.
+		 *
+		 * @param string $action Form action value
```
jom, line 200: These need to be aligned.

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -190,8 +190,20 @@ public function show_messages() {
...
+		 * @param WP_Job_Manager_Form $this
+		 *
+		 * @since @@version
```
jom, line 203: Go ahead and set this to 1.35.0

jom (changes requested):
Seems like a good idea.

includes/abstracts/abstract-wp-job-manager-form.php
tripflex: Updated and removed incorrect description, moved @since above params to match rest of plugin filters, added missing @return tag

tripflex:
@jom thanks updated, removed wrong description, moved @since above params to match rest of plugin, and added missing `@return`

jom (approved):
Looks good!

PR #2099, opened 2021-01-11
Fix issue with double slash in webpack builds

jom:

* Minor fix for new webpack builds. The file loader output would have a double-slash between the path and filename. 
* Test break is unrelated. I'll fix it in a new PR.

### Testing instructions

* Run `npm run build` and verify assets load.


gkaragia (approved):
LGTM!

PR #2102, opened 2021-01-12
job_category frontend validation fix, select2 target job_category only

### LINKED ISSUE #2095: HTML5 validation on job_category on submit form assumes only Select2 element
https://github.com/Automattic/WP-Job-Manager/blob/a0a769648572e54cc64ed6b95fc466334fb24be7/assets/js/job-submission.js#L91-L99

as well as here

https://github.com/Automattic/WP-Job-Manager/blob/a0a769648572e54cc64ed6b95fc466334fb24be7/assets/js/job-submission.js#L148-L152

Both of these are targeting the "input" created by Select2 for multi-select fields where you "search" a value.  This jQuery selector is assuming that this is the only multiselect field on the page.  Instead of directly targeting that element, we should be targeting that element INSIDE of the job_category fieldset wrapper.

I'll work on getting a PR put together to fix this, basically what is happening is if `job_category` is invalid and there's another multiselect field on the page, it will show validation error on the first Select2 field on the page.

tripflex:
Fixes #2095

* Fixes frontend validation of `job_category` field (`jobCategory.val()` always returns true because value is empty array `[]`)
* Fixes frontend validation targeting first Select2 input element, now specifically targets `job_category` select2 input element

### Testing instructions

* Fill out job submit form, leave `job_category` field empty, should validate on frontend without submitting
* After validation error, select a value, validation should clear

### DIFF
--- a/assets/js/job-submission.js
+++ b/assets/js/job-submission.js
@@ -91,7 +91,7 @@ jQuery(document).ready(function($) {
 		if ( jobCategoryFieldIsInvalid() ) {
 			$(this).find( 'input[type=submit]' ).blur();
 
-			var jobCategoryInput = $( '.select2-search__field' )[0];
+			var jobCategoryInput = $( '.fieldset-job_category .select2-search__field' )[0];
 			jobCategoryInput.setCustomValidity( job_manager_job_submission.i18n_required_field );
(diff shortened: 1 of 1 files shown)


jom (approved):
Good changes. Works well!

PR #2103, opened 2021-01-12
Fix validation and sanitization of application method

### LINKED ISSUE #1814: Submitting existing job doesn't pre-fill with posted values

### LINKED ISSUE #2067: "Application email/URL" expects the the URL protocol but is not clear in the error or description
The "Application email/URL" job submission field expects the protocol (HTTP:// or HTTPS://) when entering a URL. However this is not explicitly stated nor is it clear in the error message when one does not enter the protocol. 

The only error shown is `Application email/URL is a required field`. 

We should make it clearer by using description, placeholder text, and/or error message that the protocol is required.

3415016-zen

jom:
Fixes #2067
Fixes #1814
Replaces #1829

* Adds schema (http or https, depending on filter) to URL application methods when possible/necessary.
* Allows sanitization errors to show on required fields. To do this, added an extra `before_sanitize` callback property on the field. Implemented internally but optional to implement (default is based on emptiness of sanitized value) for custom fields.
* Invalid emails are cleared.
* Note: saving to drafts will allow invalid data to be saved, but not non-sanitized data. When job is properly submitted, validation will occur. 
* Eventually we should add field level error handling. Too large of task for this PR, though.

### Testing instructions

* Using all setting variations for Application Method (email, url, both), fill out the job submission form. 
* Fill out the URL/both frontend application form field with various invalid URLS (note, `filter_var` is pretty relaxed in its URL validation rules). Try just putting "google.com" and verifying it prefixes it with the schema.
* Fill out the email/both frontend application form field with various invalid emails. This is a bit more strict... for example `()@example.com` shouldn't work. 

### New/Updated Hooks

* `job_manager_form_assumed_url_scheme`: Allow setting some URLs to a `https` schema by default when no schema set.

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -74,16 +74,6 @@ abstract class WP_Job_Manager_Form {
 	public $form_name = '';
 
 	/**
-	 * Field names of values that had a value on submission but may have been cleared during sanitization.
-	 *
-	 * This is optional and is a helper to be used when sanitation errors should be displayed over
-	 * empty required fields.
-	 *
-	 * @var array
-	 */
-	public $values_existed = [];
-
-	/**
 	 * Cloning is forbidden.
 	 */
 	public function __clone() {
@@ -450,8 +440,13 @@ abstract class WP_Job_Manager_Form {
 					$values[ $group_key ][ $key ] = $this->get_posted_field( $key, $field );
 				}
 
-				// Set fields value.
-				$this->fields[ $group_key ][ $key ]['value'] = $values[ $group_key ][ $key ];
+				$this->fields[ $group_key ][ $key ]['empty'] = true;
+				if ( null !== $values[ $group_key ][ $key ] ) {
+					$this->fields[ $group_key ][ $key ]['empty'] = false;
+
+					// Set fields value.
+					$this->fields[ $group_key ][ $key ]['value'] = $values[ $group_key ][ $key ];
+				}
 			}
 		}
 
@@ -494,7 +489,12 @@ abstract class WP_Job_Manager_Form {
 		if ( 'url' === $sanitizer ) {
 			return esc_url_raw( $this->normalize_url( $value ) );
 		} elseif ( 'email' === $sanitizer ) {
-			return sanitize_email( $value );
+			$sanitized_value = sanitize_email( $value );
+			if ( $sanitized_value !== $value ) {
+				return '';
+			}
+
+			return $sanitized_value;
 		} elseif ( 'url_or_email' === $sanitizer ) {
 			if ( is_email( $value ) ) {
 				return sanitize_email( $value );
@@ -543,7 +543,7 @@ abstract class WP_Job_Manager_Form {
 	 *
 	 * @param  string $key
 	 * @param  array  $field
-	 * @return string|array
+	 * @return string|array|null
 	 */
 	protected function get_posted_field( $key, $field ) {
 		// Allow custom sanitizers with standard text fields.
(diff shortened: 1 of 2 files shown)


includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -651,10 +732,16 @@ protected function upload_file( $field_key, $field ) {
...
 			if ( ! empty( $field['multiple'] ) ) {
-				return $file_urls;
+				$value = $file_urls;
 			} else {
-				return current( $file_urls );
+				$value = current( $file_urls );
 			}
 		}
+
+		if ( isset( $field['before_sanitize'] ) && is_callable( $field['before_sanitize'] ) ) {
```
jom, line 741: TBH this doesn't do much and is called in a weird spot, but just thought I'd add it for consistency.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -358,7 +358,11 @@ protected function should_application_field_skip_email_url_validation() {
 	protected function validate_fields( $values ) {
 		foreach ( $this->fields as $group_key => $group_fields ) {
 			foreach ( $group_fields as $key => $field ) {
-				if ( $field['required'] && empty( $values[ $group_key ][ $key ] ) ) {
+				if (
+					$field['required']
+					&& empty( $values[ $group_key ][ $key ] )
+					&& ( ! isset( $field['empty'] ) || $field['empty'] )
```
jom, line 364: This makes the `empty` parameter optional if set by the posted field callback.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -451,31 +455,32 @@ protected function validate_fields( $values ) {
...
 		// Application method.
-		if ( ! $this->should_application_field_skip_email_url_validation() && isset( $values['job']['application'] ) && ! empty( $values['job']['application'] ) ) {
+		if ( ! $this->should_application_field_skip_email_url_validation() && isset( $values['job']['application'] ) ) {
 			$allowed_application_method   = get_option( 'job_manager_allowed_application_method', '' );
-			$values['job']['application'] = str_replace( ' ', '+', $values['job']['application'] );
+
+			$is_valid = true;
+
+			// phpcs:ignore WordPress.Security.NonceVerification.Missing -- Nonce checked earlier when required.
+			$posted_value = isset( $_POST['application'] ) ? sanitize_text_field( wp_unslash( $_POST['application'] ) ) : false;
```
jom, line 464: Relaxing the sanitization for the re-submit just so it isn't quite as jarring for the user.

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -432,6 +432,18 @@ protected function get_posted_fields() {
...
 
+				$this->fields[ $group_key ][ $key ]['empty'] = false;
+
+				if ( ! isset( $field['before_sanitize'] ) ) {
+					$field['before_sanitize'] = function ( $value ) use ( $group_key, $key ) {
+						if ( is_string( $value ) ) {
+							$value = trim( $value );
+						}
+
+						$this->fields[ $group_key ][ $key ]['empty'] = empty( $value );
```
gkaragia, line 443: Won't this report more values that it should as empty? We were checking for null only previously.

gkaragia (commented):
Looks good! There might be a possible issue.

includes/abstracts/abstract-wp-job-manager-form.php
jom: This is essentially just an earlier check from what happens in `\WP_Job_Manager_Form_Submit_Job::validate_fields` for the `x is a required field`, which uses and has always used just `empty()`.

gkaragia: Cool! Missed that!

gkaragia (approved):
Looks good, works well!

PR #2114, opened 2021-01-28
Update date methods and expire jobs at the end of the day

### LINKED ISSUE #919: Use WordPress timezone for expiring jobs and sending emails
When a user submits a job and adds a closing date, the post disappears at the start of that day (ie. Jan 1, 00:00h) as opposed to the end of the day (ie. Jan 1, 23:59). 

This is something that is an issue for my client as they have listings that are closing before the actual deadline.

jom:
Fixes #919

* Update the date methods used and standardize all date rules to use the WP timezone.
* Expire jobs at the end of the day. Introduced filter `job_manager_jobs_expire_end_of_day` to revert to the original (expire at the start of the expiration day). We can easily flip this back.

We might have some changes to make in Application Deadline, but I don't _think_ anything urgent.

### Testing instructions

* Make sure dates are saved correctly when approving jobs and editing from backend. (Note: Don't forget you can't clear a job expiration if duration is set in settings)
* Dates appear correctly in metabox and on Job Dashboard page.
* In the database directly, have a job expire (`_job_expires` post meta) yesterday or today with `add_filter( 'job_manager_jobs_expire_end_of_day', '__return_false' );`. Run the `job_manager_check_for_expired_jobs` job manually and make sure they are set to a post_status=expired.

### New/Updated Hooks
- `job_manager_jobs_expire_end_of_day`

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -898,7 +898,7 @@ class WP_Job_Manager_Post_Types {
 		$expiration_date = $this->get_job_expiration( $post );
 
 		// Clear the expiration field if it has expired.
-		if ( $expiration_date && $this->has_job_expired( $post ) ) {
+		if ( $this->has_job_expired( $post ) ) {
 			$this->set_job_expiration( $post, null );
 		}
 
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -1743,7 +1743,8 @@ class WP_Job_Manager_Post_Types {
 		}
 
 		// Checks for valid date.
-		if ( ! DateTimeImmutable::createFromFormat( 'Y-m-d', $meta_value ) ) {
+		$test_date = DateTimeImmutable::createFromFormat( 'Y-m-d', $meta_value, wp_timezone() );
+		if ( ! $test_date || $test_date->format( 'Y-m-d' ) !== $meta_value ) {
 			return '';
 		}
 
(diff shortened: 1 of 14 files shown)


includes/class-wp-job-manager-post-types.php
```diff
@@ -888,32 +895,135 @@ public function set_expiry( $post ) {
 			return;
 		}
 
-		// See if it is already set.
-		if ( metadata_exists( 'post', $post->ID, '_job_expires' ) ) {
-			$expires = get_post_meta( $post->ID, '_job_expires', true );
-			if ( $expires && strtotime( $expires ) < current_time( 'timestamp' ) ) {
-				update_post_meta( $post->ID, '_job_expires', '' );
-			}
+		$expiration_date = $this->get_job_expiration( $post );
```
alexsanford, line 898: Do we even need this? Couldn't we just use `$this->has_job_expired( $post )` on its own?

alexsanford (commented):
When I try to set the expiry date on a listing, it doesn't seem to work. When I save a listing, regardless of what I have set for the "Listing Expiry Date", it seems to clear the expiry date. I don't have any WPJM extensions activated, and I've confirmed this is a regression from `master`.

Otherwise, this is looking good 👍

includes/class-wp-job-manager-post-types.php
jom: Yup bd86cf7a86dbb7b2f228495455b2a188727b4f72

jom:
Fixed that validation bug in 8e506e68df263a35cdffc680a4fd525072bf2597. As mentioned in Slack, it was a _very good bug_.

`strtotime()` will interpret date in PHP TZ / UTC unless otherwise specified. I checked for other usages and we should be okay (one wee one in usage tracking, but not too worried about it).

alexsanford (approved):
That fixes it for me! Changes look good, all seems to work 🚀

PR #2117, opened 2021-02-08
Add help text for explaining time of day that job listing will expire

jom:

* Minor change to add help text for the time of day that a job listing will expire.

Screen Shot 2021-02-08 at 6 57 01 pm

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -1473,6 +1473,11 @@ class WP_Job_Manager_Post_Types {
 			$application_method_placeholder = __( 'https://', 'wp-job-manager' );
 		}
 
+		$job_expires_description = __( 'Job listing expires at the end of the selected day.', 'wp-job-manager' );
+		if ( ! self::instance()->jobs_expires_end_of_day() ) {
+			$job_expires_description = __( 'Job listing expires at the start of the selected day.', 'wp-job-manager' );
+		}
+
 		$fields = [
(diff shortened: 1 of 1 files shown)


includes/class-wp-job-manager-post-types.php
```diff
@@ -1473,6 +1473,11 @@ public static function get_job_listing_fields() {
 			$application_method_placeholder = __( 'https://', 'wp-job-manager' );
 		}
 
+		$job_expires_description = __( 'Job listing expires at the end of the selected day.', 'wp-job-manager' );
```
alexsanford, line 1476: I wonder if we could just say "... at the end of the day"?

includes/class-wp-job-manager-post-types.php
```diff
@@ -1473,6 +1473,11 @@ public static function get_job_listing_fields() {
...
+		if ( ! self::instance()->jobs_expires_end_of_day() ) {
+			$job_expires_description = __( 'Job listing expires at the start of the selected day.', 'wp-job-manager' );
```
alexsanford, line 1478: Same as above, could we just say "... at the start of the day"?

alexsanford (approved):
A small comment on wording, but otherwise LGTM

includes/class-wp-job-manager-post-types.php
jom: Yup, had that originally then I think I overthought it when in the interface. Changed.

PR #2140, opened 2021-04-14
check for job cat select2 search before handle validation fixes #2139

### LINKED ISSUE #2139: jQuery/JavaScript error when Job Category is single select
Regression added from this commit https://github.com/Automattic/WP-Job-Manager/commit/dc26b43a9e64edfd683fe85f0bece9219b58d959 which attempts to target specific job category select2 input.

From this PR:
https://github.com/Automattic/WP-Job-Manager/pull/2102

When the `job_category` field is changed to a single taxonomy select, the Select2 `.select2-search__field` is actually appended to the body instead of the field itself (when multiple it's the field itself).

Quick and easy fix i'll submit a PR for now

tripflex:
Fixes #2139 

* Adds check around Select2 input for `job_category` before triggering validation handling

### Testing instructions

* Goto submit listing page, and don't fill out `job_category` field, confirm it throws validation error

### DIFF
--- a/assets/js/job-submission.js
+++ b/assets/js/job-submission.js
@@ -92,8 +92,10 @@ jQuery(document).ready(function($) {
 			$(this).find( 'input[type=submit]' ).blur();
 
 			var jobCategoryInput = $( '.fieldset-job_category .select2-search__field' )[0];
-			jobCategoryInput.setCustomValidity( job_manager_job_submission.i18n_required_field );
-			jobCategoryInput.reportValidity();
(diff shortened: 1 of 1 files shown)


tripflex:
@jom can we try and get this in the 1.35.1 release please?  Numerous clients with manual patches added to fix this, afraid if not in the next release will have a ton of users with issues because they update without making sure this patch was included 😛

jom (approved):
Nice! This works now with the changes.

Issue #2127, opened 2021-03-07
Pending payment jobs not showing in the job dashboard


Gnodesign:
**Describe the bug**
Not sure if this should be handled via WC Paid Listings but since the Job Dashboard is handled by WPJM I will just submit this issue here. 

If a user decides to leave the page during job submission while they are on the checkout page the job is saved as "pending_payment" in the backend but can not be shown in the job dashboard. This is confusing for users because they think the job never got saved and they have to refill everything again.

We can add the "pending_payment" status via the `'job_manager_get_dashboard_jobs_args'` filter but I think this would be a lot better if it were supported by the plugin to allow users to continue their submission as it happens with pending jobs (when the option is enabled).

https://github.com/Automattic/WP-Job-Manager/blob/5efcfd4632ba5492f5fa480d28b375f039878a9b/includes/class-wp-job-manager-shortcodes.php#L268-L277

**To Reproduce**
1. **WC Paid Listings required**
2. Start a job submission & leave the page on checkout
3. Go to **Job Dashboard**

**Expected behavior**
- Show "pending_payment" jobs in the job dashboard
- Allow users to "continue submission" from the job dashboard (if pending edits 

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
- WordPress Version: 5.6.2
- WP Job Manager Version: 1.35.0
- PHP Version:

Hastibe:
Yes, could this please be prioritized? I reported this to WPJM support almost a year ago (in **May 2020**) and [a separate issue](https://github.com/Automattic/WP-Job-Manager/issues/2022) about this was opened by another user in **June 2020***. Is anyone working on this? Can anyone?

Gnodesign:
@Hastibe in case you need to show "pending_payment" jobs in your job dashboard you can use the following snippet:
https://gist.github.com/Gnodesign/adfa7778b1c5d246549cd57e196ad3ea

Just a work around, there should be native support for this though.

Hastibe:
[...]

PR #2141, opened 2021-04-15
Add jobs pending payment to the Jobs dashboard

yscik:

* Include jobs with `pending_payment` status in the job dashboard
* Add a `Continue Submission` action to pending jobs

### Testing instructions

* Setup WC paid listing
* Start submitting a job with a new package
* Stop when reaching the WooCommerce checkout
* Open Job dashboard, and see that the job is there
* Click `Continue Submission` to try again

### Known issues

* This just takes back the user to the start of the job submission. If they select a package, but still have one in their cart, they'll have both packages in cart at checkout. 
  — should we clear the cart?
  — should this be more clever, and direct the user to the checkout if there is a related package in the cart?
  (in both cases the action should likely move to the paid listings plugins)

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -418,6 +418,10 @@ class WP_Job_Manager_Shortcodes {
 				break;
 			case 'pending_payment':
 			case 'pending':
+			$actions['continue'] = [
+				'label' => __( 'Continue Submission', 'wp-job-manager' ),
+				'nonce' => $base_nonce_action_name,
+			];
 				if ( WP_Job_Manager_Post_Types::job_is_editable( $job->ID ) ) {
 					$actions['edit'] = [
 						'label' => __( 'Edit', 'wp-job-manager' ),
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -268,7 +268,7 @@ class WP_Job_Manager_Shortcodes {
 	private function get_job_dashboard_query_args( $posts_per_page = -1 ) {
 		$job_dashboard_args = [
 			'post_type'           => 'job_listing',
-			'post_status'         => [ 'publish', 'expired', 'pending', 'draft', 'preview' ],
+			'post_status'         => [ 'publish', 'expired', 'pending', 'draft', 'preview', 'pending_payment' ],
 			'ignore_sticky_posts' => 1,
 			'posts_per_page'      => $posts_per_page,
 			'orderby'             => 'date',
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -418,10 +418,10 @@ class WP_Job_Manager_Shortcodes {
 				break;
 			case 'pending_payment':
 			case 'pending':
-			$actions['continue'] = [
-				'label' => __( 'Continue Submission', 'wp-job-manager' ),
-				'nonce' => $base_nonce_action_name,
-			];
+				$actions['continue'] = [
+					'label' => __( 'Continue Submission', 'wp-job-manager' ),
+					'nonce' => $base_nonce_action_name,
+				];
 				if ( WP_Job_Manager_Post_Types::job_is_editable( $job->ID ) ) {
 					$actions['edit'] = [
 						'label' => __( 'Edit', 'wp-job-manager' ),


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -418,6 +418,10 @@ public function get_job_actions( $job ) {
 				break;
 			case 'pending_payment':
 			case 'pending':
+				$actions['continue'] = [
```
jom, line 421: We might need to split `pending` and `pending_payment` here. `pending` is if they've actually submitted it but `pending_payment` is before the process. `Edit` (when available based on setting) makes more sense for `pending` and `pending_payment`. 

We should probably put this in an action under `default:`, but somehow not break compatibility with older WCPL that don't hook into the new action.

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -268,7 +268,7 @@ private function is_job_available_on_dashboard( WP_Post $job ) {
 	private function get_job_dashboard_query_args( $posts_per_page = -1 ) {
 		$job_dashboard_args = [
 			'post_type'           => 'job_listing',
-			'post_status'         => [ 'publish', 'expired', 'pending', 'draft', 'preview' ],
+			'post_status'         => [ 'publish', 'expired', 'pending', 'draft', 'preview', 'pending_payment' ],
```
jom, line 271: We should probably hook into `job_manager_get_dashboard_jobs_args` from WCPL.

Hastibe:
@yscik and @jom -- really glad to see this being moved on! I've been using [Gnodesign's snippets](https://github.com/Automattic/WP-Job-Manager/issues/2127) to achieve this on a very active job board, and wanted to chime in from our experience:

> This just takes back the user to the start of the job submission. If they select a package, but still have one in their cart, they'll have both packages in cart at checkout. 
> — should we clear the cart?

Having the job package duplicated at checkout **is** problematic--it is not an unusual occurrence on our job board where our users don't notice such a duplication and place an order for twice as much as they intended to and then have to follow-up with us for a refund (leaving a bad taste in their mouth, and causing extra work on all sides). 

On the other hand, having their WooCommerce cart unexpectedly cleared of other job packages (like job posts they wanted to purchase in advance from the WooCommerce product pages) or other products they'd added to it would also be confusing to users (and job board operators who don't know that this is expected behavior).

I think an **ideal solution** would be that any job package that has an attached Job Listing with an identical title is removed--i.e. just removing what would be the duplicate job package.

> — should this be more clever, and direct the user to the checkout if there is a related package in the cart?

I think it's important that the user is able to still view and edit the submission, and that it **doesn't** just take them straight to checkout. It is common on our job board that a user abandons a job post on the "Choose a Package" page or checkout page, and when coming back to continue its submission wants to verify the details of the job post and/or edit it before submitting it. **UPDATE:** one additional reason added in comment below. 

_**Side note**, in case this is also relevant: when the "Paid Listings Flow" is "Choose a package after entering job details," as on my site, and a user takes a job post to the "Choose a Package" page and then abandons it, the job post is listed as "Pending Payment" but is **not** added to the cart (since a package hasn't been chosen yet). Pending Payment job packages are only added to the cart **after** a package is selected and the "Submit" button is hit on the "Choose a Package" page._

Lastly, we also realized that having job posts labeled as "Pending Payment" on the front-end is **alarming** to users--our users saw this, and were concerned that they were going to be getting an invoice or have their card billed, so we actually re-labeled the "Pending Payment" status to "**Draft**" on the front-end user dashboard, which I would strongly encourage doing. 

After all, an abandoned job post is **not** "Pending Payment", it really is still just a "Draft" (for the user, at least), and I think this is the correct label for it (at least on the front-end for users).

I hope this is helpful, and want to offer that I am happy to contribute in whatever way I can with this (beyond coding, which I don't have your skills at, unfortunately)!

Hastibe:
> > — should this be more clever, and direct the user to the checkout if there is a related package in the cart?
> 
> I think it's important that the user is able to still view and edit the submission, and that it **doesn't** just take them straight to checkout. It is common on our job board that a user abandons a job post on the "Choose a Package" page or checkout page, and when coming back to continue its submission wants to verify the details of the job post and/or edit it before submitting it.

Another reason that continuing submission of a "Pending Payment" job post should take the user to the job post submission form (and not skip to a later page, like going directly to checkout) is that users will often abandon a job post because they chose the wrong job package for it (or change their mind about what package to use or purchase), and so users will want to be able to choose a different package when continuing submission of the "Pending Payment" job post.

yscik:
Thanks @Hastibe for the additional insight! 

I'm thinking we could add a *Cart* section with packages already in cart to the package selection screen — that way the user could see if they already have a package, and make an explicit choice to use that or add a new one. 

I'm not sure labeling *Pending payment* submissions as *Draft* is good idea, as they'll behave differently (eg activate when the user ends up finishing the checkout).

includes/class-wp-job-manager-shortcodes.php
tripflex: Yes I agree, and IMO it should set the status back to Draft to actually have the listing go through the status transitions when they actually occur.  Reason being the listing should go into `preview` status again, then transition to `pending_payment` (instead of just remaining on `pending_payment` through the entire flow)

tripflex:
I posted a comment on Gino's open issue about this, but IMO we should add a different action than the standard `continue` as this is meant for `preview` or `draft` statuses, not for `pending` or `pending_payment`:
https://github.com/Automattic/WP-Job-Manager/issues/2127#issuecomment-828838264

I think we should add a new action that will actually set the listing back to `draft` status, so the transition of the statuses is followed as they should be `draft` -> `preview` -> `pending_payment` 

Otherwise the listing will always just stay in the `pending_payment` status and will never transition (which could cause issues with other plugins or themes)

Hastibe:
> Thanks @Hastibe for the additional insight!

You're welcome, @yscik!

> I'm thinking we could add a _Cart_ section with packages already in cart to the package selection screen — that way the user could see if they already have a package, and make an explicit choice to use that or add a new one.

Hmm! So, you would have the package selection screen look something like the following below?

**Your Packages:**
- Featured Job Post

**In Cart:**
- Regular Job Post
**Job Listing:** Example Job
- Regular Job Post
**Job Listing:** Some Other Job
- Featured Job Post

**Purchase Package:**
- Regular Job Post
- Featured Job Post

Let's say I had abandoned "Example Job" on the checkout page, but then gone back to the dashboard, clicked "Continued Submission" on it and made edits to it, and now wanted to submit it as a "Regular Job Post"--would I select it from the option under the "In Cart" section (or would this not have the edits I made to it?) or under the "Purchase Package" section, or wouldn't it matter? After making a selection on the "Choose a Package" page and continuing to the checkout page, would the cart be cleared, except for the "In Cart" or "Purchase Package" option that I'd chosen, or would the checkout page still have the three items in my cart (plus any package purchase selection I had made on the "Choose a Package" page)?

> I'm not sure labeling _Pending payment_ submissions as _Draft_ is good idea, as they'll behave differently (eg activate when the user ends up finishing the checkout).

This is a bit of an aside, since I imagine it would be easy for me to re-label job statuses to anything I want via a translate function, but from a user perspective, wouldn't a _Pending Payment_ job and a _Draft_ job behave the same, though? The only place they are going to see the status of them is on their dashboard, where they both aren't submitted yet and have a "Continue Submission" option, which if used to take them to checkout, will activate either of them. 

I guess, if the user had previously taken the _Pending Payment_ job not just to the "Choose a Package" page, but to the checkout page, and then did eventually checkout with that job in the cart, that job will activate (whereas the _Draft_ job wouldn't show-up in the cart, of course), but I still think you could have the _Pending Payment_ job labeled as a _Draft_, and it would be a more understandable status for the user to see, without causing any confusion or other problems.

This is because I think a user would reasonably assume that a "_Draft_" (actually _Pending Payment_) job that is in their cart would activate after they complete checkout for it (I mean, who wouldn't, right?), so I still don't see the existence of the _Pending Payment_ status on the front-end as making any helpful distinction from the _Draft_ status to a front-end user.

PR #2165, opened 2021-06-23
Add Job submission limit

Gnodesign:

* Just like in the Resume Manager, users can set a job submission limit via the settings.
* Adds a "post job" in the job dashboard if the **submission limit** is bigger than the submitted listings.
* Redirects user to the **Job Dashboard** from the **Job Submission** page if the submission limit has been exceeded

### Testing instructions

* Go to "Job Listings -> Settings -> Job Submission" and set a value for the "Listing limit"
* Go to **Submit Job Form Page** and you will be redirect to the dashboard if the limit has been exceeded

The behavior is exactly the same as in the **Resume Manager** plugin.


Gnodesign:
Lots of users have been asking for this feature, it would be great to have this added in the next release. Could someone review this?
@jom @donnapep @renatho @alexsanford

Gnodesign:
Since all these PRs are getting merged maybe we can merge this as well. :) 
@renatho @fjorgemota @onubrooks

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -90,6 +91,56 @@ private function is_job_dashboard_page() {
...
+			 */
+			$redirect_url = apply_filters(
+				'job_manager_redirect_url_exceeded_listing_limit',
+				$redirect_url,
+				$submission_limit,
+				$job_count
+			);
+
+			if ( $redirect_url ) {
+				wp_safe_redirect( esc_url( $redirect_url ) );
```
onubrooks, line 137: Do you think we can add a notice explaining the reason for the redirect?

onubrooks (approved):
Works for me, added a comment below though.

PR #2175, opened 2021-08-05
Release 1.35.2

renatho:
* Update changelog.
* Bump version.
* Bump tested version.
* Update pot file.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -153,6 +153,11 @@ It then creates a database based on the parameters passed to it.
 
 == Changelog ==
 
+= 1.35.2 =
+* Enhancement: Add agreement checkbox to job submission.
+* Fix: Remove unnecessary filter avoiding warning in WP 5.8.
+* Fix: Fix padding on "Apply for job" button.
+
 = 1.35.1 =
 * Change: On new installs, do not enable account registration on job submission page by default.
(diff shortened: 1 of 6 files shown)


readme.txt
```diff
@@ -4,7 +4,7 @@ Tags: job manager, job listing, job board, job management, job lists, job list,
 Requires at least: 5.5
 Tested up to: 5.7
```
donnapep, line 5: Are we good with bumping this to 5.8?

readme.txt
```diff
@@ -4,7 +4,7 @@ Tags: job manager, job listing, job board, job management, job lists, job list,
 Requires at least: 5.5
```
donnapep, line 4: 5.6 if we're bumping to 5.8.

readme.txt
renatho: Good catch! Updated here: https://github.com/Automattic/WP-Job-Manager/pull/2175/commits/5f30812b27c29dcb8a68441a67fd3f70537a04c2

Yep, I tested it and it's working well.

readme.txt
renatho: Also in https://github.com/Automattic/WP-Job-Manager/commit/5f30812b27c29dcb8a68441a67fd3f70537a04c2

donnapep (approved):

PR #2192, opened 2021-10-26
Load WP_Job_Manager_Admin if it wasn't loaded

### LINKED ISSUE #2189: Error logged in cron jobs
**Describe the bug**
We've gotten reports of an error being thrown in the logs, which appears to be happening during a cron job:

```
Uncaught Error: Class 'WP_Job_Manager_Settings' not found in /.../wp-content/plugins/wp-job-manager-resumes/includes/admin/class-wp-resume-manager-settings.php:15
```

A similar error appears if the Applications extension is being used instead of Resumes.

It appears that there are two factors to this issue:

- WPJM core and WPJM extensions load their admin code at different times (during plugin load, and during the `plugins_loaded` action, respectively).
- Some other plugin on the site is causing `is_admin` to return `true` before `plugins_loaded`, but after WPJM core is loaded.

**To Reproduce**
1. On a test site, install WPJM core along with Resumes and/or Applications.
2. Create a new file in `wp-content/plugins/test-cron/test-cron.php` with the following code:

```php
<?php
/**
 * Plugin Name: Test Cron
 * Description: Testing Cron.
 * Version: 1.0.0
 * Author: Automattic
 * Requires at least: 5.0
 * Tested up to: 5.7
 * Requires PHP: 7.0
 *
 * Copyright: 2021 Automattic
 * License: GNU General Public License v3.0
 * License URI: http://www.gnu.org/licenses/gpl-3.0.html
 */

add_action( 'plugins_loaded', function() {
	if ( wp_doing_cron() && ! defined( 'WP_ADMIN' ) ) {
		define( 'WP_ADMIN', true );
	}
}, 1 );
```

3. Activate the "Test Cron" plugin.
4. Install and Activate WP Crontrol and run any cron job.
5. Open the error logs and see the error.

**Expected behavior**

No error to be thrown.

**Workaround**

The following code snippet should remove the problem:

```php
add_action( 'plugins_loaded', function() {
 	if ( is_admin() && defined( 'JOB_MANAGER_PLUGIN_DIR' ) && ! class_exists( 'WP_Job_Manager_Admin' ) ) {
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/admin/class-wp-job-manager-admin.php';
 	}
} );
```

gkaragia:
Fixes #2189

### Overview
There were cases reported that `WP_Job_Manager_Settings` was not included although `is_admin()` was returning true. This could happen in cases where `is_admin` started to return true (probably by another plugin) after Job Manager plugin completed loading.

* Adds a hook which includes the admin PHP file if `is_admin` returns true and the file wasn't included.
* The test failures are unrelated to this PR and will be handled in a future task.

### Testing instructions

* Follow the intructions in #2189 and observe that the error isn't happening anymore.

### DIFF
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -578,7 +578,7 @@ class WP_Job_Manager {
 	 * load. See the below issue for more information: https://github.com/Automattic/evergreen/issues/136
 	 */
 	public function include_admin_files() {
-		if ( is_admin() && ! defined( 'WP_Job_Manager_Admin' ) ) {
+		if ( is_admin() && ! class_exists( 'WP_Job_Manager_Admin' ) ) {
 			include_once JOB_MANAGER_PLUGIN_DIR . '/includes/admin/class-wp-job-manager-admin.php';
 		}
 	}
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -92,6 +92,7 @@ class WP_Job_Manager {
 		add_action( 'wp_logout', [ $this, 'cleanup_job_posting_cookies' ] );
 		add_action( 'init', [ 'WP_Job_Manager_Email_Notifications', 'init' ] );
 		add_action( 'rest_api_init', [ $this, 'rest_init' ] );
+		add_action( 'plugins_loaded', [ $this, 'include_admin_files' ] );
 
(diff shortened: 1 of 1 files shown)


renatho (approved):
Besides the unrelated issues in GH actions, it looks good and works well! 🥳

PR #2193, opened 2021-10-31
Use wp_kses_post to process a job title instead of esc_html

### LINKED ISSUE #1169: Span tag For Applied Job are Not Processed and Undefined Variable
1) If A user created and he has self applied then HTML TAGS Must be Processed at at Jobs,Job Dashboard Page
2) There should not be any Undefined Variable error 
#### What Happened Instead
1) If A user created and he has self applied then HTML TAGS WONT Process at all at Job Dashboard Page
2) Undefined variable: application_ids in wp-content\plugins\wp-job-manager-applications\includes\class-wp-job-manager-applications-post-types.php on line 190
Enable WP_Debug and See there is Undefined variable error at When Employeer reviews and Add comment after that it gives this error.
-------------------------------------------------------------------
**HTML Tags Not Processing**
1)create job and apply for same
2) Go to Jobs and Jobs Dashboard Page and see there is  `<span class="job-manager-applications-applied-notice">Applied</span>` is not processed
------------------------------------------------------------------
**Undefined Variable Error**
Enable WP Debug
1)Goto Job dashboard page and Click on number of applications.
2) Then give user review and change its to Some thing Interviewed or else
3)There will be undefined variable error there.

Alt text

merkushin:
Fixes #1169 & https://github.com/Automattic/wp-job-manager-applications/issues/281

There might be added some HTML to the title in filter hooks.
To be more consistent and output the title similarly, use wp_kses_post.

* Use `wp_kses_post` instead of `html_esc` to process `wpjm_get_the_job_title`

### Testing instructions

* All steps must be done with the same user.
* Install WP Job Manager Applications
* Create a job.
* Apply for the job.
* Go to Jobs and check if there is no HTML on the page (`<span class="job-manager-applications-applied-notice">Applied</span>`).
* Go to Jobs Dashboard and click "Mark filled", check there is no HTML in the notification message.

### Screenshot / Video
Screenshot 2021-10-31 at 21 17 52

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
     "bjornjohansen/wp-pre-commit-hook": "0.3.0",
     "squizlabs/php_codesniffer": "3.5.6",
     "wp-coding-standards/wpcs": "2.3.0",
-    "sirbrillig/phpcs-variable-analysis": "^2.6"
+    "sirbrillig/phpcs-variable-analysis": "^2.6",
+    "yoast/phpunit-polyfills": "^1.0"
   },
   "archive": {
     "exclude": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c21646ece7a33b04c827cdf035465988",
+    "content-hash": "f0d1614c50f85476e9a126fe46ec1850",
     "packages": [],
     "packages-dev": [
         {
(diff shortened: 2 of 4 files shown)


renatho (commented):
Looks good and works well!

I just noticed an issue with the tests. A reference of a commit in Sensei where @gkaragia fixed it: https://github.com/Automattic/sensei/pull/4315/commits/1f22beef1f4e59c11d3bd18f1c152462671e7be6

> Go to Jobs and check if there is no HTML on the page (<span class="job-manager-applications-applied-notice">Applied</span>).

I couldn't reproduce it in `master` branch. Should it happen in `master`?

merkushin:
> I couldn't reproduce it in `master` branch. Should it happen in `master`?

Yes, with all steps done it should be possible to reproduce in `master`.

renatho (approved):
Looks good and works well! 🥳

About the issue, it was my mistake. That one I could reproduce, and I thought there was another place where it had only the "Applied".

PR #2200, opened 2021-12-14
fix not translatable string

### LINKED ISSUE #1950: String id=381 in .pot file doesn't translate
**Describe the bug**
The string with id=381 in .pot file doesn't translate to another language.

**To Reproduce**
1. Create .po from .pot
2. Translate string with id=381 "If you don't have an account you can %screate one below by entering your email address/username." and all the others needed
3. Compile to .mo and upload in the correct folder
4. Go to the "add new listing" page as not logged user
5. All the others string, which have references in this specific .mo file, get correctly translated in the front end, except the one with id=381, which stay in english.

**Screenshots**
Screenshot 2019-12-18 at 15 01 02
Screenshot 2019-12-18 at 15 03 06

**Expected behavior**
The string should be translated.
The bug is probably caused from the "%s" in the original string which I suppose not necessary 

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [ ] This bug happens with a default WordPress theme active.

**WordPress Environment**
- WordPress Version: 5.3.1
- WP Job Manager Version: 1.34.1
- PHP Version: 7.2.22
- Other important details:

masteradhoc:
Fixes #1950

* fix not translatable string as noted in #1950

### Testing instructions
* see  #1950

### DIFF
--- a/languages/wp-job-manager.pot
+++ b/languages/wp-job-manager.pot
@@ -1893,9 +1893,8 @@ msgstr ""
 msgid "Sign in"
 msgstr ""
 
-#. translators: Placeholder %s is the optionally text.
 #: templates/account-signin.php:47
-msgid "If you don't have an account you can %screate one below by entering your email address/username."
+msgid "If you don't have an account you can create one below by entering your email address/username."
 msgstr ""
 
 #. translators: Placeholder %s is the optionally text.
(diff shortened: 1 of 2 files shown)


aaronfc:
👋 @masteradhoc, thanks for working on this.

I could not reproduce the issue, but the '%s' is for sure a mistake and I can imagine that maybe the compiler from `.po` to `.mo` that you are using might be failing because of that.

Merging your fix!

PR #2212, opened 2022-02-22
Fix dependencies (npm/composer) problems

fjorgemota:

* Update NPM Lock file to v2;
* Remove/update some old dependencies that limited the project to be installed only with Composer 1;
* Add and configure Husky as a replacement for `bjornjohansen/wp-pre-commit-hook` (which was removed by this PR as it's not updated anymore and doesn't support Composer v2);
* Also update NPM minimum version to 7.0;

### Testing instructions

* Download the repository in some new folder;
* Run `composer install` using Composer v2 or newer;
* Run `npm install` using NPM v7 or newer;
* If possible, change one of the PHP files without respecting any style guides and try to commit it to check if the pre-commit hook correctly catches the change;

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
--- a/composer.json
+++ b/composer.json
@@ -6,9 +6,8 @@
   "license"    : "GPL-3.0+",
   "require-dev"    : {
     "phpunit/phpunit": "^7.0",
-    "dealerdirect/phpcodesniffer-composer-installer": "0.4.4",
+    "dealerdirect/phpcodesniffer-composer-installer": "0.7.1",
     "phpcompatibility/phpcompatibility-wp": "2.1.0",
-    "bjornjohansen/wp-pre-commit-hook": "0.3.0",
     "squizlabs/php_codesniffer": "3.5.6",
     "wp-coding-standards/wpcs": "2.3.0",
     "sirbrillig/phpcs-variable-analysis": "^2.6",
@@ -38,5 +37,10 @@
       ".*",
       "*.test.js"
     ]
+  },
+  "config": {
+    "allow-plugins": {
+      "dealerdirect/phpcodesniffer-composer-installer": true
+    }
   }
 }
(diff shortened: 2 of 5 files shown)


dadish (commented):
Thanks for all these tooling updates @fjorgemota !

This one looks all good. One thing weird that happens when I do `npm install` is that removes the `"name": "wp-job-manager",` from `package-lock.json` file.

Screen Shot 2022-02-23 at 12 14 06

<details>
  <summary>Raw Git Diff</summary>
  <pre>
diff --git a/package-lock.json b/package-lock.json
index 902f8eda..5303c927 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "wp-job-manager",
       "version": "1.35.2",
       "license": "GPL-2.0-or-later",
       "dependencies": {
  </pre>
</details>

Does it do the same thing for you? My npm version is 7.7.6 and node version is 15.14.0

fjorgemota:
No problem, @dadish! 

About the issue you found...that's...quite weird.

My Node's system version is v16.14.0 and npm's system version is v8.3.1. And there [that issue doesn't happen](https://asciinema.org/a/LQThhUMuKcgBjBQgbEpLmyz1D). Testing here locally using nvm to have the same versions as yourself, [I also couldn't reproduce the issue](https://asciinema.org/a/zKVxvg95ccwYkFP9cqQUnBKHQ).

I'm not sure how I could helphere, but I find those differences in `package-lock.json` quite...interesting. :thinking:

dadish:
> My Node's system version is v16.14.0 and npm's system version is v8.3.1. And there [that issue doesn't happen](https://asciinema.org/a/LQThhUMuKcgBjBQgbEpLmyz1D). Testing here locally using nvm to have the same versions as yourself, [I also couldn't reproduce the issue](https://asciinema.org/a/zKVxvg95ccwYkFP9cqQUnBKHQ).

Got it. It might be something happening on my machine only. Nvm.

dadish (approved):

renatho:
> > My Node's system version is v16.14.0 and npm's system version is v8.3.1. And there [that issue doesn't happen](https://asciinema.org/a/LQThhUMuKcgBjBQgbEpLmyz1D). Testing here locally using nvm to have the same versions as yourself, [I also couldn't reproduce the issue](https://asciinema.org/a/zKVxvg95ccwYkFP9cqQUnBKHQ).
> 
> Got it. It might be something happening on my machine only. Nvm.

I also tested it, and it worked well for me with npm 8.1.2. Maybe it's a difference in npm 7? 🤷‍♂️

PR #2216, opened 2022-02-24
Disable transient cache when the order is set to random or random featured

### LINKED ISSUE #1967: Featured Jobs widget: sort by random behaves unexpectedly
**Describe the bug**

In the settings for the Featured Jobs widget, if you choose sort by: random, it doesn't seem to display a random selection of featured jobs. The same jobs are displayed on refresh, and it doesn't make sense to have a 'sort direction' field for random.

**To Reproduce**
1. Create several jobs and make them all featured.
2. Add the Featured Jobs widget to any widget area and set sort by: random.
3. View a page containing the widget.
4. Results are inconsistent, but refreshing the page does not change the jobs shown.
5. Changing the sort direction displays a different selection of jobs.

**Expected behavior**
When sort by is set to random, a random selection from all featured jobs should be displayed, and that should change on refresh. Not sure if the 'sort direction' field makes sense when sort by is random, maybe it should be disabled.

**WordPress Environment**
- WordPress Version: 5.3.2
- WP Job Manager Version: 1.34.1
- PHP Version: 7.2.26

Reported in 2655100-zen

fjorgemota:
This guarantees that the set of jobs displayed is always random. The bad aspect is, of course, that it will require more requests to the database just to get the list of jobs, but the alternative here would be to get a big pool of jobs from the database and limit that on the PHP side, but guess what? That's not very efficient either, nor allow it to show a random set of the full pool of jobs registered.

As users on #1967 were already disabling cache for this use case, I think it's worth to just do the same here, but only when ordering randomly.

By the way, this also disables the shuffling of cached results, as those are not required anymore because the result set will always be randomly ordered by the database itself.

Fixes #1967

* Disable transient cache on the function `get_job_listings` when the `orderby` used is `rand` or `rand_featured`;

### Testing instructions

1. Checkout this branch;
2.  Create several jobs and make them all featured.
3.  Add the Featured Jobs widget to any widget area and set sort by random.
4. View a page containing the widget.
5. Verify that each page refresh shows a totally different set of the registered jobs;

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -162,13 +162,14 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 
 		do_action( 'before_get_job_listings', $query_args, $args );
 
+		$should_cache = 'rand_featured' !== $args['orderby'] && 'rand' !== $args['orderby'];
+
 		// Cache results.
-		if ( apply_filters( 'get_job_listings_cache_results', true ) ) {
-			$to_hash              = wp_json_encode( $query_args );
-			$query_args_hash      = 'jm_' . md5( $to_hash . JOB_MANAGER_VERSION ) . WP_Job_Manager_Cache_Helper::get_transient_version( 'get_job_listings' );
-			$result               = false;
-			$cached_query_results = true;
-			$cached_query_posts   = get_transient( $query_args_hash );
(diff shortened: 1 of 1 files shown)


aaronfc (approved):
Looks good and works well!

Tested also the shortcode `[jobs orderby=rand]` 👍

PR #2217, opened 2022-02-24
Fix button "Apply for job" when the page is translated using Google Translate

### LINKED ISSUE #1995: Google Translate extension on Chrome breaks the Apply button Javascript
Originally reported on the WPJM Applications extension, but this is a general issue in Core:

---

#### Steps to Reproduce
1. Go to a job listing
2. Translate the page to another language using Google Translate extension on Chrome
3. Click on the Apply button

#### What I Expected
Application form is opened

#### What Happened Instead
Nothing happens.

#### Browser / OS version
Chrome 80
Google Translate 2.0.9

#### Context / Source
2845441-zen

I acknowledge that is might be out of our scope. Checking if anything we can do from the code to prevent Google Translate breaking the button.

---

Looks like this is happening because of [this jQuery selector](https://github.com/Automattic/WP-Job-Manager/blob/d3a2765520494759d961a6622482515d82f214ac/assets/js/job-application.js#L8). The Google Translate extension adds some of its own wrappers so that the two elements are no longer siblings.

Note that we should not simply change this to a global selector as there may be multiple job listings shown on a single page, and we would not want to expand the Application section for all of them.

fjorgemota:
Fixes #1995

* Changes the logic to get the application details block to one that is compatible with the Google Translate extension, which changes the HTML structure of the page, by looking first on the parent job_application and then looking out for application_details.

### Testing instructions

1. Go to a job listing
2. Translate the page to another language using [Google Translate extension on Chrome](https://chrome.google.com/webstore/detail/aapbdbdomjkkjkaonfhkkikfgjllcleb)
3. Click on the Apply button
4. Check if the application details are shown;

### Screenshot / Video

https://user-images.githubusercontent.com/529864/155597739-a342121b-9e3b-4c6b-8fde-7ffb80bb94ed.mp4

### DIFF
--- a/assets/js/job-application.js
+++ b/assets/js/job-application.js
@@ -5,7 +5,7 @@ jQuery(document).ready(function($) {
 	}
 
 	$( document.body ).on( 'click', '.job_application .application_button', function() {
-		var $details = $(this).siblings('.application_details').first();
+		var $details = $(this).parents('.job_application').find('.application_details').first();
 		var $button = $(this);
 		$details.slideToggle( 400, function() {
 			if ( ! $(this).is(':visible') ) {


dadish (approved):
Works great! :shipit:

PR #2221, opened 2022-03-03
Add workflows to provide plugin building on WP Job Manager

fjorgemota:

* Add workflow similar to the ones added in sensei repository (https://github.com/Automattic/sensei/pull/4884 and https://github.com/Automattic/sensei/pull/4886), but in just one PR and changing references from Sensei core to WP Job Manager;

### Testing instructions

* Check if a "Plugin Build" check appears on this PR, click on "Details", then "Summary" and verify if the artifact "wp-job-manager-4c18a1d8b4483a846958a65806219e8b37f248b4" appears on the Artifacts list, and if it's in fact a build of this plugin.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,49 @@
+name: Plugin Build
+
+on: pull_request
+
+jobs:
+    build:
+        name: Plugin Build
+        runs-on: ubuntu-18.04
+        steps:
+            - uses: actions/checkout@v2
+            - name: Get npm cache directory
+              id: npm-cache
+              run: |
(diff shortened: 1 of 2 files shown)


renatho (approved):
Looks good and works well! 🤖

I just didn't understand why "Plugin Build Download" didn't appear too. Is there something else we need to add in order to make it work?

fjorgemota:
> I just didn't understand why "Plugin Build Download" didn't appear too. Is there something else we need to add in order to make it work?

That's expected. As I documented on #4884, the `report-build.yml` workflow only appear and run when it's on the main branch of the repository. So after merging this PR any new PRs opened should have the "Plugin Build Download" working as expected. :smiley:

PR #2230, opened 2022-04-08
Fix issue where input_select didn't detect selected value properly

fjorgemota:

* Fix issue where `input_select` didn't detect the selected value properly, causing the wrong value to be shown on the WP-Admin interface;

### Testing instructions

1. Install https://wordpress.org/plugins/custom-field-for-wp-job-manager/
2. Create a new Select field
3. Add a few options ( make sure some options have spaces and some do not eg "Legal", "Finance & Accounting")
4. Create a new job listing and select Legal -> publish -> all okay
5. Edit the job and select Finance & Accounting -> publish -> all okay
6. Open the same editor in a new tab, and verify if the first option from the select field is not shown as active even if you selected any other option when editing the job previously;

### DIFF
--- a/includes/admin/class-wp-job-manager-writepanels.php
+++ b/includes/admin/class-wp-job-manager-writepanels.php
@@ -415,6 +415,10 @@ class WP_Job_Manager_Writepanels {
 		} else {
 			$name = $key;
 		}
+		$selected_value = null;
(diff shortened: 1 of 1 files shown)


alexsanford (approved):
Nice! Works well 👍

PR #2258, opened 2022-06-07
New hook to filter empty fields

### LINKED ISSUE #1480: Form validation, consider zero (0) as a value for required fields
Consider the example, you create a field `job_salary` and configure it in PHP as a **required** field, to require user to enter a value.

If you wanted the user to be able to enter a value of 0 (zero), the form validation returns an error, because PHP's `empty` considers 0 (zero) and empty value.

http://php.net/manual/en/function.empty.php

https://github.com/Automattic/WP-Job-Manager/blob/2a7e33075ce935d6ff7719df3437a2c28f0c0c7e/includes/forms/class-wp-job-manager-form-submit-job.php#L305

My thought was to maybe create a helper function/method to replace using `empty` and maybe pass through a filter of "false" values to check against?

```php
protected function is_empty( $value ){
    //  @see http://php.net/manual/en/function.empty.php -- standard default empty values
    $false_vals = apply_filters( 'submit_job_form_validate_fields_empty_values', array( '', 0, 0.0, '0', null, false, array() );

    // strict true for type checking also
    if( in_array( $value, $false_vals, true ) ){
        return true;
    }

    return false;
}
```

Then if end user really wants to require a value but still consider 0 (zero) a true value, it's as easy as just using that filter to remove it from the array ... thoughts?

onubrooks:
Fixes #1480 

* Added a new hook to pass values that should be considered empty.
* Added `is_empty` method that checks empty fields, to replace PHPs `empty` function.

### Testing instructions

* Follow the instructions [here](https://wpjobmanager.com/document/tutorial-adding-a-salary-field-for-jobs/) to add salary field that is required.
* Try posting a job, leaving the salary field empty or filling it with zero (0).
* Notice how the form validation error displays to show salary is required.
* Use any snippet plugin like [this](https://wordpress.org/plugins/code-snippets/) to add the following snippet:
```
add_filter( 'submit_job_form_validate_fields_empty_values', 'get_null_values' );

function get_null_values( $values ) {
	return array( null, false, [] );
}
```

* Now try again filling zero in the salary field. Make sure the job listing submits.

### New/Updated Hooks

* `submit_job_form_validate_fields_empty_values`: we have a default list of empty values to check for during form validation (`[ '', 0, 0.0, '0', null, false, [] ]`), this filter can modify that list to return what we want to consider as empty values.

### Known Issues

- I noticed that when '0' is filled in the salary field, the form submits an empty string instead of the number 0. To test that is works, the empty string '', has to be excluded from the array returned in the `submit_job_form_validate_fields_empty_values` filter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ sftp-config.json
 /build/
 /assets/dist/
 /*.zip
+.idea
 
 # Ignore lib/wpjm_rest but keep folder
 lib/wpjm_rest/*
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -376,7 +376,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 			foreach ( $group_fields as $key => $field ) {
 				if (
 					$field['required']
-					&& empty( $values[ $group_key ][ $key ] )
+					&& $this->is_empty( $values[ $group_key ][ $key ] )
 					&& ( ! isset( $field['empty'] ) || $field['empty'] )
 				) {
 					// translators: Placeholder %s is the label for the required field.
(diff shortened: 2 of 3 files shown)


dadish (approved):
Works great! 🔥

PR #2260, opened 2022-06-09
Job Visibility Settings

Gnodesign:

Visibility settings for browsing job listings and viewing single job listing page. This works exactly as the WPJM Resumes visibility feature does.

### Testing instructions

* Go to "WP Dashboard -> Job Listings -> Settings -> Job Visibility"
* Add the roles to the settings
* Visit [job] page or single job page

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -397,7 +397,7 @@ class WP_Job_Manager_Settings {
 							'label' => __( 'Browse Job Capability', 'wp-job-manager' ),
 							'type'  => 'capabilities',
 							// translators: Placeholder %s is the url to the WordPress core documentation for capabilities and roles.
-							'desc'  => sprintf( __( 'Enter which <a href="%s">roles or capabilities</a> allow visitors to browse resumes. If no value is selected, everyone (including logged out guests) will be able to browse resumes..', 'wp-job-manager' ), 'http://codex.wordpress.org/Roles_and_Capabilities' ),
+							'desc'  => sprintf( __( 'Enter which <a href="%s">roles or capabilities</a> allow visitors to browse job listings. If no value is selected, everyone (including logged out guests) will be able to browse job listings.', 'wp-job-manager' ), 'http://codex.wordpress.org/Roles_and_Capabilities' ),
 						],
 						[
 							'name'  => 'job_manager_view_job_capability',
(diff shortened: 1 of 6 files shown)


onubrooks (approved):
Works as advertised! 
One small thing: is it possible to add an option to reset it back to default where no capability is set at all? I imagine people may want to tinker around those options and later still prefer to have it the default way.
I'm happy to merge it either way.

Gnodesign:
> Works as advertised! One small thing: is it possible to add an option to reset it back to default where no capability is set at all? I imagine people may want to tinker around those options and later still prefer to have it the default way. I'm happy to merge it either way.

I could add a reset button but I believe this would only bloat the plugins since it would require JS to make it work as well a new **input_button** in the `class-wp-job-manager-settings.php`, the RM doesn't have a reset feature either. 

Lets merge it as it is for now and maybe in the future we can handle both cases via WPJM to avoid any code duplication.

alexsanford:
FYI this PR was reverted in version `1.36.2`. [See this PR.
](https://github.com/Automattic/WP-Job-Manager/pull/2277)

Unfortunately, the new feature was causing breakage on several sites. Users were finding that their users could no longer view the Jobs page without being logged in, and it was not obvious how to fix the issue.

It appears that there were two problems at play:

First, the default value for the new setting was set to `administrator,employer,candidate`. While this is fine for the similar feature in Resume Manager, this was a non-backwards compatible change in this PR, as it changed the default behaviour of core significantly.

Second, the UI for this feature was a standard multi-select box, which is not very intuitive to use, especially if you are trying to unselect all of the options in order to restore the default behaviour. The UI in Resume Manager is much more intuitive.

We should definitely take another look at this and ship the feature in a new release. We greatly appreciate the contribution! In order to get this out the door, I think we'll need to make a few changes:

1. Ensure the defaults do not change the behaviour of existing sites. In this case, I believe it means that we should default this setting to _empty_, so that users who upgrade the plugin do not remove anyone's ability to see job listings.
2. NOTE: since some users may have upgraded to 1.36.1, the new options may have been set in the DB on some sites! We should use a different option name when we re-implement this, so that those users do not get the incorrect defaults.
3. We should implement the UI to match what is in Resume Manager, so it is more intuitive to users how to manage this feature.

Gnodesign:
Thanks a lot for the explanation @alexsanford. 
For me it was showing the same UI as the RM but it seems that the multiselect inputs were getting initialized due to having the RM plugin activated, should have run some tests without it. I will fix all these points and push again.

PR #2261, opened 2022-06-13
Add support for multiple location search.

### LINKED ISSUE #1217: Feature: Multiple locations in search
Can we either have the ability to add multiple locations to search queries, or apply some sort of fuzzy logic or search radius?

What I mean by this is that it would be great to be able to say to return all results within a certain radius of the primary job location... or to be able to list "city 1, city 2, city 3" in job search queries so that we can build better landing pages with pre-defined search results.

This could then also be used on the job search front end to search by radius or multiple locations.

dadish:
Fixes #1217 

* Adds support for multiple location search. We simply take the `search_location` argument and split it with `[,\s]+` regex. This allows to insert location names with one or more comma and/or space between them.

### Testing instructions

* Make sure you have job posts with different location names.
* Enter a single location name and search. Confirm the results are correct.
* Enter two or more location names separating them with one or more commas and/or spaces and search. Confirm the results are correct.

### Screenshot / Video

https://user-images.githubusercontent.com/2578542/173345120-e088940c-c3a9-4a17-bc56-c15c9238c669.mp4

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -77,11 +77,14 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
 			$location_search    = [ 'relation' => 'OR' ];
 			foreach ( $location_meta_keys as $meta_key ) {
-				$location_search[] = [
-					'key'     => $meta_key,
-					'value'   => $args['search_location'],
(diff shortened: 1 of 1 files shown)


onubrooks (approved):
🔥

tripflex:
While this has been modified since for remote positions, this has now created other issues.

If a user searches for "Atlanta, GA" .. this is now going to return results from GA regardless of whether they are in "Atlanta" or not.  No good.

PR #2263, opened 2022-06-15
Add salary field

### LINKED ISSUE #1320: Add salary field to satisfy Google's Job Search schema

onubrooks:
Fixes #1320 
Builds upon #1510 

* Adds a salary field to the frontend and admin forms.
* Adds salary information to Google's Job Search Structured Data
* Adds a setting to enable salary field. This is useful to not break sites using snippets or other plugins to add salary field.

### Testing instructions

* Copy the output HTML and paste here - [https://search.google.com/structured-data/testing-tool/u/0/](https://search.google.com/structured-data/testing-tool/u/0/)

### New/Updated Hooks

* `wpjm_job_salary_currency`: This filter provides a default currency, `USD`.
* `wpjm_job_salary_unit`: This filter provides a default unit, `YEAR`, which can be adjusted to `MONTH`, `WEEK` etc.
* `the_job_salary`: Filters the returned job salary.

### Improvements
Base on https://github.com/Automattic/WP-Job-Manager/issues/1320#issuecomment-537865339, here are a few possible enhancements:

- Setting for per job currencies with a global default. Base our function(s) off of get_woocommerce_currencies and get_woocommerce_currency_symbol.
- Interval will also need to be supported / tracked (HOUR, DAY, WEEK, MONTH, YEAR). For the default, we could probably use YEAR with a filter to adjust.
- Ideally, ranges should also be supported from the start. We could do some magic parsing of a string (20000-40000) -> 'minValue' => 20000, 'maxValue' => 40000 or two fields. Two fields would probably be best.
- UI: input currency, salary value/range (seamless way to go between a single salary and range), Interval unit (hour, day, week, month, year)

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -691,7 +691,7 @@ class WP_Job_Manager_Post_Types {
 		$location  = get_the_job_location( $post_id );
 		$company   = get_the_company_name( $post_id );
 		$job_types = wpjm_get_the_job_types( $post_id );
-		$salary = get_the_job_salary( $post_id );
+		$salary    = get_the_job_salary( $post_id );
 
 		if ( $location ) {
 			echo '<job_listing:location><![CDATA[' . esc_html( $location ) . "]]></job_listing:location>\n";
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -691,6 +691,7 @@ class WP_Job_Manager_Post_Types {
 		$location  = get_the_job_location( $post_id );
 		$company   = get_the_company_name( $post_id );
 		$job_types = wpjm_get_the_job_types( $post_id );
+		$salary = get_the_job_salary( $post_id );
 
 		if ( $location ) {
 			echo '<job_listing:location><![CDATA[' . esc_html( $location ) . "]]></job_listing:location>\n";
@@ -701,6 +702,9 @@ class WP_Job_Manager_Post_Types {
 		}
 		if ( $company ) {
 			echo '<job_listing:company><![CDATA[' . esc_html( $company ) . "]]></job_listing:company>\n";
+		}
+		if ( $salary ) {
+			echo '<job_listing:salary><![CDATA[' . esc_html( $salary ) . "]]></job_listing:salary>\n";
 		}
 
 		/**
--- a/assets/css/frontend.scss
+++ b/assets/css/frontend.scss
@@ -411,7 +411,7 @@ div.job_listings {
 		}
 
 		.salary:before {
-			.display-icon;
+			@include display-icon();
 			content: '\e800';
 		}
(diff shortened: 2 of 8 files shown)


includes/class-wp-job-manager-post-types.php
```diff
@@ -702,6 +703,9 @@ public function job_feed_item() {
 		if ( $company ) {
 			echo '<job_listing:company><![CDATA[' . esc_html( $company ) . "]]></job_listing:company>\n";
 		}
+		if ( $salary ) {
+			echo '<job_listing:salary><![CDATA[' . esc_html( $salary ) . "]]></job_listing:salary>\n";
+		}
```
dadish, line 708: Couldn't understand what's this used for.

dadish (approved):
Works great! 🔥

includes/class-wp-job-manager-post-types.php
onubrooks: Not sure either but I think it has something to do with RSS feeds

tripflex:
https://wordpress.org/support/topic/new-update-adds-salary-field-but-how-to-get-data-from-old-salary-field/

As i'm sure others will point out, this probably was not the best idea to output the rate and currency as some people use this field to define the salary in other means, and there's no easy way of disabling or removing this.

StefMattana:
There seems to be a conflict with a third-party plugin too (WPJM Extra Fields), as the core salary filed keeps being displayed despite the option being toggled off: https://wordpress.org/support/topic/conflict-with-alternate-plugin-for-showing-salary/#post-15764240

I asked the user to report this to the plugin author, but I believe that's worth flagging it here too.

StefMattana:
Upon testing 1.36.2 I found out that nothing gets displayed in my test site's job listings upon enabling the salary options. 

I filled the salary field both from the backend and the frontend, and also by adding simple numbers (es 1000) and numbers with specifications (1000/hours; 1000/year). Still, nothing got published. 

I am running JobifyVersion: 3.11.1 if that helps.

StefMattana:
We already have a request [from the forum](https://wordpress.org/support/topic/all-salaries-are-now-showing-usd-year-after-salary-how-to-remove/#post-15763297) for implementing the `yearly/monthly/daily/hourly` toggle upon adding salaries via this new feature, which ideally could be customisable for each job entry

>we have some hourly, daily or monthly pay rates and these will vary from job to job, rather than being consistent across our platform for all jobs.

Issue #2274, opened 2022-06-21
Job visibility can't deselect from drop-down


cestrian77:
Having just updated to version 1.36.0 I now can't deselect all options in the Job Visibility settings to enable non-logged users to view job listings.

cestrian77:
Ignore me. Figured out holding Command and clicking allows to deselect... but not abundantly clear.

PR #2280, opened 2022-06-22
Job Visibility feature

Gnodesign:
*The feature has been reworked due to users facing various issues after updating to version 1.36.0.*

Visibility settings for browsing job listings and viewing single job listing page. This works exactly as the WPJM Resumes visibility feature does.

### Testing instructions

* Go to "WP Dashboard -> Job Listings -> Settings -> Job Visibility"
* Add the roles to the settings
* Visit [job] page or single job page

### Fixes
https://github.com/Automattic/WP-Job-Manager/pull/2260 https://github.com/Automattic/WP-Job-Manager/issues/2274

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -51,6 +51,7 @@ class WP_Job_Manager_Settings {
 	public function __construct() {
 		$this->settings_group = 'job_manager';
 		add_action( 'admin_init', [ $this, 'register_settings' ] );
+		add_action( 'admin_action_update', [ $this, 'pre_process_settings_save' ] );
 	}
 
 	/**
@@ -429,6 +430,27 @@ class WP_Job_Manager_Settings {
 							'label' => __( 'Terms and Conditions Page', 'wp-job-manager' ),
 							'desc'  => __( 'Select the page to link when "Terms and Conditions Checkbox" is enabled. See setting in "Job Submission" tab.', 'wp-job-manager' ),
 							'type'  => 'page',
+						],
+					],
+				],
+				'job_visibility' => [
+					__( 'Job Visibility', 'wp-job-manager' ),
+					[
+						[
+							'name'  => 'job_manager_browse_job_listings_capability',
+							'std'   => '',
+							'label' => __( 'Browse Job Capability', 'wp-job-manager' ),
+							'type'  => 'capabilities',
+							// translators: Placeholder %s is the url to the WordPress core documentation for capabilities and roles.
+							'desc'  => sprintf( __( 'Enter which <a href="%s">roles or capabilities</a> allow visitors to browse job listings. If no value is selected, everyone (including logged out guests) will be able to browse job listings.', 'wp-job-manager' ), 'http://codex.wordpress.org/Roles_and_Capabilities' ),
+						],
+						[
+							'name'  => 'job_manager_view_job_listing_capability',
+							'std'   => '',
+							'label' => __( 'View Job Capability', 'wp-job-manager' ),
+							'type'  => 'capabilities',
+							// translators: Placeholder %s is the url to the WordPress core documentation for capabilities and roles.
+							'desc'  => sprintf( __( 'Enter which <a href="%s">roles or capabilities</a> allow visitors to view a single job listing. If no value is selected, everyone (including logged out guests) will be able to view job listings.', 'wp-job-manager' ), 'http://codex.wordpress.org/Roles_and_Capabilities' ),
 						],
 					],
 				],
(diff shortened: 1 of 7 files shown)


Gnodesign:
@alexsanford So I reworked the PR and these are the changes applied.

1) Default values are now set to **empty**. This will ensure that all users (logged in & non logged in) will be able to view and browse job listings unless the user changes the values. By setting it to empty users that are updating the plugin won't face any issues with listing visibilities.

2) Option names have been changed to avoid getting the previously saved defaults.

- `job_manager_browse_job_capability` has been changed to `job_manager_browse_job_listings_capability`.

- `job_manager_view_job_capability` has been changed to `job_manager_view_job_listing_capability`.

3) Job Visibility select inputs use the same UI as in the Resume Manager now.

Should be fine now but please let me know in case you catch something.

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -934,4 +956,131 @@ protected function input_multi( $option, $ignored_attributes, $values, $ignored_
...
+			'job_manager_browse_job_listings_capability',
+			'job_manager_view_job_listing_capability',
+		];
+		foreach ( $capabilities_fields as $capabilities_field ) {
+			// phpcs:disable WordPress.Security.NonceVerification.Missing -- Settings save will handle the nonce check.
+			if ( isset( $_POST[ $capabilities_field ] ) ) {
+				// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Sanitized by `WP_Resume_Manager_Settings::capabilities_array_to_string()`
+				$input_capabilities_field_value = wp_unslash( $_POST[ $capabilities_field ] );
+				if ( is_array( $input_capabilities_field_value ) ) {
+					$_POST[ $capabilities_field ] = self::capabilities_array_to_string( $input_capabilities_field_value );
```
fjorgemota, line 1022: I think that, instead of modifying `$_POST` here, we could rely on using the filter `sanitize_option_{$option}`, [documented here](https://developer.wordpress.org/reference/hooks/sanitize_option_option/), which should also achieve the same effect without messing with `$_POST`.

Also, I think you can save an array directly into the database, too, because options just need to be serializable (and an array of strings is serializable), so you could avoid things [like this](https://github.com/Automattic/WP-Job-Manager/pull/2280/files#diff-f2e92c961523255d469f2ea4ce5a6a5772b739071ce9b4f2e93e1f946636b882R1613) too.

WDYT?

templates/content-single-job_listing.php
```diff
@@ -9,42 +9,48 @@
  * @package     wp-job-manager
  * @category    Template
  * @since       1.0.0
- * @version     1.28.0
+ * @version     1.36.0
```
fjorgemota, line 12: I think we can update this version here to 1.37.0? 

Good catch, by the way. :)

wp-job-manager-functions.php
```diff
@@ -1602,3 +1602,72 @@ function job_manager_count_user_job_listings( $user_id = 0 ) {
...
+		$can_browse = false;
+		foreach ( $caps as $cap ) {
+			if ( current_user_can( $cap ) ) {
+				$can_browse = true;
+				break;
+			}
+		}
+	}
+
+	return apply_filters( 'job_manager_user_can_browse_job_listings', $can_browse );
```
fjorgemota, line 1625: Could you add a doc block here similar [to this one](https://github.com/Automattic/WP-Job-Manager/pull/2285/files#diff-5cab4b0c50d438a5a481a986095b8efac2312497af4113d47715ca479416e954R1326-R1338), for documentation purposes?

wp-job-manager-functions.php
```diff
@@ -1602,3 +1602,72 @@ function job_manager_count_user_job_listings( $user_id = 0 ) {
...
+	}
+
+	$key = get_post_meta( $job_id, 'share_link_key', true );
+
+	// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+	if ( $key && ! empty( $_GET['key'] ) && $key === $_GET['key'] ) {
+		$can_view = true;
+	}
+
+	return apply_filters( 'job_manager_user_can_view_job', $can_view, $job_id );
```
fjorgemota, line 1672: Also add documentation for this filter too.

wp-job-manager-functions.php
```diff
@@ -1602,3 +1602,72 @@ function job_manager_count_user_job_listings( $user_id = 0 ) {
...
+		}
+	}
+
+	return apply_filters( 'job_manager_user_can_browse_job_listings', $can_browse );
+}
+
+/**
+ * True if an the user can view a resume.
+ *
+ * @since 1.36.0
```
fjorgemota, line 1631: Remember to update to 1.37.0

wp-job-manager-functions.php
```diff
@@ -1602,3 +1602,72 @@ function job_manager_count_user_job_listings( $user_id = 0 ) {
...
+	if ( $job->post_author > 0 && absint( $job->post_author ) === get_current_user_id() ) {
+		$can_view = true;
+	}
+
+	$key = get_post_meta( $job_id, 'share_link_key', true );
+
+	// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+	if ( $key && ! empty( $_GET['key'] ) && $key === $_GET['key'] ) {
+		$can_view = true;
+	}
```
fjorgemota, line 1670: This block of code looks like..unused. I haven't found any other code in WPJM (even in this PR) mentioning that post meta.

Could you remove it, please?

templates/access-denied-browse-job_listings.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Access denied message when attempting to browse job listings.
+ *
+ * This template can be overridden by copying it to yourtheme/job_manager/access-denied-browse-job_listings.php.
+ *
+ * @see         https://wpjobmanager.com/document/template-overrides/
+ * @author      Automattic
+ * @package     wp-job-manager
+ * @category    Template
+ * @version     1.36.0
```
fjorgemota, line 11: 1.37.0, just to remember.

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -934,4 +956,131 @@ protected function input_multi( $option, $ignored_attributes, $values, $ignored_
...
+	 *
+	 * @param array $value Array of capabilities and roles.
+	 * @return string
+	 */
+	private static function capabilities_array_to_string( $value ) {
+		if ( ! is_array( $value ) ) {
+			return '';
+		}
+
+		$caps = array_filter( array_map( 'sanitize_text_field', $value ) );
```
fjorgemota, line 1057: I think you could try to check against the list of values returned by `get_capabilities_and_roles` here, just to avoid possible issues with invalid values being injected by the client. WDYT?

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -934,4 +956,131 @@ protected function input_multi( $option, $ignored_attributes, $values, $ignored_
...
+	private static function capabilities_string_to_array( $value ) {
+		return array_filter(
+			array_map(
+				function( $value ) {
+					return trim( sanitize_text_field( $value ) );
+				},
+				explode( ',', $value )
+			)
+		);
+	}
```
fjorgemota, line 1044: I think if we save the option as an array on the database we won't need this, right? (unless maybe a call to the same function that we will define as sanitizer for this option)

WDYT?

fjorgemota (commented):
Hey there, thanks for your PR!

I left a few comments that I noticed after reading the code. I hope you don't mind.

Testing it on WP, it seems to work pretty OK. 

Thanks again!

fjorgemota (approved):
Hey there. Disregard the last review, I'll fix that in a separate PR. Should be some quick fixes, after all. =)

Thanks for your contribution!

Gnodesign:
@fjorgemota Sorry been away and not available these. Thanks a lot for jumping in on this! 🎉

fjorgemota:
Ah, no worries, @Gnodesign. I decided to tackle the fixes based on the review they looked simple enough. Feel free to take a look at #2292 if you want, but no pressure at all. =)

PR #2285, opened 2022-07-02
Fix Salary Field Currency and Unit

### LINKED ISSUE #2278: Salary Field uses USD as the default currency
**Describe the bug**

With the latest WPJM update (v.1.36.1), when the salary field is enabled, "USD / YEAR" is applied automatically on the job display page. There isn't a way to change this text elsewhere.

**To Reproduce**
1. Go to Job Listings > Settings > Job listings
2. Scroll down to 'Salary' and enable Job Salary
3. Post a job with a salary, entering a numerical value
4. See the job posting with USD applied 

**Screenshots**
image

**Expected behavior**
I expected to be able to either select desired currency, or no currency at all.

**WordPress Environment**
- WordPress Version: 6.0
- WP Job Manager Version: 1.36.1
- PHP Version: 7.4

fjorgemota:
Fixes #2278

* Add an option to set default salary currency on the settings panel;
* Add an option to set the default salary unit (or period of time, I may need to find a better name here) on the settings panel;
* Add an option to enable or disable customization of the salary currency on the settings panel for each job;
* Add an option to enable or disable customization of the salary unit (again, of period..) on the settings panel for each job;
* Add, on the job editor, an option to set salary currency (if that's enabled on the settings);
* Add, on the job editor, an option to set salary unit (if that's enabled on the settings);
* Allow setting an empty salary unit, which disables the addition of a "/" on the salary. Might need to document this behavior better;
* If there's no salary, hide the icon in the frontend;
* Add the new fields to the frontend, so users submitting new jobs or editing already registered jobs can see it too;
* Fix an issue with non-translated job units in the frontend; 

### Testing instructions

1. Install the plugin containing this branch of the code in a WordPress install;
2. Check the new options in Job Listings -> Settings -> Job Listings related to Salary Currency and Unit;
3. Verify the new fields related to Salary Currency and Unit on the page Job Listings -> Add New and their behavior;
4. Verify the new information appearing on the Job Page;

### New Hooks

- `the_job_salary_message` - Filter the returned job salary message (with the additional formatting);
- `wpjm_job_salary_unit_display_text` - Filter the returned job salary unit display text (for translations and other messages);

### Tasks to do

- [x] Right now, if you enable the customization of the salary currency, and edit the salary currency for a given job, and then disable the customization again, the job will keep showing the customized salary currency. I'm thinking here about possible issues with this approach;
- [x] Verify the output of the structured data in each case;
- [x] Check if the new hook is indeed necessary;


alexsanford (commented):
From an initial look, this seems to work well! 🎉 

I have a few thoughts/suggestions:

- I'm not convinced that the per-listing customization is needed. It adds complexity of options, extra fields in the editor, and technically the fields should be added to the job listing add/edit form on the frontend as well (as that is where most people edit their listings). Given that we'll still have the filters (which take the `$post` as input), maybe we can just have the site-wide setting, rather than per-listing?
- I would love to see the settings show and hide based on whether "Enable Job Salary" is checked. Similar to how the "Account Creation" checkbox works on the "Job Submission" tab.
- I think calling it a Salary Period or something like that makes more sense than "Unit", personally.

tripflex:
> From an initial look, this seems to work well! 🎉
> 
> I have a few thoughts/suggestions:
> 
> * I'm not convinced that the per-listing customization is needed. It adds complexity of options, extra fields in the editor, and technically the fields should be added to the job listing add/edit form on the frontend as well (as that is where most people edit their listings). Given that we'll still have the filters (which take the `$post` as input), maybe we can just have the site-wide setting, rather than per-listing?
> * I would love to see the settings show and hide based on whether "Enable Job Salary" is checked. Similar to how the "Account Creation" checkbox works on the "Job Submission" tab.
> * I think calling it a Salary Period or something like that makes more sense than "Unit", personally.

Regarding the per-listing customization I respectfully disagree, having worked on client sites for the past 9+ years I can tell you that users who added a salary field, 90% of them are not just a "single" salary type (many support hourly, etc).  Maybe we can come up with a better way of handling it, but i don't think we should assume everyone will use "one salary type"

alexsanford:
> Regarding the per-listing customization I respectfully disagree

Fair enough! Thanks for the input. In that case, I guess we'll need to add these fields to the frontend as well.

I wonder if for the "Period" (or "Unit") we should have a dropdown, rather than a text field? Seems like salary by Year, Month, or Hour would generally cover most cases? And we could provide a filter for other options? Text field would also be fine, but a dropdown may be a better experience.

WDYT @tripflex?

tripflex:
> > Regarding the per-listing customization I respectfully disagree
> 
> Fair enough! Thanks for the input. In that case, I guess we'll need to add these fields to the frontend as well.
> 
> I wonder if for the "Period" (or "Unit") we should have a dropdown, rather than a text field? Seems like salary by Year, Month, or Hour would generally cover most cases? And we could provide a filter for other options? Text field would also be fine, but a dropdown may be a better experience.
> 
> WDYT @tripflex?

Yeah I think we should limit the available selections to those that are supported by google for structured data (HOUR/DAY/WEEK/MONTH/YEAR) https://developers.google.com/search/docs/advanced/structured-data/job-posting

That way we know how to output in the structured data based on what the user selects when submitting a listing.  

My thought we could add a setting in admin area that is like "Salary Units" in a dropdown, where the admin can decide if they want a static salary unit (and we don't show unit field on listing), or they could select the "Specified on listing" option which would then show a "unit" field when submitting a listing.

This would then cover both situations, where an admin wants to have a specific SET unit type for every listing, or they can let the user decide when submitting a listing.  IMO we would default to the "Specified on listing" one.

**Regarding issues with users who already have a `job_salary` field:**

I've also found that a lot of clients who already have this setup, the user is defining in the actual `job_salary` field what type of "unit" it is as well.

I'm thinking maybe it would be good to come up with some kind of parser to try and parse out the value from the static field and see if we can "guess" what the unit type should be.  This would mean users with this existing field can still benefit from structured data output.  My first thought was to actually just create some kind of free plugin to put on WordPress.org that would help users convert their existing `job_salary` fields to the new structure we decide on, so just mentioning this as whatever we go with that's something that will be a pain for a lot of users right now.

Also wondering if maybe it's best to have the job_salary field disabled by default because there are so many themes out there that have already created this `job_salary` field in some way which seems to have conflicts with the current implementation

tripflex:
Also thinking for the currency, maybe we could do the same thing with a dropdown, the admin can set a static currency for all listings, or select a "Specified on listing" and then we show that field on listing submission maybe?

fjorgemota:
Hey there.

> - I would love to see the settings show and hide based on whether "Enable Job Salary" is checked. Similar to how the "Account Creation" checkbox works on the "Job Submission" tab.

Done that in [faefe2a](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/faefe2a15efab035ab5c969e587e54c4593868e4)

> - I think calling it a Salary Period or something like that makes more sense than "Unit", personally.

Just to note down here: We decided to leave the name of the field as Unit because Google calls that way, and we already have functions named as `salary_unit`, so...yeah. =) 


> Yeah I think we should limit the available selections to those that are supported by google for structured data (HOUR/DAY/WEEK/MONTH/YEAR) [developers.google.com/search/docs/advanced/structured-data/job-posting](https://developers.google.com/search/docs/advanced/structured-data/job-posting)

Fixed that in [fd9dc6f](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/fd9dc6fbec420b938d73f8857e1e82a2c59852bf)

 > My thought we could add a setting in admin area that is like "Salary Units" in a dropdown, where the admin can decide if they want a static salary unit (and we don't show unit field on listing), or they could select the "Specified on listing" option which would then show a "unit" field when submitting a listing.

Technically, this is why I added both a checkbox AND a separate option to set the value. If the admin doesn't want to let users edit the salary unit, just uncheck the checkbox and set the default option as usual. Otherwise, check the checkbox and allow the user to edit the listing as usual.

> This would then cover both situations, where an admin wants to have a specific SET unit type for every listing, or they can let the user decide when submitting a listing. IMO we would default to the "Specified on listing" one.

Also, about this, I don't want to add even another change to the field. Given that the previous default was to set Currency as USD, and Unit as Year, I left it as it was, except that now you can customize using the admin panel. =) 

> I'm thinking maybe it would be good to come up with some kind of parser to try and parse out the value from the static field and see if we can "guess" what the unit type should be. This would mean users with this existing field can still benefit from structured data output.

This looks somewhat risky, so I won't consider it to be part of the scope of the PR. Having that as a separate plugin looks okay, tho.

> Also thinking for the currency, maybe we could do the same thing with a dropdown, the admin can set a static currency for all listings, or select a "Specified on listing" and then we show that field on listing submission maybe?

Technically, again, the idea of using checkboxes + field for the default value should allow for this. About adding a dropdown, that would mean adding even another option to allow the user to customize the currencies available, so I don't think it should be part of the scope of this PR, maintaining the list of currencies might lead to problems, too.

fjorgemota:
By the way, one thing that I forgot:

> I guess we'll need to add these fields to the frontend as well.

I added the new fields to the frontend on the commit [e93f819](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/e93f819f94ac1335892314b74bd3214651875c1c)

And I also improved a little bit the wording for the new fields.

I await for your reviews =)

Gnodesign:
Great job @fjorgemota 👏

> Also wondering if maybe it's best to have the job_salary field disabled by default because there are so many themes out there that have already created this `job_salary` field in some way which seems to have conflicts with the current implementation

I agree with @tripflex on leaving the job_salary field disabled by default. A lot of WPJM themes (including my Cariera theme) have added their own `job_salary` field, this created quite the issue once 1.36.0 got released. By leaving it disabled by default it will save all of us a lot of support tickets and we can update our themes with a workaround for the users to choose if the want to use the native job_salary or continue using the custom one.

fjorgemota:
Just FYI, I merged `trunk` again into this branch after the merge of PR #2288. I had to fix one conflict, but it should still be working in the same way as before: [`1143fe1` (#2285)](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/1143fe16852979d443a366a2d5c3693c8138ec66) (with code styles here [d73a2e4](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/d73a2e435c5a544fa489ae7d8193c4422d3979e5) and fixes for the unit tests here [89f5e27](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/89f5e275a95505d7aabf1903d0712daf48be45da))

fjorgemota:
@tripflex and @Gnodesign One question: What exactly do you mean "leaving the job_salary field disabled by default"?

Right now, all the new field's settings have its `std` set as `'0'`, [including the field `job_manager_enable_salary`](https://github.com/Automattic/WP-Job-Manager/blob/1143fe16852979d443a366a2d5c3693c8138ec66/includes/admin/class-wp-job-manager-settings.php#L227-L235), which means that they are disabled by default. 

Testing here on the [latest plugin build](https://github.com/Automattic/WP-Job-Manager/suites/7279692577/artifacts/293176430) on a new WP installation, the salary field doesn't appear by default for me, so...what do you mean by leaving that field disabled by default? :) 

Thanks!

alexsanford (commented):
This is looking good! Works well for me. Will await some testing from others, though.

Could we change the frontend template to use the display string, rather than the all-caps string? E.g. "Week" instead of "WEEK"?

Screen Shot 2022-07-08 at 16 10 57

wp-job-manager-template.php
```diff
@@ -1371,5 +1400,41 @@ function get_the_job_salary_unit( $post = null ) {
...
+ * Get the display text for the job salary unit
+ *
+ * @since 1.37.0
+ * @param int|WP_Post|null $post (default: null).
+ * @return string|null
+ */
+function get_the_job_salary_unit_display_text( $post = null ) {
+	$job_unit = get_the_job_salary_unit( $post );
+
+	$translated_options = [
```
alexsanford: I'm realizing that this is the 4th place where this list has been duplicated. I wonder if we should extract them out? As a bonus, we could even make it filterable 🤷‍♂️

alexsanford (commented):
It's a little bit weird to see the salary icon on the frontend for jobs that don't have any salary information...should we hide this?

Screen Shot 2022-07-08 at 16 44 39

wp-job-manager-template.php
fjorgemota: Yep, I'd really like to avoid duplicating that list, but, at the same time, I don't have an idea of where to put it.

Do you have any suggestions? Maybe a new class or something?

alexsanford: I'm not sure it warrants a whole new class. What about just throwing it in `wp-job-manager-functions.php`? 🤷‍♂️

fjorgemota:
> It's a little bit weird to see the salary icon on the frontend for jobs that don't have any salary information...should we hide this?

I agree! I adjusted the code to hide it in [454c819](https://github.com/Automattic/WP-Job-Manager/pull/2285/commits/454c81985434e25acd9743f59c924ee3700bf3ea)!

wp-job-manager-template.php
fjorgemota: Good idea, done in 81e8252cc7ef622b9e4de2d38dfb4de1a316536c.

Gnodesign:
> Right now, all the new field's settings have its `std` set as `'0'`, [including the field `job_manager_enable_salary`](https://github.com/Automattic/WP-Job-Manager/blob/1143fe16852979d443a366a2d5c3693c8138ec66/includes/admin/class-wp-job-manager-settings.php#L227-L235), which means that they are disabled by default.
> 
> Testing here on the [latest plugin build](https://github.com/Automattic/WP-Job-Manager/suites/7279692577/artifacts/293176430) on a new WP installation, the salary field doesn't appear by default for me, so...what do you mean by leaving that field disabled by default? :)
> 
> Thanks!

@fjorgemota Just tested it and it works perfectly fine for me. Guess most of the tickets were coming from users that were enabling the salary option by themselves and were facing issues because of the 2 different `_job_salary`. Will handle everything else from my theme. Sorry for the confusion!

fjorgemota:
Ah, nice to hear that, @Gnodesign! I was worried for a moment here. =) Thanks for the reply and no worries!

alexsanford (commented):
This is looking great! I just found _one_ more issue haha.

Other than the below, the code looks good, and everything else worked well in my testing 🎉 

---

When I edit a job listing on the frontend (through the Job Dashboard), the Salary Unit doesn't default to the current custom value for that listing.

### In the Block Editor:

Screen Shot 2022-07-11 at 10 53 52

### On the frontend edit page:

Screen Shot 2022-07-11 at 10 54 13

alexsanford (approved):
We discussed the above and discovered that it was just the browser caching the form fields, which is expected behaviour. Typically, users should not see this.

🚀

AshcloudMac:
Is this available for everyone yet?

fjorgemota:
@AshcloudMac It's already merged into the default branch, but we haven't released it yet. We're still planning some details around the release, so it's not available for everyone yet. 

I'd recommend keeping track of the [releases page](https://github.com/Automattic/WP-Job-Manager/releases/) to be informed about any new release. :)

PR #2290, opened 2022-07-07
search_location back to original handling + remote_listing field in filters

### LINKED ISSUE #2287: Multiple locations search causes incorrect results to be returned
https://github.com/Automattic/WP-Job-Manager/pull/2261

If a user searches "Atlanta, GA" this is going to return any result in "GA" which is not what the user is looking for specifically.  This merged PR basically splits the location query into multiple queries now split by comma or spaces.

What about when someone wants to do an exact search?

The only solution I can think of for now is to add a filter to disable this functionality and use the original search handling as before, or coming up with a way to determine when the user wants to search as an "OR" type query vs an exact query.  It's naive to assume everyone wants it as an "OR" query, especially when locations sometimes need to be very specific.

https://wordpress.org/support/topic/locations-dropdown-no-longer-specific/#post-15799430

**I feel like this PR was merged very quickly without any discussion around it (implementation wise) even though it was a major change on how location queries are actually made.**

Especially because since 2017 there's only been 3 upvotes on wanting the actual feature:
https://github.com/Automattic/WP-Job-Manager/issues/1217

tripflex:
Fixes #2287 

- [x] Revert `search_location` back to original handling prior to 1.36.x+
- [x] Add `remote_position` to filtering
- [x] Add shortcode support for `remote_position`
- [x] Add support for `remote_position` in `get_job_listings`
- [x] Add support for `remote_position` in `WP_Job_Manager_Ajax` `get_listings`
- [x] Add support for `remote_position` in Recent Jobs Widget
- [x] Add setting to enable/disable `remote_position`
- [x] Add mobile support (stacking) for keyword and location fields

### Remote Position Functionality
- Shortcode, Recent Jobs Widget (via select), and search method (function) support 3 values for `remote_position`:
    - `true` to show only remote, `false` to hide remote, `null` to show both (default)

- By default when `remote_position` parameter is specified in search, and search is for listings that are NOT remote positions, two meta queries will be added.  One for `!= 1` and one for `NOT EXIST` for backwards compatibility of sites prior to 1.36.x that have listings without this meta saved on them.  The `NOT EXIST` meta query can be disabled by returning false to `job_manager_get_job_listings_remote_position_check_not_exists` filter

- Remote position displaying on frontend search & filtering can be disabled via setting in admin area (disabling remote position field) or by passing false to `job_manager_job_filters_show_remote_position` filter

- Remote position field works in conjunction with the standard location search.  If a value is passed for both `search_location` and `remote_position` it will be combined as an `AND` query (ie you search `Atlanta` and `true` for `remote_position`, it will only return results that match against `Atlanta` AND have remote position checked) as listings can be both remote and have specific location values set on them.

### DIFF
--- a/assets/js/ajax-filters.js
+++ b/assets/js/ajax-filters.js
@@ -367,8 +367,8 @@ jQuery( document ).ready( function( $ ) {
 					location = $location.val();
 				}
 
-				if( $remote_position.length && $remote_position.is( ':checked' ) ) {
-					remote_position = 'true';
+				if( $remote_position.length ) {
+					remote_position = $remote_position.is( ':checked' ) ? 'true' : null;
 				}
 
 				data = {
--- a/assets/js/ajax-filters.js
+++ b/assets/js/ajax-filters.js
@@ -312,6 +312,7 @@ jQuery( document ).ready( function( $ ) {
 			var order = $target.data( 'order' );
 			var featured = $target.data( 'featured' );
 			var filled = $target.data( 'filled' );
+			var remote_position = $target.data( 'remote_position' );
 			var job_types = $target.data( 'job_types' );
 			var post_status = $target.data( 'post_status' );
 			var index = $( 'div.job_listings' ).index( this );
@@ -378,6 +379,7 @@ jQuery( document ).ready( function( $ ) {
 					page: page,
 					featured: featured,
 					filled: filled,
+					remote_position: remote_position,
 					show_pagination: $target.data( 'show_pagination' ),
 					form_data: $form.serialize(),
 				};
@@ -406,6 +408,7 @@ jQuery( document ).ready( function( $ ) {
 					page: page,
 					featured: featured,
 					filled: filled,
+					remote_position: remote_position,
 					show_pagination: $target.data( 'show_pagination' ),
 				};
 			}
--- a/assets/js/ajax-filters.js
+++ b/assets/js/ajax-filters.js
@@ -356,6 +356,7 @@ jQuery( document ).ready( function( $ ) {
 				location = '';
 				var $keywords = $form.find( ':input[name="search_keywords"]' );
 				var $location = $form.find( ':input[name="search_location"]' );
+				var $remote_position = $form.find( ':input[name="remote_position"]' );
 
 				// Workaround placeholder scripts
 				if ( $keywords.val() !== $keywords.attr( 'placeholder' ) ) {
@@ -364,6 +365,10 @@ jQuery( document ).ready( function( $ ) {
 
 				if ( $location.val() !== $location.attr( 'placeholder' ) ) {
 					location = $location.val();
+				}
+
+				if( $remote_position.length && $remote_position.is( ':checked' ) ) {
+					remote_position = 'true';
 				}
 
 				data = {
@@ -453,7 +458,7 @@ jQuery( document ).ready( function( $ ) {
 		} );
 
 	$(
-		'#search_keywords, #search_location, .job_types :input, #search_categories, .job-manager-filter'
+		'#search_keywords, #search_location, #remote_position, .job_types :input, #search_categories, .job-manager-filter'
 	)
 		.change( function() {
 			var $target = $( this ).closest( 'div.job_listings' );
@@ -487,6 +492,10 @@ jQuery( document ).ready( function( $ ) {
 				.find( ':input[name="filter_job_type[]"]' )
 				.not( ':input[type="hidden"]' )
 				.prop( 'checked', true );
+			$form
+				.find( ':input[name="remote_position"]' )
+				.not( ':input[type="hidden"]' )
+				.prop( 'checked', false );
 
 			$target.triggerHandler( 'reset' );
 			$target.triggerHandler( 'update_results', [ 1, false ] );
--- a/templates/job-filters.php
+++ b/templates/job-filters.php
@@ -8,7 +8,7 @@
  * @author      Automattic
  * @package     wp-job-manager
  * @category    Template
- * @version     1.33.0
+ * @version     1.37.1
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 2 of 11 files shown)


tripflex:
In hindsight looking at this, i'm not sure if having the ability to define multiple "Locations" is the way to go.  Checking monster.com, indeed.com, and careerbuilder.com -- none of them support multiple "Locations" in a search.  Most likely due to the complications around using auto suggest and handling multiple entries (is my guess).

For now i'm thinking it's best to not make this something as a setting in default system.  We can easily add support for an array of location values (if that's passed by theme or plugin), but i think it's best we leave that up to plugins or themes and just add basic support for array of locations.

Hastibe:
Just wanted to chime in that I agree with @tripflex -- I also thought this was strange, as I couldn't find a single significant job search site that does this, and I don't think users would expect this (and haven't encountered any meaningful demand expressed for it, here or elsewhere). If visitors want to search for multiple locations, it's very quick and easy to do that the same way it is on any other significant job search site out there, by doing a second search.

Gnodesign:
@alexsanford Any chance we can get this added to 1.37.0 as well? The multiple location search has only been causing issues.

tripflex:
What is you guys thoughts in regards to adding the "remote position" on the filtering page?  Should it be a "remote positions only" kind of filter or an "include remote positions"?

Seems that Monster does it as "Remote jobs only" and allows you to still enter a location value, CareerBuilder does it as remote-only and does NOT allow location values, ZipRecruiter same as CareerBuilder.  I do like how monster does it though you can search for a job that is remote while also narrowing it down to a specific location on the remote job as well (most likely where home office is located kind of thing) and since we do allow the submission of listing that has a Location value and remote position checked ... makes me thing monster flow type may be better suited.

https://www.monster.com/jobs/search?q=&where=Atlanta%2C+GA&page=1&et=REMOTE&so=m.s.lh

Or we could do it like a "Include remote positions" so you can search for specific location but also return remote listings at the same time, and maybe just add a setting or a filter the site admin can set to define if it should be "include" or "only" in regards to remote positions

Hastibe:
If just talking about a "remote" search option, I think it should be a "Remote positions only" filter that can be checked (in addition to filling out keywords and/or location, if desired). Most of the remote positions we see still have a geographical constraint (i.e. are remote in our state or are remote in a portion of the state, or are remote in our region or country), so being able to see the location specified for the remote position (and have it come up in searches for that location, not to mention keyword) is still important. 

Being able to pre-filter it by shortcode would be important, so visitors can be directed to a "Remote jobs" landing page, and I also think any "Remote jobs" filtering option should be able to be disabled by the site admin, if not wanted / relevant.

A more comprehensive solution (see the "Location & Remote Preferences" filter for [Idealist](https://www.idealist.org/en/jobs?q=) and the "Listing Type" filter for [Mac's List](https://jobs.macslist.org/search)), which I think would be better, would essentially be a separate, second row of filtering options, just like how Job Types ("Full-Time," "Part-Time," etc.) work, where they are all pre-checked, but can be unchecked to filter down, and would include these options:

- On-Site
- Hybrid
- Remote

...In fact, just adding these options as additional Job Types (in addition to "Full-Time," Part-Time," etc.) would work perfectly, except that having the two different Job Type types, so to speak, combined would be confusing to employers (as they already miss the note that they can select multiple job types, and there's currently no way -without custom work- to enforce a minimum selection of at least two different types, etc.), as well as confusing to job seekers, as the two different Job Type types would be listed in a mixed-up (alphabetical) way on the front-end.

Hastibe:
To follow-up on my previous comment with a TL;DR, it seems to me, as I think it'd have the ability to be both the most comprehensive and most flexible, that adding an extra taxonomy for job post (on-site, remote, hybrid, etc.) filtering options could be the best option (we have "Job Type," but we also need something like a "Location Type" or "Position Type" now).

tripflex:
Thanks for the input @Hastibe!

> If just talking about a "remote" search option, I think it should be a "Remote positions only" filter that can be checked (in addition to filling out keywords and/or location, if desired). Most of the remote positions we see still have a geographical constraint (i.e. are remote in our state or are remote in a portion of the state, or are remote in our region or country), so being able to see the location specified for the remote position (and have it come up in searches for that location, not to mention keyword) is still important.
> 

Yup okay that's what I was thinking as well, so that's the way I will work on building it out (at least as the default way of handling things)

> Being able to pre-filter it by shortcode would be important, so visitors can be directed to a "Remote jobs" landing page, and I also think any "Remote jobs" filtering option should be able to be disabled by the site admin, if not wanted / relevant.
> 

Good call, i'll add this to the list to have as a shortcode option.  I'll also make sure to add a filter to disable showing the remote positions on search/filtering page 

> A more comprehensive solution (see the "Location & Remote Preferences" filter for [Idealist](https://www.idealist.org/en/jobs?q=) ....

Yeah this would be much more work outside the scope of this current PR

assets/js/ajax-filters.js
```diff
@@ -365,6 +367,10 @@ jQuery( document ).ready( function( $ ) {
 					location = $location.val();
 				}
 
+				if( $remote_position.length && $remote_position.is( ':checked' ) ) {
+					remote_position = 'true';
+				}
```
dadish: This allows only to enable "Remote positions only", but not disable it. For example if the shortcode is `[jobs remote_position=true]` then the "Remote positions only" is checked by default, but unchecking it does not clear the value. It always shows the remote positions only.

templates/job-filters.php
```diff
@@ -36,6 +36,13 @@
 			<input type="text" name="search_location" id="search_location" placeholder="<?php esc_attr_e( 'Location', 'wp-job-manager' ); ?>" value="<?php echo esc_attr( $location ); ?>" />
 		</div>
 
+		<?php if( apply_filters( 'job_manager_job_filters_show_remote_position', get_option('job_manager_enable_remote_position', true ), $atts ) ) : ?>
```
dadish, line 39: Shouldn't the default value for `get_option('job_manager_enable_remote_position', true )` be `false` here?

dadish (commented):
This is a very good work @tripflex! Please address some small comments below and it should be good to go.

templates/job-filters.php
tripflex: So the reason I left it as true is because the original release with the `remote_position` field added had no way to disable it, so I was matching the actual setting default value (which is remote position enabled by default).

Reason being, say someone has installed the original version that added the `remote_position` field but never went into settings and saved them at all.  When they update to this release, the field will not show on the form even though they "assume" it's enabled based on the default value.

Figured users would want this field to show on the filters by default, if they want to disable it can be done via the settings or filter if you specifically don't want to show on the filters page (but default it shows)

assets/js/ajax-filters.js
tripflex: Ahh good catch, okay so I think this would work instead, as this would set it to null value if the input exists (as in they haven't disabled showing it via a filter) and the checkbox is **unchecked**

```
if( $remote_position.length ) {
    remote_position = $remote_position.is( ':checked' ) ? 'true' : null;
}
```

Passing `null` which will basically override the shortcode "default" value, what do you think?

Sharkspm:
Hi,  With regards to the Remote Position - How will this work with existing WPJM sites  that use GEO Job Manager.  If a job is based in Manchester but also has a Remote Position ticked and someone searches for a job in London (hundreds of miles away) that matches the Manchester Job would the Job show in the search even though it maybe outside the users original search radius

Thanks :)

assets/js/ajax-filters.js
dadish: Yes, this should work.

templates/job-filters.php
dadish: Good call. I agree.

dadish:
@Sharkspm 

Nope. The job in Manchester will not show up when user searches for London, even if it has Remote Position ticked.

assets/js/ajax-filters.js
tripflex: @dadish 👍 just pushed the updated to this PR

dadish (approved):
lgtm

PR #2292, opened 2022-07-08
Some fixes for the Job Visibility feature

fjorgemota:
Some fixes to #2280, mostly based on [this review](https://github.com/Automattic/WP-Job-Manager/pull/2280#pullrequestreview-1032266561).

* Update PHPDocs to refer to correct next version;
* Refactor code to not change `$_POST` on the Settings page, instead relying on sanitize callbacks;
* Move code related to Select2 from admin.js to settings.php, so we can use the same event handler to detect when the tab changed;
* Save options of capabilities as an array, avoiding conversions between strings and arrays all the time;
* Remove check for `share_link_key`, which is a post meta not used in WPJM;

### Testing instructions

Verify that the Job Visibility Feature still works, by following the testing instructions on PR #2280.

### Questions

- Right now, in the capabilities field, we show the list of capabilities that the current user has. However, we allow the user to add new "capabilities" to the list pretty freely, so, here's the question: Should we block that and allow the user to only choose what's available on the list? That should be pretty easy to do, but I'm not 100% sure it's needed. :upside_down_face:

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -1044,7 +1044,7 @@ class WP_Job_Manager_Settings {
 		}
 		$result = [];
 		foreach ( $value as $item ) {
-			$item = strtolower( trim( sanitize_text_field( $item ) ) );
+			$item = trim( sanitize_text_field( $item ) );
 			if ( $item ) {
 				$result[] = $item;
 			}
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -534,16 +534,48 @@ class WP_Job_Manager_Settings {
 				return false;
 			});
 
+			if ( jQuery.isFunction( jQuery.fn.select2 ) ) {
+
+				if ( jQuery( '.settings-role-select' ).length > 0 ) {
+					// This fixes a issue where backspace on role just turns it into search.
+					// @see https://github.com/select2/select2/issues/3354#issuecomment-277419278 for more info.
+					jQuery.fn.select2.amd.require(
+						['select2/selection/search' ],
+						function ( Search ) {
+							Search.prototype.searchRemoveChoice = function (decorated, item) {
+								this.trigger(
+									'unselect',
+									{
+										data: item
+									}
+								);
+
+								this.$search.val( '' );
+								this.handleSearch();
+							};
+						},
+						null,
+						true
+					);
+				}
+			}
+
+			var job_listings_admin_select2_settings = {
+				'tags': true // Allows for free entry of custom capabilities.
+			};
+
 			jQuery('.nav-tab-wrapper a').click(function() {
 				if ( '#' !== jQuery(this).attr( 'href' ).substr( 0, 1 ) ) {
 					return false;
 				}
 				jQuery('.settings_panel').hide();
 				jQuery('.nav-tab-active').removeClass('nav-tab-active');
-				jQuery( jQuery(this).attr('href') ).show();
+				var $content = jQuery( jQuery(this).attr('href') );
+				$content.show();
 				jQuery(this).addClass('nav-tab-active');
 				window.location.hash = jQuery(this).attr('href');
 				jQuery( 'form.job-manager-options' ).attr( 'action', 'options.php' + jQuery(this).attr( 'href' ) );
+				$content.find( '.settings-role-select' ).select2( job_listings_admin_select2_settings );
 				window.scrollTo( 0, 0 );
 				return false;
 			});
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -437,20 +437,22 @@ class WP_Job_Manager_Settings {
 					__( 'Job Visibility', 'wp-job-manager' ),
 					[
 						[
-							'name'  => 'job_manager_browse_job_listings_capability',
-							'std'   => '',
-							'label' => __( 'Browse Job Capability', 'wp-job-manager' ),
-							'type'  => 'capabilities',
+							'name'              => 'job_manager_browse_job_listings_capability',
+							'std'               => '',
+							'label'             => __( 'Browse Job Capability', 'wp-job-manager' ),
+							'type'              => 'capabilities',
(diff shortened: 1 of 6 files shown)


includes/admin/class-wp-job-manager-settings.php
```diff
@@ -995,68 +1032,24 @@ class="regular-text settings-role-select"
...
-			// phpcs:disable WordPress.Security.NonceVerification.Missing -- Settings save will handle the nonce check.
-			if ( isset( $_POST[ $capabilities_field ] ) ) {
-				// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Sanitized by `WP_Resume_Manager_Settings::capabilities_array_to_string()`
-				$input_capabilities_field_value = wp_unslash( $_POST[ $capabilities_field ] );
-				if ( is_array( $input_capabilities_field_value ) ) {
-					$_POST[ $capabilities_field ] = self::capabilities_array_to_string( $input_capabilities_field_value );
-				}
+		$result = [];
+		foreach ( $value as $item ) {
+			$item = strtolower( trim( sanitize_text_field( $item ) ) );
```
alexsanford: I don't think `strtolower` is safe in this case. Custom roles can be [arbitrary strings](https://developer.wordpress.org/reference/classes/wp_roles/add_role/). Even though the default roles are lower case, I don't think they have to be.

templates/access-denied-single-job_listing.php
```diff
@@ -8,7 +8,8 @@
  * @author      Automattic
  * @package     wp-job-manager
  * @category    Template
- * @version     1.36.0
+ * @since 		1.37.0
```
alexsanford: This indentation looks off

alexsanford (commented):
Works well! A couple comments below

templates/access-denied-single-job_listing.php
fjorgemota: Fixed in [a269ffc](https://github.com/Automattic/WP-Job-Manager/pull/2292/commits/a269ffc94967748db82e6dbcb5d39808ffd87288) (I just forgot to type correctly...I can fix the message by force pushing if you want)

includes/admin/class-wp-job-manager-settings.php
fjorgemota: Ah, nice. Fixed in 385164801139cb67af25072a6d2f84b1f85b1a2c.

wp-job-manager-functions.php
```diff
@@ -1662,12 +1662,5 @@ function job_manager_user_can_view_job_listing( $job_id ) {
 		$can_view = true;
 	}
 
-	$key = get_post_meta( $job_id, 'share_link_key', true );
```
renatho, line 1665: I checked that this meta exists in the Resumes plugin. Maybe it had some purpose related to the resumes? Or it really shouldn't be here? If it had some purpose, probably we'd need a filter and expand it through Resumes. 🤔

alexsanford: I haven't looked closely, but I'm guessing it's related to a way to share Resumes that aren't publicly visible by sending a URL that has a key in it. If we want that for WPJM I think we would need a separate PR.

This is a guess, though.

fjorgemota: Yeah, reading the Resumes plugin it look related to only resumes, not job advertisements. 

So it's kinda out of scope for the original feature, IMHO.

fjorgemota:
I just noticed that I haven't added a PHPDoc for the added filters as suggested [here](https://github.com/Automattic/WP-Job-Manager/pull/2280#discussion_r916360251). So I just pushed the commit 4626e658 fixing that.

wp-job-manager-functions.php
tripflex: Yes when a resume is used for applications a special link is sent to the listing owner using that share link key so they can view it regardless of the visibility settings.   May not be a bad idea to keep it integrated though, in case a theme or addon plugin is using it in some way.

tripflex: Removing this could have adverse side effects we're not aware of and i see no real benefit of removing it with the potential downsides of doing so

fjorgemota: Hey there @tripflex, thanks for your comments. 

Some points:

- Testing here, the WP Job Manager Resumes still works successfully (I have tested it locally), even after having this change removed, so it shouldn't break that functionality.
- Sorry, but this phrase:  
> May not be a bad idea to keep it integrated though, in case a theme or addon plugin is using it in some way.

Doesn't "match" well with this one: 

> Removing this could have adverse side effects we're not aware

Given that having this post meta being checked here is, by definition, a side-effect, given that we're not using the `share_link_key` anywhere else in the plugin nor allowing the user to change it here.

- With that said, if some theme or plugin wants to add that feature anyway, it can always use filter `job_manager_user_can_view_job`, without any issues. 
- And...finally, as mentioned previously, if we want to add that specific feature for WPJM, it's no problem, we just need to add this a separate PR instead, because it will need some additional changes besides having that checked here. 

In other words, I still prefer to keep this out of this feature, as it makes more sense than just having this snippet here, unused. If we want to add that feature (in a more complete way, with changes to the UI and so on) back in the future, we can do that, but in the meantime, I don't think it's worth keeping it here given that we're not using this post meta anywhere else in this plugin.

alexsanford: > In other words, I still prefer to keep this out of this feature, as it makes more sense than just having this snippet here, unused.

I agree. Omitting this incomplete feature should not have any adverse effects, I think. Note that it was only added in [this previous PR](https://github.com/Automattic/WP-Job-Manager/pull/2280), which has not been released yet. The similar feature within the Resume Manager plugin should not be affected.

@tripflex does that make sense? I just want to be sure we're not missing a case that you're seeing...

alexsanford (approved):
From my perspective, this is ready to go. Would like to hear back from @tripflex on the above discussion, just to make sure we're understanding properly and not missing something. Otherwise, let's 🚢 it!

alexsanford:
I'm going to go ahead and merge this. We'll create a prerelease package today so @tripflex feel free to test and leave any feedback 🙂

tripflex:
Sorry guys just saw these notifications, i guess yeah we will have to see how things go now with this merged :P

PR #2297, opened 2022-07-12
Fix warnings

alexsanford:
Fixes some warnings when testing the 1.37.0 package.

* Remove unused hook.
* Check for empty value before looping.

### Testing instructions

* Go to the Settings page for WPJM, and on the Job Listings tab, change the settings for the Salary field.
* Check the PHP error log and ensure there are no new messages.

Previously there were the messages:

```
[12-Jul-2022 17:18:57 UTC] PHP Warning:  call_user_func_array() expects parameter 1 to be a valid callback, class 'WP_Job_Manager_Settings' does not have a method 'pre_process_settings_save' in /.../wp-includes/class-wp-hook.php on line 307

...

[12-Jul-2022 17:18:57 UTC] PHP Warning:  Invalid argument supplied for foreach() in /.../plugins/wp-job-manager/includes/admin/class-wp-job-manager-settings.php on line 1092
```

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -1088,12 +1088,16 @@ class WP_Job_Manager_Settings {
 			$value = explode( ',', $value );
 		}
 		$result = [];
-		foreach ( $value as $item ) {
-			$item = trim( sanitize_text_field( $item ) );
-			if ( $item ) {
-				$result[] = $item;
+
+		if ( ! empty( $value ) ) {
+			foreach ( $value as $item ) {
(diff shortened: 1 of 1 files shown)


fjorgemota (approved):
Looks good and works well! 

Testing here, I also found the following warning in the PHP logs:

```[12-Jul-2022 17:44:02 UTC] PHP Warning:  in_array() expects parameter 2 to be array, string given in /var/www/html/wp-content/plugins/WP-Job-Manager/includes/admin/class-wp-job-manager-settings.php on line 1069```

However, I wasn't able to find a way to reproduce this warning consistently, and while testing your branch it didn't happen anymore either, so...with that, I'm approving this PR.

PR #2298, opened 2022-07-12
Add support for empty salary unit field

fjorgemota:

* Add PHPDoc to document field related to options of the salary unit field;
* Include the empty option in all salary unity fields, so we can have the empty option as a salary unit [as requested by some users](https://github.com/Automattic/WP-Job-Manager/issues/2278#issuecomment-1167320597);

### Testing instructions

* Verify that the empty "option" (`--`) appears in every salary unit field. The default for the settings page should still be the "Year" option, by the way;
* Verify that if you choose that option (the empty one), but choose a currency and a salary, the salary unit nor the "slash" will appear on the frontend;

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -1621,7 +1621,7 @@ class WP_Job_Manager_Post_Types {
 				'label'         => __( 'Salary Unit', 'wp-job-manager' ),
 				'type'          => 'select',
 				'data_type'     => 'string',
-				'options'       => job_manager_get_salary_unit_options( true ),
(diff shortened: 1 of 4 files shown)


alexsanford (commented):
Hmm. This seems ok, although it's missing a weird case. If we have a non-empty unit set in the global settings, then there is no way to have an empty unit for individual listings. If I choose the "blank" option, then this falls back to the default from the settings, instead of actually being blank.

Is this the intention? Maybe this is an edge case that is more work than it's worth?

alexsanford:
Discussed, and we're ok with it as-is for now.

alexsanford (approved):

PR #2313, opened 2022-08-24
Allow `0` as input during job submission

### LINKED ISSUE #2312: Allow job submission input of `0` (zero) on standard input fields
**Describe the bug**
We don't allow '0' on standard input fields due to a logic error.

**To Reproduce**
1. Enable the salary field.
2. Submit a job with 0 salary.
3. Observe that the field is cleared.

**Expected behavior**
Zero to be preserved.

Similar to #1480, but not entirely about required field logic.

### Context
https://wordpress.org/support/topic/text-input-fields-not-saving-0-as-value/#post-15906460
5470259-zd-woothemes

jom:
Fixes #2312

* Fixes logic on `get_posted_jobs` and `get_posted_textarea_field `. We really just want to check for strict `false` from input check.
* Also added `is_array()` checks when we expect an array and are about to pass to `array_map()`.
* Moved `is_empty` method to parent class. Added `$key` param. Used this when populating `empty` for most fields (used in required checks).

### Testing instructions

* Make sure you can enter 0 for currency and it saves.
* With the snippet below, ensure you can enter `0` for a job title and it saves (this causes lots of other problems, but it is just to test the functionality for other custom fields).

### New/Updated Hooks

* `submit_job_form_validate_fields_empty_values` : Added `$key` parameter to allow people to customize what the field is.

### Snippet for Testing
```php
add_filter('submit_job_form_validate_fields_empty_values', function( $empty_vals, $key ) {
  if ($key === 'job_title') {
	return array_diff( $empty_vals, [ '0' ] );
  }
  
  return $empty_vals;
}, 10, 2);
```

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -504,7 +504,7 @@ abstract class WP_Job_Manager_Form {
 							$value = trim( $value );
 						}
 
-						$this->fields[ $group_key ][ $key ]['empty'] = empty( $value );
+						$this->fields[ $group_key ][ $key ]['empty'] = $this->is_empty( $value, $key );
 					};
 				}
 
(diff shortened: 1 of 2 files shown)


includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -533,6 +533,35 @@ protected function get_posted_fields() {
...
 	}
 
+	/**
+	 * Checks whether a value is empty.
+	 *
+	 * @param string|numeric|array|boolean $value The value that is being checked.
+	 * @param string                       $key   The key of the field that is being checked.
+	 * @return bool True if value is empty, false otherwise.
+	 */
+	protected function is_empty( $value, $key = '' ) {
```
jom, line 543: Doubt anyone is using this, but setting default to `$key` because it isn't part of initial implementation. We could add a deprecation notice for ``, but not sure if necessary.

renatho (approved):
Works as described!

PR #2323, opened 2022-09-07
Fix fatal error when using Sensei's learning mode

### LINKED ISSUE #2322: Conflict with Sensei's Learning Mode
**Describe the bug**
On a WP installation with both WPJM and Sensei installed and activated, if you create a course with a few lessons, and enable Learning Mode for that course, you can't get to access any lesson page with Learning Mode theme. 

**To Reproduce**
1. On a WP installation, install and activate both WPJM and Sensei;
2. Create a course on Sensei and enable Learning Mode;
3. Create and publish a lesson on the course that you just created;
4. Try to access the lesson, it will crash with a critical error;

**Screenshots**

There are no screenshots (aside from WP's "critical error" page), but I have this stacktrace:

```
PHP Fatal error:  Uncaught Error: Call to undefined function job_manager_count_user_job_listings() in /var/www/html/wp-content/plugins/WP-Job-Manager/includes/class-wp-job-manager-shortcodes.php:105
Stack trace:
#0 /var/www/html/wp-includes/class-wp-hook.php(307): WP_Job_Manager_Shortcodes->handle_redirects(Object(WP))
#1 /var/www/html/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters('', Array)
#2 /var/www/html/wp-includes/plugin.php(522): WP_Hook->do_action(Array)
#3 /var/www/html/wp-includes/class-wp.php(771): do_action_ref_array('wp', Array)
#4 /var/www/html/wp-includes/functions.php(1310): WP->main('')
#5 /var/www/html/wp-content/plugins/sensei/includes/course-theme/class-sensei-course-theme.php(129): wp()
#6 /var/www/html/wp-includes/class-wp-hook.php(307): Sensei_Course_Theme->maybe_override_theme('')
#7 /var/www/html/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#8 /var/www/html/wp-includes/plugin.php(474): WP_Hook->do_action(Array)
#9 /var/www/html/wp-settings.php(519): do_action('se in /var/www/html/wp-content/plugins/WP-Job-Manager/includes/class-wp-job-manager-shortcodes.php on line 105
```

**Expected behavior**
The lessons using Learning Mode should still be acessible even if WPJM is enabled;

- ~[ ] This bug happens with a default WordPress theme active.~ Not relevant, really.

**WordPress Environment**
- WordPress Version: WP 5.9, but also happens with 6.0;
- WP Job Manager Version: 1.38.0 (it started happening on 1.36.0...1.35.3 looks 100% fine);
- PHP Version: 7.2, but also happens with 8.0; 
- Other important details:  N/A;

fjorgemota:
Fixes #2322

* Moves `is_page` checking earlier in the method `handle_redirects`, so we can detect whether we should call WPJM functions or not;

### Testing instructions

Verify if the functionality added on #2165 still works, and verify that Sensei's learning mode is available (or, in other words, verify if the behavior reported in #2322 doesn't happen anymore).

### Notes

I discovered the issue happens because the method `handle_redirects` loads very early in the code. So early, in fact, that sometimes the WPJM functions are not even defined yet, like when we're using Sensei's learning mode. This PR fixes the code only to call those functions when the page being accessed is, in fact, the "submit job" page defined by the user.

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -97,10 +97,9 @@ class WP_Job_Manager_Shortcodes {
 	public function handle_redirects() {
 		$submit_job_form_page_id = get_option( 'job_manager_submit_job_form_page_id' );
 
-		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Input is used safely.
-		if ( ! is_user_logged_in() ||
-			 ! is_page( $submit_job_form_page_id ) ||
-			 ( ! empty( $_REQUEST['job_id'] ) && job_manager_user_can_edit_job( intval( $_REQUEST['job_id'] ) ) )
+		if ( ! is_user_logged_in() || ! is_page( $submit_job_form_page_id ) ||
+			 // phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Input is used safely.
+			( ! empty( $_REQUEST['job_id'] ) && job_manager_user_can_edit_job( intval( $_REQUEST['job_id'] ) ) )
 		) {
 			return;
 		}
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -95,20 +95,23 @@ class WP_Job_Manager_Shortcodes {
 	 * Handle redirects
 	 */
 	public function handle_redirects() {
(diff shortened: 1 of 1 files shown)


fjorgemota:
Maybe you'll want to take a look at this PR too, @Gnodesign. :)

renatho (approved):
Looks good! Fix the issue, and the feature that you mentioned continues working properly! 🎉

PR #2325, opened 2022-09-08
only perform application validation when required or not empty value

### LINKED ISSUE #2324: Application field throws error (can't submit form) when required is false
**Describe the bug**
If a user decides to set the `application` field as optional (setting required to `false`) the form will not allow them to submit at all, due to the specific validation handling found below, that does not check the `required` flag in any way:
https://github.com/Automattic/WP-Job-Manager/blob/trunk/includes/forms/class-wp-job-manager-form-submit-job.php#L516-L547

Users **can** just use the `job_manager_application_field_skip_email_url_validation` filter to return true, but this would then remove all validation regardless of whether the field is required or not.

**To Reproduce**
1. Set `required` to false for `application` field through filter or Field Editor
2. Goto submit page, leave Application field blank
3. Form can not be submitted and "Please enter a valid application email address or URL" error is shown

Basically we need to add a check in the `validate_fields` method to see if `application` field is required or not
https://github.com/Automattic/WP-Job-Manager/blob/0a2d4e337663067106331cb86fe5a033d02220cc/includes/forms/class-wp-job-manager-form-submit-job.php#L516

tripflex:
Fixes #2324

* Only perform application field validation when required OR value is not empty

### Testing instructions

* Try submit listing with invalid value while `required` is still `true` (default)
* Should show validation error

* Set application field `required` to `false` using filter or Field Editor
* Attempt to submit a listing with no value in `application` field, should show preview step

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -515,6 +515,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		// Application method.
 		if ( ! $this->should_application_field_skip_email_url_validation() && isset( $values['job']['application'] ) ) {
 			$allowed_application_method = get_option( 'job_manager_allowed_application_method', '' );
+			$application_required       = isset( $this->fields['job']['application']['required'] ) && $this->fields['job']['application']['required'];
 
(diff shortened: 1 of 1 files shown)


yscik (approved):
Makes sense, thank you for the fix!

PR #2328, opened 2022-09-29
Fix deprecated error in case $post=null in the function the_company_twitter()

MPolleke:
Fix deprecated error in case $post=null in the function the_company_twitter()

This way, if $post=null, it will not run the second test with strlen().

Error for reference: Deprecated: strlen():
Passing null to parameter #1 ($string) of type string is deprecated in /wp-job-manager/wp-job-manager-template.php on line 1137

Tested on PHP 8.1, not sure if that makes a difference.

### DIFF
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -1134,7 +1134,7 @@ function get_the_company_tagline( $post = null ) {
 function the_company_twitter( $before = '', $after = '', $echo = true, $post = null ) {
 	$company_twitter = get_the_company_twitter( $post );
 
-	if ( 0 === strlen( $company_twitter ) ) {
+	if ( ! $company_twitter || 0 === strlen( $company_twitter ) ) {
 		return null;
 	}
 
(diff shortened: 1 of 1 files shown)


wp-job-manager-template.php
```diff
@@ -1134,7 +1134,7 @@ function get_the_company_tagline( $post = null ) {
 function the_company_twitter( $before = '', $after = '', $echo = true, $post = null ) {
 	$company_twitter = get_the_company_twitter( $post );
 
-	if ( 0 === strlen( $company_twitter ) ) {
+	if ( ! $company_twitter || 0 === strlen( $company_twitter ) ) {
```
yscik: I think we can simplify this to:
```suggestion
	if ( empty( $company_twitter ) ) {
```

MPolleke:
Hi,
The pull request is now updated. 

The change suggested is fine with me. I saw both variations in the existing code and just picked the one I liked :)

Also, the getter function has now been changed.

yscik (approved):
Sounds good. Thanks for the fix!

PR #2332, opened 2022-10-12
Add support for WP.com Marketplace Addon purchase flow

### LINKED ISSUE #67551: Marketplace: Add integration to WP Job Manager add-ons
Add Dotcom Marketplace integration and generic filters to control the WP Job Manager's built-in licensing notices.

To be more specific, the open-source WP Job Manager plugin will need a WPCOM specific implementation [in the third-party directory](https://github.com/Automattic/WP-Job-Manager/tree/trunk/includes/3rd-party) that would listen to the WPCOMSH `wpcom_marketplace_webhook_response_$product_slug` licensing filter.

When this filter will be executed, the code would grab the license payload and update its settings.

BogdanUngureanu:
When an addon is purchased through the WP.com Marketplace, the new integration would automatically configure the addon with the licensing information.

This is done by using a WordPress.com specific filter that's executed after the provision_license event is triggered. This filter contains the licensing information needed by the plugin to connect to WPJobManager.com back-end API.

Fixes https://github.com/Automattic/wp-calypso/issues/67551

* Add an integration for WP.com Marketplace that automatically configures the licensing on a target site

### Testing instructions

* Checkout this branch locally and upload the plugin on one of your test sites
* If you have WP CLI installed, execute `wp shell` and execute the following snippet
```
$payload = [ 'license_code' => 'invalid-license-key', 'email_address' => 'example@example.com', 'wpjm_product_slug' => 'wp-job-manager-applications' ];
$result = apply_filters( 'wpcom_marketplace_webhook_response_wp-job-manager-applications', true, $payload, 'provision_license' );
```
- This should return an error.
- To execute it with a valid license with an associated valid email address, please contact me.

### DIFF
--- a/includes/3rd-party/wpcom.php
+++ b/includes/3rd-party/wpcom.php
@@ -63,7 +63,7 @@ foreach ( WPJM_WPCOM_PRODUCTS as $wpjm_wpcom_product ) {
  * @return false|mixed
  */
 function wpjm_hide_addon_license_form_for_purchases_on_wpcom( $status, $product_slug ) {
-	$subscriptions = get_option( 'wpcom_active_subscriptions', array() );
+	$subscriptions = get_option( 'wpcom_active_subscriptions', [] );
 	if ( isset( $subscriptions[ $product_slug ] ) ) {
 		return false;
 	}
@@ -81,12 +81,12 @@ add_filter( 'wpjm_display_license_form_for_addon', 'wpjm_hide_addon_license_form
  * @return false|void
  */
 function wpjm_display_managed_by_wpcom_notice_for_addon( $product_slug ) {
-	$subscriptions = get_option( 'wpcom_active_subscriptions', array() );
+	$subscriptions = get_option( 'wpcom_active_subscriptions', [] );
 	if ( ! isset( $subscriptions[ $product_slug ] ) ) {
 		return false;
 	}
 
-	esc_html_e( 'The license for this product is automatically managed by WordPress.com.', 'wpjm' );
+	esc_html_e( 'The license for this product is automatically managed by WordPress.com.', 'wp-job-manager' );
 }
 
 add_action( 'wpjm_manage_license_page_after_license_form', 'wpjm_display_managed_by_wpcom_notice_for_addon' );
--- a/includes/3rd-party/wpcom.php
+++ b/includes/3rd-party/wpcom.php
@@ -8,9 +8,9 @@
 /**
  * Configure license configuration for WP Job Manager when purchased from WP.com Marketplace.
  *
- * @param bool|WP_Error $result The result of the licensing configuration.
- * @param array $payload        The payload receivced from WPJobManager.com back-end API.
- * @param string $event_type    The event type that triggers this filter.
+ * @param bool|WP_Error $result     The result of the licensing configuration.
+ * @param array         $payload    The payload receivced from WPJobManager.com back-end API.
+ * @param string        $event_type The event type that triggers this filter.
  *
  * @return bool
  */
--- a/includes/3rd-party/wpcom.php
+++ b/includes/3rd-party/wpcom.php
@@ -86,7 +86,7 @@ function wpjm_display_managed_by_wpcom_notice_for_addon( $product_slug ) {
 		return;
 	}
 
-	esc_html_e( 'The license for this product is automatically managed by WordPress.com.', 'wp-job-manager' );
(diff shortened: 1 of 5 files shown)


includes/3rd-party/wpcom.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Configure license configuration for WP Job Manager when purchased from WP.com Marketplace.
+ *
+ * @param bool|WP_Error $result     The result of the licensing configuration.
+ * @param array         $payload    The payload receivced from WPJobManager.com back-end API.
+ * @param string        $event_type The event type that triggers this filter.
+ *
+ * @return bool
+ */
+function dotcom_marketplace_configure_license_for_wp_job_manager_addon( $result, $payload, $event_type ) {
```
mmtr: I think this should be prefixed with `wpjm_`

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -511,7 +511,7 @@ private function handle_request() {
 	 * @param string $licence_key
 	 * @param string $email
 	 */
-	private function activate_licence( $product_slug, $licence_key, $email ) {
+	public function activate_licence( $product_slug, $licence_key, $email ) {
```
mmtr, line 514: Was this method previously private for any particular reason? Just wondering if this change can open up any security breach.

mmtr (commented):
Feels good to see how finally all pieces start fitting together 🙂

includes/3rd-party/wpcom.php
BogdanUngureanu: Sure!

includes/helper/class-wp-job-manager-helper.php
BogdanUngureanu: The functionality was used just inside this helper, but now we'll need it in the context of WPCOM too. I don't see any security issue here. In the end, the code is public and if someone wants to use it, they can simply copy the code and alter it the way they want.

The implementation just makes an API call to wpjobmanager.com and, if it succeeds, saves the information in an option.

mmtr:
> This should return an error.

I got a null value. Is that expected?

Screen Shot 2022-10-13 at 17 07 03

BogdanUngureanu:
> > This should return an error.
> 
> I got a null value. Is that expected?
> 
> Screen Shot 2022-10-13 at 17 07 03

Can you try to run each line separately? Also, do you have the plugin installed on your site? I've run it again on my test site and it worked.

mmtr:
> Can you try to run each line separately?

That was it!

Screen Shot 2022-10-13 at 17 20 26

mmtr (approved):
This looks good to me! It'd be great though if someone else can double check this, since I'm not very familiar with this plugin and I might be missing something!

jom (approved):
Looks like a nice solution 🚢

BogdanUngureanu:
Thanks for the approval! I think we should wait a bit more with merging this branch into main until I find a way to test the entire flow.

I also would like to make another review (from a user standpoint) of how the licensing information is presented to the user from the WPJM plugin. We might need to make some additional UI changes, and releasing another new version is... probably not the best way to do it 😅

includes/3rd-party/wpcom.php
```diff
@@ -53,3 +53,40 @@ function ( $message ) {
...
+ * Display a notice after the license form for each addon.
+ *
+ * @param string $product_slug
+ *
+ * @return false|void
+ */
+function wpjm_display_managed_by_wpcom_notice_for_addon( $product_slug ) {
+	$subscriptions = get_option( 'wpcom_active_subscriptions', [] );
+	if ( ! isset( $subscriptions[ $product_slug ] ) ) {
+		return false;
```
mmtr: Nitpick: This technically doesn't need to return anything.
```suggestion
		return;
```

includes/3rd-party/wpcom.php
```diff
@@ -53,3 +53,40 @@ function ( $message ) {
...
+	}
+
+	esc_html_e( 'The license for this product is automatically managed by WordPress.com.', 'wp-job-manager' );
```
mmtr: Nitpick: I think "add-on" is more inline with the plugin.
```suggestion
	esc_html_e( 'The license for this add-on is automatically managed by WordPress.com.', 'wp-job-manager' );
```

mmtr (approved):
Thanks for adding changes to hide the license information when it is provided by WP.com. It works great!
Screen Shot 2022-10-21 at 13 04 42

jom (approved):
Looks good!

PR #2338, opened 2022-11-15
#2337: Make salary value string translatable

### LINKED ISSUE #2337: Salary Placeholder isn't Translatable
**Describe the bug**

In the job submission form, the placeholder string for salary (`e.g. 20000`) isn't translatable.

**Expected behavior**

- Like other placeholders, this one should be translatable.

**WordPress Environment**
- WordPress Version: v6.1
- WP Job Manager Version: v1.38.1
- PHP Version: v8.1
- Other important details:

bobmatyas:
Fixes #

* Makes salary value placeholder translatable. Fixes #2337 

### Testing instructions

* Use a plugin such as "Say What?" to translate try to translate the placeholder (`e.g. 20000`) on the job posting page.
* It doesn't work before this PR, but works afterwards

#### Before:

Annotation on 2022-11-15 at 15-43-59

#### After:

Annotation on 2022-11-15 at 15-44-35

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -1599,7 +1599,7 @@ class WP_Job_Manager_Post_Types {
 			'_job_salary'          => [
 				'label'         => __( 'Salary', 'wp-job-manager' ),
 				'type'          => 'text',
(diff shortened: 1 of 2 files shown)


jom (approved):
Looks good! Thanks for the PR.

PR #2342, opened 2022-11-25
Download language packs for add-ons

jom:

* Downloads language packs from [WordPress.com's GlotPress](https://translate.wordpress.com/projects/wp-job-manager).
* Passes locales to plugin update check.

### Testing instructions

* Ensure you have at least one WP Job Manager official add-on installed (doesn't have to be activated).
* Switch site over to a language that has language packs available on GlotPress (click on project and see language pack button on far-right).
* Go to WP Admin > Dashboard > Updates.
* Refresh a few times if you don't see translation updates at the bottom.
* Install the translation updates.
* Ensure translations are loaded for that plugin/language.

Other things to check:
- Using Query Monitor, ensure these translations aren't checked unnecessarily and are cached correctly.
- Set the site to a different language first and ensure all languages are downloaded. You can also check `wp-content/languages/plugins`

### New Filters
- `wp_job_manager_helper_disable_language_packs`: Filter to true to disable this new functionality.

### DIFF
--- a/includes/helper/class-wp-job-manager-helper.php
+++ b/includes/helper/class-wp-job-manager-helper.php
@@ -91,6 +91,22 @@ class WP_Job_Manager_Helper {
 	}
 
 	/**
+	 * Get the locales used in the site.
+	 *
+	 * @return string[]
+	 */
+	private function get_site_locales() {
+		$locales = array_values( get_available_languages() );
+
+		/** This action is documented in WordPress core's wp-includes/update.php */
+		// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+		$locales = apply_filters( 'plugins_update_check_locales', $locales );
+		$locales = array_unique( $locales );
+
+		return $locales;
+	}
+
+	/**
 	 * Handles special tasks on admin requests.
 	 */
 	private function handle_admin_request() {
@@ -182,11 +198,13 @@ class WP_Job_Manager_Helper {
 
 		$response = $this->api->plugin_update_check(
 			[
-				'plugin_name'    => $plugin_data['Name'],
-				'version'        => $plugin_data['Version'],
-				'api_product_id' => $product_slug,
-				'licence_key'    => $licence['licence_key'],
-				'email'          => $licence['email'],
+				'plugin_name'       => $plugin_data['Name'],
+				'version'           => $plugin_data['Version'],
+				'api_product_id'    => $product_slug,
+				'licence_key'       => $licence['licence_key'],
+				'email'             => $licence['email'],
+				'locale'            => get_locale(),
+				'available_locales' => implode( ',', $this->get_site_locales() ),
 			]
 		);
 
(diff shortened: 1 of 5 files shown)


includes/helper/class-wp-job-manager-helper.php
```diff
@@ -594,7 +655,7 @@ private function handle_api_errors( $product_slug, $response ) {
 		}
 
 		$errors         = ! empty( $response['errors'] ) ? $response['errors'] : [];
-		$allowed_errors = [ 'no_activation', 'expired_key', 'expiring_soon' ];
+		$allowed_errors = [ 'no_activation', 'expired_key', 'expiring_soon', 'update_available' ];
```
jom, line 649: This will allow us to notify the users if they don't have a license but there is a pending update available. For now, it will just show in the license area, but eventually we could use this to hint on the plugins screen.

Note: This isn't implemented on WPJobManager.com.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -85,11 +93,64 @@ public function init() {
...
+
+		$this->language_pack_helper = new WP_Job_Manager_Helper_Language_Packs( $this->get_plugin_versions(), $this->get_site_locales() );
+	}
+
+	/**
+	 * Get the versions for the installed managed plugins, keyed with the plugin slug.
+	 *
+	 * @return string[]
+	 */
+	public function get_plugin_versions() {
```
renatho: I think this method could be `private` for now, right?

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -85,11 +93,64 @@ public function init() {
...
+
+		return $locales;
+	}
+
+	/**
+	 * Get the language packs helper.
+	 *
+	 * @return WP_Job_Manager_Helper_Language_Packs
+	 */
+	public function get_language_pack_helper() {
```
renatho: Are we using this method?

renatho (approved):
Looks good and works well!

I just added some small observations about the code, but nothing critical.

> ensure these translations aren't checked unnecessarily and are cached correctly.

Just confirming, from this you just wanted to confirm if the transient was working properly to avoid the fetch without changing the versions and locales, right?

includes/helper/class-wp-job-manager-helper.php
jom: I meant to add a review comment. Since this isn't a singleton, I wanted to ensure people were able to remove the filter added in the class by accessing the instance. Alternatively, we could apply a filter in the class that would allow people to disable this functionality.

includes/helper/class-wp-job-manager-helper.php
jom: Changed in cce6c6a24f1d7302c926e668155573af70c70260

includes/helper/class-wp-job-manager-helper.php
jom: For now, I got rid of it and added the filter in d3fd874b8d380147867ecba35c059778907347a5 that makes it a bit easier if someone wanted to disable this functionality.

renatho (approved):
Looks good! 🎉

PR #2356, opened 2022-12-20
Add "required" support on Term select/multi select templates

Gnodesign:

This will add a **`required`** tag on the term select and multiselect templates if the actual field is required.


gikaragia:
Hey @Gnodesign I am not sure I understand why this feature is needed. Currently, we already check for required fields in JS, namely in `specialFieldsAreInvalid` function. Can you specify some use cases where this is useful?

Also the solution won't work for the multi-select template, I have added some changes to the `add/required-support` branch to fix that.

gikaragia:
I'll resolve this until the above questions are answered. Please reopen if we should take any actions about it.

PR #2360, opened 2023-01-09
Use github actions from Automattic org instead of from my personal profile

fjorgemota:

* Use `github-action-report-artifact` action from the Automattic org instead of from my own personal profile;

### Testing instructions

Verify if the action still works fine (with the message `Click on "Details" to download the plugin zip file` on the Checks section below, and also with the Details link redirecting to the ZIP file of the artifact).

### Context

p1673279045277219-slack-C8QJAM52B


jom (approved):
Looks good! Thanks for making this change.

PR #2362, opened 2023-01-12
Add check for job listing limit being 0

### LINKED ISSUE #2333: Listing Limit Option Not Working As Expected
**Describe the bug**

The job listing limit doesn't work as expected. If you enter 0, you are allowed to enter an unlimited number of jobs. This should either not allow you to enter any jobs or the field description should state that 0 means unlimited jobs are allowed.

**To Reproduce**
1. Add more than 2 test jobs.
2. Go to 'Settings > Job Submission'
3. Add a "Listing Limit" of 0.
4. Try adding another job, you are allowed to add an additional job.
5. Set the listing limit to 1.
6. Try adding another job. You should be redirected to the Jobs dashboard.
7. Remove the "Listing Limit" entirely -- you can add jobs.

**Expected behavior**

- I would expect that if I enter "0", I wouldn't be able to add additional jobs.

**WordPress Environment**
- WordPress Version:  v6.0.3
- WP Job Manager Version: v1.3.8
- PHP Version:  v8.1
- Other important details:

mikeyarce:
Fixes #2333

* Add a check for job listing limit being `0`.  Currently if it's set to `0` it works the same as it being unlimited, which I don't think is intended.

### Testing instructions

* Create a few job listings
* Set the limit to 0
* Verify that you can't add more job listings

### DIFF
--- a/templates/job-dashboard.php
+++ b/templates/job-dashboard.php
@@ -87,7 +87,7 @@ $submit_job_form_page_id	= get_option( 'job_manager_submit_job_form_page_id' );
 				<?php endforeach; ?>
 			<?php endif; ?>
 		</tbody>
-		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || 'false' === $submission_limit ) ) : ?>
+		<?php if ( job_manager_user_can_submit_job_listing() ) : ?>
 			<tfoot>
 				<tr>
 					<td colspan="<?php echo count( $job_dashboard_columns ); ?>">
--- a/templates/job-dashboard.php
+++ b/templates/job-dashboard.php
@@ -87,7 +87,7 @@ $submit_job_form_page_id	= get_option( 'job_manager_submit_job_form_page_id' );
 				<?php endforeach; ?>
 			<?php endif; ?>
 		</tbody>
-		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || ! $submission_limit ) ) : ?>
+		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || ! $submission_limit ) && '0' !== $submission_limit ) : ?>
 			<tfoot>
 				<tr>
 					<td colspan="<?php echo count( $job_dashboard_columns ); ?>">
--- a/templates/job-dashboard.php
+++ b/templates/job-dashboard.php
@@ -87,14 +87,7 @@ $submit_job_form_page_id	= get_option( 'job_manager_submit_job_form_page_id' );
 				<?php endforeach; ?>
 			<?php endif; ?>
 		</tbody>
-		<?php
-
-var_dump($submit_job_form_page_id );
-var_dump(  job_manager_count_user_job_listings() < $submission_limit );
-var_dump( $submission_limit );
-
-		?>
-		<?php if ( $submit_job_form_page_id && job_manager_count_user_job_listings() < $submission_limit && false !== $submission_limit ) : ?>
+		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || 'false' === $submission_limit ) ) : ?>
 			<tfoot>
 				<tr>
 					<td colspan="<?php echo count( $job_dashboard_columns ); ?>">
--- a/templates/job-dashboard.php
+++ b/templates/job-dashboard.php
@@ -23,7 +23,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
 
-$submission_limit			= get_option( 'job_manager_submission_limit' );
+$submission_limit			= ! empty( get_option( 'job_manager_submission_limit' ) ) ? absint( get_option( 'job_manager_submission_limit' ) ) : false;
 $submit_job_form_page_id	= get_option( 'job_manager_submit_job_form_page_id' );
 ?>
 <div id="job-manager-job-dashboard">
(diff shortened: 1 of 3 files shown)


templates/job-dashboard.php
```diff
@@ -87,7 +87,7 @@
 				<?php endforeach; ?>
 			<?php endif; ?>
 		</tbody>
-		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || ! $submission_limit ) ) : ?>
+		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || ! $submission_limit ) && '0' !== $submission_limit ) : ?>
```
yscik: Could it be clearer to cast `$submission_limit` via `absint` if it's not empty and set it to `false` otherwise? With that the condition can be just `( job_manager_count_user_job_listings() < $submission_limit || false === $submission_limit`.

yscik (commented):
Looks like this option is also used here where the same change should happen.
https://github.com/Automattic/WP-Job-Manager/blob/7411e91b491475e841a65c3b13b31132727c69f3/includes/class-wp-job-manager-shortcodes.php#L110-L114

And is there a similar check on the actual job submission page and endpoint?

templates/job-dashboard.php
mikeyarce: I updated the code to add these checks, let me know what you think!

templates/job-dashboard.php
```diff
@@ -87,7 +87,14 @@
 				<?php endforeach; ?>
 			<?php endif; ?>
 		</tbody>
-		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || ! $submission_limit ) ) : ?>
+		<?php
+
+var_dump($submit_job_form_page_id );
+var_dump(  job_manager_count_user_job_listings() < $submission_limit );
+var_dump( $submission_limit );
```
yscik: There is an oopsie here :)

templates/job-dashboard.php
```diff
@@ -87,7 +87,14 @@
...
+
+		?>
+		<?php if ( $submit_job_form_page_id && job_manager_count_user_job_listings() < $submission_limit && false !== $submission_limit ) : ?>
```
yscik: I think the logic is not right? It should be
- if there is a `$submit_job_form_page_id`
- AND ( the job count is smaller than the limit, OR there is no limit )

Not sure if you saw my other comment:

> Looks like this option is also used here where the same change should happen.
https://github.com/Automattic/WP-Job-Manager/blob/7411e91b491475e841a65c3b13b31132727c69f3/includes/class-wp-job-manager-shortcodes.php#L110-L114
> 
> And is there a similar check on the actual job submission page and endpoint?

mikeyarce:
@yscik - I tried the logic you are suggesting but it won't work if the Listing Limit is `0`.

The logic I have now is:
- Is this the job submit form page?
- AND ( Are we below the limit OR is there a submission limit in place?)

### Use Cases
No Limit
- `$submission_limit` will be `false` 
- `job_manager_count_user_job_listings() < $submission_limit ` will be false

0 Limit
- `$submission_limit` will be `false`
- `job_manager_count_user_job_listings() < $submission_limit` will be false

So in both cases we have both false, however they need to work in the opposite way.  

I thought about adding this 
```
if ( '0' == get_option( 'job_manager_submission_limit' ) ) {
	$submission_limit = true;
}
```

But, it seemed cleaner to just have a `false === $submission_limit` check.  What do you think?

templates/job-dashboard.php
```diff
@@ -23,7 +23,7 @@
 	exit; // Exit if accessed directly.
 }
 
-$submission_limit			= get_option( 'job_manager_submission_limit' );
+$submission_limit			= ! empty( get_option( 'job_manager_submission_limit' ) ) ? absint( get_option( 'job_manager_submission_limit' ) ) : false;
```
yscik, line 26: `empty(0)` and `empty('0')` is also `true`. I think we need `'' !== get_option( 'job_manager_submission_limit', '' )` to test if it doesn't have any value set.

The `if` below checks for `'false'` (string) instead of the `false` (bool) set here.

templates/job-dashboard.php
```diff
@@ -87,7 +87,7 @@
 				<?php endforeach; ?>
 			<?php endif; ?>
 		</tbody>
-		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || ! $submission_limit ) ) : ?>
+		<?php if ( $submit_job_form_page_id && ( job_manager_count_user_job_listings() < $submission_limit || 'false' === $submission_limit ) ) : ?>
 			<tfoot>
```
yscik, line 91: There are other places in the plugin where this check needs to be updated. Might make sense to add a new `job_manager_can_user_submit_job()` to handle it consistently everywhere.

https://github.com/Automattic/WP-Job-Manager/blob/7411e91b491475e841a65c3b13b31132727c69f3/includes/class-wp-job-manager-shortcodes.php#L110-L114

templates/job-dashboard.php
mikeyarce: Fixed in https://github.com/Automattic/WP-Job-Manager/pull/2362/commits/a8d3b34116c82318f4cdefea0633f6aa0a8c2138

mikeyarce:
@yscik Thanks for the suggestions!  I made a helper function `job_manager_user_can_submit_job_listing` to address this.  I also added the check into the shortcode in e7b5f42b2dfda35ad76d4b58e7a244f8edbee492

wp-job-manager-functions.php
```diff
@@ -1707,3 +1707,29 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+ *
+ * @return bool
+ * @since $$next-version$$
+ */
+function job_manager_user_can_submit_job_listing() {
+	$can_submit       = true;
+	$submission_limit = get_option( 'job_manager_submission_limit', '' );
+	$job_count        = job_manager_count_user_job_listings();
+
+	if ( '' !== $submission_limit &&
```
gikaragia: You can set the $can_submit variable only once with:

```
	$can_submit = '' === $submission_limit || $submission_limit > $job_count;
```

gikaragia (commented):
Suggested a small improvement. Other than that it looks good!

wp-job-manager-functions.php
mikeyarce: Thanks @gkaragia !  Done in https://github.com/Automattic/WP-Job-Manager/pull/2362/commits/cc64267ca8d43a68a744acdba11890cc6604017a

gikaragia (approved):
Looks good!

PR #2363, opened 2023-01-13
Fix "Hide Expired Listings Content" setting

### LINKED ISSUE #2359: Job visibility feature bypassing setting expired listing setting
**Describe the bug**

When unchecking "Hide Expired Listings Content" setting, everyone should see the job content on the job page, but it's not happening because `job_manager_user_can_view_job_listing` function is bypassing this setting.

**To Reproduce**
1. Go to wp-admin > Job Listings > Settings.
2. Make sure the setting "Hide Expired Listings Content" is unchecked.
3. Add a new job.
4. Edit the job to expire in the past, so you force that the job is expired.
5. Log in with a different user (or logged out).
6. Go to the job page.
7. Check that you see a message that the job is expired and you can't see the job details.

**Screenshots**

Screen Shot 2023-01-06 at 15 33 35

Screenshot 2023-01-06 at 19 33 15

**Expected behavior**

Since the setting "Hide Expired Listings Content" is unchecked, we should see the content.

**WordPress Environment**
- WordPress Version: WP 6.1
- WP Job Manager Version: 1.39.0
- PHP Version: 7.4
- Original report: p1673031316220179-slack-C6GGX896G

**Workaround**

If not using the Job Visibility feature, it can be bypassed with the following snippet:

```php
add_filter( 'job_manager_user_can_view_job', '__return_true' );
```

renatho:
Fixes #2359

* It fixes the "Hide Expired Listings Content" setting by allowing the template to handle expired jobs. [See more details about the issue](https://github.com/Automattic/WP-Job-Manager/issues/2359#issuecomment-1374225762).

### Testing instructions

1. Go to wp-admin > Job Listings > Settings.
2. Make sure the setting "Hide Expired Listings Content" is unchecked.
3. Add a new job.
4. Edit the job to expire in the past, so you force that the job is expired.
5. Log in with a different user (or logged out).
6. Go to the job page.
7. Check that you see the job details.

### DIFF
--- a/templates/access-denied-single-job_listing.php
+++ b/templates/access-denied-single-job_listing.php
@@ -15,9 +15,6 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
+?>
 
-if ( $post->post_status === 'expired' ) : ?>
-	<div class="job-manager-info"><?php _e( 'This listing has expired', 'wp-job-manager' ); ?></div>
-<?php else : ?>
-	<p class="job-manager-error"><?php _e( 'Sorry, you do not have permission to view this job listing.', 'wp-job-manager' ); ?></p>
-<?php endif; ?>
+<p class="job-manager-error"><?php _e( 'Sorry, you do not have permission to view this job listing.', 'wp-job-manager' ); ?></p>
--- a/templates/access-denied-single-job_listing.php
+++ b/templates/access-denied-single-job_listing.php
@@ -9,7 +9,7 @@
  * @package     wp-job-manager
  * @category    Template
  * @since       1.37.0
- * @version     1.37.0
+ * @version     1.39.1
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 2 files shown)


renatho:
Hi @Gnodesign! 👋

We received a report that the "Hide Expired Listings Content" setting wasn't working properly. I fixed it by allowing the template to handle the expiration. It means that if the user shouldn't have access through the `job_manager_user_can_view_job_listing` check, it will always show the message "Sorry, you do not have permission to view this job listing.", otherwise, it will allow the job listing template (`templates/content-single-job_listing.php`) to handle the expiration check or show the job details.

If you have the opportunity to check this PR, I want to confirm with you if I didn't miss anything from [Job Visibility feature](https://github.com/Automattic/WP-Job-Manager/pull/2280), breaking something.

Gnodesign:
Hi @renatho,

Sorry for the late response been unavailable these days and couldn't test the issue https://github.com/Automattic/WP-Job-Manager/issues/2359 you opened.

The visibility feature for WPJM was handled the same way as in the Resume Manager. Since the RM doesn't have a **Hide Expired Listings Content** option I have overlooked this part. I run some tests and it works fine for me.

renatho:
Nice! Thank you for checking @Gnodesign!

m1r0 (approved):
Works as described. 👍

PR #2367, opened 2023-02-17
Update language; hide expired job listings from archives

### LINKED ISSUE #1884: WordPress search returns expired job
Reported in 2331742-zen

If we look at https://github.com/Automattic/WP-Job-Manager/pull/1570 , we should not be seeing Filled and Expired jobs in the WordPress search any longer. However, they do appear now (with WPJM 1.33.5).

Aside: how can we exclude jobs entirely from WordPress search, without breaking the plugin's search capability? Is that possible?

interlineae:
Fixes #1884

* Updates language between the 'Filled' and 'Expired' options to be more consistent.
* Excludes expired and filled jobs from WP Search and Jobs' archives, widgets, and feed.

Search exclusions are built off of the changes in reverted [PR 1570](https://github.com/Automattic/WP-Job-Manager/pull/1570). In this instance, additional `meta_query` checks were added to ensure that specified meta were not only the checked value, but that the meta existed at all.

This latter check, missing originally, seems to be why 'normal' posts and pages were excluded from search. With the changes made here, that content returns in search as expected, even if filled and hidden posts are hidden.

### Testing instructions

* Check out the branch.
* Navigate to **Job Listings --> Settings --> Job Listings** and enable or disable the `Filled Positions` and `Expired Positions` options, checking WP Search, Sitemap, and widgets like the 'Recent Jobs' widget.


includes/class-wp-job-manager-post-types.php
```diff
@@ -651,6 +662,71 @@ public function job_feed() {
...
+					'value'   => '0',
+					'compare' => '=',
+				),
+				array(
+					'key'     => '_filled',
+					'compare' => 'NOT EXISTS',
+				),
+			);
+
+			if ( ! empty( $meta_query ) ) {
```
jom: `! empty( $meta_query )` will always be true at this point (same below in the expiration block) because of the addition to the array above.

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -153,26 +153,26 @@ protected function init_settings() {
...
 						],
 						[
 							'name'       => 'job_manager_hide_expired_content',
 							'std'        => '1',
-							'label'      => __( 'Hide Expired Listings Content', 'wp-job-manager' ),
-							'cb_label'   => __( 'Hide content in expired single job listings', 'wp-job-manager' ),
-							'desc'       => __( 'Your site will display the titles of expired listings, but not the content of the listings. Otherwise, expired listings display their full content minus the application area.', 'wp-job-manager' ),
+							'label'      => __( 'Hide Expired Positions Content', 'wp-job-manager' ),
+							'cb_label'   => __( 'Hide content in expired single job position', 'wp-job-manager' ),
+							'desc'       => __( 'Your site will display the titles of expired position, but not the content of the positions Otherwise, expired positions display their full content minus the application area.', 'wp-job-manager' ),
```
aaronfc: Some plural is wrong (I think) and there is a dot missing.

```suggestion
							'desc'       => __( 'Your site will display the titles of expired positions, but not the content of the positions. Otherwise, expired positions display their full content minus the application area.', 'wp-job-manager' ),
```

includes/class-wp-job-manager-post-types.php
```diff
@@ -651,6 +662,71 @@ public function job_feed() {
 		remove_filter( 'posts_search', 'get_job_listings_keyword_search' );
 	}
 
+	/**
+	 * Modifies WordPress Query of public search.
+	 *
+	 * @param WP_Query $query Query being processed.
+	 */
+	public function maybe_hide_filled_expired_job_listings_from_search( Object $query ): void {
```
aaronfc: This `Object` parameter type is a little bit odd in here it should be `WP_Query` **but** we [recently discovered](https://senseip2.wordpress.com/2023/02/21/sensei-pro-1-11-0-post-mortem/) that typing the parameters for hook callbacks might give problems.

So better removing the type completely.

includes/class-wp-job-manager-post-types.php
```diff
@@ -1373,6 +1449,40 @@ public function sitemaps_maybe_hide_filled( $query_args, $post_type ) {
...
+	 *
+	 * @access private
+	 * @since 1.39.0
+	 *
+	 * @param array  $query_args Array of WP_Query arguments.
+	 * @param string $post_type  Post type name.
+	 *
+	 * @return array
+	 */
+	public function sitemaps_maybe_hide_expired( Array $query_args, String $post_type ): array {
```
aaronfc: Not sure where is this method being used. Is this some leftover code?

aaronfc (commented):
Looking good! Left a couple of comments and I couldn't test it. If @jom hasn't tested it by tomorrow I will give it a go.

includes/class-wp-job-manager-post-types.php
```diff
@@ -609,6 +610,16 @@ public function job_feed() {
 			];
 		}
 
+		// Hide expired positions from the job feed.
```
jom: The hourly scheduled job that runs should change the post status for these to `expired`, which should take care of this path and the change in `get_job_listings()`. Were you able to reproduce an issue where these weren't hidden?

includes/class-wp-job-manager-post-types.php
```diff
@@ -651,6 +662,71 @@ public function job_feed() {
...
+				'relation' => 'OR',
+				array(
+					'key'     => '_filled',
+					'value'   => '0',
+					'compare' => '=',
+				),
+				array(
+					'key'     => '_filled',
+					'compare' => 'NOT EXISTS',
+				),
```
jom: Nice catch with these `NOT EXISTS` additions. Seems to work nicely.

jom (commented):
Everything seems to be working well! Nice job investigating this. I've left a few minor comments, in addition to @aaronfc's review. 

There are a few PHPCS issues (`npm run lint-php`). Most can probably be fixed with `./vendor/bin/phpcbf`.

includes/admin/class-wp-job-manager-settings.php
interlineae: Good catch, yep. Committing your suggested edits.

includes/class-wp-job-manager-post-types.php
interlineae: Ah, I see that now. Will remove it as unneeded 👍🏻 .

includes/class-wp-job-manager-post-types.php
interlineae: Ah, yes, I realised that expired posts are already excluded from sitemaps, so I will remove this.

interlineae:
@jom @aaronfc The above feedback has been addressed in commit 31306e0. Going forward, I'll aim to make my commits more atomic. 

Edit: there was an update to `package-lock.json` included as well. I thought I had excluded it but apparently not. What's usually done with these lock file updates?

package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "wp-job-manager",
-      "version": "1.38.0",
+      "version": "1.39.0",
```
jom, line 9: Review note: @anaemnesis and I talked about this and are sneaking it into this PR.

includes/class-wp-job-manager-post-types.php
jom: What do you think about the check added in `get_job_listings()`?

interlineae: I'm wondering if I'm following along as I should be, but here we go.

 I added the check in `get_job_listings()` as originally, the **Hide expired positions** option was _not_ hiding expired posts from the Recent Jobs widget. Or at least, that's the behaviour I was seeing at first.

When I was working on the check, I had added to the `$args` array with:

```
'featured'          => null,
'filled'            => null,
'expired'           => null,
```

And then checked with:

```
! is_null( $args['expired'] ) 
```

I had also thought about this:

```
'expired' === ($args['post_status'] )
```

But I decided that the second check wasn't great, or least as an OR, as it would simply hide `expired` posts and never make it to the option check (although that seems to be the behaviour anyway), e.g:

```
if ( 'expired' === ($args['post_status'] ) || 1 === absint( get_option( 'job_manager_hide_expired' ) ) ) {
```

In my testing, yesterday and today, however, expired posts _are_ being excluded from the Recent Jobs widget, even with the check and related `meta_query` manipulation commented out. So I don't know why I was seeing the behaviour I was originally, but I'm unable to repeat it. 

**But also**

The check doesn't work as I'd expected. If the comparison is reversed, all jobs are hidden from the Recent Jobs widget, not just those that are expired (i.e. have a past date). It seems to be excluding based simply on the existence of the `_job_expires` key, and so isn't properly taking (1) the key and doing (2) a date comparison (i.e., if the value date is less than today's date, the post is expired). What seems to be happening, is that the value is being set to today's date... but I'm unclear on that, mostly.

It seems though that removing the check is fine, as I'm not able to replicate any issues, and the Recent Jobs widget behaves as expected with the option to hide expired posts enabled.

jom: Sorry, I'm a little lost about which checks you're referring to here. The [post status option check](https://github.com/Automattic/wp-job-manager/blob/c2ef474d5b01ca51e1567dc7fe434870d782ee4d/wp-job-manager-functions.php#L49-L55) should catch the jobs already switched to expired (manually or by the job). 

I'm not sure what was happening with expired jobs showing up in your Recent Jobs widget. Results do get cached, so I'm not sure if that was related. 

> The check doesn't work as I'd expected. If the comparison is reversed, all jobs are hidden from the Recent Jobs widget, not just those that are expired (i.e. have a past date). It seems to be excluding based simply on the existence of the _job_expires key

Which comparison are you reversing here?

interlineae: > The [post status option check](https://github.com/Automattic/wp-job-manager/blob/c2ef474d5b01ca51e1567dc7fe434870d782ee4d/wp-job-manager-functions.php#L49-L55) should catch the jobs already switched to expired (manually or by the job).

That check works as expected for me. Thinking out loud --

~~Jobs listings that are beyond their expiration date are correctly given an `expired` post status, or `publish` post status if the expiration date changes to a new future date. Posts are then shown, or hidden, from areas like the Recent Jobs widget based on that status alone.~~

Actually, seeing the `post_status` value change from `publish` to `expired` is handled elsewhere. [There is the `check_for_expired_jobs()` method](https://github.com/Automattic/WP-Job-Manager/blob/da58379b8798022267ca19ad9b5615517fc915dd/includes/class-wp-job-manager-post-types.php#L721), as well as the [`save_job_listing_data()` method](https://github.com/Automattic/WP-Job-Manager/blob/afb358e3b3bd81bdc6bd57a219dea8c6799c9726/includes/admin/class-wp-job-manager-writepanels.php#L673).

The [`job_manager_check_for_expired_jobs` cron job](https://github.com/Automattic/WP-Job-Manager/blob/caacfa7cdb1cfb7766e86828655930c6f12d2189/includes/class-wp-job-manager.php#L216) uses `check_for_expired_jobs()` to set the post status to `expired` when run, and that's [here](https://github.com/Automattic/WP-Job-Manager/blob/c7c4c3f168ee185ae1774da2615f1d0243e7228e/includes/class-wp-job-manager-post-types.php#L58). Otherwise, `check_for_expired_jobs()` handles the post status on its own via `wp_update_post()`.

`get_job_listings()` is setting `$post_status` to retrieve posts either with a `publish` or `expired` value, or just a `publish` value. _Ohhh_, that's exactly it. If the `job_manager_hide_expired` option is enabled, it retrieves only posts with a `publish` `post_status`, otherwise, it will retrieve both `publish` and `expired` `post_status` posts. 

> Results do get cached, so I'm not sure if that was related.

I think that's what was happening.

> Which comparison are you reversing here?

Sorry, [the comparison here](https://github.com/Automattic/WP-Job-Manager/blob/c2ef474d5b01ca51e1567dc7fe434870d782ee4d/wp-job-manager-functions.php#L143). If I switch `compare` to check for less than/equal to `date`, then _all_ Job listings are removed from the Recent Jobs widget, not just those with a `_job_expires` meta key with date value in the past. I think what would have been needed is a date comparison like the one exampled [here](https://github.com/Automattic/WP-Job-Manager/blob/da58379b8798022267ca19ad9b5615517fc915dd/includes/class-wp-job-manager-post-types.php#L743).

jom: > If I switch compare to check for less than/equal to date, then all Job listings are removed from the Recent Jobs widget, not just those with a _job_expires meta key with date value in the past.

These greater-than and less-than always make my brain melt when in the `meta_value` query syntax, but could it be that reversing the query is asking for job listings that expired in the past **and** still have a `publish` status (from the earlier `post_status` check that you mentioned above)?

jom (approved):
This looks great! Nice job investigating. 🚢 🚢 🚢

PR #2380, opened 2023-03-20
Add/email notification settings

mikeyarce:

* Currently in WPJM, you can have emails be sent to the main Admin account.  This is not always ideal as the Admin account might be someone who doesn't deal with jobs.  
* This PR adds a new setting for each "Admin" Email notification so you can customize the email address it sends to.
* If no email is set, it will default to the Admin email.

### Testing instructions

* Check out this PR
* Go to Job Listings > Settings > Email Notifications
* Enter email addresses for the 3 new settings (New Job, Updated Job, Expired Job)
* Use a plugin like Email Log to capture what emails are sent
* Trigger each email
* Ensure you see the emails sent to the new email addresses

### Screenshot / Video

Screenshot 2023-03-20 at 1 34 28 PM


includes/class-wp-job-manager-email-notifications.php
```diff
@@ -636,6 +636,18 @@ private static function get_email_setting_fields( $email_notification_key ) {
 				],
 			],
 		];
+		if ( 'employer_expiring_job' !== $email_notification_key ) {
```
gikaragia, line 640: I think in general that we should add empty lines before 'if's to make them a bit easier to read.

includes/emails/class-wp-job-manager-email-admin-updated-job.php
```diff
@@ -78,7 +78,8 @@ public function get_from() {
 	 * @return string|array
 	 */
 	public function get_to() {
-		return get_option( 'admin_email', false );
+		$option = get_option( 'job_manager_email_admin_updated_job', false );
+		return ! $option['email_to'] ? get_option( 'admin_email', false ) : $option['email_to'];
```
gikaragia: You can use the elvis operator to simplify these:
```suggestion
		return get_option( 'job_manager_email_admin_updated_job', false ) ?: get_option( 'admin_email', false );
```

You need to be a bit careful as the check will fail for all 'falsy' values but I think this is ok in this case.

tests/php/tests/includes/test_class.wp-job-manager-email-notifications.php
```diff
@@ -425,10 +425,20 @@ public function test_add_email_settings() {
 			$email_key                = $email_keys[ $key ];
 			$email_settings           = call_user_func( [ $email_class, 'get_setting_fields' ] );
 			$email_is_default_enabled = call_user_func( [ $email_class, 'is_default_enabled' ] );
-			$defaults                 = [
-				'enabled'    => $email_is_default_enabled ? '1' : '0',
-				'plain_text' => '0',
-			];
+			if ( 'job_manager_email_employer_expiring_job' === $setting['name'] ) {
+				$defaults                 = [
```
gikaragia: You can write this like this:
```
			$defaults                 = [
				'enabled'    => $email_is_default_enabled ? '1' : '0',
				'plain_text' => '0',
			];
			$email_settings_count = count( $email_settings ) + 1;

			if ( in_array( $setting['name'], [ 'job_manager_email_admin_updated_job', 'job_manager_email_admin_new_job', 'job_manager_email_admin_expiring_job'] ) ) {
				$email_settings_count++;
				$defaults['email_to'] = '';
			}
```

This makes it a bit more robust as it covers future changes of the settings in the test and makes it a bit easier to read as an 'if' is always easier to read than an 'if/else'

gikaragia (dismissed):
In my environment the settings do not look nice as in your screenshot. Any ideas in what is the issue?
Screenshot 2023-03-31 at 13 08 34

gikaragia (commented):
Sorry approved by mistake

includes/emails/class-wp-job-manager-email-admin-updated-job.php
mikeyarce: Good suggestion!  It looks like our PHPCS rules don't allow this though? What are your thoughts?  
```
 81 | ERROR | Using short ternaries is not allowed (WordPress.PHP.DisallowShortTernary.Found)
```

mikeyarce:
Thanks @gkaragia !

I've addressed your feedback in a few commits.  I had to update the test logic a bit more because of the order of the items in the array.  I also left a note about PHPCS complaining about short ternaries.

> In my environment the settings do not look nice as in your screenshot. Any ideas in what is the issue?

Sometimes this happens if you haven't run the build scripts yet?

includes/emails/class-wp-job-manager-email-admin-updated-job.php
gikaragia: Ah yes forgot about this one. In general we could add a comment to ignore the rule but I don't think we should it in this case. Can we at least switch the order to avoid the negation?
```
$option['email_to'] ?  $option['email_to'] : get_option( 'admin_email', false );
```

gikaragia (commented):
> Sometimes this happens if you haven't run the build scripts yet?

Yeah I tried this a couple of times but didn't work still. No worries, probably something with my environment I'll retest once the PR gets merged.

There are a few tests failing.

includes/emails/class-wp-job-manager-email-admin-updated-job.php
mikeyarce: Done in 6e6bc581ed0f4972abb1f6eb7d4ab5881049302f

mikeyarce:
Thanks @gkaragia !  Locally the tests are not failing for me, but it looks like there might be a problem with some of our tests not being able to run.

gikaragia:
Ah right. Yeah it wasn't your PR that caused this probably. I raised a PR to fix the tests: https://github.com/Automattic/WP-Job-Manager/pull/2389

Is it possible to review and merge the PR and then rebase to make sure that the tests run succesfully?

tests/php/tests/includes/test_class.wp-job-manager-email-notifications.php
```diff
@@ -429,10 +429,15 @@ public function test_add_email_settings() {
 				'enabled'    => $email_is_default_enabled ? '1' : '0',
 				'plain_text' => '0',
 			];
+			$email_settings_count = count( $email_settings ) + 1;
+
+			if ( in_array( $setting['name'], [ 'job_manager_email_admin_updated_job', 'job_manager_email_admin_new_job', 'job_manager_email_admin_expiring_job', 'job_manager_email_valid_email' ] ) ) {
+				$email_settings_count++;
+				$defaults = array_merge( array_splice( $defaults, 0, 1 ), [ 'email_to' => '' ], $defaults );
```
gikaragia, line 436: I am not sure what this line tries to achieve here. Can we write this as
```
$defaults['email_to'] = '';
```

gikaragia (commented):
Hey @mikeyarce I rebased your branch with trunk to be able to run the tests.

includes/emails/class-wp-job-manager-email-admin-new-job.php
```diff
@@ -78,7 +78,8 @@ public function get_from() {
 	 * @return string|array
 	 */
 	public function get_to() {
-		return get_option( 'admin_email', false );
+		$option = get_option( 'job_manager_email_admin_new_job', false );
+		return ! $option['email_to'] ? get_option( 'admin_email', false ) : $option['email_to'];
```
gikaragia: Can we update similarly this line too?

includes/emails/class-wp-job-manager-email-admin-expiring-job.php
```diff
@@ -50,7 +50,8 @@ public static function get_description() {
 	 * @return string|array
 	 */
 	public function get_to() {
-		return get_option( 'admin_email', false );
+		$option = get_option( 'job_manager_email_admin_expiring_job', false );
+		return ! $option['email_to'] ? get_option( 'admin_email', false ) : $option['email_to'];
```
gikaragia: Same here

tests/php/tests/includes/test_class.wp-job-manager-email-notifications.php
mikeyarce: Good question!  It has to be in a specific order for the PHPUnit tests to not fail because the unit tests expect the arrays to be exactly the same - so this is just my way of making the arrays identical.

gikaragia (approved):

Issue #2370, opened 2023-03-02
Select2 overlap over fields in Job Submission form


mikeyarce:
**Describe the bug**
When you go submit a Job, the Select2 multi-select form overlaps the field when going below and has too much margin when going above.

Below:
![tJcXDP.png](https://user-images.githubusercontent.com/3220162/222562710-f8a50ced-f62f-4088-a417-bca558e32ec5.png)

Above:
![VFL3eN.png](https://user-images.githubusercontent.com/3220162/222563543-ef67eefd-0334-457e-8ce8-faa245127c26.png)

I tried with about 5 different themes and it always does this regardless of what theme it is.

**To Reproduce**
1. Go to the Job Submission page
2. Click on Job Type
3. See the dropdown overlap the input field

**Edit:  This only happens when logged in and you have the Admin bar present.**

tripflex:
Are you logged in as admin with admin bar showing?

mikeyarce:
@tripflex Yup!  Good catch, I'll edit the report to mention it only happens when the Admin bar is present.

tripflex:
https://gist.github.com/tripflex/7f4a10fb27c7e8c95c86010f93ac250b

Yup been an issue for a while also seems to depend on the theme as well

gikaragia:
This is now fixed

PR #2381, opened 2023-03-23
Fix Select2 Padding for Frontend Job Listing

mikeyarce:
Fixes #https://github.com/Automattic/WP-Job-Manager/issues/2370

* Add some padding for the Select2 inputs when the WP Admin Bar is present

### Testing instructions

* Apply this Patch
* Go to enter a Job from the frontend
* Make sure you are logged in and the Admin Bar is present
* Try the Job Type field, make sure the padding is correct
* Try with many different themes
* Do the same on a screen below 782 pixels wide

### Screenshot / Video
Before
Screenshot 2023-03-23 at 2 36 56 PM

After
Screenshot 2023-03-23 at 2 36 49 PM

### DIFF
--- a/assets/css/frontend.scss
+++ b/assets/css/frontend.scss
@@ -979,3 +979,14 @@ nav.job-manager-pagination,
 		}
 	}
 }
+
+// Admin Bar Select2 overlap fixes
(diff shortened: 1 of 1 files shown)


gikaragia (approved):
Looks good!

Issue #2382, opened 2023-03-24
Modernise design for admin notices.


aaronfc:
Hello @aaronfc and friends 👋 I took a look at this and I was wondering if we could make them look similar to the Sensei ones, so we get some consistency going. 

Here's what I have for you:
![image](https://user-images.githubusercontent.com/10038976/227241821-4994c5e3-4445-417f-8a72-dc8880bc4986.png)

![image](https://user-images.githubusercontent.com/10038976/227242218-857a23c6-ef11-4ec3-8377-956484aa7cdd.png)

![image](https://user-images.githubusercontent.com/10038976/227242259-c21214b1-82c3-47d4-80de-853c06305344.png)

Not sure if I missed anything, so do let me know. Figma file is [here](NfoFCYVLE46Wu8mEyLQhm4-fi-20018242%3A334&t=3WXeJp4h2PPPBgUg-1). 

@burtrw

_Originally posted by @Luchadores in https://github.com/Automattic/WP-Job-Manager/issues/2379#issuecomment-1481336928_

**Success Criteria**:
- [ ] All different styles are supported for the wpjm admin notices.
- [ ] All existing notices have been migrated to the new design.
- [ ] All existing notices are dismissible – including the one about [addon updates](https://github.com/Automattic/WP-Job-Manager/issues/2379).

aaronfc:
Related PR: https://github.com/Automattic/WP-Job-Manager/pull/2383

PR #2383, opened 2023-03-24
Add core addon outdated notices.

### LINKED ISSUE #2379: Notices for out of date WPJM addons.
We want to add new notices when WPJM addons are out of date.

We would like those notices to be consolidated (if multiple) since there are users with many addons at the same time.

**Success Criteria**:
- [x] When a user has an out of date addon gets a notification.
- [x] When shown multiple notifications they consolidate in a good and friendly way.
- [x] We show an "Update" action pointing to plugins page ~(licensed addons) or licensing page (unlicensed addons)~

aaronfc:
Fixes #2379

* Add ground for new notifications (will be completed in #2382).
* Save pending updates info in a transient for later use and save notice.
* Render correct messages depending on the amount of addons pending to update.

### Testing instructions
* Checkout to the branch.
* Purchase a paid plugin from wpjobmanager.com (you can use a promo code).
* Download **an outdated** copy of your plugin (you can use the one from the private repositories). 
* Go to `Job Listings > Add-ons > Licenses` and activate your license.
* Go to `Job Listings > All Jobs` or some other WPJM screen and you should see the notice to update.
* Repeat the steps with a different plugin and confirm the multi-plugin case.

### New/Updated Hooks
* `job_manager_admin_notice_addon_update_available`: New action that is triggered to display the notice. 

### Known bugs
- The update action brings you always to the plugins page – in the original ticket we wanted to bring you to plugins page or licensing page depending on if the plugin was licensed or unlicensed. If we want to have this behaviour **we need to decide** what to do on multiple plugins case when some are licensed and some are unlicensed. **Opinions?**
- The design is a _little bit different_ than the one reported in the original ticket, but we have a separate ticket to polish design (icons, dismissible, etc).
- The notice is not dismissible.

#### Single
image

#### Multiple
image

### DIFF
--- a/includes/admin/class-wp-job-manager-admin-notices.php
+++ b/includes/admin/class-wp-job-manager-admin-notices.php
@@ -215,7 +215,7 @@ class WP_Job_Manager_Admin_Notices {
 	}
 
 	/**
-	 * Displays the setup wizard notice when WPJM is first activated.
+	 * Displays the notice that informs about WPJM addon updates available.
 	 *
 	 * Note: For internal use only. Do not call manually.
 	 */
--- a/includes/admin/class-wp-job-manager-admin-notices.php
+++ b/includes/admin/class-wp-job-manager-admin-notices.php
@@ -260,7 +260,8 @@ class WP_Job_Manager_Admin_Notices {
 
 	/**
 	 * Given an array of updates, generate a notice.
-	 * @param $updates
+	 *
+	 * @param array $updates The contents of the addon updates transient.
 	 * @return array|null
 	 */
 	private static function generate_notice_from_updates( $updates ) {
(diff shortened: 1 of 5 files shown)


includes/helper/class-wp-job-manager-helper.php
```diff
@@ -207,10 +208,20 @@ public function check_for_updates( $check_for_updates_data ) {
 				&& isset( $response['new_version'] )
 				&& version_compare( $response['new_version'], $plugin_data['Version'], '>' )
 			) {
+				$available_addon_updates[ $product_slug ] = $response;
 				$check_for_updates_data->response[ $plugin_data['_filename'] ] = (object) $response;
 			}
 		}
 
+		// Enable or disable notices.
```
renatho, line 216: I realized that it's called only when we access the Plugins or the Dashboard > Updates page. Not sure if it's called in some job too. So it just creates the notice after that.
And it will also just clear the notice only after accessing that pages when updated.

Is it something expected? Or should we implement it on a different hook?

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -202,6 +214,24 @@ public static function display_core_setup() {
 		include dirname( __FILE__ ) . '/views/html-admin-notice-core-setup.php';
 	}
 
+	/**
+	 * Displays the setup wizard notice when WPJM is first activated.
```
renatho: We need to fix this PHPDoc.

renatho (approved):
In general, it looks good and works well!

There's just one important comment about the hook that we check for updates. But if it's expected, feel free to go ahead. 😉

includes/helper/class-wp-job-manager-helper.php
aaronfc: Mmm, good catch!
My logic was that I didn't want to change when this addon update check is being called.

Showing the notice only after those pages are accessed makes sense to me. And also in my tests when you update the plugins it will also be called, so I think that the removal will behave correctly.

Additionally: We will make this notice dismissible when improving the design for these notices in [here](https://github.com/Automattic/WP-Job-Manager/issues/2382). So in case it got "stuck" users would be able to manually remove it.

For the moment I am going to merge this, but will take a note to review this during testing period to see if there is any other case I am missing.

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Good catch! Fixed b8ff7669

PR #2385, opened 2023-03-28
Delete jm_wpjmcom_add_on_messages on install

interlineae:
Required for: https://github.com/Automattic/wpjobmanager.com/issues/205
cc @aaronfc

**Without this change, add-on notices will misbehave.** 

* Deletes the `jm_wpjmcom_add_on_messages` transient whenever WP Job Manager is installed or updated.

### Testing instructions

- Apply the one-line change to an older version of WPJM, e.g. with `git checkout 1.38.0`
- Make sure your wpjobmanager.com local env is running https://github.com/Automattic/wpjobmanager.com/tree/add/outdated-message-display
- Clear transients, e.g. with `wp transient delete -all`
- Visit **Job Listings** --> Addons in WP Admin to see the below notice (`/wp-admin/edit.php?post_type=job_listing&page=job-manager-addons`)
- Remove the `wp_job_manager_version` option from the database if it shows v1.39.0
- Update WP Job Manager
- The notice should now be removed

wpjm-update-notice

### DIFF
--- a/includes/class-wp-job-manager-install.php
+++ b/includes/class-wp-job-manager-install.php
@@ -69,6 +69,7 @@ class WP_Job_Manager_Install {
 		}
 
 		delete_transient( 'wp_job_manager_addons_html' );
+		delete_transient( 'jm_wpjmcom_add_on_messages' );
 		update_option( 'wp_job_manager_version', JOB_MANAGER_VERSION );
 	}
 


renatho (approved):
Good catch! Looks good and works well!
I wonder if we also need the same in Sensei LMS. I'll check it.

**Update:** In Sensei we don't need it because the version check is local on the site.

aaronfc (approved):
Looks good and works well! 🚀

PR #2387, opened 2023-03-31
Return plugin info for addon even when there's no license configured

### LINKED ISSUE #2376: Check for addon updates even when license is invalid.
In the same vibe as we plan to do on Sensei (2204-gh-Automattic/sensei-pro) we want to check for addon updates availability even when the license is not valid yet.

This depends on the equivalent work on the API (212-gh-Automattic/wpjobmanager.com).

**Success Criteria**:
- [x] When no license is set or license is invalid the plugins page shows there is an update available.
- [x] Download is not available, though, unless the license is active and valid.
- [x] A link to "Activate" or purchase a license is shown.
- [ ] Optional: Consider bulk check for update information instead of one request per addon.

fjorgemota:
Fixes #2376 

- Remove the check to return early when the addon doesn't have a license, so the addon now appear on the list of plugins to be updated even when it is not licensed;

### Testing instructions

On a WordPress installation with this branch of the plugin installed, and a local version of WPJobManager.com installed and running:

1. Install a  WPJobManager addon;
2. Make sure its version is older than the one registered on your local WPJobManager.com environment, which should be running this branch too:  214-gh-Automattic/wpjobmanager.com
3. Make sure the addon isn't licensed;
4. Verify if the addon appears on the list of plugins to be updated (the `/wp-admin/update-core.php` on your installation);
5. Verify that, if you select to update the addon, it actually cannot be updated and fails with the message "An error occurred while updating WP Job Manager - YOUR ADDON: Update package not available."

Here's a snippet to use your local WPJobManager.com dev environment with a WordPress installation running WPJM:

```php

		$new_url = str_replace(
			$scheme . '://' . $host,
			'http://YOUR_IP:YOUR_PORT',
			$url
		);

	}
	return $preempt;
}, 10, 3 );
	return true;
});

add_filter( 'http_allowed_safe_ports', function($ports) {
	$ports[] = YOUR_PORT;
	return $ports;
});
```

Remember to replace `YOUR_IP` with the host/IP for your dev environment, and `YOUR_PORT` with the port used in your dev environment.

### DIFF
--- a/includes/helper/class-wp-job-manager-helper.php
+++ b/includes/helper/class-wp-job-manager-helper.php
@@ -362,9 +362,6 @@ class WP_Job_Manager_Helper {
 			return false;
 		}
 		$args = $this->get_plugin_licence( $product_slug );
-		if ( empty( $args['licence_key'] ) || empty( $args['email'] ) ) {
-			return false;
-		}
 		$args['api_product_id'] = $product_slug;
 
 		$response = $this->api->plugin_information( $args );
--- a/includes/helper/class-wp-job-manager-helper.php
+++ b/includes/helper/class-wp-job-manager-helper.php
@@ -361,7 +361,7 @@ class WP_Job_Manager_Helper {
 		if ( ! $this->is_product_installed( $product_slug ) ) {
 			return false;
 		}
-		$args = $this->get_plugin_licence( $product_slug );
+		$args                   = $this->get_plugin_licence( $product_slug );
 		$args['api_product_id'] = $product_slug;
 
 		$response = $this->api->plugin_information( $args );


renatho (approved):
> so the addon now appear on the list of plugins to be updated even when it is not licensed;

I think this part was already working for me before this PR, but with this PR I can open the modal of "view version details.". Did I understand correctly?

If yes, it looks good and works well!

fjorgemota:
Hey there, thanks for the review!

> I think this part was already working for me before this PR, but with this PR I can open the modal of "view version details.". Did I understand correctly?

Before this PR, if you visit `/wp-admin/update-core.php` on your WP installation with an outdated WPJM addon, that addon will not appear on the list. However, with this PR, that addon will now appear on that list. Remember to clear all the transients of the WP installation on each test as everything is cached on a transient, and remember to use the WPJobManager.com PR here: 214-gh-Automattic/wpjobmanager.com to make sure that the data will be returned even when the license is invalid and/or doesn't exist.

Could you please make sure you're seeing the same thing as me? I'll wait before merging this just to make sure that everything is alright. :)

renatho:
> Could you please make sure you're seeing the same thing as me?

Sure! I tested `feature/reengaging-customers`, and the plugins are displayed in the Updates list. While debugging, I only see the function changed here being called when I open the modal with the details. 🤔

Let me know if you want me to check anything specific in my env. 😉

fjorgemota:
Actually, testing again, you're right! 

This is definitely unexpected for me (I was expecting it to only show the addon on the plugin list when this PR was used), but anyway, it indeed fixes the issue that you reported, which isn't something I was even aware about.. :laughing: 

I'm gonna merge this and consider the given issue as fixed, which it should be, anyway.

Thanks! Both for the review and for testing again!

PR #2389, opened 2023-04-04
Update npm dependencies

gikaragia:

* Updates dependencies and fixes tests
* The errors in tests are caused by a deprecation in core that happened in 6.2. As we want to test in both 6.1 and 6.2, it isn't possible to migrate to the new class so I disabled the check instead.

### Testing instructions

* Just make sure that the actions run successfully


yscik (approved):
🍀

PR #2390, opened 2023-04-07
Add support for activating multiple licenses at once

### LINKED ISSUE #2375: Improve licensing page experience for activation of multiple licenses.
As part of the Re-engaging Customers project we want to improve the experience for WPJM users who have purchased a bundle with multiple addons.

Currently the experience is quite cumbersome, by needing to activate each of the licenses individually.

Hopefully a simple "single save button" would be feasible and make it easier for the users.

We depend on 210-gh-Automattic/wpjobmanager.com for making a single request in bulk.

Design proposal: p6r3EZ-1lK-p2

Design i2: p6r3EZ-1mQ-p2

**Success Criteria**:
- [x] On the licensing page multiple licenses can be activated by a single "save" button.
- [x] On error the error message from server is shown (including link to renew or purchase licenses).
- [x] `invalid_product` errors are ignored (not shown) unless there wasn't any success
- [x] WP.com managed licenses management is still disabled. More context: here (p1680262423094569-slack-C04URKMQHA9) and [here](https://github.com/Automattic/WP-Job-Manager/pull/2332).

fjorgemota:
Fixes #2375
This depends on 221-gh-Automattic/wpjobmanager.com

- Add support for bulk licensing multiple WPJM addons in a single request;
- Improve design of the Licenses page for WPJM addons;

### Testing instructions

Using a local version of WPJobManager.com running the latest code on 221-gh-Automattic/wpjobmanager.com, and this version of the WPJobManager plugin (and with the core addons enabled too):

1. Verify if you can activate a plugin;
2. Verify if you can bulk activate all the plugins;
3. Verify if the bulk activation form disappears when every plugin is disabled;
4. Verify what happens when there's an error connecting to your local WPJobManager.com server;
5. Verify if you can deactivate a plugin;
6. Verify what happens when you type a invalid code in the bulk activation form;
7. Verify what happens when you type a invalid code in the activation form for a plugin;
8. Verify if the design matches what was specified;
9. Try different variations of tests. it is very important to have this code well tested;

**Note:** You can use the following snippet to use a local WPJobManager.com version in your PC:

```php

		$new_url = str_replace(
			$scheme . '://' . $host,
			'http://YOUR_IP:YOUR_PORT',
			$url
		);

	}
	return $preempt;
}, 10, 3 );
	return true;
});

add_filter( 'http_allowed_safe_ports', function($ports) {
	$ports[] = YOUR_PORT;
	return $ports;
});
```

### Notes

- I didn't add search on this PR nor added the separate view of inactive vs. active plugins. I'll add such tasks in smaller PRs, as this one is already quite big!
- ~The php unit checks here at github are failing because they're rely on an outdated mock. I'll checkout on fixing them later~ The tests are fixed;

### Screenshots

image

image


aaronfc:
Hey @fernando, I am struggling a little testing this.

When I do the requests from my local environment to my local wpjobmanager.com I get the following error on server side:
```
[12-Apr-2023 11:58:25 UTC] WC:LOG:fatal-errors CRITICAL Uncaught Error: Call to undefined function Never5\LicenseWP\Shortcode\wc_add_notice() in /Users/aaron/workspace/automattic/wpjobmanager.com/plugins/license-wp/src/Shortcode/UpgradeLicenseForm.php:75
Stack trace:
#0 /Users/aaron/workspace/automattic/wpjobmanager.com/plugins/license-wp/src/Shortcode/UpgradeLicenseForm.php(26): Never5\LicenseWP\Shortcode\UpgradeLicenseForm->load_license()
#1 /Users/aaron/workspace/automattic/wpjobmanager.com/plugins/license-wp/src/Plugin.php(133): Never5\LicenseWP\Shortcode\UpgradeLicenseForm->__construct()
#2 /Users/aaron/Local Sites/wpjmcom/app/public/wp-includes/class-wp-hook.php(308): Never5\LicenseWP\Plugin->Never5\LicenseWP\{closure}('')
#3 /Users/aaron/Local Sites/wpjmcom/app/public/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array)
#4 /Users/aaron/Local Sites/wpjmcom/app/public/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#5 /Users/aaron/Local Sites/wpjmcom/app/public/wp-settings.php(617): do_action('init')
#6 /Users/aaron/Local Sites/wpjmcom/app/public/wp-config.php(111): require_once('/Users/aaron/Lo...')
#7 /Users/aaron/Local Sites/wpjmcom/app/public/wp-load.php(50): require_once('/Users/aaron/Lo...')
#8 /Users/aaron/Local Sites/wpjmcom/app/public/wp-blog-header.php(13): require_once('/Users/aaron/Lo...')
#9 /Users/aaron/Local Sites/wpjmcom/app/public/index.php(17): require('/Users/aaron/Lo...')
#10 {main}
  thrown in /Users/aaron/workspace/automattic/wpjobmanager.com/plugins/license-wp/src/Shortcode/UpgradeLicenseForm.php on line 75
```

But I don't see that error when sending the following curl:
```
curl --request POST \
  --url https://wpjmcom.local/wp-json/wpjmcom-licensing/v1/activate \
  --header 'Content-Type: application/json' \
  --data '{
	"license_key": "290B-FE73-B637-4F2D",
	"site_url": "https://test3.local",
	"product_slugs": [ "wp-job-manager-resumes", "something"]
}'
```

Any idea what I might be missing?

includes/helper/class-wp-job-manager-helper-api.php
```diff
@@ -67,44 +67,66 @@ public function plugin_information( $args ) {
...
+	public function bulk_activate( $licence_key, $product_slugs ) {
+		return $this->request_endpoint(
+			'wp-json/wpjmcom-licensing/v1/activate',
+			[
+				'method' => 'POST',
+				'body'   => [
+					'site_url'      => $this->get_site_url(),
+					'license_key'   => $licence_key,
+					'product_slugs' => $product_slugs,
+				],
```
aaronfc: We need to `json_encode` the body.

includes/helper/class-wp-job-manager-helper-api.php
```diff
@@ -127,29 +149,33 @@ protected function request( $args, $return_error = false ) {
...
+	 * @param array  $args The arguments to pass to the request.
+	 * @param bool   $return_error If we should return the error details or not.
+	 *
+	 * @return array|false The response, an error if $return_error is true, or false if the request failed and $return_error is false.
+	 */
+	protected function request_endpoint( $endpoint, $args, $return_error = false ) {
+		$defaults = [
+			'timeout' => 10,
+			'headers' => [
+				'Accept' => 'application/json',
```
aaronfc: We need also the `'Content-Type' => 'application/json'` header.

includes/helper/views/html-licences.php
```diff
@@ -12,11 +12,46 @@
 <h1 class="screen-reader-text"><?php esc_html_e( 'Licenses', 'wp-job-manager' ); ?></h1>
 <div class="wpjm-licences">
 	<?php if ( ! empty( $licenced_plugins ) ) : ?>
+		<?php if ( ! empty( $show_bulk_activate ) ) : ?>
+		<div class="wpjm-bulk-activate">
+			<b class="wpjm-bulk-activate--title">
+				Activate Job Manager Licenses
```
aaronfc: Missing internationalisation in all these copies.

aaronfc (dismissed):
I debugged a little my problems and found the root cause for them.
I am not sure why you could test it without this... maybe a missing commit?

I will do more testing once this is addressed :)

includes/helper/views/html-licences.php
fjorgemota: Good catch! Fixed in 154266c97b5afbc23348943a034c7ce72a90839f, thanks!

includes/helper/class-wp-job-manager-helper-api.php
fjorgemota: Great catch! Fixed in e8784cea3e361479145af5c88038571bb41b23e8, thanks!

includes/helper/class-wp-job-manager-helper-api.php
fjorgemota: Nice catch! Fixed in 8a07a35917f33fe6ab617c84d0c251365e4f9cb1 (but using `wp_json_encode` instead), thanks!

fjorgemota:
Oh, great catches! I have no idea how it worked on my end. Is it something related to WP 5.9 or some filter I'm using? No idea. Or maybe it is related to LocalWP? :thinking: 

Either way, I fixed the points you mentioned. Thanks!

includes/helper/class-wp-job-manager-helper-api.php
```diff
@@ -67,44 +67,70 @@ public function plugin_information( $args ) {
...
-		if ( false === $response ) {
+		$args         = wp_parse_args( $args );
+		$product_slug = $args['api_product_id'];
+		$response     = $this->bulk_activate( $args['licence_key'], [ $product_slug ] );
+		if ( false === $response || ! array_key_exists( $product_slug, $response ) ) {
 			return false;
 		}
-		return $response;
+		$item = $response[ $product_slug ];
+		// Add keys used previously for backwards compatibility.
```
aaronfc, line 81: Nice catch!

includes/helper/class-wp-job-manager-helper-api.php
```diff
@@ -67,44 +67,70 @@ public function plugin_information( $args ) {
...
+		$item['activated'] = $item['success'];
+		unset( $item['success'] );
+		if ( ! $item['activated'] ) {
+			$item['error'] = $item['error_message'];
+			unset( $item['error_message'] );
+		}
```
aaronfc: Another option could be to directly map the new structure to the old one. Instead of adding/removing the activated/success/error/error_message items.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -504,10 +504,28 @@ public function licence_output() {
...
+	 * Return if we should show or not the bulk activation form.
+	 *
+	 * @param array $licenced_plugins The list of licensed plugins to handle.
+	 *
+	 * @return bool If we should show the bulk activation form or not.
+	 */
+	private function show_bulk_activation_form( $licenced_plugins ) {
+		foreach ( array_keys( $licenced_plugins ) as $product_slug ) {
+			$licence = self::get_plugin_licence( $product_slug );
+			if ( empty( $licence['licence_key'] ) && apply_filters( 'wpjm_display_license_form_for_addon', true, $product_slug ) ) {
```
aaronfc, line 522: Maybe we should add the hook/filter documentation?

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -533,46 +551,133 @@ public function licence_error_notices() {
...
 	 */
 	private function handle_request() {
+		if (
+			empty( $_POST['action'] )
+			|| empty( $_POST['_wpnonce'] )
+			|| ! check_admin_referer( 'wpjm-manage-licence' )
+		) {
+			return;
+		}
+		if ( str_starts_with( sanitize_text_field( wp_unslash( $_POST['action'] ) ), 'bulk_' ) ) {
```
aaronfc, line 566: I think this method `str_starts_with` is only part of PHP 8.0? And I think we are still expected to support PHP 7.2 based on https://github.com/Automattic/WP-Job-Manager/blob/trunk/wp-job-manager.php#L11

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -533,46 +551,133 @@ public function licence_error_notices() {
...
+				}
+			}
+		}
+		foreach ( $product_slugs as $product_slug ) {
+			$result = $response && isset( $response[ $product_slug ] ) ? $response[ $product_slug ] : false;
+			if ( $skip_invalid_product && false !== $result && ! $result['success'] && 'invalid_product' === $result['error_code'] ) {
+				continue;
+			}
+			$this->handle_product_activation_response( $result, $product_slug, $licence_key, false );
+		}
```
aaronfc: I think this logic is fine but I found it a little bit difficult to follow.
What do you think about this approach (pseudo-code)?
```php
$response = ...;
$errors = [];
$any_success = false;
foreach ($response as $product_slug => $product_response) {
   if ($product_response['success']) {
     $any_success = true;
     // Maybe process the successful response here?
   } else {
      $errors[$product_response['error_code']] = $product_reponse['error_message'];
   }
}
if ($any_success) {
  unset($errors['invalid_product']); // Removing invalid product error because there was at least a success
}
// Process all $errors to add the messages.
// If successes were not processed above, process them now like you are doing.
```

Let me know if you find this confusing or I am missing anything.

aaronfc (commented):
Looking good and my tests in local were successful!

I left a comment about a method that I am not sure is available in PHP 7.2 – but if that's not wrong or you fix it, I am ok with approving the rest.

Left some other comments but are more suggestions that really problems.

includes/helper/class-wp-job-manager-helper.php
fjorgemota: This filter isn't new, it was added in #2332, and as far as I know it is intended to be kinda of "secret". I can document it, sure, but honestly I'm not sure if that's something we really want to do. :thinking:

includes/helper/class-wp-job-manager-helper.php
fjorgemota: It is, but WP also provides a polyfill here: https://github.com/WordPress/WordPress/blob/master/wp-includes/compat.php#L445-L465

In #2392, I bumped the minimum WP version to 6.0, so it should already include that polyfill. :)

fjorgemota: It is, but WP also provides a polyfill here: https://github.com/WordPress/WordPress/blob/master/wp-includes/compat.php#L445-L465

In #2392, I bumped the minimum WP version to 6.0, so it should already include that polyfill. :)

includes/helper/class-wp-job-manager-helper-api.php
fjorgemota: Good one! Done in 33c20ac, thanks!

includes/helper/class-wp-job-manager-helper.php
fjorgemota: I think I improved the logic significantly in b4e315d, could you take a look?

Now the `$skip_invalid_product` should be in fact used well, at least..

includes/helper/class-wp-job-manager-helper.php
aaronfc: Awesomeeee!

aaronfc (approved):
Looks good and works well! 🚀

PR #2392, opened 2023-04-11
Add Search functionality to the Licenses tab

### LINKED ISSUE #2394: Add search bar to the Licenses for WPJM Addons page
### Steps to Reproduce
1. Go to the Jobs Listings -> Add-ons -> Licenses page;
2. Verify if there's a way to filter the list of addons;

### What I Expected

To have a search bar there to filter for the addons

### What Happened Instead

No search bar :(

### PHP / WordPress / WP Job Manager Version

Any.

### Browser / OS Version

Any.

fjorgemota:
Fixes #2394 
Based on #2390 

* Add logic for searching licensed plugins on the Licenses page;
* Add form to search licensed plugins;
* Bump WP versions required by the plugin so we can use some useful polyfills;

### Testing instructions

Verify if the search works on the WPJM Licenses page.


includes/helper/views/html-licences.php
```diff
@@ -45,6 +45,10 @@
 			?>
 		</div>
 		<?php endif; ?>
+		<form method='post' class='plugin-licence-search'>
```
renatho: WDYT to use double quotes here? Just to follow the standard of the rest.

renatho (approved):
Looks good and works well!
Just added a small suggestion.

includes/helper/views/html-licences.php
fjorgemota: Sure, I also changed the code to use double quotes in the rest of the HTML too: 9a083b8. 

Thanks!

PR #2397, opened 2023-04-13
Add sections active and inactive on the WPJM Licenses page

fjorgemota:
Fixes #2396
Based on #2392 

### Changes proposed in this Pull Request

* Add sections "active" and "inactive" for the WPJM Licenses page;

### Testing instructions

1. Verify if the list of addons for WPJM is grouped by active and inactive;
2. Verify if the buttons to activate a license and deactivate a licence still work, and that the product is moved to the correct place;

### Screenshots

image

PR #2400, opened 2023-04-14
Design Adjustments for the Licenses page for WPJM addons

fjorgemota:
This PR is the result of a call that I had with @Luchadores regarding possible improvements/adjustments based on the work in #2397. It contains some small design fixes and adjustments for the WPJM Addons License page.

* Adjust "Inactive" section not being moved correctly to align vertically more with the Search input;
* Remove notice style from the message "Lost your license key" in the bottom of the page;
* Changed checkmark icon to match the one from Gutenberg;
* Added  `aria-hidden` to the logo of the product, which isn't an interactive element anyway; (well, this isn't a design change, but I included here anyway)
* Added red border color to the licence field when there's a notice showing an error;
* Changed notices background from white to the color that Gutenberg uses for each notice type;

### Testing instructions

Verify if the WPJM Addons Licenses page design is more aligned with the one from the Figma design.

Check:

- Both the error and success notices
- If the first section is more vertically aligned with the search input
- If the checkmark for an activated product looks good
- If the border of the field of an addon with an error becomes red

### Screenshot / Video

image

image

image

image


aaronfc (approved):
Looks good and works well!

I found some notices that I think were already fixed by you in some other branch, but if you could double check? They appear when using a notice to activate some products.

image

fjorgemota:
> I found some notices that I think were already fixed by you in some other branch, but if you could double check? They appear when using a notice to activate some products.

Ah, yes, that actually happens because of the bulk licensing logic for ignoring the `invalid_product` errors, and it indeed was fixed in another PR/branch, more specifically, it was fixed in eea90243dc323ff2dc64dadc33790816fc48407f, which wasn't included in this PR until I merged the base branch back into this branch in c5f8128.

Given that the issue was already fixed and reviewed, I'm going ahead and merging this PR, then. Thanks, @aaronfc, for all the reviews!

PR #2398, opened 2023-04-13
Consume remote notices

aaronfc:
Fixes #2374 

### Changes proposed in this Pull Request
* Started using `\WP_Job_Manager` namespace with new classes.
* Added `\WP_Job_Manager\WP_Job_Manager_Com_API` class to encapsulate all requests to wpjobmanager.com – currently only implementing notices consumption.
* Consuming notices endpoint and displaying those using the existing `render_notice` method.
* Added some `@deprecated` annotations to previous notices mechanism.
* Implemented `\WP_Job_Manager\Admin\Notices_Conditions` to encapsulate all conditions checking (added unit tests).

### Testing instructions
There is a couple of ways of testing this:
* Testing using your local wpjobmanager.com environment.
  * Check out your local wpjm.com environment to `feature/reengaging-customers`.
  * Apply the patch from [here](227-gh-Automattic/wpjobmanager.com) – you can edit the values to test the use case you want.
  * Make sure you are pointing to your local wpjobmanager.com environment.
  * Clean your local transients.
* Mocking the api responses with a snippet.
  * Use the following snippet:
```
add_filter( 'pre_http_request', function($preempt, $r, $url) {
	$parsed = parse_url( $url );

	if ( 'wpjobmanager.com' === $parsed['host'] ) {
		if ( '/wp-json/wpjmcom-notices/1.0/notices' === $parsed['path']) {
			return array(
            'headers'     => array(),
            'cookies'     => array(),
            'filename'    => null,
            'response'    => [ 'code' => 200 ],
            'status_code' => 200,
            'success'     => 1,
            'body'        => <<<EOQ
{
	"notices": {
		"test-notice-1": {
			"type": "site-wide",
			"icon": "wpjm",
			"heading": "Awesome Testing Header 1",
			"message": "This is a test notice message.",
			"actions": [
				{
					"label": "Read more",
					"url": "https:\/\/wpjobmanager.com",
					"primary": true,
					"target": "_blank"
				},
				{
					"label": "Other",
					"url": "https:\/\/wpjobmanager.com",
					"primary": false,
					"target": "_blank"
				}
			],
			"conditions": [
				{
					"type": "screens",
					"screens": [
						"wpjm*"
					]
				}
			]
		},
		"test-notice-2": {
			"type": "site-wide",
			"level": "warning",
			"icon": "wpjm",
			"heading": "Awesome Testing Header 2",
			"message": "This is a test warning message.",
			"actions": [
				{
					"label": "Read more",
[...]

PR #2401, opened 2023-04-14
Admin notice modernisation

### LINKED ISSUE #2382: Modernise design for admin notices.

interlineae:
Fixes #2382

* Different styling for different notice types added (error, warning, info, success)
* Setup for different notice classes in place
* Setup for icon retrieval in place
* Setup for Sensei's notice dismissal JS in place
* Setup for `get_dissmised_notices_`, `saved_dismissed_notices`, and `handle_notice_dimiss` is in place.

After discussing with @aaronfc about the scope of #2382 , we determined that setting up this PR, and the issue in question, would entail (1) firming up admin styling while (2) setting up persistent notice dismissal in a following issue. This was determined after realising that it would be a mark larger task than initially realised to have notice dismissal be persistent while also moving over all notice types to the new implementation.

To that end, this PR, like the issue it fixes:

* Adds support for different WPJM notice styling in admin
* Provides a foundation for notices to be migrated to the new notice design
* Makes notices dismissable, with persistence to be added in a latter, separate issue

### Testing instructions 

* Check out branch
* Ensure that you have an older version of a paid WPJM add-on active. If not, use a discount code and grab a license for an addon from https://wpjobmanger.com, then download an older copy of the addon from GH.
* You should then see a notice like so:

notice


includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -317,16 +332,40 @@ private static function generate_notice_from_updates( $updates ) {
...
-		// TODO Handle different levels of notices.
-		echo '<div class="notice wpjm-admin-notice wpjm-admin-notice--info">';
+	private static function render_notice( $notice_id, $notice ) {
+		if ( empty( $notice['actions'] ) || ! is_array( $notice['actions'] ) ) {
+			$notice['actions'] = [];
+		}
+
+		$notice_class = '';
+		if ( ! empty( $notice['style'] ) ) {
+			$notice_class = 'wpjm-notice-' . $notice['style'];
```
aaronfc: I think it would be nice to check if `style` is info, error, warning, success and fall back to info.

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -317,16 +332,40 @@ private static function generate_notice_from_updates( $updates ) {
...
+		}
+
+		$is_dismissible       = $notice['dismissible'];
```
aaronfc:
```suggestion
		$is_dismissible       = $notice['dismissible'] ?? true;
```

Not sure if true or false by default. Since false is more restrictive I would say true... but I am open for discussion.

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -317,16 +332,40 @@ private static function generate_notice_from_updates( $updates ) {
...
+		$notice_wrapper_extra = '';
+		if ( $is_dismissible ) {
+			wp_enqueue_script( 'job_manager_notice_dismiss' );
+			$notice_class        .= ' is-dismissible';
+			$notice_wrapper_extra = sprintf( ' data-dismiss-action="wpjm_dismiss_notice" data-dismiss-notice="%1$s" data-dismiss-nonce="%2$s"', esc_attr( $notice_id ), esc_attr( wp_create_nonce( self::DISMISS_NOTICE_NONCE_ACTION ) ) );
+		}
+
+		// TODO Remove hard-coded 'is-dismissable' CSS class once all notices are converted to use this class.
+		echo '<div class="notice wpjm-admin-notice wpjm-admin-notice--info is-dismissible' . esc_attr( $notice_class ) . '"' . esc_attr__( $notice_wrapper_extra ). '>';
```
aaronfc: I think that if we do the above change we can already remove the `is-dismissible` already.

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -317,16 +332,40 @@ private static function generate_notice_from_updates( $updates ) {
...
 
 		echo '<div class="wpjm-admin-notice__top">';
-		// TODO Add icon.
+
+		// TODO Implement get_icon method.
+		if ( ! empty( $notice['icon'] ) ) {
+			// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped -- Dynamic parts escaped in the function.
+			echo get_icon( $notice['icon'], 'wpjm-notice__icon' );
```
aaronfc: Not sure I understand this. What is this `get_icon` method? Is a method that does not exist?
We need to implement this. I am happy if we initially only add a method like:

```
private function get_icon( $icon_name ) {
    switch ($icon_name) {
        case 'wpjm':
            return '';// path to the actual image
    }
    return ''; // default image (maybe the wpjm one?)
}
```

We can talk about this if you want over slack.

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -317,16 +332,40 @@ private static function generate_notice_from_updates( $updates ) {
...
+		}
+
 		echo '<div class="wpjm-admin-notice__message">';
-		echo wp_kses( $notice['message'], self::ALLOWED_HTML );
+		echo '<strong>' . wp_kses( $notice['message'], self::ALLOWED_HTML ) . '</strong>';
```
aaronfc, line 381: maybe we want to add `font-weight: bold` to the `wpjm-admin-notice__message` better?

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -350,6 +389,37 @@ private static function render_notice( $notice ) {
...
+
+	/**
+	 * Save dismissed notices.
+	 *
+	 * @param array $dismissed_notices Array of dismissed notices.
+	 * @param bool  $is_user_notification True if we are setting user notifications (vs site-wide notifications).
+	 */
+	private function save_dismissed_notices( $dismissed_notices, $is_user_notification ) {
+		// TODO Implement from Sensei.
+	}
```
aaronfc: Can you remove these? Since we are not using them we don't really need these skeletons here.

includes/admin/class-wp-job-manager-admin.php
```diff
@@ -89,7 +89,7 @@ public function admin_enqueue_scripts() {
 		WP_Job_Manager::register_select2_assets();
 
 		$screen = get_current_screen();
-		if ( in_array( $screen->id, apply_filters( 'job_manager_admin_screen_ids', [ 'edit-job_listing', 'plugins', 'job_listing', 'job_listing_page_job-manager-settings', 'job_listing_page_job-manager-addons' ] ), true ) ) {
+		if ( in_array( $screen->id, apply_filters( 'job_manager_admin_screen_ids', [ 'edit-job_listing', 'plugins', 'job_listing', 'job_listing_page_job-manager-settings', 'job_listing_page_job-manager-addons', 'edit-job_listing_type', 'dashboard' ] ), true ) ) {
```
aaronfc, line 92: Is this needed?

webpack.config.js
```diff
@@ -11,13 +11,14 @@ const files = {
...
 	'js/term-multiselect': 'js/term-multiselect.js',
+	'js/admin/wpjm-notice-dismiss': 'js/admin/wpjm-notice-dismiss.js',
 	'css/admin': 'css/admin.scss',
 	'css/admin-notices': 'css/admin-notices.scss',
 	'css/frontend': 'css/frontend.scss',
 	'css/job-listings': 'css/job-listings.scss',
 	'css/job-submission': 'css/job-submission.scss',
 	'css/setup': 'css/setup.scss',
-	'css/menu': 'css/menu.scss',
+	'css/menu': 'css/menu.scss'
```
aaronfc: I think it's a good practice to leave the trailing comma

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: If you think this will take a lot of time, we can just leave the todo (without the non-existing method) and create a separate ticket to implement this.

aaronfc (commented):
Looks good! Left some comments.

I am adding @renatho so he can give feedback if there is a better way to do something in frontend (assets loading, icon, etc).

includes/admin/class-wp-job-manager-admin.php
interlineae: With the notices as they currently display, if this isn't added, then navigating to **Job Listings --> Job Types** or WP Admin home shows the notice as so:

styles

Did we want to hide the notices on these pages instead?

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: I had gone with `<strong>` for the semantic significance (i.e. this is an important notice for screen readers). Maybe that's a misplaced concern, and if so, I can switch this out to the `font-weight` property.

Or, is there another consideration between the two that I'm leaving out?

assets/css/admin-notices.scss
```diff
@@ -7,10 +11,22 @@
 
 	// Spacing.
 	gap: 10px;
-	padding: 18px 24px;
+	padding: 10px 12px;
+
+	&.wpjm-admin-notice--error {
```
renatho: Just a suggestion to make use of the sass features, keeping it a little cleaner. The same applies to the next classes. ;)

```suggestion
	&--error {
```

assets/css/admin-notices.scss
```diff
@@ -7,10 +11,22 @@
...
+		border-left-color: $notice-error;
+	}
+
+	&.wpjm-addon-update--warning {
```
renatho: Was this accidental to have a different name?

assets/js/admin/wpjm-notice-dismiss.js
```diff
@@ -0,0 +1,41 @@
...
+		formData.append( 'action', element.dataset.dismissAction );
+		formData.append( 'nonce', element.dataset.dismissNonce );
+
+		fetch( ajaxurl, {
+			method: 'POST',
+			body: formData,
+		} );
+	};
+
+	document.body.addEventListener( 'click', ( event ) => {
```
renatho: I imagine you used this approach because the button is being added later for some reason, right?
If yes, I wonder if we could reduce the `body` scope to something more specific. So we prevent running this function on every click on the page. WDYT?

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -43,14 +46,15 @@ public static function init() {
...
 	}
 
 	/**
 	 * Add a notice to be displayed in WP admin.
 	 *
-	 * @since 1.32.0
-	 *
 	 * @param string $notice Name of the notice.
+	 *
+	 * @since 1.32.0
```
renatho, line 70: Out of curiosity, is there any special reason to move the `@since` to the last item?

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -317,16 +332,40 @@ private static function generate_notice_from_updates( $updates ) {
...
 	 */
-	private static function render_notice( $notice ) {
-		// TODO Handle different levels of notices.
-		echo '<div class="notice wpjm-admin-notice wpjm-admin-notice--info">';
+	private static function render_notice( $notice_id, $notice ) {
+		if ( empty( $notice['actions'] ) || ! is_array( $notice['actions'] ) ) {
+			$notice['actions'] = [];
+		}
+
+		$notice_class = '';
```
renatho: Maybe it could be an array? So we don't worry about the spacings while when adding other classes. :)
So when using, we just `implode`.

includes/admin/class-wp-job-manager-admin.php
renatho: About this, I'd say that it would be nice to add only the needed assets, otherwise, we add a lot of unneeded assets to the page.

But I see that it's the current standard here and it seems to be including all the assets used in the admin. And since it's in the admin, I think it's fine to keep that for now to avoid a big refactor.

renatho (commented):
Hey! Adding some comments here too. 😉

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: At the moment, `$is_dismissable` is `null`. If the endpoint sets that to an empty string it'll be `false`/`0`, but with the understanding that we want these notices to _always_ be dismissible, then I agree with defaulting to `true`. That way, a non-dismissible notice would also have to be intentional.

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: That we can!

webpack.config.js
interlineae: This was a JSHint suggestion, but looking into it further, it seems to have been made only out of respect to old versions of IE. Reading up on 'final commas' (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas), trailing commas are indeed good practice!

assets/css/admin-notices.scss
interlineae: Ah, yep, glossed over that. With the change above this has been removed.

assets/js/admin/wpjm-notice-dismiss.js
interlineae: I used the approach in Sensei (copied from Sensei), so if we do updates here, we'll want to take a look at Sensei as well. To tighten this up, are we thinking something like this?

```
const wpjmNotice = document.querySelector('.wpjm-admin-notice' );
if ( wpjmNotice ) {
    wpjmNotice.addEventListener( 'click', ( event ) => {
    ...
}
```

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: I haven't been able to find anything suggesting the order is important.

This was done by PHPStorm's linter, and I decided to keep it as it made sense to me in terms of consistency. When I compare, for instance, with Sensei's [class-sensei-admin-notices.php](https://github.com/Automattic/sensei/blob/trunk/includes/admin/class-sensei-admin-notices.php), the `@since` isn't consistently placed in doc blocks (that I can see): sometimes it's last, sometimes it's first.

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/4cc876d2ea4b6c27f8079a8886540a29cfd44466

assets/css/admin-notices.scss
interlineae: Commited https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/56386a5274e65a9c5abae0ac8e5e8f7ff62b1401

assets/css/admin-notices.scss
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/56386a5274e65a9c5abae0ac8e5e8f7ff62b1401

webpack.config.js
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/229e6a4f410620d5e9b495cc725f8ca6e7e3a1d0

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/f82515acd805ecddeac7434e5f7a66de48f4c4ca

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/adc630bf257f5104f22ae576597dc46e719f916b

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: I've committed a change for this at https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/4c798a9f95176ca5b17fa66c93fa1d2ab6157a44.

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: @aaronfc Let's chat about this, as thinking over it, I'm not sure that a `get_icon` method is required, but let's chat.

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Yeah, I would disable these "automatic" changes in your PHPStorm configuration. Imagine that VSCode would do the same but putting the since first. We would be "fighting" each other in each change.

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: 🤔 I didn't thought about the semantic significance... makes sense to me now :)

includes/admin/class-wp-job-manager-admin.php
aaronfc: Thanks for the explanation @anaemnesis ! I am good with this approach too

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Talked over slack and we will implement a private method for this and add the new icon asset.

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: @aaronfc I've made another change to hopefully improve handling of this, so that we don't risk nonsense CSS classes. https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/03898a95f2a118ce6de45554cbc7950238abc153

includes/admin/class-wp-job-manager-admin-notices.php
interlineae: @aaronfc I've committed for this in https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/722258ccd6c7dc40366f9fcafaf1e200c343d0c0. I've kept the WPJM as a fallback icon for branding/recognisability. We can always return an empty string instead if we thought against that.

assets/js/admin/wpjm-notice-dismiss.js
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/eb913659a1f603c37f759cab5c791f44ea082b3a

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: I reverted these `@since` changes in my PR [here](https://github.com/Automattic/WP-Job-Manager/pull/2398) (still in draft at the time of writing this).

aaronfc (approved):
Looking good! I tested this in conjunction with my [branch](https://github.com/Automattic/WP-Job-Manager/pull/2398).

I might add there any extra changes if I find anything else that needs to be tweaked.

Good joooob! 🚀

assets/js/admin/wpjm-notice-dismiss.js
renatho: Hi @anaemnesis! 👋

Yep! Thank you for the change. But in this case, maybe we don't need the conditional anymore since we're only getting clicks under `.wpjm-admin-notice`, right? Not something that would cause a bug though. 🙂

PR #2407, opened 2023-04-19
Change addon update notices to new system.

### LINKED ISSUE #2406: Use new notice system for addon update notices.
We added some notices in #2379 and later we improved the notices system in #2374 .

We want to move these notices to the new system based on the new `wpjm_admin_notices` filter.

aaronfc:
Fixes #2406 

* Move addon update notices to new `wpjm_admin_notices` filter.

### Testing instructions
* Either install an outdated addon or go to `includes/helper/class-wp-job-manager-helper.php` and reverse the `version_compare` condition.
* Clean transients.
* Confirm the outdated addons notice appears in dashboard and the wpjm default screens.
* If you see wrong addon names in the notice you might need to checkout to this branch (229-gh-Automattic/wpjobmanager.com) in your local wpjobmanager.com environment.

### Screenshot / Video

image

### DIFF
--- a/includes/admin/class-wp-job-manager-admin-notices.php
+++ b/includes/admin/class-wp-job-manager-admin-notices.php
@@ -405,7 +405,8 @@ class WP_Job_Manager_Admin_Notices {
 		if ( ! empty( $updates ) ) {
 			$notice = self::generate_notice_from_updates( $updates );
 			if ( ! is_null( $notice ) ) {
-				$notices[ self::NOTICE_ADDON_UPDATE_AVAILABLE ] = $notice;
+				$notice_id             = self::generate_notice_id_from_updates( $updates );
+				$notices[ $notice_id ] = $notice;
 			}
 		}
 		return $notices;
@@ -570,6 +571,20 @@ class WP_Job_Manager_Admin_Notices {
 		}
 		echo '</div>';
 
+	}
+
+	/**
+	 * Generate unique notice ID based on the updates available.
+	 *
+	 * @param array $updates The updates available.
+	 * @return string The notice ID.
+	 */
+	private static function generate_notice_id_from_updates( $updates ) {
+		$notice_id = self::NOTICE_ADDON_UPDATE_AVAILABLE;
+		foreach ( $updates as $update ) {
+			$notice_id .= '-' . $update['plugin'] . '@' . $update['new_version'];
+		}
+		return md5( $notice_id );
 	}
 }
 
--- a/includes/admin/class-wp-job-manager-admin-notices.php
+++ b/includes/admin/class-wp-job-manager-admin-notices.php
@@ -405,7 +405,7 @@ class WP_Job_Manager_Admin_Notices {
 			return $notices;
 		}
 
-		$updates = get_transient( 'wpjm_addon_updates_available', [] );
+		$updates = get_site_transient( 'wpjm_addon_updates_available', [] );
 		if ( ! empty( $updates ) ) {
 			$notice = self::generate_notice_from_updates( $updates );
 			if ( ! is_null( $notice ) ) {
--- a/includes/admin/class-wp-job-manager-admin-notices.php
+++ b/includes/admin/class-wp-job-manager-admin-notices.php
@@ -401,6 +401,10 @@ class WP_Job_Manager_Admin_Notices {
 	 * @return mixed
 	 */
 	public static function maybe_add_addon_update_available_notice( $notices ) {
+		if ( ! current_user_can( 'install_plugins' ) ) {
+			return $notices;
+		}
+
 		$updates = get_transient( 'wpjm_addon_updates_available', [] );
 		if ( ! empty( $updates ) ) {
 			$notice = self::generate_notice_from_updates( $updates );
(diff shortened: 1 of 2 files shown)


includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -144,13 +144,7 @@ public static function has_notice( $notice ) {
...
 		// core_setup: Notice is used when first activating WP Job Manager.
 		add_action( 'job_manager_admin_notice_' . self::NOTICE_CORE_SETUP, [ __CLASS__, 'display_core_setup' ] );
-		add_action(
-			'job_manager_admin_notice_' . self::NOTICE_ADDON_UPDATE_AVAILABLE,
-			[
-				__CLASS__,
-				'display_addon_update_available',
-			]
-		);
+		add_filter( 'wpjm_admin_notices', [ __CLASS__, 'maybe_add_addon_update_available_notice' ], 10, 1 );
```
jom: I think this should be moved up into `init`. I think `init_core_notices` is only loaded when the legacy system is and the legacy notices aren't loaded on Ajax calls (for the dismiss action, for example). This is making it so the dismiss action doesn't work because it isn't a known notice when it gets to `handle_notice_dismiss`.

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -398,22 +392,23 @@ public static function display_core_setup() {
...
+	public static function maybe_add_addon_update_available_notice( $notices ) {
 		$updates = get_transient( 'wpjm_addon_updates_available', [] );
-
 		if ( ! empty( $updates ) ) {
-			$notice    = self::generate_notice_from_updates( $updates );
-			$notice_id = self::NOTICE_ADDON_UPDATE_AVAILABLE;
-			self::render_notice( $notice_id, $notice );
+			$notice = self::generate_notice_from_updates( $updates );
+			if ( ! is_null( $notice ) ) {
+				$notices[ self::NOTICE_ADDON_UPDATE_AVAILABLE ] = $notice;
```
jom: As is, if they ever dismiss the notice, it will never be shown again. Do we want this to be the behavior or should we generate the notice ID using the plugin updates available?

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -398,22 +392,23 @@ public static function display_core_setup() {
...
+	 *
+	 * @return mixed
 	 */
-	public static function display_addon_update_available() {
-		if ( ! self::is_admin_on_standard_job_manager_screen( [ 'dashboard' ] ) ) {
-			return;
-		}
-
+	public static function maybe_add_addon_update_available_notice( $notices ) {
 		$updates = get_transient( 'wpjm_addon_updates_available', [] );
```
jom: I'm not sure if this requires a change, but I was trying to think of the consequences of storing this `wpjm_addon_updates_available` transient as a normal transient when it is hooked into and updating a _site_ transient. My brain overloaded and shut down, but it might be something to think about.

jom (commented):
This is looking good! I've added a few comments/questions.

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Fixed in 54154a2c

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Yeah, I thought that showing it for every release available could be too much... But on a second thought that's the main purpose of this project so...  Changed here 16f25dbf

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Humm, that's a very interesting point.
Not sure there is any consequences since, AFAIK, they are handled separately... Aren't they?

But it's true that we are using a site_option in the multisite case for the license information. In that case, it would make sense to have this transient made a "per_site" transient too. If we don't, we could have this transient contain addon updates based on updates for a different license. Does it make sense to you? I mean, moving this `wpjm_addon_updates_available` to a site transient.

aaronfc: After our slack conversation:
- Added notice only for users who can install plugins e130656f
- Moved addon updates availability transient to site_transient 575322e6

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -569,6 +576,20 @@ private static function render_notice( $notice_id, $notice ) {
...
+	 *
+	 * @param array $updates The updates available.
+	 * @return string The notice ID.
+	 */
+	private static function generate_notice_id_from_updates( $updates ) {
+		$notice_id = self::NOTICE_ADDON_UPDATE_AVAILABLE;
+		foreach ( $updates as $update ) {
+			$notice_id .= '-' . $update['plugin'] . '@' . $update['new_version'];
+		}
+		return md5( $notice_id );
```
jom: Really tiny thing: what do you think about using standard prefix and then hashing the plugin/version combos? Only benefit is being able to see them together in the dismissed actions array.

jom (approved):
Works well and looks good. Left one comment about maintaining the prefix for the notice ID, but not critical.

includes/admin/class-wp-job-manager-admin-notices.php
aaronfc: Done in 81c35791

PR #2409, opened 2023-04-21
Update historical notices to use the new `render_notice` method.

### LINKED ISSUE #2378: Revamp unlicensed notices design.
During the Re-engaging Customers project it was raised that the current notices when unlicensed are quite old-fashioned and messy.

The idea would be to revamp the UI as well as the UX (consolidating multiple of them in a single one, for example)

This is how it looks currently:
image

After #2383 and #2382 we should be in a good place to do something better with these.

interlineae:
Fixes #2378

This PR updates historical licenses to use the new `WP_Job_Manager_Admin_Notices::render_notice` method.

This change changes the `render_notice` notice method from a `private static function` to a `public static function` and then updates the notice display per the renders at https://github.com/Automattic/WP-Job-Manager/issues/2382.

One item I have chosen to forgo for this change is the grouped update notification. I've decided this as the immediate request is to update historical notices to use the `render_notice` method, whereas the grouped update notifications would require the creation of new, non-historical notices.

One additional item I noticed is that the update notices don't inherit styling on non-WPJM pages:

Screenshot from 2023-05-04 11-26-57

To address this I could either move the relevant `wp_enqueue_style` and `wp_register_style` outside of the `if ( in_array( $screen->id, apply_filters(....` conditional, or, add more screen IDs to the check. I've opted to move the style as it's not possible to anticipate all of the potential screens the notices might display on, or even if it was, the check would be so long as to be pointless.

To prevent a conflict with WC styling, I've also made use of a `! important` declaration. I don't know if, instead, we want to copy into WPJM the relevant `notice` styles and have our own `wpjm-notice` class instead. I'm not keen on `! important` so if we want to go to in that direction, I'm all for it. The declaration does seem the most straightforward though, and it's specific to our notices.

Without this change, notices lose their padding on certain WC pages (like WooCommerce --> Orders). For example:

Screenshot from 2023-05-04 11-59-24

### Testing instructions

* Checkout branch
* Install a couple of paid WPJM plugins, but don't input a license key
* Notices should display with the updated formatting as shown below
* If you've tested this prior and an update notice isn't showing, use `wp option delete wp_job_manager_dismissed_notices` through WP CLI (or otherwise delete this option) to force the notice to display.

Screenshot from 2023-05-04 11-03-16

* To show an error notice, comment out the code to show the missing license notice https://github.com/Automattic/WP-Job-Manager/blob/15393a2e1a84d2d194f0097d2023e2940ac38e4e/includes/helper/class-wp-job-manager-helper.php#L569-L576
* change the `elseif` into an `if`, and then the check to show an error notice from `! empty` to `empty` https://github.com/Automattic/WP-Job-Manager/blob/15393a2e1a84d2d194f0097d2023e2940ac38e4e/includes/helper/class-wp-job-manager-helper.php#L576-L583


assets/css/admin-notices.scss
```diff
@@ -11,7 +11,7 @@ $notice-success: #43af99;
 
 	// Spacing.
 	gap: 10px;
-	padding: 10px 12px;
+	padding: 10px 12px !important;
```
interlineae: I'm not totally cool with this, but after testing a few other options, it's the most straightforward option and limited to WPJM notice styling.

Another option is to rename `notice` to `wpjm-notice` and copy the relevant styling over to `admin-notices.scss`.

includes/admin/class-wp-job-manager-admin.php
```diff
@@ -128,6 +125,9 @@ public function admin_enqueue_scripts() {
 
 		WP_Job_Manager::register_style( 'job_manager_admin_menu_css', 'css/menu.css', [] );
 		wp_enqueue_style( 'job_manager_admin_menu_css' );
+
+		WP_Job_Manager::register_style( 'job_manager_admin_notices_css', 'css/admin-notices.css', [] );
+		wp_enqueue_style( 'job_manager_admin_notices_css' );
```
interlineae, line 133: I've moved this outside of the conditional above as I was adding so many different screen IDs that the check was becoming pointless. It's also not possible to anticipate IDs as users will have very many different screens based on very many different plugins, and loading the styles globally makes the most sense to me.

interlineae:
Noting that since the feature branch has been closed by @aaronfc, approval of this PR will result in a merge directly to `trunk`, triggering a deployment !!

assets/css/admin-notices.scss
renatho: I realized an issue that we're also overriding an style that increases the padding in order to not have the dismiss button over the text.

Screenshot 2023-05-05 at 16 28 33

Maybe a solution could be to add the `!important` only on the `padding-top` and `padding-bottom`?

includes/admin/class-wp-job-manager-admin.php
renatho: It makes sense to me too! Just to confirm, I noticed you also removed the `dashboard` from the conditional. Was it there only for the notices?

includes/helper/views/html-licence-key-error.php
```diff
@@ -8,9 +8,18 @@
...
-?>
-<div class="error">
-	<p class="wpjm-updater-dismiss" style="float:right;"><a href="<?php echo esc_url( wp_nonce_url( add_query_arg( 'dismiss-wpjm-licence-notice', $product_slug ), 'dismiss-wpjm-licence-notice', '_wpjm_nonce' ) ); ?>"><?php esc_html_e( 'Hide notice', 'wp-job-manager' ); ?></a></p>
-	<?php // translators: %1$s is the plugin name, %2$s is the license setting page URL. ?>
-	<p><?php printf( wp_kses_post( __( 'There is a problem with the license for "%1$s". Please <a href="%2$s">manage the license</a> to check for a solution and continue receiving updates.', 'wp-job-manager' ) ), esc_html( $plugin_data['Name'] ), esc_url( admin_url( 'edit.php?post_type=job_listing&page=job-manager-addons&section=helper#' . sanitize_title( $product_slug . '_row' ) ) ) ); ?></p>
-</div>
+
+$notice = [
+	'level'       => 'error',
+	'dismissible' => true,
```
renatho: I noticed that with this change, the dismissal is not being persisted anymore.

It seems we have a method checking the `wpjm_hide_notice` query string, but in this case, it's dismissing the notice without adding the query string or running any request.

Do you know if it's working in any other place to see what is different?

renatho (commented):
Hey! In general, it's looking very good!
I just left comments about 2 things I noticed.

> Noting that since the feature branch has been closed by @aaronfc, approval of this PR will result in a merge directly to `trunk`, triggering a deployment !!

About this, I think it's not the case for the Job Manager plugin. :)

assets/css/admin-notices.scss
interlineae: Ah, I see what's happening there, yeah. I've gone with the separated `padding` route and I'm not seeing any issues with that.

Committed here https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/b63ae05c8050adb24085c5ed9fdf743fca25abeb

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -545,39 +545,41 @@ function ( $plugin ) use ( $search_term ) {
...
 	 */
-	public function licence_error_notices() {
-		$screen = get_current_screen();
-		if (
-			null === $screen ||
-			in_array( $screen->id, [ 'job_listing_page_job-manager-addons' ], true ) ||
-			! current_user_can( 'update_plugins' )
-		) {
-			return;
-		}
```
interlineae: After speaking with @aaronfc we realised that `null` was being encountered when it ought not have been. This was removed as we decided that (1) error notices and (2) missing license notices should be displayed on the add-ons page.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -545,39 +545,41 @@ function ( $plugin ) use ( $search_term ) {
...
+						'wpjm_admin_notices',
+						function ( $notices ) use ( $product_slug, $plugin_data ) {
+							return $this->add_missing_licence_notice( $notices, $product_slug, $plugin_data );
+						}
+					);
 				} elseif ( ! empty( $licence['errors'] ) ) {
-					include 'views/html-licence-key-error.php';
+					add_filter(
+						'wpjm_admin_notices',
+						function ( $notices ) use ( $product_slug, $plugin_data ) {
```
interlineae, line 591: I tried to avoid the anonymous functions here, but I'm not aware of another way to pass `$product_slug` and `$plugin_data` to the methods called.

includes/admin/class-wp-job-manager-admin.php
interlineae: That's right. I added that here https://github.com/Automattic/WP-Job-Manager/pull/2401/commits/50a8a711471ebed61d76c9222f458c1524530dce.

interlineae:
@aaronfc @renatho After speaking with Aaron, I completely refactored the notice displays to use the new `wpjm_admin_notice` filter. Testing instructions have been updated, and I worked through a couple of hangups with Aaron's help. But we're good for another round of testing.

https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/15393a2e1a84d2d194f0097d2023e2940ac38e4e

Edit: oh hang on, apparently we have unit tests now 👀 
Edit: Fixed up. Failing tests now pertain to PHP version conflicts in the test suite itself

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -545,39 +545,41 @@ function ( $plugin ) use ( $search_term ) {
...
 			}
 		}
 		return false;
 	}
 
 	/**
-	 * Outputs unset license key notices.
+	 * Outputs unset licence key notices.
 	 */
-	public function licence_error_notices() {
```
renatho, line 565: In the plugins open for the users we should take care to not remove the public functions to avoid errors with any type of integration.

It seems just a rename right? So in this case, we could deprecate this method, and inside it, call the `maybe_add_licence_error_notices`. So if someone is using it, they will see the deprecation warning instead of an error, and the function will continue working.

renatho (commented):
Hi Katja!

Thank you for working on that!

In my new test, it didn't show the notices. Maybe it's related to the order of hooks where we're using the `wpjm_admin_notices` now? Let me know if you need any help with that. ;)

Also, as we talked, if we replace the "license" with "licence" in this PR, we could try to avoid the `apply_filters` and the `log_event` functions to avoid breaking filters and losing metrics. Does it make sense? I'd advocate to not rename the strings too to not lose translations of what is already translated. But it's less critical if we really want this change.

includes/helper/class-wp-job-manager-helper.php
interlineae: @aaronfc and I had discussed renaming, so it's just a rename that's right.

I've not deprecated a method before, but that sounds like a plan to me. I'll get to looking at that.

interlineae:
> Also, as we talked, if we replace the "license" with "licence" in this PR, we could try to avoid the apply_filters and the log_event functions to avoid breaking filters and losing metrics. Does it make sense? I'd advocate to not rename the strings too to not lose translations of what is already translated. But it's less critical if we really want this change.

We did, and completely agree. I've gone ahead and hard-reset and reapplied changes:

- https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/8755fa325110855241e9e5ee8e0c4a8570696cec
- https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/aeba146ebb5702c7f49fdd8f7981ec9cb541e5be

The notices are showing for me though, so let me run through them again and see if I can replicate.

includes/helper/views/html-licence-key-error.php
interlineae: This was addressed in the refactor https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/15393a2e1a84d2d194f0097d2023e2940ac38e4e

includes/helper/class-wp-job-manager-helper.php
interlineae: Deprecated here https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/ed89a13d0a2f23d86fb65b376bfef5dae6322a28

interlineae:
> In my new test, it didn't show the notices. Maybe it's related to the order of hooks where we're using the `wpjm_admin_notices` now? Let me know if you need any help with that. ;)

I think this is due to previous testing. Can you delete the `wp_job_manager_dismissed_notices` option and see if that forces the notice to display anew for you? In WP CLI that would be `wp option delete wp_job_manager_dismissed_notices`.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -10,14 +10,14 @@
...
- * Helper functions used in WP Job Manager regarding addons and licenses.
+ * Helper functions used in WP Job Manager regarding addons and licences.
  *
  * @package wp-job-manager
  * @since   1.29.0
  */
 class WP_Job_Manager_Helper {
 	/**
-	 * License messages to display to the user.
+	 * Licence messages to display to the user.
```
aaronfc: I would leave this as they were (with s).

includes/helper/class-wp-job-manager-helper.php
aaronfc: Looking nice to me. Good catch, @renatho !

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -559,25 +559,39 @@ private function show_bulk_activation_form( $licenced_plugins ) {
...
 	public function licence_error_notices() {
-		$screen = get_current_screen();
-		if (
-			null === $screen ||
-			in_array( $screen->id, [ 'job_listing_page_job-manager-addons' ], true ) ||
-			! current_user_can( 'update_plugins' )
-		) {
-			return;
-		}
+		_deprecated_function( __METHOD__, '1.33.0', 'maybe_add_license_error_notices()' );
```
aaronfc, line 569: Here you call the method 'license' but the real name you set is 'licence'.

I would go with 's' everywhere (except from the deprecated method, which obviously has to be kept like it was before).

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -559,25 +559,39 @@ private function show_bulk_activation_form( $licenced_plugins ) {
...
+		$this->maybe_add_license_error_notices();
```
aaronfc, line 570: I think there is a typo here. The method is actually called "licence" (but I think you should rename it to "license").

includes/helper/class-wp-job-manager-helper.php
aaronfc: Mmm good call!

I am fine with using an anonymous function here. The idea to not use an anonymous function is to be able to "unhook" it programatically. But in this case we can always "hook" later and remove the notice_id that we want to remove.

I can think of alternative ways like storing in a private attribute the data needed or something like that, but I don't like it. Good to go with the anonymous function 🚀

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -895,6 +909,71 @@ private function handle_product_activation_response( $response, $product_slug, $
...
+
+	/**
+	 * Add a notice to the admin if a licence key is missing.
+	 *
+	 * @param array  $notices     Notices to be displayed.
+	 * @param string $product_slug The plugin slug.
+	 * @param array  $plugin_data The plugin data.
+	 * @return array
+	 */
+	public function add_missing_licence_notice( $notices, $product_slug, $plugin_data ) {
```
aaronfc: this can be private, right?

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -895,6 +909,71 @@ private function handle_product_activation_response( $response, $product_slug, $
...
+
+	/**
+	 * Add a notice to the admin if a licence key is missing.
+	 *
+	 * @param array  $notices     Notices to be displayed.
+	 * @param string $product_slug The plugin slug.
+	 * @param array  $plugin_data The plugin data.
+	 * @return array
+	 */
+	public function add_error_licence_notice( $notices, $product_slug, $plugin_data ) {
```
aaronfc: same here, can be private :)

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -895,6 +909,71 @@ private function handle_product_activation_response( $response, $product_slug, $
...
+			'message'     => sprintf(
+				wp_kses_post(
+				// translators: %1$s is the URL to the licence key page, %2$s is the plugin name.
+					__( '<a href="%1$s">Please enter your licence key</a> to get updates for "%2$s".', 'wp-job-manager' )
+				),
+				esc_url( admin_url( 'edit.php?post_type=job_listing&page=job-manager-addons&section=helper#' . sanitize_title( $product_slug . '_row' ) ) ),
+				esc_html( $plugin_data['Name'] )
+			),
+		];
+		$notices[ 'wpjm_licence_notice_' . $product_slug ] = $notice;
```
aaronfc: I would add here "missing" to the license notice ID. Also, since the dismissing mechanism is different I think we can rename it freely (and also name it "license" with an S).

Final name: `'wpjm_missing_license_notice_' . $product_slug`

aaronfc (approved):
Works great! 🚀 

I left some comments but are mostly about naming. Once that's fixed I am happy with this being merged.

Thanks for working on this! <3

includes/helper/class-wp-job-manager-helper.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/d9e812b90a5b4c38a01c1fb96ed29c40c2c1cad0

includes/helper/class-wp-job-manager-helper.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/d9e812b90a5b4c38a01c1fb96ed29c40c2c1cad0

includes/helper/class-wp-job-manager-helper.php
interlineae: Committed (for here and above) https://github.com/Automattic/WP-Job-Manager/pull/2409/commits/ea758b00beec40385a238ad15d3a41ccaccb878a

aaronfc (approved):
Looks good, works well! 🚀

includes/helper/class-wp-job-manager-helper.php
renatho: Thank you for changing that, Katja! And sorry for the late delay on reviewing it!

Just instead of `maybe_add_license_error_notices()` we should use `WP_Job_Manager_Helper::maybe_add_license_error_notices` to reference the correct method. Not something critical and urgent though. 😉

PR #2413, opened 2023-05-05
Release 1.40.0

aaronfc:
Release branch.

## Includes
- Bump version to 1.40.0
- Bump minimum php version to 7.4
- Regenerate pot files.
- Regenerate changelogs.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,15 @@
+= 1.40.0 =
+* Enhancement: Improve licensing page experience
+* Enhancement: New email notifications setting
+* Enhancement: Improve discoverability of addon updates
+* Enhancement: New admin notices sytem with redesign
+* Fix: "Hide Expired Listings Content" setting
+* Fix: Hide expired job listings from archives
+* Fix: Show license notice only to users who can manage licenses
+* Fix: Select menu overlapping in Job Listing frontend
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "wp-job-manager",
-  "version": "1.39.0",
+  "version": "1.40.0",
```
renatho, line 3: I think we need to run an `npm install` to also update the `package-lock`.

changelog.txt
```diff
@@ -1,5 +1,18 @@
 # WP Job Manager
 
+## 1.40.0 - 2023-05-11
```
renatho, line 3: Is it this date on purpose because we plan to release this date?

renatho (approved):
Just one small tweak and a question about the release date. 😉
I'm already approving it, but I think it's worth waiting for a double check from @Automattic/zen.

yscik (approved):
🙌

changelog.txt
aaronfc: That's right :)

package.json
aaronfc: Done! [a28122f](https://github.com/Automattic/WP-Job-Manager/pull/2413/commits/a28122f7bc329835219f15d905c43f8d484730b4)

PR #2416, opened 2023-05-11
WIP: Add Relist Feature for Expiring Listings

### LINKED ISSUE #848: Provide relist option before expiration

onubrooks:
Fixes #848

* Adds the re-list link to the actions when a listing is about to expire.
* Adds 'Expires soon' label to listings that are about to expire.
* Adds an option (setting) to allow re-listed listing to be published automatically.

### Testing instructions

* Go to the Job Dashboard page. Notice the option to Relist the job listing even though it is not expired.
* Relist the listing. The new expiry date should be calculated based on the old expiry date, not the current date.

### New/Updated Hooks

* `job_manager_expiring_soon_days`: Number of days that determine if a listing is expiring soon.
* `job_manager_should_publish_relisting`: Automatically publish relisted listings without going through approval process.

### Screenshot / Video

Screenshot 2023-05-12 at 13 27 02

Screenshot 2023-05-12 at 13 28 46

### TODO

- Make it work nicely with WC Paid Listings and Simple Paid Listings

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -346,15 +346,6 @@ class WP_Job_Manager_Settings {
 							'attributes' => [],
 						],
 						[
-							'name'       => 'job_manager_publish_relistings',
-							'std'        => '1',
-							'label'      => __( 'Publish Relisted Listings', 'wp-job-manager' ),
-							'cb_label'   => __( 'Automatically publish relisted listings', 'wp-job-manager' ),
-							'desc'       => __( 'Sets all relisted submissions to "publish." They will not need admin approval.', 'wp-job-manager' ),
-							'type'       => 'checkbox',
-							'attributes' => [],
-						],
-						[
 							'name'       => 'job_manager_expiring_soon_days',
 							'std'        => 5,
 							'label'      => __( 'Expiring Soon Days', 'wp-job-manager' ),
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -346,6 +346,15 @@ class WP_Job_Manager_Settings {
 							'attributes' => [],
 						],
 						[
+							'name'       => 'job_manager_publish_relistings',
+							'std'        => '1',
+							'label'      => __( 'Publish Relisted Listings', 'wp-job-manager' ),
+							'cb_label'   => __( 'Automatically publish relisted listings', 'wp-job-manager' ),
+							'desc'       => __( 'Sets all relisted submissions to "publish." They will not need admin approval.', 'wp-job-manager' ),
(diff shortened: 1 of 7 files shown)


onubrooks:
Some Thoughts/Notes

This is a work in progress but it's important to get people's thoughts early on to determine the direction we go on this.
Two possible ways to implement re-listing:

1. Go through the edit flow again for the soon expiring listing. This is how it works for an already expired listing but the listing now has to go through the approval process again if the moderation setting is on. I've added a new setting to not go through the approval again, as well as a filter for the same. This actually works but the risk is that in the edit flow when re-listing, the employer can edit and add things that would normally not get through moderation. How do we feel about this?
2. Skip the Edit flow when re-listing. Just extend the expiry date. This would likely mean more custom work in the Paid Listings plugins. It may, however, mean that we can skip approval, without the moderation concerns from the approach in this PR.

Is there any other approach we should take? Please share your thoughts.

CC: @gkaragia @yscik @mikeyarce @jom @kraftbj @tripflex @Gnodesign @burtrw

gikaragia:
I think that the main benefit of relisting is that the listing will always be available so we shouldn't unpublish the listing in any case. Additionally, there is already a UI in the job dashboard and settings in admin about editing a listing that covers all the cases (edits are not allowed, edits are automatically approved, edits require approval) so we shouldn't mix relisting with editing. The only benefit of the user going through the edit flow again is that the process will be the same with expired listings.

So I think that we should go with option 2 unless there are serious technical implications. The feature should probably be named 'Extend' instead of 'Relist'. Finally the only option that we could introduce as part of this is to allow users to disable extension entirely, although maybe this should be better done with a filter?

gikaragia:
Just wanted to say a couple more things about this. Following option 1 is ok too if it is going to be hard to implement the 'Choose a package' workflow for paid listings. If we go with option 1 we should:
- Still not add any new settings
- Edits in general should follow the `Allow Published Edits` setting even when extending the listing
- Is it possible to check if the user has actually edited the listing when relisting and if he didn't then approve automatically?

yscik:
For paid listings, I think the expectation for re-listing/extending is that it needs payment/consumes another job package. So it would need to be its own explicit action (option 2), which paid listings can hook into. The flow there would be something like:
- [WCPL]: Relist → Select job package → Go through checkout & purchase if its a new package → Submit
- [SPL]: Relist → Pay → Submit 

On 'Extend' vs 'Relist', maybe we can allow both? (Allow relisting both before or after expiring)

gikaragia:
> On 'Extend' vs 'Relist', maybe we can allow both? (Allow relisting both before or after expiring)

Yep 'Relist' already exists at the moment but a job needs to expire in order to become available. The way 'Relist' works is that it takes you to the post job form and you follow the same flow with posting a new listing.

onubrooks:
Closing in favor of #2422

PR #2422, opened 2023-05-16
Introduce listing renewals

### LINKED ISSUE #848: Provide relist option before expiration

onubrooks:
Fixes #848 
Based on discussion in https://github.com/Automattic/WP-Job-Manager/pull/2416

* Adds the extend link to the actions when a listing is about to expire, extends the listing expiry date.
* Adds 'Expires soon' label to listings that are about to expire.
* Adds a new setting enabling or disabling extending a listing's expiry date.

### Testing instructions

* In Job Submissions settings, enable extending job listings (Extend job listing expiry), it's turned off by default.
* Go to the Job Dashboard page. Notice the option to 'Extend' the job listing action.
* Click on 'Extend' and follow the steps. The new expiry date should be calculated based on the old expiry date, not the current date.

### New/Updated Hooks

* ~`job_manager_expiring_soon_days`~ `job_manager_renewal_days`: Number of days that determine if a listing is expiring soon.
* `renew_job_steps`: Fires when listing renewal steps have been added.

### Screenshot / Video

https://github.com/Automattic/WP-Job-Manager/assets/26160845/99a48604-dfb5-4b10-bc67-3f4cba96086c

### TODO

- ~Add new hooks for integrating with WC Paid Listings and Simple Paid Listings~ Done

### Update
*Extend* is now *Renew* which seems like a better description for the feature.


includes/admin/class-wp-job-manager-settings.php
```diff
@@ -345,6 +345,28 @@ protected function init_settings() {
 							'type'       => 'checkbox',
 							'attributes' => [],
 						],
+						[
+							'name'       => 'job_manager_expiring_soon_days',
+							'std'        => 5,
+							'label'      => __( 'Expiring Soon Days', 'wp-job-manager' ),
+							'cb_label'   => __( 'Number of days to expiry', 'wp-job-manager' ),
```
gikaragia: `cb_label` stands for checkbox label I think and it is only used in checkboxes.

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -345,6 +345,28 @@ protected function init_settings() {
...
+							'desc'       => __( 'Sets the number of days to expiry, after which a listing can be extended.', 'wp-job-manager' ),
+							'type'       => 'number',
+							'attributes' => [],
+						],
+						[
+							'name'       => 'job_manager_can_extend_listing_expiry',
+							'std'        => 'no',
+							'label'      => __( 'Extend job listing expiry', 'wp-job-manager' ),
+							'cb_label'   => __( 'Number of days to expiry', 'wp-job-manager' ),
```
gikaragia: Same here

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -345,6 +345,28 @@ protected function init_settings() {
...
+							'desc'       => __( 'Deternines whether extending job listing expiry is allowed.', 'wp-job-manager' ),
+							'type'       => 'select',
+							'attributes' => [],
+							'options' => [
+								'yes' => __( 'Allow extending job expiry date', 'wp-job-manager' ),
```
gikaragia: I think we should better use a checkbox here.

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -345,6 +345,28 @@ protected function init_settings() {
 							'type'       => 'checkbox',
 							'attributes' => [],
 						],
+						[
+							'name'       => 'job_manager_expiring_soon_days',
```
gikaragia: I think a native English speaker could help with this one cc @burtrw 

I am thinking that we should use 'Renewal' everywhere for this functionality to separate from 'Relisting'. 

We are also thinking to add two new options as part of this, one that will enable the functionality and one to specify the number of days before expiry that users can renew their listings. Maybe we should use a single setting and have 0 days to disable the functionality?

For the text itself can you have a look below? We can modify it accordingly if we chose a single setting.
Screenshot 2023-05-17 at 12 10 42

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -253,13 +253,19 @@ public function job_dashboard_handler() {
 
 						break;
 					case 'relist':
+					case 'extend':
```
gikaragia: I think 'renew' is a better term for this. Sorry for not thinking about it earlier

templates/job-dashboard.php
```diff
@@ -49,6 +49,9 @@
 								<?php if ('job_title' === $key ) : ?>
 									<?php if ( $job->post_status == 'publish' ) : ?>
 										<a href="<?php echo esc_url( get_permalink( $job->ID ) ); ?>"><?php wpjm_the_job_title( $job ); ?></a>
+										<?php if ( job_manager_job_can_be_extended( $job ) ) : ?>
+											<small>(<?php echo esc_html_e('Expires soon', 'wp-job-manager'); ?>)</small>
```
gikaragia: `esc_html_e` already does an echo

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1209,4 +1211,75 @@ public function track_job_submission( $post_id ) {
...
+	}
+
+	/**
+	 * Handle steps for extending a listing's expiry.
+	 *
+	 * @access private
+	 * @param array $steps
+	 * @return array
+	 */
+	public function maybe_extend_job_expiry( $steps ) {
```
gikaragia: Maybe we could name this `remove_edit_steps` to make it more clear?

templates/job-extend-preview.php
```diff
@@ -0,0 +1,48 @@
+<?php
```
gikaragia: I am not sure if adding a new template is the right approach here. Have you considered updating the existing 'preview' one and remove and rename the existing buttons?

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1209,4 +1211,75 @@ public function track_job_submission( $post_id ) {
...
+			$extending = $this->is_extend_action() && job_manager_job_can_be_extended( $this->job_id );
+
+			if ( $extending ) {
+				$old_expiry = date_create_immutable_from_format( 'Y-m-d', get_post_meta( $job->ID, '_job_expires', true ) );
+				$new_expiry = calculate_job_expiry( $job->ID, false, $old_expiry );
+				update_post_meta( $job->ID, '_job_expires', $new_expiry );
+			}
+
+			$requires_approval = get_option( 'job_manager_submission_requires_approval' );
+			$post_status       = $extending || ! $requires_approval ? 'publish' : 'pending';
```
gikaragia: Can you explain this part a bit? The way that I understand the code, `$extending` is always going to be true.

So AFAIU the only way `extend_preview_handler` is going to be called is if the user clicked the 'Extend' link which only appears if `job_manager_job_can_be_extended` is true. And we also call `is_extend_action` when we add the handler in the first place. 

Am I missing something?

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,35 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+ * @since $$next_version$$
+ *
+ * @param int|WP_Post $job The job or job ID.
+ *
+ * @return bool
+ */
+function job_manager_job_can_be_extended( $job ) {
+	$job    = get_post( $job );
+	$status = get_post_status( $job );
+	$expiry = strtotime( get_post_meta( $job->ID, '_job_expires', true ) );
```
gikaragia: I think we need to call `get_job_expiration` here as it has some more logic into it.

gikaragia (commented):
Good work overall! I like how this feature is starting to materialize!

includes/admin/class-wp-job-manager-settings.php
onubrooks: Thanks, didn't know that! https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/1e0a04553fcc46a875d813c11cded48c03d7e346

includes/admin/class-wp-job-manager-settings.php
onubrooks: 👍

includes/admin/class-wp-job-manager-settings.php
onubrooks: Agreed, also addressed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/1e0a04553fcc46a875d813c11cded48c03d7e346

includes/admin/class-wp-job-manager-settings.php
onubrooks: I like your suggestions, I'm not quite great at naming and giving labels but I can spot a good one when some else does 😄. Also addressed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/1e0a04553fcc46a875d813c11cded48c03d7e346

includes/class-wp-job-manager-shortcodes.php
onubrooks: Yeah, it's more descriptive! 👍 
Fixed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/fb1244e94979c13c938fca46a70d985e594e3742

includes/forms/class-wp-job-manager-form-submit-job.php
onubrooks: Addressed with the other renamings to 'renew' in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/fb1244e94979c13c938fca46a70d985e594e3742

templates/job-extend-preview.php
onubrooks: Hmmm, I went back and forth between using a new template and using the 'preview' template. I didn't want to add too much conditional logic in that template since there is none at all currently. I do however think it is better to use the same template now because simple paid listings kinda depends on it.
I'm working on a fix for this now...

includes/forms/class-wp-job-manager-form-submit-job.php
onubrooks: Yeah my bad. This handler was initially part of the `preview_handler` logic in the other PR raised last week and then moved to it's own method here. Fixed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/3fa0f119c7b4ae7164a162474cc5ba21516ebefd

I still wonder though, even with paid listings after successful payment, the status can still be set to 'pending' if the option `job_manager_submission_requires_approval` is set to true. Do you think that logic should also apply here?

wp-job-manager-functions.php
onubrooks: So we have these options:

1. `WP_Job_Manager_Post_Types::instance()->get_job_expiration( $job )->getTimestamp();`
2. `strtotime( get_post_meta( $job->ID, '_job_expires', true ) );` 

Both produce a timestamp with a day difference, I find the second one more accurate because it calculates the exact days (eg if we set 5 days to renewal, a date that is 5 days to the expiration returns true) while the `get_job_expiration` method seems a day late (It only returns true from 4 days to expiration).

templates/job-extend-preview.php
onubrooks: Fixed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/bb49c7a4b68b4ffab516267d914797a70853b34e

includes/admin/class-wp-job-manager-settings.php
gikaragia: The phrasing should be `Sets the number of days before expiration, that users can renew the listing.`. Should we also have a single setting?  In that case, setting the days to 0 would disable the functionality. We could also add a `Use 0 to disable renewals` in the description.

templates/job-extend-preview.php
gikaragia: Yep, this was the main problem with it, that users that have done customizations must define a second template while simply updating the existing one should be easier.

includes/forms/class-wp-job-manager-form-submit-job.php
gikaragia: So I am thinking that generally, the admin approval is required to make sure that no spammy content is published. As long as we don't allow editing as part of this flow, it should be ok to simply republish it. Do you think of any scenarios that renewing content that was approved before might need this extra approval?

wp-job-manager-functions.php
gikaragia: So the thing is that `get_job_expiration` uses the `job_manager_jobs_expire_end_of_day` filter which can be modified by users and return different results. For this reason, the `_job_expires` meta is only read inside `get_job_expiration` and we use this method everywhere else. I think we still need to use `get_job_expiration` and make sure that the functionality still works if someone has modified this filter.

includes/admin/class-wp-job-manager-settings.php
onubrooks: Good idea, implemented in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/bc089c29876c26ce810655f0203bcf858324d769

includes/forms/class-wp-job-manager-form-submit-job.php
onubrooks: I agree, there's no edit so it should be fine to maintain prior approval.

wp-job-manager-functions.php
onubrooks: Done in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2aca812b5280a411e90d3c1521ceb1bb21c87b9a with a comment.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1209,4 +1213,94 @@ public function track_job_submission( $post_id ) {
...
+
+	/**
+	 * Filters submit button text for renew-listing step.
+	 *
+	 * @param int $text The button text.
+	 * @return string
+	 */
+	public function submit_button_text_renewal( $text ) {
+		if ( $this->get_step_key( $this->step ) === 'renew-listing' ) {
+			return __( 'Renew Listing &rarr;', 'wp-job-manager' );
```
gikaragia: Was the arrow intentional? Not sure if it is a good idea

onubrooks: Yeah it was intentional, why do you think it's not a good idea?

gikaragia: It feels a bit out of place and it isn't consistent with the other submit button. Seems more like a customization that some sites might want to do by updating the filter. But let's check with @yscik or @Luchadores 

So we are looking if we want to add an arrow in the submit button which will be the last step of renewing a listing. You can have a look at the video in the description to see the screen that is after that step. Do you think that we should add it?

Screenshot 2023-05-23 at 11 13 45

For reference, this is the same step when we relist a previously expired listing:

Screenshot 2023-05-23 at 11 17 11

gikaragia: We had a discussion about this one [here](https://a8c.slack.com/archives/C036R3Y5ET0/p1684832723675199) and we are going to keep the button.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,40 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+ *
+ * @since $$next_version$$
+ *
+ * @param int|WP_Post $job The job or job ID.
+ *
+ * @return bool
+ */
+function job_manager_job_can_be_renewed( $job ) {
+	$job    = get_post( $job );
+	$status = get_post_status( $job );
```
gikaragia: Generally, you should define a variable close to where is used for improved code readability.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,40 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+
+	/*
+	 * `WP_Job_Manager_Post_Types::instance()->get_job_expiration( $job )->getTimestamp()` seems off by a day.
+	 * The following expression is equivalent to: `strtotime( get_post_meta( $job->ID, '_job_expires', true ) )`
+	 */
```
gikaragia: I don't think that this comment adds any value. Instead of explaining what the following code block does, it adds more confusion as it mentions a possible problem without explaining it. I would expect problems with the code to belong to a bug report or maybe a TODO, but comments should in general be helpful and save time.

So for the issue at hand, `get_job_expiration` is not off by a day, it just returns the exact timestamp (in seconds) that the expiration will happen. So if for example, a job has an expiration set to `2023/05/27` the exact timestamp of the expiration is going to be `2023/05/27 23:59:59` which means that the expiration day is included. Some users might find this confusing and expect that for the same date, the exact expiration timestamp to be `2023/05/27 00:00:00`. This is the reason that we added the `job_manager_jobs_expire_end_of_day` filter I think.

In any case, your code shouldn't make any assumptions and work regardless of the value of the filter, so it shouldn't remove a day. For example if I set a job with an expiry of `2023/05/27`, set the renewal days to 3 and call this function it will return true which isn't correct. If I set 3 days I would expect these days to be 27, 26 and 25 of May. 

I think we should also add at least a few test cases for this function, testing different values of the filter as time differences are always difficult to test.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,40 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+	$expiry = WP_Job_Manager_Post_Types::instance()->get_job_expiration( $job )->getTimestamp() - DAY_IN_SECONDS;
+
+	// If there is no expiry, then relisting is not necessary.
+	if ( ! $expiry ) {
```
gikaragia: `$expiry` can be `DateTimeImmutable` or `false`. Although it works, I think we should avoid implicit type conversion as they can cause hard-to-track bugs.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,40 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+	/**
+	 * Number of days before a job expires to allow relisting it.
+	 *
+	 * @since $$next_version$$
+	 *
+	 * @param int $expiring_soon_days The default number of days.
+	 */
+	$expiring_soon_days = apply_filters( 'job_manager_renewal_days', get_option( 'job_manager_renewal_days', 5 ) );
+
+	return 'publish' === $status && $expiry - time() < $expiring_soon_days * DAY_IN_SECONDS;
```
gikaragia: We should use `current_datetime()` instead of `time()` as we should take into account the WP timezone, same as we do in other places.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,40 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+	}
+
+	/**
+	 * Number of days before a job expires to allow relisting it.
+	 *
+	 * @since $$next_version$$
+	 *
+	 * @param int $expiring_soon_days The default number of days.
+	 */
+	$expiring_soon_days = apply_filters( 'job_manager_renewal_days', get_option( 'job_manager_renewal_days', 5 ) );
```
gikaragia: Why do you think that we need a filter as we already have a setting for this value?

gikaragia (changes requested):

wp-job-manager-functions.php
onubrooks: Hmmm, I usually find it more organised when all the variables are defined at the top of the function and then the logic comes after, just my personal opinion though.

But since there are some other variables that will be generated later, it's fine to move it down. https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/b42c439429cefa978335297df2635830c92b30f3

wp-job-manager-functions.php
onubrooks: Okay I get it now, expiration includes the entire day of the expiry date. Thanks for the explanation, fixed also in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/b42c439429cefa978335297df2635830c92b30f3

Added tests for the function in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/4b505cd66d3c912e483e6cd4cbb091c91f289b16

wp-job-manager-functions.php
onubrooks: Fixed in this refactor: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/b42c439429cefa978335297df2635830c92b30f3

wp-job-manager-functions.php
onubrooks: Also done in this refactor: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/b42c439429cefa978335297df2635830c92b30f3

wp-job-manager-functions.php
onubrooks: Yeah maybe we don't, I was thinking it will benefit developers somehow but I really can't tell how yet. Removed it for now.

wp-job-manager-functions.php
gikaragia: [Here's](https://wiki.c2.com/?DeclareVariablesAtFirstUse) a good article that explains the reasoning behind this.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1209,4 +1213,94 @@ public function track_job_submission( $post_id ) {
...
+	 */
+	public function remove_edit_steps_for_renewal( $steps ) {
+		unset( $steps['submit'], $steps['preview'] );
+		$steps['renew-listing'] = [
+			'name'     => 'Renew Listing',
+			'view'     => [ $this, 'preview' ],
+			'handler'  => [ $this, 'renew_preview_handler' ],
+			'priority' => 29, // This action should be just before 'done' whose priority is 30.
+		];
+		return apply_filters( 'renew_job_steps', $steps, $this );
```
gikaragia: Can we document the filter?

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1209,4 +1213,94 @@ public function track_job_submission( $post_id ) {
...
+				$update_job                  = [];
+				$update_job['ID']            = $job->ID;
+				$update_job['post_status']   = $post_status;
+				$update_job['post_date']     = current_time( 'mysql' );
+				$update_job['post_date_gmt'] = current_time( 'mysql', 1 );
+
+				wp_update_post( $update_job );
+			}
+
+			$this->step ++;
```
gikaragia:
```suggestion
			$this->step++;
```

gikaragia (commented):
So I did some testing with WC Paid listings and it seems that there are a couple of issues. 

When completing the 'Renew' flow, if I choose to buy a new package then when I place the order, a weird '... has been submitted succesfully message' although the order is not completed yet.
Screenshot 2023-05-29 at 17 28 43

Then when I mark the order as completed in admin, the listing doesn't actually renew. I would expect that:
1. The listing is renewed when the order is completed.
2. No changes are done to the listing until the order is completed.

If I choose to use one of the existing packages, in that case, although the flow works, the remaining submissions count doesn't decrease which essentially means that I get a free renewal.

I believe that this means that we are going to need changes to WC paid listings for the above to work?

includes/forms/class-wp-job-manager-form-submit-job.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/fb33140815923ce9339feabeb8bca4a7a5f98fb2

onubrooks:
> Then when I mark the order as completed in admin, the listing doesn't actually renew. I would expect that:

>The listing is renewed when the order is completed.
>No changes are done to the listing until the order is completed.
>If I choose to use one of the existing packages, in that case, although the flow works, the remaining >submissions count doesn't decrease which essentially means that I get a free renewal.

>I believe that this means that we are going to need changes to WC paid listings for the above to work?

Yes. I have made some changes to the other paid listings repository. More details there. Meanwhile I added some refactoring [here](https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/495dbb1addece6b3cdae584a63e72ce88364f1d9) and [here](https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/3051e2e795259ac4f38ff0c236e7b7c24d4800cf) to for the paid listings fixes to work.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1239,12 +1243,14 @@ public function is_renew_action() {
 	 */
 	public function remove_edit_steps_for_renewal( $steps ) {
 		unset( $steps['submit'], $steps['preview'] );
-		$steps['renew-listing'] = [
-			'name'     => 'Renew Listing',
-			'view'     => [ $this, 'preview' ],
-			'handler'  => [ $this, 'renew_preview_handler' ],
-			'priority' => 29, // This action should be just before 'done' whose priority is 30.
-		];
+		if ( ! class_exists( 'WC_Paid_Listings' ) && ! class_exists( 'WP_Job_Manager_Simple_Paid_Listings' ) ) {
```
onubrooks: Here we add preview step only if there's no paid listing plugin activated. If there's a paid listing, renewing the listing will happen in the plugin (WC paid listings) or in the `done_before` step (Simple paid listing).

wp-job-manager-functions.php
```diff
@@ -1763,3 +1763,14 @@ function job_manager_renew_job_listing( $job ) {
...
+
+/**
+ * Checks if the wc paid listings has the minimum version when installed and activated.
+ *
+ * @since $$next-version$$
+ *
+ * @return bool
+ */
+function job_manager_is_wcpl_renew_compatible() {
+	return ! class_exists( 'WC_Paid_Listings' ) || version_compare( JOB_MANAGER_WCPL_VERSION, '2.9.9', '>' );
```
onubrooks: This disables the feature, to test, change the version number to something less eg `2.9.8`

onubrooks, line 1775: This disables the feature, to test, change the version number to something less eg `2.9.8`

gikaragia (commented):
So there is an issue with this approach. Currently we have the core plugin and WCPL which depends on core.

The code in core should in general be agnostic of the code in plugins. Core should provide entry points (hooks) for extensions but other than that it should not mention code from other plugins. This means that any version checks should exist in WCPL.

So basically the way this should work should be:
- WCPL (and possibly simple listings, not sure if it applies) should check that WPJM's version supports renewal and if it is, add the actions and filters that would extend the functionality.
- If WPJM is on an old version and doesn't support renewals, then display a notice to upgrade core.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,58 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+
+	return 'publish' === $status && $expiry - $current_time_stamp < $expiring_soon_days * DAY_IN_SECONDS;
+}
+
+/**
+ * Renew a job listing.
+ *
+ * @param WP_Post $job The job to renew.
+ */
+function job_manager_renew_job_listing( $job ) {
```
gikaragia: This is now called twice once from the preview handler and once from the before_done step. This causes the job to be renewed twice.

gikaragia (commented):
Found an issue.

As discussed, shouldn't we add checks that if old versions of WCPL and SPL are installed, we should disable the renew functionality?

wp-job-manager-functions.php
onubrooks: Fixed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/37dec87bce4cc9552faf81129bec9a6161d477dd.
In done_before it was a fall-back for SPL but I found a hook on SPL: `wp_job_manager_spl_payment_complete` to do the renewal.

wp-job-manager-functions.php
```diff
@@ -1703,3 +1708,80 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+
+/**
+ * Checks if the Simple Paid Listings has the minimum version when installed and activated.
+ *
+ * @since $$next-version$$
+ *
+ * @return bool
+ */
+function job_manager_is_spl_renew_compatible() {
+	return ! class_exists( 'WP_Job_Manager_Simple_Paid_Listings' ) || version_compare( JOB_MANAGER_SPL_VERSION, '1.4.4', '>' );
```
onubrooks: Same here. This disables the feature, to test, change the version number to something less eg 1.4.3

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+	 * Submit job form instance.
+	 *
+	 * @var WP_Job_Manager_Form_Submit_Job
+	 */
+	public $form = null;
+
+	/**
+	 * Stores static instance of class.
+	 *
+	 * @access protected
```
gikaragia: I don't think this adds any value as the attribute is already protected

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+ * @since $next-version$$
+ */
+class WP_Job_Manager_Helper_Listing_Renewal {
+
+	/**
+	 * Submit job form instance.
+	 *
+	 * @var WP_Job_Manager_Form_Submit_Job
+	 */
+	public $form = null;
```
gikaragia: I don't think this should be public, especially since it's used in this class only.

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Stores static instance of class.
+	 *
+	 * @access protected
+	 * @var WP_Job_Manager_Form_Submit_Job The single instance of the class
+	 */
+	protected static $instance = null;
```
gikaragia: I am not sure I understand why this should be protected and not private

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Handles the renewal of Job Listings.
+ *
+ * @since $next-version$$
```
gikaragia: We don't currently replace these during our releases. Maybe we should just use the next version for now?

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+			return false;
+		}
+		return isset( $_GET['action'] ) && sanitize_text_field( wp_unslash( $_GET['action'] ) ) === 'renew';
+	}
+
+	/**
+	 * Handle steps for renewing a listing before expiry, removes edit steps and overrides preview step.
+	 *
+	 * @access private
+	 * @param array $steps
```
gikaragia: Missing the description

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+			'view'     => [ $this->form, 'preview' ],
+			'handler'  => [ $this, 'renew_preview_handler' ],
+			'priority' => 20,
+		];
+		/**
+		 * Filter the steps for renewing a listing before expiry.
+		 *
+		 * @since $$next-version$$
+		 *
+		 * @param array $steps
```
gikaragia: Same here

includes/helper/class-wp-job-manager-helper-listing-renewal.php
```diff
@@ -0,0 +1,145 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Handles the renewal of Job Listings.
+ *
+ * @since $next-version$$
+ */
+class WP_Job_Manager_Helper_Listing_Renewal {
```
gikaragia: As this class seems more like a collection of renewal-related functionality, maybe we should name it something like `WP_Job_Manager_Helper_Renewals`? 

The way it is named now, it hints that it is a helper for a specific listing renewal. Also, it will fit the name of the rest of the classes in this directory.

wp-job-manager-functions.php
```diff
@@ -1712,3 +1717,80 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+ * Checks if the job expiry can be extended.
+ * This is true if the job is public, the option to extend is set and the job expires within 5 days.
+ *
+ * @since $$next_version$$
+ *
+ * @param int|WP_Post $job The job or job ID.
+ *
+ * @return bool
+ */
+function job_manager_job_can_be_renewed( $job ) {
```
gikaragia: As you introduced a new class, maybe it makes sense to move these global functions inside the class?

gikaragia (commented):
Liked the idea of introducing a renewals class. Left some comments

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -592,6 +593,36 @@ private static function generate_notice_id_from_updates( $updates ) {
...
+	 *
+	 * @return array Notices.
+	 */
+	public static function paid_listings_renewal_notice( $notices ) {
+		if ( ! job_manager_is_wcpl_renew_compatible() ) {
+			$notices['wcpl_listing_renewal'] = [
+				'level'       => 'info',
+				'dismissible' => true,
+				'message'     => wp_kses_post(
+					__( 'Job listing renewal feature requires the latest version of WC paid listings, please update the plugin to use the feature.', 'wp-job-manager' )
```
gikaragia:
```suggestion
					__( 'Listing renewals require the latest version of WC Paid Listings. Please update the plugin to enable the feature.', 'wp-job-manager' )
```

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -592,6 +593,36 @@ private static function generate_notice_id_from_updates( $updates ) {
...
+				),
+			];
+		}
+		if ( ! job_manager_is_spl_renew_compatible() ) {
+			$notices['spl_listing_renewal'] = [
+				'level'       => 'info',
+				'dismissible' => true,
+				'message'     => wp_kses_post(
+					__( 'Job listing renewal feature requires the latest version of Simple paid listings, please update the plugin to use the feature.', 'wp-job-manager' )
```
gikaragia:
```suggestion
					__( 'Listing renewals require the latest version of Simple Paid Listings. Please update the plugin to enable the feature.', 'wp-job-manager' )
```

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -40,6 +40,13 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	 */
 	protected $preview_job;
 
+	/**
+	 * Renew listing helper instance.
+	 *
+	 * @var WP_Job_Manager_Form_Submit_Job
+	 */
+	public $renew_listing_helper;
```
gikaragia: I don't think that this is needed. We can access the instance from everywhere with `WP_Job_Manager_Helper_Listing_Renewal::instance`. I

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2eb330f440bcd4bf0dbb1d1d54de5306448ba041

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: Also fixed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2eb330f440bcd4bf0dbb1d1d54de5306448ba041

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2eb330f440bcd4bf0dbb1d1d54de5306448ba041

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: Yeah been thinking about that. Going with 1.41.0 for now since it's a new feature release. I had `$$next-version$$` in other places too. https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/df5a472d8a6edf285e5f2155b36af6235b2727d5

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2eb330f440bcd4bf0dbb1d1d54de5306448ba041

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2eb330f440bcd4bf0dbb1d1d54de5306448ba041

includes/helper/class-wp-job-manager-helper-listing-renewal.php
onubrooks: Agreed. Renamed the class and the file in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2eb330f440bcd4bf0dbb1d1d54de5306448ba041

includes/forms/class-wp-job-manager-form-submit-job.php
onubrooks: Ok. Changed in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/c8c7b1ac8a5ab88988bbc446b185fd4f373dbe41

wp-job-manager-functions.php
onubrooks: Hmmm, wanted them to be available everywhere cleanly.
Anyways, moved them to static methods in https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/0c324fa774dc02eb518c0c08223dd5e4be940fdf

includes/admin/class-wp-job-manager-admin-notices.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2e14c190f26c6ceef0944ae1cca25f1de849191f

includes/admin/class-wp-job-manager-admin-notices.php
onubrooks: https://github.com/Automattic/WP-Job-Manager/pull/2422/commits/2e14c190f26c6ceef0944ae1cca25f1de849191f

gikaragia (approved):
LGTM!

gikaragia:
### Documentation

This work needs to add a section to the [job dashboard page](https://wpjobmanager.com/document/the-job-dashboard/). We should mention the differences between 'Renew' and 'Relist':
- Renew extends the expiry date of a published listing by the job listing duration or if WCPL is installed by the package duration. The link to renew is available X amount of days before expiry where X is the 'Renewal Days' setting in Job Submission. When following the renewal flow, users can't edit the listing and there is no admin approval required.
- Relist happens when a listing is relisted after expiry. It has a different flow as the user has the chance to edit the listing and it requires admin approval to be completed if the appropriate setting is set.

We should also mention that the feature will be disabled if the user has a version of WCPL or SPL that is older than the latest one (3.0.0 and 2.0.0) - this is to not allow users to renew paid listings for free. Additionally, renewals are supported by WCPL and SPL and users can use their job packages or pay through PayPal/Stripe to renew.

PR #2427, opened 2023-05-22
Fix tests for PHP 8.0 and 8.1.

aaronfc:
Fixes test failures on GitHub for PHP 8.0 and 8.1

* Remove the `--1` which was forcing an old composer version.
* Add support for 8.x in composer.json
* Update phpunit to version supporting 8.x (required updating some tests to add `void` return type)
* Update `composer.lock`.

### Testing instructions
* See that the tests on this PR are green again for PHP 8.0 and 8.1.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -35,7 +35,7 @@ jobs:
           coverage:    none
 
       - name: Install PHP dependencies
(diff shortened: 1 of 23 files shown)


jom (approved):
Yay! 🥳 

I'm going to now work on one for `phpcs` on PHP 8.

edit: done in https://github.com/Automattic/WP-Job-Manager/pull/2428

PR #2430, opened 2023-05-25
Update `$company_tagline` check for PHP8+

### LINKED ISSUE #2424: Modifying `company_tagline` function for compatibility with PHP 8.1
We heard from a user [in the forums](https://wordpress.org/support/topic/php-8-1-compatibility-23/) that there is a deprecated PHP warning for the following code in in `wp-job-manager/wp-job-manager-template.php` line 1093:

```
if ( 0 === strlen( $company_tagline ) ) {
		return;
	}
```

The PHP warning states: 

`Deprecated: strlen(): Passing null to parameter #1 ($string) of type string, please use ! empty function for this`

Thanks!

interlineae:
Fixes # https://github.com/Automattic/WP-Job-Manager/issues/2424

As of PHP8, passing `null` to `strlen()` is deprecated. The `$company_tagline` var is assigned to the output of the `get_the_company_tagline` function, which returns `null` under certain conditions.

With this change, we're checking that `$company_tagline` exists: continue if it does, but if it's `null` or doesn't exist, return early.

### Testing instructions

* Check out branch
* Create a new job listing with a company tagline
* Ensure no tagline shows if one is not present, and a tagline shows if one _is_ present

### DIFF
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -1090,7 +1090,7 @@ function get_the_company_website( $post = null ) {
 function the_company_tagline( $before = '', $after = '', $echo = true, $post = null ) {
 	$company_tagline = get_the_company_tagline( $post );
 
-	if ( ! $company_tagline ) {
+	if ( is_null( $company_tagline ) || 0 === strlen( $company_tagline ) ) {
 		return;
 	}
 
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -1090,7 +1090,7 @@ function get_the_company_website( $post = null ) {
 function the_company_tagline( $before = '', $after = '', $echo = true, $post = null ) {
 	$company_tagline = get_the_company_tagline( $post );
 
-	if ( 0 === strlen( $company_tagline ) ) {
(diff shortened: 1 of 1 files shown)


aaronfc (commented):
Hey! It is looking good!

I am not sure if "company_tagline" could be "0" (a string with just a 0). In that case this condition wouldn't behave the same.

(Check [here](https://3v4l.org/UnNhh) if you are curious).

If that's something that we don't expect to happen I am ok with the fix. But if this could ever happen I would suggest doing a check like `is_null($company_tagline) || 0 === strlen($company_tagline)`.

Or maybe even `0 === strlen($company_tagline ?? '')`.

What do you think?

interlineae:
@aaronfc I checked for (string) `0` and couldn't replicate a use-case where it happened, but just in case, I'll go with

```
if ( is_null( $company_tagline ) || 0 === strlen($company_tagline) ) {
	return;
}
```

This way, we're checking for `null` or `0` specifically rather than existence.

Committed https://github.com/Automattic/WP-Job-Manager/pull/2430/commits/a0556d60393b8eaab6669282d02c4113e0f77756

PR #2431, opened 2023-05-26
Return all jobs in admin

### LINKED ISSUE #2423: Search in admin not working as expected. Results missing.
**Describe the bug**
The search feature from the admin is not giving the expected number of search results since version 1.40. Everything is good when rolling back to v1.39

**To Reproduce**
1. In the admin area, click on **Job listings** in the left menu
2. Enter a search string in the search field and click **Search Jobs**
3. Count the number of results
4. Clear the search and sort all jobs by alphabetical order of **Position**
5. Scroll down to your searched string and manually count the expected result

**Screen shots**
5 results when using the search:
Capture d’écran, le 2023-05-17 à 09 47 51

More than 5 results when sorting and checking manually:
Capture d’écran, le 2023-05-17 à 09 50 16

**Expected behavior**

Search results should include all instances of the string, as v1.39 did.

- [X] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [X] This bug happens with a default WordPress theme active.
- [X] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
- WordPress Version: 6.2.1
- WP Job Manager Version: 1.40.1
- PHP Version: 8.1.18
- Other important details: currently 1037 jobs in database

gikaragia:
Fixes #2423

* Adds a check if we are in admin and returns all jobs in case we are

### Testing instructions

* In WPJM have some jobs that are filled or expired
* Enable the 'Hide filled positions in job archives/search' or  'Hide expired positions in job archives/search' in settings
* Search the filled/expired jobs in admin
* Observe that the jobs are returned

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -658,7 +658,7 @@ class WP_Job_Manager_Post_Types {
 	 * @param WP_Query $query Query being processed.
 	 */
 	public function maybe_hide_filled_expired_job_listings_from_search( $query ) {
-		if ( ! $query->is_search() ) {
+		if ( ! $query->is_search() || is_admin() ) {
 			return;
 		}
 


gikaragia:
Hey @anaemnesis since it seems like you did a lot of investigation for #2367, I am now thinking that as part of #2367 we basically modify every query, even if it isn't related to WPJM. Do you think that we should also add other checks, to ensure that this modification to the query happens only to specific pages?

I think it is unlikely to have other side effects but maybe we should play it safe.

interlineae:
@gkaragia Per our chat, we're going to nix this and revert #2367 for an alternative implement. Is that right?

gikaragia:
Indeed we are going to look into a better approach for https://github.com/Automattic/WP-Job-Manager/issues/1884 in another PR. Resolving.

PR #2433, opened 2023-05-26
Allow searching multiple job locations at once

### LINKED ISSUE #2373: Search for Multiple Locations
Currently not able to search for multiple locations in the job board, search is limited to just one location (which also limits features like the Job Alerts and Embeddable Job Widget for showing more than one location)

Be able to search multiple locations, either comma or semicolon separated

### Additional context
Requested here: 6039635-zen

interlineae:
Fixes #2373

This change allows a job searcher to input multiple locations in a semi-colon-delimited list, where spaces are stripped, so "London;Denver" works, as does "London; Denver".

This change impacts any functionality that uses the `get_job_listings()` function, like the Job dashboard (frontend), Job widget, etc.

### Testing instructions

* Checkout branch
* Search for multiple job listings, single job listings, etc
* Look out for any performance issues. In my local environment, I used [`wp post generate`](https://developer.wordpress.org/cli/commands/post/generate/) to generate 1,100 job entries.

### Screenshot / Video

Screen Recording 2023-05-26 at 01 17 37 pm

### Documentation changes

This change introduces new functionality in WPJM core. The following KB URL will need an update to mention this new functionality:

- [ ] https://wpjobmanager.com/document/understanding-job-and-resume-search/

The section **Search by Location** needs additional explanatory text, which I think works best as a third paragraph:

> **_Search by Location_**
>
> This field works similarly to the Keywords search but only searches the information provided in the Location field of the Job Listing post.
> 
>  This field also acts as a filter to shortlist the results from the Keyword search. For example, If you search for “iOS Developer”, and Location “New York” then the page will show all the active job listings marching iOS or Developer or iOS Developer & New York as the location.
> 
> [New paragraph]
> 
> Multiple locations can be searched at once using a semi-colon delimited list. For example, if you wanted to search for positions available in "London, UK" and "New York, USA" at the same time, you would use `London, UK; New York, USA`.

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -98,7 +98,8 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 		if ( ! empty( $args['search_location'] ) ) {
 			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
 			$location_search    = [ 'relation' => 'OR' ];
-			$locations          = preg_split( '/"/', $args['search_location'] );
+			$locations          = explode( ';', $args['search_location'] );
+
 			foreach ( $locations as $location ) {
 				$location = trim( $location );
 				if ( ! empty( $location ) ) {
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -98,18 +98,20 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 		if ( ! empty( $args['search_location'] ) ) {
 			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
 			$location_search    = [ 'relation' => 'OR' ];
-			$locations          = explode( ',', $args['search_location'] );
+			$locations          = preg_split( '/"/', $args['search_location'] );
 			foreach ( $locations as $location ) {
-				$location          = trim( $location );
-				$location_subquery = [ 'relation' => 'OR' ];
-				foreach ( $location_meta_keys as $meta_key ) {
-					$location_subquery[] = [
-						'key'     => $meta_key,
-						'value'   => $location,
-						'compare' => 'like',
-					];
+				$location = trim( $location );
+				if ( ! empty( $location ) ) {
+					$location_subquery = [ 'relation' => 'OR' ];
+					foreach ( $location_meta_keys as $meta_key ) {
+						$location_subquery[] = [
+							'key'     => $meta_key,
+							'value'   => $location,
+							'compare' => 'like',
+						];
+					}
+					$location_search[] = $location_subquery;
 				}
-				$location_search[] = $location_subquery;
 			}
 
 			if ( $remote_position_search ) {
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -98,12 +98,18 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 		if ( ! empty( $args['search_location'] ) ) {
 			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
 			$location_search    = [ 'relation' => 'OR' ];
-			foreach ( $location_meta_keys as $meta_key ) {
-				$location_search[] = [
-					'key'     => $meta_key,
-					'value'   => $args['search_location'],
-					'compare' => 'like',
-				];
+			$locations          = explode( ',', $args['search_location'] );
+			foreach ( $locations as $location ) {
+				$location          = trim( $location );
+				$location_subquery = [ 'relation' => 'OR' ];
+				foreach ( $location_meta_keys as $meta_key ) {
+					$location_subquery[] = [
+						'key'     => $meta_key,
+						'value'   => $location,
+						'compare' => 'like',
+					];
+				}
+				$location_search[] = $location_subquery;
 			}
 
 			if ( $remote_position_search ) {
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -592,7 +592,7 @@ class WP_Job_Manager_Post_Types {
 		if ( ! empty( $input_search_location ) ) {
 			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
 			$location_search    = [ 'relation' => 'OR' ];
-			$locations          = preg_split( '/"/', $input_search_location );
+			$locations          = explode( ';', $input_search_location );
 			foreach ( $locations as $location ) {
 				$location = trim( $location );
 				if ( ! empty( $location ) ) {
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -592,12 +592,20 @@ class WP_Job_Manager_Post_Types {
 		if ( ! empty( $input_search_location ) ) {
 			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
 			$location_search    = [ 'relation' => 'OR' ];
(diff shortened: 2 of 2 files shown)


wp-job-manager-functions.php
```diff
@@ -98,12 +98,18 @@ function get_job_listings( $args = [] ) {
...
-				$location_search[] = [
-					'key'     => $meta_key,
-					'value'   => $args['search_location'],
-					'compare' => 'like',
-				];
+			$locations          = explode( ',', $args['search_location'] );
+			foreach ( $locations as $location ) {
+				$location          = trim( $location );
+				$location_subquery = [ 'relation' => 'OR' ];
+				foreach ( $location_meta_keys as $meta_key ) {
```
interlineae: With this change, we have a `foreach` within a `foreach`. We're looping through locations, and for each location, looping through each of the `$location_meta_keys`. In my testing, this was fine performance-wise, but I'm also not sure about the scale of some sites using WPJM, e.g., do we have sites with tens of thousands of job listings?

If this *is* a performance concern, I was also thinking of storing results in arrays and then doing any results/comparisons there. I think there would still be two `foreach` somewhere, though.

interlineae:
One other thought after speaking with Jake: I don't _think_ it's a concern, but if someone tries to search for, say, "London, Ontario" (Meaning the city of London, in the province of Ontario), the comma-delimited nature of the implementation means that it will search "London", and "Ontario" as separate locations.

The reason I don't _think_ this is a concern is that in my testing, it's not possible to search "London, Ontario" anyway, but maybe it is (and I've just not quite done something right). Something to think about when testing.

Gnodesign:
@anaemnesis Check #2261 #2287 to avoid making the same mistakes. When this multi-location search feature was merged, it caused more problems than it solved.

wp-job-manager-functions.php
fjorgemota: Yeah, I don't think this will be a performance problem, at least not on the PHP side, unless the user starts typing hundreds/thousands of search locations, all split with `,`, and even then it would be a problem mostly on the database side, because we would have the number of locations specified by the user multiplied by 3.

And eventually that would be a problem... for the database, both because of the big number of conditions to compare (imagine thousands of `LIKE` conditions in the database..it doesn't seen fun), as well as because of the size of the query, given that MySQL limits the size of the query (however, this number is quite big..by default it is 64MB on MySQL 8, and 4MB on MySQL 5.7). 

Looking here, for `London, Denver`, this is the query that WP generates:

```sql
SELECT SQL_CALC_FOUND_ROWS  wp_posts.ID
					FROM wp_posts  INNER JOIN wp_postmeta ON ( wp_posts.ID = wp_postmeta.post_id )  INNER JOIN wp_postmeta AS mt1 ON ( wp_posts.ID = mt1.post_id )
					WHERE 1=1  AND ( 
  ( 
    ( 
      ( wp_postmeta.meta_key = 'geolocation_formatted_address' AND wp_postmeta.meta_value LIKE '%Denver%' ) 
      OR 
      ( wp_postmeta.meta_key = '_job_location' AND wp_postmeta.meta_value LIKE '%Denver%' ) 
      OR 
      ( wp_postmeta.meta_key = 'geolocation_state_long' AND wp_postmeta.meta_value LIKE '%Denver%' )
    ) 
    OR 
    ( 
      ( mt1.meta_key = 'geolocation_formatted_address' AND mt1.meta_value LIKE '%London%' ) 
      OR 
      ( mt1.meta_key = '_job_location' AND mt1.meta_value LIKE '%London%' ) 
      OR 
      ( mt1.meta_key = 'geolocation_state_long' AND mt1.meta_value LIKE '%London%' )
    )
  )
) AND wp_posts.post_type = 'job_listing' AND ((wp_posts.post_status = 'publish' OR wp_posts.post_status = 'expired'))
					GROUP BY wp_posts.ID
					ORDER BY wp_posts.menu_order ASC, wp_posts.post_date DESC, wp_posts.ID DESC
					LIMIT 0, 10
```

So I think we're pretty safe in this regard.

fjorgemota (commented):
Looks good and it is working well! However, I found a small related issue that I think it would be nice for us to solve: On the frontend, you can get an RSS feed for a given list of queries, and that RSS feed also includes the search location, naturally.

With this change, we're supporting getting the list of filters on the UI, but not on the RSS feed, and this can be considered an issue ("why is 'London, Denver' working on the UI but not on the RSS feed???"). 

To quickly fix this, you can apply the same change you applied here in the `job_feed` method in `includes/class-wp-job-manager-post-types.php`. Don't ask me why that feed doesn't use the same logic implemented here.. I don't know. :P

interlineae:
@Gnodesign @fjorgemota I've committed a change that delimits by double quotation mark rather than comma https://github.com/Automattic/WP-Job-Manager/pull/2433/commits/d04771fd65d340d6533be53831370c2fd249c928. This change does _not_ delimit by single quotation mark, so searching a location like `L'Austin, United Kingdom` works without having to escape characters.

With this change:

- Entering a search like `London, United Kingdom` without double quotes will search for that as a single location (it won't return `London` as one result and `United Kingdom` as another
- Entering a search term like `"London, United Kingdom", "Edinburgh, United Kingdom"` will only return results for both locations, with each as a single location.
- The `,` separating the two entries is optional. As this delimits by `"` it's possible to use anything as a visual separator, including a space. A comma just looks nice. Here's a gif to demonstrate:

Screen Recording 2023-06-01 at 04 33 30 pm

interlineae:
@fjorgemota after speaking with @Luchadores we're going to mention the `;` delimiter for multiple location search in the docs, which is here https://wpjobmanager.com/document/understanding-job-and-resume-search/.

Just needs a further review from you and should we not notice anything else awry, we should be good.

wp-job-manager-functions.php
```diff
@@ -98,12 +98,21 @@ function get_job_listings( $args = [] ) {
...
 			$location_search    = [ 'relation' => 'OR' ];
-			foreach ( $location_meta_keys as $meta_key ) {
-				$location_search[] = [
-					'key'     => $meta_key,
-					'value'   => $args['search_location'],
-					'compare' => 'like',
-				];
+			$locations          = explode( ';', $args['search_location'] );
+
+			foreach ( $locations as $location ) {
```
rowasc, line 103: 🛸 why do we need to copy the same code here as in includes/class-wp-job-manager-post-types.php ? (I am 100% ignorant about this codebase, just passing by and curious!)

fjorgemota: That is a good question. 

Looking here, I think it is because the `get_job_listings` necessarily return the `WP_Query` AND has some extra actions and hooks for customization, while the logic in `includes/class-wp-job-manager-post-types.php` uses `query_post` to change the global object. 

I think we can eventually replace it in the future...but for now, I'd not recommend doing it because I think this might cause more problems than anything else. These sorts of changes are..complicated.

fjorgemota (approved):
Looks good and works well! Good job! :D

wp-job-manager-functions.php
interlineae: > Looking here, I think it is because the get_job_listings necessarily return the WP_Query AND has some extra actions and hooks for customization, while the logic in includes/class-wp-job-manager-post-types.php uses query_post to change the global object.

This is what I understood as well, looking at it last night. There are things specific to the RSS Feed rendering that results in the difference between the two:

https://github.com/Automattic/WP-Job-Manager/blob/b1184b2836bba567218281c0a13f68694c10dd55/wp-job-manager-functions.php#L101

https://github.com/Automattic/WP-Job-Manager/blob/b1184b2836bba567218281c0a13f68694c10dd55/includes/class-wp-job-manager-post-types.php#L595

Namely, `query_posts` in the one instance.

rowasc: ohhh ok, that makes sense to me :) thank you!

gikaragia:
### Documentation

We should probably mention that users can search for multiple locations using `;` in [this page](https://wpjobmanager.com/document/understanding-job-and-resume-search/).

PR #2434, opened 2023-05-26
Remove custom config for Webpack loaders

renatho:

* @gkaragia  realized an issue while building the package for the [last release](https://github.com/Automattic/WP-Job-Manager/releases/tag/1.40.2) that it wasn't importing the fonts correctly.
  * I built it in my env (without running a new `npm install`) and it worked, we saved the package for the plugin release.
  * After running `npm install` I started seeing the same problem.
  * @gkaragia noticed that it happened after [updating the `@wordpress/scripts`](https://github.com/Automattic/WP-Job-Manager/pull/2389/files) (which was a big jump).
* So after an investigation, I fixed it by removing our custom config for Webpack loaders, which was probably fixing something for old versions.

### Testing instructions

* Build the plugin, and make sure no errors happen.
* Install it on a WordPress site.
* Make sure the fonts are working (an example is the Job Manager icon in WP Admin).
* Make sure the `gif` file is working (On the jobs page, when you filter, a spinner should be displayed).
* As a bonus, you could download the last release package, and compare to see if you see any problem with the new package.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,5 @@
-const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
-const path = require( 'path' );
+const defaultConfig = require('@wordpress/scripts/config/webpack.config');
+const path = require('path');
 
 const files = {
 	'js/admin': 'js/admin.js',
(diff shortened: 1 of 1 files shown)


gikaragia (approved):
🚀

PR #2446, opened 2023-06-02
Remove feature github template

gikaragia:
To be able to triage new issues, we need every new issue to have a label. As I reviewed the "Feature Request" template, I noticed that it doesn't apply any labels.

Instead of adding a label, I am thinking that maybe we don't need the template at all. We should probably just have bug reports and enhancements.


onubrooks (commented):
I'm thinking that a feature request and an enhancement really mean different things right? Do you suggest that there should be no feature requests or every feature request should be filled under enhancement?

Screenshot 2023-06-05 at 14 15 59

gikaragia:
Yes they are a bit different but they can be grouped into a single flow. Normally, users will use [our form](https://wpjobmanager.com/2022/05/10/requesting-features-for-wp-job-manager/) in  WPJM.com so I would expect that this flow would be mainly used by HEs.

I updated the phrasing a bit to include new features too.

onubrooks (approved):
> I updated the phrasing a bit to include new features too.

Nice!

PR #2449, opened 2023-06-06
Add methods to hide filled and expired posts

### LINKED ISSUE #1884: WordPress search returns expired job

interlineae:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/1884

* Adds three new methods: (1) `get_filled_job_listings()`, (2), `get_expired_job_listings` and (3) `maybe_hide_filled_expired_job_listings_from_search`
* The first two methods only run if the "Hide Filled Jobs" or "Hide Expired Jobs" options are enabled. If they aren't, an empty array is returned.
* The third method returns early if the "Hide Filled Jobs" and "Hide Expired Jobs" options are disabled. If they aren't, a `post__not_in` parameter is applied to `$query->set` when **not** in WP Admin, and when on a WP search results page.

Unlike the previous implementation, this one relies on `get_posts()` and `meta_query` but without the expensive `OR` relation.

Because `expired` jobs already have their post status updated to `expired` if a post's expiry date is set in the past or elapses, we don't need to `meta_query` for expired posts but can instead look for `job_listings` with an `expired` post status.

We don't have the same luxury with `filled` posts, but we can do a singular `meta_query` without any relations to retrieve these.

My understanding is that these changes should be far less expensive than the previous implementation.

To further help with performance, the results of `get_filled_job_listings` and `get_expired_job_listings` are stored in a transient that is expired every 24 hours, which lines up with the various [job expiration methods](https://github.com/Automattic/WP-Job-Manager/blob/acb96b9e0ff499338f144732519f6d51b1a4dbca/includes/class-wp-job-manager-post-types.php#L1018).

Either transient is then cleared on post update if: (1) we're looking at a job listing, (2) the '_filled' post meta has been updated, or (3) the post status has changed from `publish` to `expired`, `trash` or vice-versa.

### Testing instructions

* Check out branch
* Create a few job posts, setting one or some to `Expired` with an expiry date in the past, or `Filled` by checking the `Filled` checkbox.
* Check transients to ensure `hide_filled_jobs_transient` and `hide_expired_jobs_transient` is not set. With WP CLI these can be checked with `wp transient get hide_filled_jobs_transient` and `wp transient get hide_expired_jobs_transient`
* Navigate to **Job Listings --> Settings --> Job Listings** and enable (as desired), the **Hide Filled Positions** and **Hide expired listings in your job archives/search** options
* Check your transient again, which should now be populated with the post IDs of the posts you marked as `expired` or `filled`
* Check WP Search results to make sure these posts, and only these posts, are excluded

If you would like to test on a large dataset, WP CLI can be used to bulk create posts:

**Bulk create `Expired` posts**

```
wp post generate --count=1000 --post_type=job_listing --post_status=expired
```

**Bulk generate `Filled` posts**

`Filled` posts are a little more troublesome because we're working with meta values, and `wp post generate`, as well as `wp post create`, do not touch these values.

So, you could do something like:

```
# generate our posts
wp post generate --count=1000 --post_type=job_listing --post_status=publish

# grab the IDs of some arbitrary number of posts, say, 100, and update the post meta
for post_id in $(wp post list --post_type=job_listing --post_status=publish --numberposts=100 --fields=ID); do
    wp post meta update \
        $post_id \
        _filled \
        1 \
       && echo "Post ID $post_id changed"
done
```

The above can be added to a file on your environment, like `create-filled-posts.sh`. Make sure to `chmod +x $file` and then `./$file.sh` to execute the `for` loop and create your posts.

**Note** that when creating posts programmatically you may need to manually clear the transients.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -722,11 +722,10 @@ class WP_Job_Manager_Post_Types {
 	 * @return void
 	 */
 	public function maybe_hide_filled_expired_job_listings_from_search( WP_Query $query ): void {
-		if ( ! get_option( 'job_manager_hide_filled_positions' ) ) {
-			return;
-		}
+		$hide_filled_positions = get_option( 'job_manager_hide_filled_positions' );
+		$hide_expired          = get_option( 'job_manager_hide_expired' );
 
-		if ( ! get_option( 'job_manager_hide_expired' ) ) {
+		if ( ! $hide_filled_positions && ! $hide_expired ) {
 			return;
 		}
 
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -723,32 +723,6 @@ class WP_Job_Manager_Post_Types {
 	}
 
 	/**
-	 * Hide expired listings from the search results.
-	 *
-	 * @param array $listings
-	 *
-	 * @return array
-	 */
-	public function hide_expired( $listings ) {
-
-		// Get all expired listings.
-		$expired_listings = get_posts(
-			[
-				'post_type'   => 'job_listing',
-				'post_status' => 'expired',
-			]
-		);
-
-		// Remove expired listings from the search results.
-		foreach ( $expired_listings as $listing ) {
-			unset( $listings[ $listing->ID ] );
-		}
-
-		return $listings;
-
-	}
-
-	/**
 	 * Adds query arguments in order to make sure that the feed properly queries the 'job_listing' type.
 	 *
 	 * @param WP_Query $wp
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -88,7 +88,8 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'transition_post_status', [ $this, 'track_job_submission' ], 10, 3 );
 
 		add_action( 'parse_query', [ $this, 'add_feed_query_args' ] );
-		add_action( 'pre_get_posts', [ $this, 'maybe_hide_filled_expired_job_listings_from_search' ] );
+		add_action( 'pre_get_posts', [ $this, 'maybe_hide_filled_job_listings' ] );
+		add_action( 'pre_get_posts', [ $this, 'maybe_hide_expired_job_listings' ] );
 
 		// Single job content.
 		$this->job_content_filter( true );
@@ -691,17 +692,16 @@ class WP_Job_Manager_Post_Types {
 	}
 
 	/**
-	 * Maybe exclude expired and/or filled job listings from search and archive pages.
+	 * Maybe exclude filled job listings from search and archive pages.
 	 *
 	 * @param $query WP_Query $query
 	 *
 	 * @return void
 	 */
-	public function maybe_hide_filled_expired_job_listings_from_search( WP_Query $query ): void {
+	public function maybe_hide_filled_job_listings( WP_Query $query ): void {
 		$hide_filled_positions = get_option( 'job_manager_hide_filled_positions' );
-		$hide_expired          = get_option( 'job_manager_hide_expired' );
 
-		if ( ! $hide_filled_positions && ! $hide_expired ) {
+		if ( ! $hide_filled_positions ) {
 			return;
 		}
 
@@ -711,11 +711,64 @@ class WP_Job_Manager_Post_Types {
 			&& ( $query->is_search() || $query->is_archive() )
 		) {
 
-			if ( $hide_expired ) {
-				$query->set( 'post_status', 'publish' );
-			}
-
 			$query->set( 'post__not_in', $this->get_filled_job_listings() );
+		}
+	}
+
+	/**
+	 * Maybe exclude expired job listings from search and archive pages.
+	 *
+	 * @param $query WP_Query $query
+	 *
+	 * @return void
+	 */
+	public function maybe_hide_expired_job_listings( WP_Query $query ): void {
+		$hide_expired = get_option( 'job_manager_hide_expired' );
+
+		if ( ! $hide_expired ) {
+			return;
+		}
+
+		if (
+			! is_admin()
+			&& $query->is_main_query()
+			&& ( $query->is_search() || $query->is_archive() )
+		) {
+
+			$this->make_expired_private();
+
+			add_action( 'posts_selection', [ $this, 'make_expired_public' ] );
+		}
+	}
+
+	/**
+	 * Make the expired post status public.
+	 *
+	 * @return void
+	 * @internal
+	 */
+	public function make_expired_public(): void {
+
+		global $wp_post_statuses;
+
+		if ( isset( $wp_post_statuses['expired'] ) ) {
+			$wp_post_statuses['expired']->public = true;
+		}
+
+	}
+
+	/**
+	 * Make the expired post status private.
+	 *
+	 * @return void
+	 * @internal
+	 */
+	public function make_expired_private() {
+
+		global $wp_post_statuses;
+
+		if ( isset( $wp_post_statuses['expired'] ) ) {
+			$wp_post_statuses['expired']->public = false;
 		}
 	}
 
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -87,8 +87,7 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'transition_post_status', [ $this, 'track_job_submission' ], 10, 3 );
 
 		add_action( 'parse_query', [ $this, 'add_feed_query_args' ] );
-		add_action( 'pre_get_posts', [ $this, 'maybe_hide_filled_job_listings_from_search' ] );
-		add_action( 'pre_get_posts', [ $this, 'maybe_hide_expired_job_listings_from_search' ] );
+		add_action( 'pre_get_posts', [ $this, 'maybe_hide_filled_expired_job_listings_from_search' ] );
 
 		// Single job content.
 		$this->job_content_filter( true );
@@ -662,63 +661,80 @@ class WP_Job_Manager_Post_Types {
 	}
 
 	/**
-	 * Maybe hide filled job listings from search.
+	 * Get filled jobs.
 	 *
-	 * @param WP_Query $query Query object.
+	 * @return array
 	 */
-	public function maybe_hide_filled_job_listings_from_search( $query ) {
-		if ( ! is_admin() && $query->is_search() && 1 === absint( get_option( 'job_manager_hide_filled_positions' ) ) ) {
-			$meta_query = $query->get( 'meta_query' );
-
-			if ( ! is_array( $meta_query ) ) {
-				$meta_query = [];
-			}
-
-			$meta_query[] = [
-				'relation' => 'OR',
-				[
-					'key'     => '_filled',
-					'value'   => '0',
-					'compare' => '=',
-				],
-				[
-					'key'     => '_filled',
-					'compare' => 'NOT EXISTS',
-				],
-			];
-
-			$query->set( 'meta_query', $meta_query );
+	public function get_filled_job_listings(): array {
+		if ( ! get_option( 'job_manager_hide_filled_positions' ) ) {
+			return [];
 		}
+
+		$filled_jobs_transient = get_transient( 'hide_filled_jobs_transient' );
+		if ( false === $filled_jobs_transient ) {
+			$filled_jobs_transient = get_posts(
+				[
+					'post_type'  => 'job_listing',
+					'fields'     => 'ids',
+					'meta_query' => [
+						[
+							'key'     => '_filled',
+							'value'   => '1',
+							'compare' => '=',
+						],
+					],
+				]
+			);
+			set_transient( 'hide_filled_jobs_transient', $filled_jobs_transient, DAY_IN_SECONDS );
+		}
+		return $filled_jobs_transient;
 	}
 
 	/**
-	 * Maybe hide expired job listings from search.
+	 * Get expired jobs.
 	 *
-	 * @param WP_Query $query Query object.
+	 * @return array
 	 */
-	public function maybe_hide_expired_job_listings_from_search( $query ) {
-		if ( ! is_admin() && $query->is_search() && 1 === absint( get_option( 'job_manager_hide_expired' ) ) ) {
-			$meta_query = $query->get( 'meta_query' );
+	public function get_expired_jobs_listings(): array {
+		if ( ! get_option( 'job_manager_hide_expired' ) ) {
+			return [];
+		}
 
-			if ( ! is_array( $meta_query ) ) {
-				$meta_query = [];
+		$expired_jobs_transient = get_transient( 'hide_expired_jobs_transient' );
+		if ( false === $expired_jobs_transient ) {
+			$expired_jobs_transient = get_posts(
+				[
+					'post_type'   => 'job_listing',
+					'post_status' => 'expired',
+					'fields'      => 'ids',
+				]
+			);
+			set_transient( 'hide_expired_jobs_transient', $expired_jobs_transient, DAY_IN_SECONDS );
+		}
+		return $expired_jobs_transient;
+	}
+
+	/**
+	 * Maybe exclude expired and/or filled job listings from search.
+	 *
+	 * @param $query WP_Query $query
+	 *
+	 * @return void
+	 */
+	public function maybe_hide_filled_expired_job_listings_from_search( WP_Query $query ): void {
+		if ( ! get_option( 'job_manager_hide_filled_positions' ) ) {
+			return;
+		}
+
+		if ( ! get_option( 'job_manager_hide_expired' ) ) {
+			return;
+		}
+
+		if ( ! is_admin() && $query->is_search() ) {
+			$jobs_to_exclude = array_merge( $this->get_filled_job_listings(), $this->get_expired_jobs_listings() );
+			if ( ! empty( $jobs_to_exclude ) ) {
+				$query->set( 'post__not_in', $jobs_to_exclude );
 			}
-
-			$meta_query[] = [
-				'relation' => 'OR',
-				[
-					'key'     => '_job_expires',
-					'value'   => gmdate( 'Y-m-d' ),
-					'compare' => '>=',
-					'type'    => 'DATE',
-				],
-				[
-					'key'     => '_job_expires',
-					'compare' => 'NOT EXISTS',
-				],
-			];
-
-			$query->set( 'meta_query', $meta_query );
 		}
 	}
 
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -87,7 +87,7 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'transition_post_status', [ $this, 'track_job_submission' ], 10, 3 );
 
 		add_action( 'parse_query', [ $this, 'add_feed_query_args' ] );
-		add_action( 'pre_get_posts', [ $this, 'maybe_hide_filled_job_listings_from_search' ] );
+		add_action( 'pre_get_posts', [ $this, 'maybe_hide_filled_expired_job_listings_from_search' ] );
 
 		// Single job content.
 		$this->job_content_filter( true );
@@ -391,28 +391,13 @@ class WP_Job_Manager_Post_Types {
 		add_feed( self::get_job_feed_name(), [ $this, 'job_feed' ] );
 
 		/**
-		 * This code checks if (1) we're on a search or (2) the option to hide expired job listings is enabled.
-		 * If either condition is false, we're going to set the register_post_status() 'public' arg to false,
-		 * otherwise it will be true.
-		 *
-		 * This will prevent single expired job listings from showing a '404', instead of an
-		 * 'Expired' notice, when viewing the single job listing and not logged in.
-		 *
-		 * This will also address historical issues stemming from addressing the issue raised here
-		 * https://github.com/Automattic/WP-Job-Manager/issues/1884.
-		 */
-		$is_site_search    = ( isset( $_GET['s'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		$is_expired_hidden = (bool) get_option( 'job_manager_hide_expired' );
-		$is_expired_public = ! $is_site_search || ! $is_expired_hidden;
-
-		/**
 		 * Post status
 		 */
 		register_post_status(
 			'expired',
 			[
 				'label'                     => _x( 'Expired', 'post status', 'wp-job-manager' ),
-				'public'                    => $is_expired_public,
+				'public'                    => true,
 				'protected'                 => true,
 				'exclude_from_search'       => true,
 				'show_in_admin_all_list'    => true,
@@ -711,10 +696,11 @@ class WP_Job_Manager_Post_Types {
 	 *
 	 * @return void
 	 */
-	public function maybe_hide_filled_job_listings_from_search( WP_Query $query ): void {
+	public function maybe_hide_filled_expired_job_listings_from_search( WP_Query $query ): void {
 		$hide_filled_positions = get_option( 'job_manager_hide_filled_positions' );
+		$hide_expired          = get_option( 'job_manager_hide_expired' );
 
-		if ( ! $hide_filled_positions ) {
+		if ( ! $hide_filled_positions && ! $hide_expired ) {
 			return;
 		}
 
@@ -723,6 +709,10 @@ class WP_Job_Manager_Post_Types {
 			&& $query->is_main_query()
 			&& ( $query->is_search() || $query->is_archive() )
 		) {
+
+			if ( $hide_expired ) {
+				$query->set( 'post_status', 'publish' );
+			}
 
 			$query->set( 'post__not_in', $this->get_filled_job_listings() );
 		}
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -401,9 +401,9 @@ class WP_Job_Manager_Post_Types {
 		 * This will also address historical issues stemming from addressing the issue raised here
 		 * https://github.com/Automattic/WP-Job-Manager/issues/1884.
 		 */
-		$is_site_search        = ( isset( $_GET['s'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		$is_expired_searchable = (bool) get_option( 'job_manager_hide_expired' );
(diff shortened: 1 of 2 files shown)


wp-job-manager-functions.php
```diff
@@ -1712,3 +1712,26 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+		}
+
+		$post_status = [ 'publish', 'expired', 'trash' ];
+		if ( in_array( $post->post_status, $post_status, true ) ) {
+			delete_transient( 'hide_expired_jobs_transient' );
+		}
+	}
+
+}
+add_action( 'save_post', 'delete_job_listings_transients_on_save', 10, 2 );
```
interlineae: These transients need to be cleared whenever:

- A post is marked as `Filled` (edit: or unset from `filled`)
- A post is changed from `publish` to `expired` or `trash` post status

I'm uncertain about this implementation and am quite open to feedback about improvements or changes to this implementation.

It's just struck me that the `filled` transient also needs to be cleared if a job is set from `filled` to not filled, but I'll wait on review before making any further changes in case this isn't the right direction as a whole.

includes/class-wp-job-manager-post-types.php
```diff
@@ -659,6 +660,83 @@ public function job_feed() {
...
+	 */
+	public function maybe_hide_filled_expired_job_listings_from_search( WP_Query $query ): void {
+		$hide_filled_positions = get_option( 'job_manager_hide_filled_positions' );
+		$hide_expired          = get_option( 'job_manager_hide_expired' );
+
+		if ( ! $hide_filled_positions && ! $hide_expired ) {
+			return;
+		}
+
+		if ( ! is_admin() && $query->is_search() ) {
```
yscik: We should add a post type check `'job_listing' === $query->get( 'post_type' )`, and also best to add `$query->is_main_query()`.

It'd be nice if this also worked for the archive pages (docs: https://wpjobmanager.com/document/enabling-full-template-support/, recent report: p1687165209914009-slack-C6GGX896G). The check for that would be `$query->is_archive()`

For the expired listings, do you think it's worth doing the transient and dedicated query? Seems like it would be simpler to filter the post status on the main query with something like `$query->set( 'post_status', 'publish' );` 

Also I'm not sure if we have some other filter adding back the `expired` status for this query, as it's [registered with `'exclude_from_search' => true`](https://github.com/Automattic/WP-Job-Manager/blob/38019efba98ce55b2c881f30003a3b8059f9e1e1/includes/class-wp-job-manager-post-types.php#L395). In WP_Query that's respected when the queried post status is `'any'` — might be that one is not set by default for search/archive queries?

interlineae: @yscik 

> We should add a post type check 'job_listing' === $query->get( 'post_type' ), and also best to add $query->is_main_query().

> It'd be nice if this also worked for the archive pages (docs: https://wpjobmanager.com/document/enabling-full-template-support/, recent report: p1687165209914009-slack-C6GGX896G). The check for that would be $query->is_archive()

Ah right, to avoid secondary (or other) queries from being impacted! 

> For the expired listings, do you think it's worth doing the transient and dedicated query? Seems like it would be simpler to filter the post status on the main query with something like $query->set( 'post_status', 'publish' );

That... would be far simpler. Too much tunnel vision.

> Also I'm not sure if we have some other filter adding back the expired status for this query, as it's [registered with 'exclude_from_search' => true](https://github.com/Automattic/WP-Job-Manager/blob/38019efba98ce55b2c881f30003a3b8059f9e1e1/includes/class-wp-job-manager-post-types.php#L395). In WP_Query that's respected when the queried post status is 'any' — might be that one is not set by default for search/archive queries?

You would think, but there's a bug where this isn't the behaviour that happens https://core.trac.wordpress.org/ticket/44737. As a result, what should be a simple `exclude_from_search` isn't. It's possible to get this option to work by changing the parameter on the post type registration:

https://github.com/Automattic/WP-Job-Manager/blob/a09f7fad82dd3b7080a4c848fc08d91afbaf5755/includes/class-wp-job-manager-post-types.php#L369

But this isn't great, and @aaronfc and I decided to go in the direction we did originally, which I've continued with here.

One other thing I noticed is that the `get_filled_job_listings()` method was limited to something like 5 posts without the `posts_per_page` parameter, so I've set that hesitantly to `-1` and kept in the transient code for it (but removed for expired posts). I'm not aware of a way to avoid this as the `_filled` meta is, well, meta. And, if we're excluding all posts, we need all posts...

I'm not sure if maybe it would be worth running the `meta_query` directly on the `$query->set`, but my feeling is that this then borks any attempt at caching:

```
$query->set( 'meta_query', array(
        array(
            'key'     => '_filled',
            'value'   => '1',
            'compare' => '!='
        ),
    )
);
```

Changes committed here:

- https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/d1d04d450117cb0c207f0aa4476dc63b8b2206f0
- https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/3fedb2bc7c13d6254316e4a6a0dd4e69be6ccdb1

includes/class-wp-job-manager-post-types.php
```diff
@@ -659,6 +660,69 @@ public function job_feed() {
...
+
+		if ( ! $hide_filled_positions && ! $hide_expired ) {
+			return;
+		}
+
+		if (
+			! is_admin()
+			&& $query->is_main_query()
+			&& $query->is_search()
+			|| $query->is_archive() && 'job_listing' === $query->get( 'post_type' )
```
yscik: I think this needs some grouping for the logic to be right:

```suggestion
			! is_admin()
			&& $query->is_main_query()
			&& ( $query->is_search() || $query->is_archive() ) 
			&& 'job_listing' === $query->get( 'post_type' )
```

wp-job-manager-functions.php
yscik: The approach makes sense to me. I'm a bit tempted to add/remove the post ID directly to the cache instead of re-running the query after every change. That could be faster but a bit more risky.

[Looks like the meta is set with `update_post_meta`](https://github.com/Automattic/WP-Job-Manager/blob/38019efba98ce55b2c881f30003a3b8059f9e1e1/includes/class-wp-job-manager-shortcodes.php#L214), so the `save_post` hook won't be enough and we should hook into the `updated_post_meta` action.

Also found this class that could be a good home for some of the caching:
[WP_Job_Manager_Cache_Helper](https://github.com/Automattic/WP-Job-Manager/blob/edfd346146b624b13f09b6a8c1e3881b98a40187/includes/class-wp-job-manager-cache-helper.php)

And I don't know much about it, but could using the object cache via [`wp_cache_add/delete`](https://developer.wordpress.org/reference/functions/wp_cache_add/) be a better fit? Though looks like the cache helper above also uses transients.

includes/class-wp-job-manager-post-types.php
```diff
@@ -659,6 +660,69 @@ public function job_feed() {
...
+	public function get_filled_job_listings(): array {
+		if ( ! get_option( 'job_manager_hide_filled_positions' ) ) {
+			return [];
+		}
+
+		$filled_jobs_transient = get_transient( 'hide_filled_jobs_transient' );
+		if ( false === $filled_jobs_transient ) {
+			$filled_jobs_transient = get_posts(
+				[
+					'post_type'      => 'job_listing',
```
yscik, line 677: Another potential optimization here could be filtering by post status, so we don't add posts that wouldn't be displayed anyway by the main query. (Eg trashed/draft/expired? ones). Not sure if there are sites keeping around a large number of non-active listings where it could matter.

interlineae: It looks like `publish` is the default (or `inherit` if we're looking at an attachment) if no argument is passed for `post_status`.

https://github.com/WordPress/wordpress-develop/blob/bea4307daa21a3f97d159898b0ac676332e0e7de/src/wp-includes/post.php#L2397

```
$parsed_args = wp_parse_args( $args, $defaults );
if ( empty( $parsed_args['post_status'] ) ) {
	$parsed_args['post_status'] = ( 'attachment' === $parsed_args['post_type'] ) ? 'inherit' : 'publish';
}
```

However, to be safe, because you never know what sites a WPJM user might have, I've gone ahead and set `get_posts` specifically to pull only `post_status => publish`.

Committed https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/fc8e201e12ce8dff5d75e59beced538f93b315c8

includes/class-wp-job-manager-post-types.php
interlineae: This breaks the method on WP search results pages, as the `post_type` returned is `any` rather than `job_listing` (i.e., we aren't necessarily searching within WP Job Manager itself, but using WordPress' default search and excluding results from its query.

That said, when testing in Jobify, I see that `expired` and `filled` jobs are hidden from taxonomy pages, and archive pages, like `/jobs`, hide these items by default as well. I'm not sure that these additions are actually needed, on further testing -- were you seeing behaviour on your end where they were showing, but shouldn't have been?

wp-job-manager-functions.php
interlineae: @yscik Do you have any more context on this?

> I'm a bit tempted to add/remove the post ID directly to the cache instead of re-running the query after every change. That could be faster but a bit more risky.

Namely, retrieving the post ID for a newly updated post without running a query. My thinking about this is, wouldn't we need to have at least 1 query running? Say:

1. One initial query to cache post IDs that we could add to an array
2. Another query to retrieve the ID of the post that was updated (if the `_filled` meta key value was changed)
3. An `array_push()` or something to add any new IDs to the array in step 1

Or doing some kind of check to see if an ID exists in the step 1 array, or just some kind of array comparison generally. Basically, I'm unsure how we know what new IDs to add if we aren't comparing against something and without at least 1 query running.

includes/class-wp-job-manager-post-types.php
```diff
@@ -390,14 +390,29 @@ public function register_post_types() {
...
+
 		/**
 		 * Post status
 		 */
 		register_post_status(
 			'expired',
 			[
 				'label'                     => _x( 'Expired', 'post status', 'wp-job-manager' ),
-				'public'                    => true,
+				'public'                    => $is_expired_public,
```
interlineae: @jom Here's the refactored implementation for setting `'public` on the post status registration. The logic is slightly different than discussed but noted in the comment. In testing:

- Job listings in WP Admin show and are counted as expected
- Expired Jobs are hidden from search results (or not, if the option is disabled)
- Expired Jobs, when viewed logged out, show the 'Expired' notice instead of a 404

wp-job-manager-functions.php
interlineae: After a further look into the class and some discussion with Jake, I'm going to keep to the current transient implementation. It doesn't look like there's anything too useful in the class for the caching that we want to do.

yscik: Yeah updating just the affected ID in the transient array is probably too complicated to be worth it.

includes/class-wp-job-manager-post-types.php
yscik: Makes sense, didn't think about a generic search. I did see expired listings in the jobs archive page without the extra post status filter.

I also still see them with the latest approach of setting `expired` to non-public on post status registration. Seems like Jobify has a lot of custom code that might interfere? It's possible to test with any theme by enabling the templates with: `add_theme_support( 'job-manager-templates' );`. And might need saving on the Permalinks page to flush the rules.

includes/class-wp-job-manager-post-types.php
```diff
@@ -389,14 +390,29 @@ public function register_post_types() {
...
+		 * This code checks if (1) we're on a search or (2) the option to hide expired job listings is enabled.
+		 * If either condition is false, we're going to set the register_post_status() 'public' arg to false,
+		 * otherwise it will be true.
+		 *
+		 * This will prevent single expired job listings from showing a '404', instead of an
+		 * 'Expired' notice, when viewing the single job listing and not logged in.
+		 *
+		 * This will also address historical issues stemming from addressing the issue raised here
+		 * https://github.com/Automattic/WP-Job-Manager/issues/1884.
+		 */
```
jom: Thanks for this comment. Really helped with figuring out the logic.

includes/class-wp-job-manager-post-types.php
```diff
@@ -389,14 +390,29 @@ public function register_post_types() {
...
+		$is_site_search        = ( isset( $_GET['s'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$is_expired_searchable = (bool) get_option( 'job_manager_hide_expired' );
```
jom: This is `true` if we're to hide expired listings, so not searchable; inverse of what the variable name implies. Is that correct? I maybe need more coffee.

includes/class-wp-job-manager-post-types.php
jom: The variable name threw me off a bit, but I think the ultimate logic is correct!

includes/class-wp-job-manager-post-types.php
```diff
@@ -659,6 +675,62 @@ public function job_feed() {
...
 
+	/**
+	 * Get filled jobs.
+	 *
+	 * @return array
+	 */
+	public function get_filled_job_listings(): array {
+		if ( ! get_option( 'job_manager_hide_filled_positions' ) ) {
+			return [];
+		}
```
jom: I think this logic is done in the _usage_ of this method. It doesn't quite match the described purpose of the method.

includes/class-wp-job-manager-post-types.php
```diff
@@ -659,6 +675,62 @@ public function job_feed() {
 		remove_filter( 'posts_search', 'get_job_listings_keyword_search' );
 	}
 
+	/**
+	 * Get filled jobs.
+	 *
+	 * @return array
```
jom: Might be helpful here to be explicit that this will return the IDs of the job listings.

wp-job-manager-functions.php
```diff
@@ -1712,3 +1712,21 @@ function job_manager_get_salary_unit_options( $include_empty = true ) {
...
+
+	if ( 'job_listing' === $post->post_type ) {
+
+		if ( '1' === get_post_meta( $post_id, '_filled', true ) ) {
+			delete_transient( 'hide_filled_jobs_transient' );
+		}
+	}
+
+}
+add_action( 'save_post', 'delete_filled_job_listing_transient_on_post_meta_update', 10, 2 );
```
jom: I'm curious about why this is separated from the other logic in this PR? Do we think this function is useful in the public API?

There is also the [possibly edge] case where a job went from filled to not filled. Do you think that is worth flushing this transient after each job listing safe?

jom (commented):
I'll do some more testing later, but this is looking good! I added a few minor comments.

includes/class-wp-job-manager-post-types.php
interlineae: Committed here https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/d23b3a3242ab91c54286d4fccf57cd5b725ecb43.

I've renamed the variable from `$is_expired_searchable` to `$is_expired_hidden`. That's more readable to my brain, but let me know if it's still a bit muddy.

includes/class-wp-job-manager-post-types.php
interlineae: Makes sense, committed here https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/ea63e6609cebe6de0d564deedd71abef6867bfc6

includes/class-wp-job-manager-post-types.php
interlineae: Good point, and after further testing, the method returns an empty array anyway if there are no posts ID, so this check is redundant.

Committed https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/3a210fb7152c0fb9a40c5393e60c33b13ee06db1

wp-job-manager-functions.php
interlineae: @jom Ah, I hadn't realised the functions file was dedicated to the API. I'll move this to post-types.php.

One thing I'm having difficulty with is getting the method to trigger on `updated_post_meta`. What I'd want to do ideally, following Peter's comment above, is:

1. Set the transient
2. If a job has the `_filled` meta updated, either add that ID to the transient or remove it

Clearing it on every post save is far too much and makes the transient pointless imo.

jom: I always get a little nervous about manipulating cached data (besides flushing it) because there are so many weird scenarios that could affect it. However, I suppose the weird ones would just happen for the life of the transient. In this case, one possible strangeness could happen in post status changes because to populate the transient you're querying published posts. While the query is inefficient generally, I think it isn't so bad for it to run it again when a possible change has been detected. 

What issues were you experiencing with `updated_post_meta`? It be possible to do a slightly lightweight flush on that and `transition_post_status`.

includes/class-wp-job-manager-post-types.php
interlineae: Gotcha. I ended up reverting to the previous `expired` implementation as the new one isn't able to check conditionals like `is_archive()`. 

Committed https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/1a7b7da26d47c33255a23ea5e95da70773f6454a

includes/class-wp-job-manager-post-types.php
interlineae: I ended up reverting this due to issues checking conditionals like `is_archive()`. Committed https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/1a7b7da26d47c33255a23ea5e95da70773f6454a

wp-job-manager-functions.php
interlineae: @jom So I've learned that I was doing a check like:

```
if ( '_filled' !== $meta_key || 'job_listing' !== get_post_type( $object_id ) ) {
	return;
}
```

But in debugging, I noticed that `$meta_key` is only ever returning `_edit_lock`, so what actually works is:

```
if ( '_edit_lock' !== $meta_key || 'job_listing' !== get_post_type( $object_id ) ) {
	return;
}
```

But I don't imagine checking against `_edit_lock` is all that useful. I'm stumped on how to check that the `_filled` meta was changed. I've committed what I have now at https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/de746c49aacb3a010e644be3b12e6329b68401ac

includes/class-wp-job-manager-post-types.php
jom: The main issue we were avoiding with the other strategy was `$query->set( 'post_status', 'publish' )` removing other custom _public_ statuses from the search. I actually wonder if this was fixed in core (5.9) with https://core.trac.wordpress.org/changeset/49830. It should now be respecting our `exclude_from_search` when registering the `expired` post status. Did you find in testing that expired posts were showing up in search results still?

I wonder if we should be setting `exclude_from_search` based on our setting. Since that fix in core, I'm guessing we're hiding expired results from search even when we shouldn't be.

interlineae: Ah, gotcha.

`exclude_from_search` is already set to `true`:

https://github.com/Automattic/WP-Job-Manager/blob/0568f04b6e27b585a17529095359038887bb10c6/includes/class-wp-job-manager-post-types.php#L395-L407

This does _not_ exclude expired jobs from WP Search.

This seems to be getting overwritten by the `exclude_from_search` parameter on the post-type registration:

https://github.com/Automattic/WP-Job-Manager/blob/0568f04b6e27b585a17529095359038887bb10c6/includes/class-wp-job-manager-post-types.php#L369

If this is set to `true`, then posts with an `expired` post status are, as expected, excluded from search -- but so is any `job` type post.

When I was testing the `'public' =>` manipulation, it worked for `$GET_['s']` but did not work for archive pages as helper methods were always returning `false`. Maybe there's a way around that, that I haven't thought of?

I'm wondering if another option would be to keep the `$query->set()`, but instead of setting it to `publish`, retrieve all post statuses with `get_post_stati()`, remove what we don't want, and then pass everything else. Though, that seems a bit... hacky?

```
$post_statuses     = get_post_stati();
$excluded_statuses = [ 'future', 'draft', 'pending', 'trash', 'auto-draft', 'preview', 'private', 'request-pending', 'request-confirmed', 'request-failed', 'request-completed', 'expired' ];
$valid_statuses    = array_diff( $post_statuses, $excluded_statuses );

$query->set( 'post_status', $valid_statuses );
```

IIUC, this is more to do with the issue report here https://core.trac.wordpress.org/ticket/44737, which is sitting stale. Making the suggested fix _does_ fix the behaviour, but then that's a whole core-related thing.

cc @yscik

jom: > This seems to be getting overwritten by the exclude_from_search parameter on the post-type registration:

I saw https://core.trac.wordpress.org/ticket/44737, but then saw the lines get changed in https://core.trac.wordpress.org/ticket/48556 and thought that would fix it, but now I can see what it is doing in the `} elseif ( ! $this->is_singular ) {` block. 

> I'm wondering if another option would be to keep the $query->set(), but instead of setting it to publish, retrieve all post statuses with get_post_stati(), remove what we don't want, and then pass everything else. Though, that seems a bit... hacky?

This is what I was going to suggest before I went down the 44737 rabbit hole because it seems like WP is doing this in a way that would work for us:
https://github.com/WordPress/WordPress/blob/32e94b4de108fb090ebc678b48113b8954f7f188/wp-includes/class-wp-query.php#L2587-L2593

I think if we did this we could just removed `expired` (I'm guessing)? It is a bit hack-y, but might be okay. 

The only other thought I had is modifying the global that holds the post status objects here after our conditionals and then restore `public => true` in the `posts_selection` action, but that might feel more hacky.

interlineae: @jom Here's the latest approach, which we talked about

https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/82e75e67cb8136e5489f78f86872141e5155abbd

includes/class-wp-job-manager-post-types.php
```diff
@@ -81,13 +81,15 @@ public function __construct() {
 		add_filter( 'wp_insert_post_data', [ $this, 'fix_post_name' ], 10, 2 );
 		add_action( 'add_post_meta', [ $this, 'maybe_add_geolocation_data' ], 10, 3 );
 		add_action( 'update_post_meta', [ $this, 'update_post_meta' ], 10, 4 );
+		add_action( 'updated_post_meta', [ $this, 'delete_filled_job_listing_transient' ], 10, 4 );
```
aaronfc, line 84: I believe this should be `update_post_meta` and not update**d** ? I couldn't find the `updated_post_meta` hook.

includes/class-wp-job-manager-post-types.php
```diff
@@ -691,49 +692,83 @@ public function get_filled_job_listings(): array {
...
-					'post_type'   => 'job_listing',
-					'post_status' => 'expired',
-					'fields'      => 'ids',
-				]
-			);
-			set_transient( 'hide_expired_jobs_transient', $expired_jobs_transient, DAY_IN_SECONDS );
+		if (
+			! is_admin()
+			&& $query->is_main_query()
+			&& ( $query->is_search() || $query->is_archive() )
```
aaronfc, line 723: It would be nice to add a comment here clarifying the reason for this logic: We want the filtering to apply only to search and archive public facing queries. We don't want it to apply in the admin panel.

includes/class-wp-job-manager-post-types.php
```diff
@@ -691,49 +692,83 @@ public function get_filled_job_listings(): array {
...
-		if ( ! is_admin() && $query->is_search() ) {
-			$jobs_to_exclude = array_merge( $this->get_filled_job_listings(), $this->get_expired_job_listings() );
-			if ( ! empty( $jobs_to_exclude ) ) {
-				$query->set( 'post__not_in', $jobs_to_exclude );
-			}
+		if (
+			! is_admin()
+			&& $query->is_main_query()
+			&& ( $query->is_search() || $query->is_archive() )
+		) {
```
aaronfc, line 753: Same here about the comment explaining the logic.

includes/class-wp-job-manager-post-types.php
```diff
@@ -691,49 +692,83 @@ public function get_filled_job_listings(): array {
...
+	 */
+	public function make_expired_public(): void {
+
+		global $wp_post_statuses;
+
+		if ( isset( $wp_post_statuses['expired'] ) ) {
+			$wp_post_statuses['expired']->public = true;
+		}
+
+	}
```
aaronfc, line 775: I couldn´t find any information about this global. Can you make a brief summary on what is it and what is this supposed to do?

includes/class-wp-job-manager-post-types.php
jom: It is a sneaky dynamic one: http://developer.wordpress.org/reference/hooks/updated_(meta_type)_meta/

includes/class-wp-job-manager-post-types.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/27b3be7732a15d5d95d1981dbbb6f7ddceb4a009

includes/class-wp-job-manager-post-types.php
interlineae: Committed https://github.com/Automattic/WP-Job-Manager/pull/2449/commits/27b3be7732a15d5d95d1981dbbb6f7ddceb4a009

includes/class-wp-job-manager-post-types.php
interlineae: @jom and I discussed this global in a call. AFAIK it's first mentioned here in `post.php`:

https://github.com/WordPress/wordpress-develop/blob/bea4307daa21a3f97d159898b0ac676332e0e7de/src/wp-includes/post.php#L1287-L1292

```
function register_post_status( $post_status, $args = array() ) {
	global $wp_post_statuses;

	if ( ! is_array( $wp_post_statuses ) ) {
		$wp_post_statuses = array();
	}
```

It is an array of all post statuses. What we're doing here is checking that the `expired` post status is registered, and if it is, we either set its `public` argument to `true` or `false` via the `maybe_hide_expired_job_listings()` method. 

From there, if `$hide_expired` is true, `$this->make_expired_private();` sets the `public` argument to `false` during query generation, after which it sets it back to true with `add_action`:

https://github.com/Automattic/WP-Job-Manager/blob/82e75e67cb8136e5489f78f86872141e5155abbd/includes/class-wp-job-manager-post-types.php#L732-L741

interlineae:
Closing as ~2 years stale. At this point a new PR would be more appropriate for taking this forward.

PR #2450, opened 2023-06-08
Add schema for promoted-jobs (including Swagger setup).

### LINKED ISSUE #2436: Define a feed schema to expose all promoted jobs
The jobs should contain at least:
- Title
- Description
- Permalink
- Salary (if it is activated)
- Company name

It also includes the basic setup for Swagger in the plugin. In the future, we could also add some documentation related to the core WordPress. So it's something to keep in mind while adding it. See context: p1685544391400649-slack-C036R3Y5ET0

**Notice that if we can do it directly through WP core API, using, filtering by promoted jobs (flag in post meta), we can close this issue and https://github.com/Automattic/WP-Job-Manager/issues/2437.**

### Success criteria

- [x] Check the approach (WP core API or new endpoint).
- [x] Consider performance while querying through post metas.
- [x] A basic setup with Swagger is implemented.
- [x] Promoted Jobs schema is accessible through Swagger UI.

aaronfc:
Fixes #2436 

* Added GH Workflow to generate Swagger UI and publish under Github Pages.
* Added swagger file for internal API including the `/wp-json/wpjm-internal/v1/promoted-jobs` endpoint.
* Added README to `docs/api/`.

### Testing instructions
* Check the Github Pages: https://automattic.github.io/WP-Job-Manager/api/internal/
* Alternatively, download the OpenAPI specification and check it on https://editor.swagger.io/ (for example)

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - trunk
-      - add/* # Temporary addition for testing.
+      - workflow_dispatch
 
 jobs:
   deploy:
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -1,0 +1,26 @@
+name: Deploy GitHub Pages
+
+on:
+  push:
+    branches:
+      - trunk
+      - add/* # Temporary addition for testing.
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v3
+      - name: Generate Swagger UI
+        uses: Legion2/swagger-ui-action@v1
+        with:
+          output: swagger-ui
+          spec-file: ./docs/api/internal.json
+          version: "^4.18.0"
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: swagger-ui
+          destination_dir: api/internal
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -91,6 +91,10 @@
                               "temporary",
                               "other"
                             ]
+                          },
+                          "salary":  {
+                            "type": "string",
+                            "example": "$50,000 - $100,000"
                           }
                         }
                       }
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -1,0 +1,108 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "WP Job Manager APIs (internal)",
+    "version": "1.0",
+    "description": "Internal WP Job Manager endpoints. These endpoints are not subject to any backwards compatibility guarantees. They may change at any time without warning since are considered as internal APIs.",
+    "contact": {
+      "name": "WPJobManager",
+      "url": "https://wpjobmanager.com/"
+    }
+  },
+  "servers": [
+    {
+      "url": "http://wpjm.local"
+    }
+  ],
+  "tags": [
+    {
+      "name": "promoted-jobs"
+    }
+  ],
+  "paths": {
+    "/wp-json/wpjm-internal/v1/promoted-jobs": {
+      "get": {
+        "summary": "Promoted Jobs - Get list of promoted jobs",
+        "description": "Get promoted jobs list",
+        "operationId": "get-promotedjobs-list",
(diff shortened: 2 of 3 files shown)


.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,26 @@
+name: Deploy GitHub Pages
+
+on:
+  push:
+    branches:
+      - trunk
+      - add/* # Temporary addition for testing.
```
renatho: Should we remove it now? I'm also fine to keep it if we want since the link is not shared anywhere.

renatho: Thinking more, since it's internal, maybe we could have it for all branches? And maybe rename the workflow name to make sure we'll only add internal things here? WDYT?

docs/api/internal.json
```diff
@@ -0,0 +1,108 @@
...
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "jobs": {
+                      "type": "array",
+                      "items": {
+                        "type": "object",
+                        "properties": {
```
renatho, line 54: Maybe we could already add the salary field here? Just in case we decide to iterate and use it in the future.

docs/api/internal.json
```diff
@@ -0,0 +1,108 @@
...
+                          },
+                          "location": {
+                            "type": "string",
+                            "example": "New York, NY, USA"
+                          },
+                          "company_name": {
+                            "type": "string",
+                            "example": "ACME Inc."
+                          },
+                          "is_remote":  {
```
renatho, line 80: Just adding that for now we won't use this field with JobTarget integration because they don't have a field for that in their API, but I think it's nice to have it here already, just in case we want to use it for something.

renatho (approved):
Looks great!

I added some questions while already approving it. So if needed, we can iterate later, or also change it here. It's up to you. ;)

And just to confirm, have you considered https://github.com/Automattic/WP-Job-Manager/issues/2436#issuecomment-1574138041 and it's better to create the new endpoint instead of extending the core?

aaronfc:
Hey @renatho thanks for your review!

> And just to confirm, have you considered https://github.com/Automattic/WP-Job-Manager/issues/2436#issuecomment-1574138041 and it's better to create the new endpoint instead of extending the core?

Good point I should have said something in the PR description. Yes, I thought about that. But quickly discarded it when I thought about future iterations. I think it is easier to maintain our own endpoint that we can version and keep track of all changes ourselves. Let me know if you have any strong opinions on using the REST API – I might have missed something.

.github/workflows/github-pages.yml
aaronfc: > Should we remove it now? I'm also fine to keep it if we want since the link is not shared anywhere.

I left it because it's also there in wpjobmanager.com, but I am happy removing it too.

> Thinking more, since it's internal, maybe we could have it for all branches? And maybe rename the workflow name to make sure we'll only add internal things here? WDYT?

That's a good point. I just thought of another alternative: We leave only "trunk" and we add the "workflow_dispatch" event too (so we can trigger it manually from any branch if we want to through the Gihub Actions interface). WDYT? (Done in cae02910)

Not sure what you mean with "rename the workflow name". Do you mean the file? `github-pages.yml` to something like `internal-docs-gh-pages.yml`?

docs/api/internal.json
aaronfc: Added 808c8896

.github/workflows/github-pages.yml
renatho: > That's a good point. I just thought of another alternative: We leave only "trunk" and we add the "workflow_dispatch" event too (so we can trigger it manually from any branch if we want to through the Gihub Actions interface). WDYT? (Done in https://github.com/Automattic/WP-Job-Manager/commit/cae029105cbfd04d30cbfd80cb4dc593c9579da2)

Nice! It makes sense to me!

> Not sure what you mean with "rename the workflow name". Do you mean the file? github-pages.yml to something like internal-docs-gh-pages.yml?

Yep! But I'm not remembering now why I thought to have only internal things in that workflow. 🤔 So ignore it, and if I remember something or we realize anything we could iterate later.

renatho (approved):
> Good point I should have said something in the PR description. Yes, I thought about that. But quickly discarded it when I thought about future iterations. I think it is easier to maintain our own endpoint that we can version and keep track of all changes ourselves. Let me know if you have any strong opinions on using the REST API – I might have missed something.

Makes sense to me! =)

PR #2452, opened 2023-06-13
Fix Pending Payment

onubrooks:
Fixes #https://github.com/Automattic/wpjm-addons/issues/210

This issue really affects paid listings extensions but the fix has to be here in core. There was a partial fix [here](https://github.com/Automattic/wp-job-manager-wc-paid-listings/pull/176).

* Enable 'Continue Submission' when job status is `pending_payment`.

### Testing instructions

* Create a new Job Listing
* Enter job details and click on [Preview]
* Choose a package
* Do not proceed, go back to jobs dashboard.
* Check to see that the new job is listed as `pending_payment` and there's a 'Continue Submission' link on hover.

### Some Thoughts
There's a suggestion to change pending payments back to draft in [this comment](https://github.com/Automattic/WP-Job-Manager/pull/2141/files#r622625262) but I think this solution covers it.

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -469,7 +469,6 @@ class WP_Job_Manager_Shortcodes {
 					];
 				}
 				break;
-			case 'pending_payment':
 			case 'pending':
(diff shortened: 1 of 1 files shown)


yscik (approved):
Makes sense!

PR #2454, opened 2023-06-13
Always show search actions

### LINKED ISSUE #1749: The "RSS", "Reset" and "Add Alert" links when no search result
The "Reset" and "Add Alert" button appears when the search query has some results (see top right corner).

image

But it does not appear when it has no result.

image

Ideally, for consistent and better user experience these buttons should appear in both cases.

yscik:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/1749

* Show the links below search even when there are no results, since they are useful in that case too.

### Testing instructions

* Search for jobs
* Check that the links for RSS and Reset are there even if there are 0 results.

### Screenshot / Video

image

### DIFF
--- a/includes/class-wp-job-manager-ajax.php
+++ b/includes/class-wp-job-manager-ajax.php
@@ -186,7 +186,7 @@ class WP_Job_Manager_Ajax {
 			'max_num_pages' => $jobs->max_num_pages,
 		];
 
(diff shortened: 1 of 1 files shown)


onubrooks (approved):

gitgreek-of:
Unfortunately this was not fixed in the Companies area:

Just checked with 1.41:

When a search has no results, the search reset link is missing. And the user is somehow stuck in the search form.

PR #2457, opened 2023-06-16
Trigger search when submitting search form

### LINKED ISSUE #2304: Submit Button on Filters page does not do anything
While working on the remote position handling on the listings page, I noticed that the submit button does not actually do anything.  

Search gets triggered when a change is made in one of the default fields, and the form is wrapped in a `<form>` element, but clicking on the submit button does nothing.

This can be replicated on the demo site:
https://demo.wpjobmanager.com/jobs/

yscik:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2304

* When using AJAX search, which submits the search whenever an input changes, also do the search when clicking the 'Search Jobs' button

### Testing instructions

* Go to the Jobs page and use the search form

### Notes

There is a bit of trade-off here. The search so far triggers on `change` events, which occurs when unfocusing an input. And that happens when the users moves to click the submit button. So in most cases the search would happen anyway. And in fact, with this change, the search request is duplicated, because both the change event and the submit event triggers it. 

The code cancels the ongoing search request when a new one is issued, but in this case it might make more sense to only trigger the search on submit when it's not already going on?

### DIFF
--- a/assets/js/ajax-filters.js
+++ b/assets/js/ajax-filters.js
@@ -457,14 +457,16 @@ jQuery( document ).ready( function( $ ) {
 			} );
 		} );
 
+	function triggerSearch() {
+		var $target = $( this ).closest( 'div.job_listings' );
+		$target.triggerHandler( 'update_results', [ 1, false ] );
+		store_state( $target );
+	}
+
 	$(
 		'#search_keywords, #search_location, #remote_position, .job_types :input, #search_categories, .job-manager-filter'
 	)
-		.change( function() {
-			var $target = $( this ).closest( 'div.job_listings' );
-			$target.triggerHandler( 'update_results', [ 1, false ] );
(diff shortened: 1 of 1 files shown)


mikeyarce:
I tested this out but on my test site I never see another request happen when clicking "Search Jobs".  I can see `triggerSearch` is called, but it never calls `update_results`.

gikaragia:
Hey @mikeyarce and @yscik, I added a commit which does the following:
- It passes the selected element to `triggerSearch` to fix the issue that Mikey mentioned above
- It checks if there is a pending search before triggering a new search.

Let me know your thoughts.

yscik:
Thanks @gkaragia, that's a great approach. Code looks good, but I can't approve as the PR author :)

gikaragia (approved):

Issue #2445, opened 2023-06-01
Basic setup


renatho:
- It's a basic setup containing folders and basic classes to initialize Promoted Jobs feature.

### Success Criteria

- [ ] We have a basic setup to allow starting other features from that.

mikeyarce:
PR: https://github.com/Automattic/WP-Job-Manager/pull/2459

renatho:
https://github.com/Automattic/WP-Job-Manager/pull/2460

Issue #2438, opened 2023-05-31
Create promote and deactivate buttons on job listings opening modals


renatho:
- Add promote and deactivate buttons to Job listing. Design: 6yDn5aNJyeJ4Xgp0lcDjDc-fi-20018640_4432
- Create the promote modal. Design: 6yDn5aNJyeJ4Xgp0lcDjDc-fi-20018663_2142
- Create the promotion deactivation modal. Design: 6yDn5aNJyeJ4Xgp0lcDjDc-fi-20018640_4216

This issue doesn't include the integration of the promote and deactivate buttons from the modals.

### Success Criteria

- [x] Clicking on promote button opens a modal with more details and a button to promote.
- [x] We have a way to set the link we'll have in the promote button of the modal for every job (currently, it can be a static link with the job ID in the query string as an example).
- [x] Clicking on deactivate button opens a confirmation modal that user can skip or proceed.
- [x] We have a way to set the link we'll have in the deactivate button of the modal for every job (currently, it can be a static link with the job ID in the query string as an example).

mikeyarce:
PR: https://github.com/Automattic/WP-Job-Manager/pull/2459

renatho:
https://github.com/Automattic/WP-Job-Manager/pull/2460

PR #2460, opened 2023-06-19
Add Promote button and modal

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2445 https://github.com/Automattic/WP-Job-Manager/issues/2438

* In the Job Listings Admin page, it moves all the Actions from buttons to links under the job titles
* Adds a new Column called "Promote" which has a button for promoting jobs
* When you click on `Promote`, it opens a dialog box which contains a shadow-dom template that's currently populated by hand.  This is just temporary until we have a way to populate it from wpjobmanager.com
* The dialog is added as an `admin-notice` just as a way to have the HTML in the DOM, that seemed the simplest way
* There's only one dialog element, but we can use JS to dynamically change the link.  Right now, as an example, the `href` attribute is set to the POST ID - we can change this to wahtever we need.

### Next tasks
- Get the HTML fragment from wpjobmanager.com 
- Store the fragment in some kind of cache
- Add proper links to the `Promote your job` and `Learn more` buttons

### Testing instructions

* Apply this PR
* Run `npm run start` to build assets
* Go to Job Listings -> All Jobs 
* Make sure the actions work (Edit, View, Approve, Delete)
* Click on "Promote" and see the modal
* Make sure you can click the `X` to leave the dialog
* Make sure you can leave the dialog by pressing `ESC`
* Set the Post Meta of a job listing to `_promoted => 1` and make sure you see the links to Edit and Deactivate the promotion (the links don't work yet)

### Screenshot / Video
Screenshot 2023-06-17 at 3 44 07 PM

promote-dialog

### DIFF
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -79,7 +79,7 @@ promote-job-modal {
 		padding: 0px 25px 25px;
 	}
 }
-.promote-jobs-dialog {
+.wpjm-dialog {
 	border: 0;
 	border-radius: 8px;
 	form.dialog {
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -33,7 +33,7 @@ promote-job-modal {
 		width: 80%;
 		margin-bottom: 0px;
 	}
-	.column-left{
+	.promote-job-modal-column-left {
 		display: flex;
 		justify-content: space-between;
 		flex-direction: column;
@@ -49,12 +49,16 @@ promote-job-modal {
 		background: url("data:image/svg+xml,<svg width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" fill=\"none\" xmlns=\"http://www.w3.org/2000/svg\"><mask id=\"mask0_20018663_2259\" style=\"mask-type:luminance\" maskUnits=\"userSpaceOnUse\" x=\"3\" y=\"5\" width=\"18\" height=\"14\"><path d=\"M8.75685 15.9L4.57746 11.7L3.18433 13.1L8.75685 18.7L20.698 6.69999L19.3049 5.29999L8.75685 15.9Z\" fill=\"white\"/></mask><g mask=\"url%28%23mask0_20018663_2259%29\"><rect width=\"23.8823\" height=\"24\" fill=\"%232270B1\"/></g></svg>") no-repeat 0 -3px;
 		font-size: 14px;
 		list-style: none;
-		padding-left: 28px;
+		padding-left: 32px;
 	}
-	.price {
-		font-size: 12px;
+	.promote-job-modal-price {
 		display: flex;
 		flex-direction: column;
+		.price-text {
+			font-size: 12px;
+			text-transform: uppercase;
+			color: #787C82;
+		}
 		span {
 			margin-top: 10px;
 			font-size: 36px;
@@ -72,7 +76,7 @@ promote-job-modal {
 	}
 }
 @media screen and ( max-width: 1000px ) {
-	.column-right {
+	.promote-job-modal-column-right {
 		display: none;
 	}
 }
@@ -81,7 +85,7 @@ promote-job-modal {
 		padding: 0px 25px 25px;
 	}
 }
-dialog {
+.promoteJobsDialog {
 	border: 0;
 	border-radius: 8px;
 	form {
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -22,10 +22,10 @@ promote-job-modal {
 	display: grid;
 	grid-template-columns: repeat(auto-fit, minmax(240px, 1fr));
 	padding: 30px 80px 50px 80px;
-	img {
+	img.promote-jobs-image {
 		width: 100%;
 	}
-	h2 {
+	h2.promote-jobs-heading {
 		font-size: 36px;
 		font-weight: 300;
 		line-height: 105%;
@@ -37,15 +37,8 @@ promote-job-modal {
 		display: flex;
 		justify-content: space-between;
 		flex-direction: column;
-
-		button {
-			padding: 16px 10px;
-		}
 	}
-	ul {
-		margin: 24px 0;
-	}
-	li {
+	li.promote-list-item {
 		background: url("data:image/svg+xml,<svg width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" fill=\"none\" xmlns=\"http://www.w3.org/2000/svg\"><mask id=\"mask0_20018663_2259\" style=\"mask-type:luminance\" maskUnits=\"userSpaceOnUse\" x=\"3\" y=\"5\" width=\"18\" height=\"14\"><path d=\"M8.75685 15.9L4.57746 11.7L3.18433 13.1L8.75685 18.7L20.698 6.69999L19.3049 5.29999L8.75685 15.9Z\" fill=\"white\"/></mask><g mask=\"url%28%23mask0_20018663_2259%29\"><rect width=\"23.8823\" height=\"24\" fill=\"%232270B1\"/></g></svg>") no-repeat 0 -3px;
 		font-size: 14px;
 		list-style: none;
@@ -89,9 +82,9 @@ promote-job-modal {
 .promote-jobs-dialog {
 	border: 0;
 	border-radius: 8px;
-	form {
+	form.dialog {
 		display: flex;
-		button {
+		button.dialog-close {
 			margin: 10px 15px auto auto;
 			content: '';
 			background: none;
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -86,7 +86,7 @@ promote-job-modal {
 		padding: 0px 25px 25px;
 	}
 }
-.promoteJobsDialog {
+.promote-jobs-dialog {
 	border: 0;
 	border-radius: 8px;
 	form {
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -17,6 +17,92 @@ a.wpjm-activate-licence-link:active {
 	color: orangered;
 }
 
+/** Promote Job Modal **/
+promote-job-modal {
+	display: grid;
+	grid-template-columns: repeat(auto-fit, minmax(240px, 1fr));
+	padding: 30px 80px 50px 80px;
+	img {
+		width: 100%;
+	}
+	h2 {
+		font-size: 36px;
+		font-weight: 300;
+		line-height: 105%;
+		margin-top: 0;
+		width: 80%;
+		margin-bottom: 0px;
+	}
+	.column-left{
+		display: flex;
+		justify-content: space-between;
+		flex-direction: column;
+
+		button {
+			padding: 16px 10px;
+		}
+	}
+	ul {
+		margin: 24px 0;
+	}
+	li {
+		background: url("data:image/svg+xml,<svg width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" fill=\"none\" xmlns=\"http://www.w3.org/2000/svg\"><mask id=\"mask0_20018663_2259\" style=\"mask-type:luminance\" maskUnits=\"userSpaceOnUse\" x=\"3\" y=\"5\" width=\"18\" height=\"14\"><path d=\"M8.75685 15.9L4.57746 11.7L3.18433 13.1L8.75685 18.7L20.698 6.69999L19.3049 5.29999L8.75685 15.9Z\" fill=\"white\"/></mask><g mask=\"url%28%23mask0_20018663_2259%29\"><rect width=\"23.8823\" height=\"24\" fill=\"%232270B1\"/></g></svg>") no-repeat 0 -3px;
+		font-size: 14px;
+		list-style: none;
+		padding-left: 28px;
+	}
+	.price {
+		font-size: 12px;
+		display: flex;
+		flex-direction: column;
+		span {
+			margin-top: 10px;
+			font-size: 36px;
+			font-weight: 700;
+		}
+	}
+	.promote-buttons-group {
+		.button {
+			padding: 5px 16px;
+			margin-right: 18px;
+		}
+		.button-secondary {
+			background: #ffffff;
+		}
+	}
+}
+@media screen and ( max-width: 1000px ) {
+	.column-right {
+		display: none;
+	}
+}
+@media screen and ( max-width: 782px ) {
+	promote-job-modal {
+		padding: 0px 25px 25px;
+	}
+}
+dialog {
+	border: 0;
+	border-radius: 8px;
+	form {
+		display: flex;
+		button {
+			margin: 10px 15px auto auto;
+			content: '';
+			background: none;
+			border: 0;
+			font-size: 0;
+			&:after {
+				content: "\2715";
+				font-size: 20px;
+			}
+		}
+	}
+	&::backdrop {
+		background: rgba(0, 0, 0, 0.5);
+	}
+}
+
 .wpjm-licences {
 	margin-top: 10px;
 
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -50,6 +50,7 @@ promote-job-modal {
 		font-size: 14px;
 		list-style: none;
 		padding-left: 32px;
+		margin: 12px 0;
 	}
 	.promote-job-modal-price {
 		display: flex;
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -179,11 +179,10 @@ jQuery(document).ready(function($) {
 	});
 });
 
-// Select all elements with the class "promote_job"
-const promote_job = document.querySelectorAll('.promote_job');
+const promoteJob = document.querySelectorAll('.promote_job');
 const promoteDialog = document.getElementById('promoteDialog');
 
-promote_job.forEach(function (element) {
+promoteJob.forEach(function (element) {
 		element.addEventListener('click', function ( event ) {
 			event.preventDefault();
 			promoteDialog.showModal();
@@ -194,13 +193,13 @@ promote_job.forEach(function (element) {
 });
 
 customElements.define('promote-job-modal',
-  class extends HTMLElement {
-    constructor() {
-      super();
-      const promoteJobs = document.getElementById('promote-job-template').content;
-      const shadowRoot = this.attachShadow({
-        mode: 'open'
-      });
-      shadowRoot.appendChild(promoteJobs.cloneNode(true));
-    }
-  });
+class extends HTMLElement {
+	constructor() {
+		super();
+		const promoteJobs = document.getElementById('promote-job-template').content;
+		const shadowRoot = this.attachShadow({
+			mode: 'open'
+		});
+		shadowRoot.appendChild(promoteJobs.cloneNode(true));
+	}
+});
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -178,3 +178,29 @@ jQuery(document).ready(function($) {
 		$('#in-' + taxonomy + '-' + id + ', #in-popular-' + taxonomy + '-' + id).prop( 'checked', c );
 	});
 });
+
+// Select all elements with the class "promote_job"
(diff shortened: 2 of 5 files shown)


assets/css/admin.scss
```diff
@@ -17,6 +17,92 @@ a.wpjm-activate-licence-link:active {
...
+	.column-right {
+		display: none;
+	}
+}
+@media screen and ( max-width: 782px ) {
+	promote-job-modal {
+		padding: 0px 25px 25px;
+	}
+}
+dialog {
```
renatho: It's worth adding a class here instead of `dialog`, otherwise it could apply to external elements that we don't want.

assets/js/admin.js
```diff
@@ -178,3 +178,29 @@ jQuery(document).ready(function($) {
 		$('#in-' + taxonomy + '-' + id + ', #in-popular-' + taxonomy + '-' + id).prop( 'checked', c );
 	});
 });
+
+// Select all elements with the class "promote_job"
+const promote_job = document.querySelectorAll('.promote_job');
```
renatho: I think we should also use camelCase in this variable, right?

assets/css/admin.scss
```diff
@@ -17,6 +17,92 @@ a.wpjm-activate-licence-link:active {
...
+		margin: 24px 0;
+	}
+	li {
+		background: url("data:image/svg+xml,<svg width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" fill=\"none\" xmlns=\"http://www.w3.org/2000/svg\"><mask id=\"mask0_20018663_2259\" style=\"mask-type:luminance\" maskUnits=\"userSpaceOnUse\" x=\"3\" y=\"5\" width=\"18\" height=\"14\"><path d=\"M8.75685 15.9L4.57746 11.7L3.18433 13.1L8.75685 18.7L20.698 6.69999L19.3049 5.29999L8.75685 15.9Z\" fill=\"white\"/></mask><g mask=\"url%28%23mask0_20018663_2259%29\"><rect width=\"23.8823\" height=\"24\" fill=\"%232270B1\"/></g></svg>") no-repeat 0 -3px;
+		font-size: 14px;
+		list-style: none;
+		padding-left: 28px;
+	}
+	.price {
+		font-size: 12px;
```
renatho: I noticed that the styles of the "Starting from" text is not matching the design. In order to make it simpler, I'd suggest separating it into 2 classes.

assets/css/admin.scss
```diff
@@ -17,6 +17,92 @@ a.wpjm-activate-licence-link:active {
...
+	ul {
+		margin: 24px 0;
+	}
+	li {
+		background: url("data:image/svg+xml,<svg width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" fill=\"none\" xmlns=\"http://www.w3.org/2000/svg\"><mask id=\"mask0_20018663_2259\" style=\"mask-type:luminance\" maskUnits=\"userSpaceOnUse\" x=\"3\" y=\"5\" width=\"18\" height=\"14\"><path d=\"M8.75685 15.9L4.57746 11.7L3.18433 13.1L8.75685 18.7L20.698 6.69999L19.3049 5.29999L8.75685 15.9Z\" fill=\"white\"/></mask><g mask=\"url%28%23mask0_20018663_2259%29\"><rect width=\"23.8823\" height=\"24\" fill=\"%232270B1\"/></g></svg>") no-repeat 0 -3px;
+		font-size: 14px;
+		list-style: none;
+		padding-left: 28px;
+	}
+	.price {
```
renatho: I'm not sure about new code we're writing in WP Job Manager, so I'd like to ask you if we're not using [BEM](https://getbem.com/naming/) similar to Sensei.
If not, I'd suggest some more specific class here. Maybe `promote-job-modal-price`? So we avoid getting `price` class style from other plugins, since it's a very common name. The same I think also applies to the `column-left` and `column-right` classes.

assets/css/admin.scss
```diff
@@ -17,6 +17,92 @@ a.wpjm-activate-licence-link:active {
...
+		flex-direction: column;
+
+		button {
+			padding: 16px 10px;
+		}
+	}
+	ul {
+		margin: 24px 0;
+	}
+	li {
```
renatho: The `li` seems a little more close here than in the design. Maybe we could add some extra margin?

assets/js/admin.js
```diff
@@ -178,3 +178,29 @@ jQuery(document).ready(function($) {
 		$('#in-' + taxonomy + '-' + id + ', #in-popular-' + taxonomy + '-' + id).prop( 'checked', c );
 	});
 });
+
+// Select all elements with the class "promote_job"
```
renatho: Since it's written in `admin.css` and `admin.js`, and we're creating a custom element, maybe we could use generic names? So we could use it in the future for other things.

And we'd need to add only the specific Promoted Jobs logic separated (update the `href`)?

includes/admin/class-wp-job-manager-cpt.php
```diff
@@ -534,18 +533,57 @@ public function primary_column( $column, $screen ) {
...
+				}
+				if ( current_user_can( 'delete_post', $post->ID ) ) {
+					$admin_actions['delete'] = [
+						'action' => 'delete',
+						'name'   => __( 'Delete', 'wp-job-manager' ),
+						'url'    => get_delete_post_link( $post->ID ),
+					];
+				}
+			}
+			$admin_actions = apply_filters( 'job_manager_admin_actions', $admin_actions, $post );
```
renatho, line 580: I wonder if it could be a concern for other plugins using this filter, trying to use it with the previous styles. Have you checked the other WP Job Manager extensions to see if it's not being used?

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,183 @@
...
+	 */
+	public function columns( $columns ) {
+		$columns['promoted_jobs'] = __( 'Promote', 'wp-job-manager' );
+		return $columns;
+	}
+
+	/**
+	 * Check if a job is promoted.
+	 *
+	 * @param [string] $post_id
```
renatho: I think it's supposed to be `int`, right?

```suggestion
	 * @param int $post_id
```

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,183 @@
...
+
+	/**
+	 * Output the promote jobs template
+	 *
+	 * @return void
+	 */
+	public function my_admin_notice() {
+		echo '
+			<template id="promote-job-template">
+			<style>
```
renatho: Do we need this `<style>` part? (I'm not used to `template`, so please bear with me with the questions here 🤭).

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,183 @@
...
+
+			</style>
+				<slot name="column-left" class="column-left">
+					<slot name="promote-heading">
+					Promote Your Job on our Partner Network
```
renatho: I think it's missing a tab here. Also, maybe it's worth adding `i18n` functions here?

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,183 @@
...
+
+					<slot name="price">
+						Starting from
+						<span>$80.00</span>
+					</slot>
+
+					<slot name="promote-list">
+						<ul>
+							<li>Your ad will get shared on our Partner Network</li>
+							<li>Featured on jobs.blog for 7 days</li>
```
renatho: Since it's a default content that could be used in some situation, maybe we should only use more generic information?
If it makes sense, we would need to be a little creative here to write some stuff that will never change, like _"Promote your job on external job boards"_.
It would also apply to the price, maybe having something like "$ --".

assets/js/admin.js
```diff
@@ -178,3 +178,29 @@ jQuery(document).ready(function($) {
...
+			event.preventDefault();
+			promoteDialog.showModal();
+			let promoteID = event.target.dataset.post;
+			let promote_job_dialog = document.querySelector('.promote-buttons-group .button-primary');
+			promote_job_dialog.setAttribute('href', promoteID );
+		});
+});
+
+customElements.define('promote-job-modal',
+  class extends HTMLElement {
```
renatho: I imagine you used spaces instead of tabs accidentally here, right?

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,183 @@
...
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_filter( 'manage_edit-job_listing_columns', [ $this, 'columns' ] );
+		add_action( 'manage_job_listing_posts_custom_column', [ $this, 'custom_columns' ], 2 );
+		add_action( 'admin_notices', [ $this, 'my_admin_notice' ] );
```
renatho: I wonder if another hook more generic, like `admin_footer` would make more sense here.

Also, I imagine this class is something generic for many things related to promoted jobs, right? If yes, I'd suggest to use some more specific name for these methods, just to avoid confusion when the class grows.

renatho (commented):
Hi @mikeyarce! It's looking very nice with the `dialog` and the `template`. =)

I just added a few suggestions and some comments to confirm. Also, I'd suggest to add Zen team as reviewer since I don't work too much on WP Job Manager and maybe I could miss something.

includes/admin/class-wp-job-manager-cpt.php
mikeyarce: I tried to copy the use from the previous implementation so that it won't be different for other plugins that previous used the filters.  What do you think could happen in this case?

mikeyarce: Also, I think we should absolutely check with third-party devs ahead of this release.

assets/js/admin.js
mikeyarce: 😂   Sometimes I write things in a codepen or something and then move it over and forget to check for this stuff...

includes/admin/class-wp-job-manager-cpt.php
renatho: > What do you think could happen in this case?

I thought about weird things like having the links, and buttons as icons mixed. But maybe it's not a big deal. I'd just confirm if our official extensions aren't using it. And if some is using it, we could also update the HTML, if needed. :)

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: So one weird caveat I found is that you can't have a `dialog` inside of a `template` - because of how the Shadow DOM works and how components are mounted.  So we need to have a fallback in `get_promote_jobs_template` that has the dialog in it.  Just wanted to make a note if it here!

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: I like that idea!

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Do we need to add `i18n` functions since it's being generated from our side?

assets/js/admin.js
mikeyarce: I'm not sure I totally understand what you mean here - do you mind explaining a bit more? 😄

assets/css/admin.scss
```diff
@@ -81,7 +85,7 @@ promote-job-modal {
 		padding: 0px 25px 25px;
 	}
 }
-dialog {
+.promoteJobsDialog {
```
renatho: In this one, since it's a class, we could use the format `.promote-jobs-dialog`. ;)

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,181 @@
...
+
+	/**
+	 * Output the promote jobs template
+	 *
+	 * @return void
+	 */
+	public function promoted_jobs_admin_footer() {
+		echo '
+			<template id="promote-job-template">
+				<slot name="column-left" class="column-left">
```
renatho: We need to also update the classes here for the fallback.

assets/css/admin.scss
```diff
@@ -17,6 +17,96 @@ a.wpjm-activate-licence-link:active {
 	color: orangered;
 }
 
+/** Promote Job Modal **/
+promote-job-modal {
+	display: grid;
+	grid-template-columns: repeat(auto-fit, minmax(240px, 1fr));
+	padding: 30px 80px 50px 80px;
+	img {
```
renatho: Another thing I was thinking here is that it would be interesting to give more CSS classes to the elements. Usually, it's a good practice, but mainly here because we're doing it a little more dynamic getting the content from wpjobmanager.com. So if we decide to add any other image to any part, for example, it would have a `width: 100%`.

The same would apply to the button, list, and so on.

includes/admin/class-wp-job-manager-promoted-jobs.php
renatho: Yep. Since it's the fallback content, probably it's worth having it translated.

assets/js/admin.js
renatho: Sorry! Sure!

I mean that we could separate a little more the classes and the functions of the generic part of the modal, and add with other classes and functions the specific details of the promote button.

An application example: if we decide later to create a confirmation modal for the job deletion, we could just use the same modal, applying the extra logic only for the deletion.

In practice, it would mean using a more generic class to the dialog, (maybe something like `wpjm-modal`?). And separate functions to open/close the modal (the generic part) and update the `href` button (the specific promote button part).

As a final result, we'd be able to easily create other modals.

I think it would also help with the deactivate modal logic.

Does it make sense to you?

assets/css/admin.scss
renatho: I'm marking it as unsolved because it still seems to miss some spacing based on the design, okay?

mikeyarce: OH I see what you mean.  I thought you meant the Checkmark and the Item, but you mean each item!  
Fixed in 3973a8885bb9b4139d5bebe92c92b8d1d94ad93b

assets/js/admin.js
mikeyarce: Makes sense!  Fixed in 11d6fd03e117f5b3ccc2d208b168c8308f2f6e6f

includes/admin/class-wp-job-manager-cpt.php
mikeyarce: The idea I had last night:  Once this is merged, I am going to ping our WPJM Friends in Slack and let them know about this coming change and see if they have questions or concerns.  How does that sound?

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Fixed in 25af99885951b8916910f87b96c62472bb50f77e

assets/css/admin.scss
mikeyarce: Fixed in acafdbb6aee91bd404fd30cfc7e34803509c4389

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: The template doesn't include the dialog - so I don't think we do in this case?  I found out that you can't have dialogs in templates, they just don't seem to work.

assets/css/admin.scss
mikeyarce: Done in f4cad966c8698eddfaa04a6b5425e2a08e32d047

includes/admin/class-wp-job-manager-cpt.php
renatho: Not sure if it's so important in order to make some type of "noise". I think it would be more important for us to check the plugins we maintain for now. I took the opportunity to run a search all in the plugins here for `job_manager_admin_actions`, and I didn't find anything. So I think we're fine here. :)

includes/admin/class-wp-job-manager-promoted-jobs.php
renatho: Ah, I meant about classes like the `class="column-left"` that we updated.

mikeyarce: Gotcha!  Done in a44c9bcd56323db227551d8426c1467637876878

renatho:
@mikeyarce, @yscik brought an idea in today's meeting to also bring the styles from wpjobmanager.com. Do you think we could easily extract the styles related to the modal content? If not, we could create a separate issue to do it later.

I'd suggest also adding it through `get_promote_jobs_template` in a `<style>` tag until we fetch it.

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -0,0 +1,186 @@
...
+	 *
+	 * TODO: we need to fetch this from wpjobmanager.com and store in cache for X amount of time
+	 * We should also have a fallback in case the API call fails.
+	 * We need to have a fallback here because we can't use a `dialog` element inside the template
+	 *
+	 * @return string
+	 */
+	public function get_promote_jobs_template() {
+		return '
+		<dialog class="promote-jobs-dialog" id="promoteDialog">
```
renatho: Just a final thing and I think we're ready to merge it.

We also need to fix the `promoteDialog` id to `promote-dialog`.

And for the `promote-jobs-dialog` class, in that idea to make it a little more generic to be able to re-use in the future in other contexts, we could call it as `wpjm-dialog`.

renatho (commented):
Hey @mikeyarce! Thank you for all the tweaks.
I sent a last thing and I think we're ready to merge after that.

Also, about my previous comment, we can open a separate issue in order to merge it earlier. 😉

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Done in cbe1c8ebe577384a0a4f9b3ec53817e425895ba5

mikeyarce:
> Do you think we could easily extract the styles related to the modal content? If not, we could create a separate issue to do it later.
> I'd suggest also adding it through `get_promote_jobs_template` in a `<style>` tag until we fetch it.

Yeah the styles are easy to extract - and the nice thing about using `<style>` in a template is that it's scoped only to that template so no need to worry about overriding other styles.

I think a new issue for this would be good 👍🏽

renatho:
> Yeah the styles are easy to extract - and the nice thing about using <style> in a template is that it's scoped only to that template so no need to worry about overriding other styles.

TIL! That's great!

renatho (approved):
Thank you for the changes @mikeyarce!

I added to [this issue](https://github.com/Automattic/WP-Job-Manager/issues/2464) the part of extracting the styles.

yscik:
Hey @mikeyarce, sorry for the late review, but I think there is some misunderstanding with the template stuff.

The role of the template tag and the custom element would be a means of filling in the row-specific items, on the client side, setting the main button to point to the right URL. Ideally it would work something like this:

- In PHP, get the content of the modal from wpjobmanager.com (or cache)
- In PHP, output it in PHP inside a `<template>` tag
- In JS, define a custom element from the template tag's content
- In JS, when clicking the Promote button, create a new instance of that custom element, and fill the button slot with the proper URL for that row. Something like:

```js
dialog.innerHTML = `<promote-job-modal>
	<div slot="buttons">
		<a href="...?postId=${postId}" class="...">Promote your job</button>
		...
	</div>
<promote-job-modal>`
```

PHP doesn't need to output an instance of `<promote-job-modal>`, we would be creating those dynamically when showing a modal. 

I don't think we need any fixed structure for the content of the modal, and not sure that having a template in the plugin that the remote content fills adds any value. The modal content can be a freeform HTML, at some point even an editable post in the remote website, that right now only needs a single slot defined for the dynamic link.

Issue #2442, opened 2023-05-31
Update job status after job is promoted or expired in JobTarget


renatho:
Create an endpoint in WP Job Manager plugin that will update a job to promoted or not promoted. Mikey raised some PRs with the `_promoted` post meta to represent it in the plugin.

Protect this endpoint will be part of a [separate issue](https://github.com/Automattic/wpjobmanager.com/issues/433).

### Success criteria

- [ ] It's possible to call an endpoint in WP Job Manager plugin in order to update the promoted status.
- [ ] Schema is documented in `/docs/api/internal.json`

fjorgemota:
Just documented here the approach that I'm thinking for protecting this endpoint: https://github.com/Automattic/wpjobmanager.com/issues/433#issuecomment-1612222137

PR #2466, opened 2023-06-20
Add class to generate and validate tokens for site authentication

fjorgemota:
Fixes 413-gh-Automattic/wpjobmanager.com
Sibling PR to 445-gh-Automattic/wpjobmanager.com

- Add class responsible for:
    -  creating and validating tokens on WPJM, associated with a post OR user
    -  validate the tokens created 

### Testing instructions
 
Run the unit tests, and experiment creating tokens using WP-CLI Shell:

```php
$instance = WP_Job_Manager_Com_Auth_Token::instance(); // To get the instance of the class
$token = $instance->generate( 'user', 1 ); // Create a token associated with the User ID 1
// Maybe await a minute using sleep(60)?
var_dump($instance->validate( 'user', 1, $token));
// Make the same test with a valid post instead of user
```

### DIFF
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -85,8 +85,10 @@ class WP_Job_Manager_Com_Auth_Token {
 	 */
 	private function generate_new_token() {
 		try {
-			$hash = random_bytes( 32 );
-			return bin2hex( $hash );
+			$hash = random_bytes( 48 );
+			//phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+			$base64 = base64_encode( $hash );
+			return str_replace( [ '+', '/', '=' ], '', $base64 );
 		} catch ( Exception $e ) {
 			return new WP_Error( 'wpjobmanager-com-token-not-generated', __( 'Token could not be generated', 'wp-job-manager' ) );
 		}
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -25,7 +25,7 @@ class WP_Job_Manager_Com_Auth_Token {
 	/**
 	 * Generates the site token associated with an object type and object id
 	 *
-	 * @param string $object_type Type of object metadata is for. Accepts 'post' or 'user'.
+	 * @param string $object_type Type of the object associated with the token. Accepts 'post' or 'user'.
 	 * @param int    $object_id The ID of the object to associate the token with.
 	 * @return string|WP_Error The token generated, or a WP_Error if the token could not be generated/persisted
 	 */
@@ -71,4 +71,72 @@ class WP_Job_Manager_Com_Auth_Token {
 		];
 	}
 
+
+
+	/**
+	 * Validate if a token is valid or not, and might remove it if it is valid or expired.
+	 *
+	 * @param string $object_type Type of object metadata is for. Accepts 'post' or 'user'.
+	 * @param int    $object_id  The ID of the object to associate the token with.
+	 * @param string $token The token to validate.
+	 * @return bool True if the token is valid, false otherwise.
+	 */
+	public function validate( $object_type, $object_id, $token ) {
+		if ( ! in_array( $object_type, self::ACCEPTED_OBJECT_TYPES, true ) ) {
+			return false;
+		}
+		$metadatas = get_metadata( $object_type, $object_id, self::META_KEY );
+		foreach ( $metadatas as $metadata ) {
+			if ( ! $this->is_valid( $metadata ) ) {
+				// If the metadata structure isn't valid, just ignore it.
+				continue;
+			}
+			if ( $this->is_expired( $metadata ) ) {
+				// If the token is expired, remove it.
+				delete_metadata( $object_type, $object_id, self::META_KEY, $metadata );
+				continue;
+			}
+			if ( $this->is_valid_token( $metadata, $token ) ) {
+				delete_metadata( $object_type, $object_id, self::META_KEY, $metadata );
+				return true;
+			}
+		}
+		return false;
+	}
+
+
+	/**
+	 * Checks if a value persisted in the database has the correct format for us.
+	 *
+	 * @param array $value The value persisted in the database.
+	 * @return bool True if the token is valid, false otherwise.
+	 */
+	public function is_valid( $value ) {
+		return is_array( $value ) &&
+			array_key_exists( 'token', $value ) &&
+			array_key_exists( 'ts', $value ) &&
+			is_string( $value['token'] ) &&
+			is_int( $value['ts'] );
+	}
+
+	/**
+	 * Checks if a token is expired or not.
+	 *
+	 * @param array $value The value persisted in the database.
+	 * @return bool True if the token is expired, false otherwise.
+	 */
+	private function is_expired( $value ) {
+		return $value['ts'] + self::EXPIRATION_TIME < time();
+	}
+
+	/**
+	 * Checks if a token is valid or not.
+	 *
+	 * @param array  $value The value persisted in the database.
+	 * @param string $token The token to analyse.
+	 * @return bool True if the token is valid, false otherwise.
+	 */
+	private function is_valid_token( $value, $token ) {
+		return wp_check_password( $token, $value['token'] );
+	}
 }
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -16,10 +16,19 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since   $$next-version$$
  */
 class WP_Job_Manager_Com_Auth_Token {
+	/**
+	 * The meta key used to store the token.
+	 */
 	const META_KEY = 'wpjmcom_site_auth_token';
 
+	/**
+	 * The accepted object types to be associated with the token.
+	 */
 	const ACCEPTED_OBJECT_TYPES = [ 'post', 'user ' ];
 
+	/**
+	 * The expiration time for the token, in seconds.
+	 */
 	const EXPIRATION_TIME = MINUTE_IN_SECONDS;
 
 	/**
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -32,6 +32,30 @@ class WP_Job_Manager_Com_Auth_Token {
 	const EXPIRATION_TIME = MINUTE_IN_SECONDS;
 
 	/**
+	 * The singleton instance of the class.
+	 *
+	 * @var WP_Job_Manager_Com_Auth_Token
+	 */
+	private static $instance;
+
+	/**
+	 * WP_Job_Manager_Com_Auth_Token constructor.
+	 */
+	private function __construct() {}
+
+	/**
+	 * Returns the singleton instance of the class.
+	 *
+	 * @return self
+	 */
+	public static function instance() {
+		if ( null === self::$instance ) {
+			self::$instance = new WP_Job_Manager_Com_Auth_Token();
+		}
+		return self::$instance;
+	}
+
+	/**
 	 * Generates the site token associated with an object type and object id
 	 *
 	 * @param string $object_type Type of the object associated with the token. Accepts 'post' or 'user'.
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -24,7 +24,7 @@ class WP_Job_Manager_Com_Auth_Token {
 	/**
 	 * The accepted object types to be associated with the token.
 	 */
-	const ACCEPTED_OBJECT_TYPES = [ 'post', 'user ' ];
+	const ACCEPTED_OBJECT_TYPES = [ 'post', 'user' ];
 
 	/**
 	 * The expiration time for the token, in seconds.
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -124,7 +124,7 @@ class WP_Job_Manager_Com_Auth_Token {
 			return false;
 		}
 		foreach ( $metadatas as $metadata ) {
-			if ( ! $this->is_valid( $metadata ) ) {
+			if ( ! $this->is_valid_format( $metadata ) ) {
 				// If the metadata structure isn't valid, just ignore it.
 				continue;
 			}
@@ -148,7 +148,7 @@ class WP_Job_Manager_Com_Auth_Token {
 	 * @param array $value The value persisted in the database.
 	 * @return bool True if the token is valid, false otherwise.
 	 */
-	private function is_valid( $value ) {
+	private function is_valid_format( $value ) {
 		return is_array( $value ) &&
 			array_key_exists( 'token', $value ) &&
 			array_key_exists( 'ts', $value ) &&
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -88,7 +88,11 @@ class WP_Job_Manager_Com_Auth_Token {
 			$hash = random_bytes( 48 );
 			//phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
 			$base64 = base64_encode( $hash );
-			return str_replace( [ '+', '/', '=' ], '', $base64 );
+			$result = str_replace( [ '+', '/', '=' ], '', $base64 );
+			if ( empty( $result ) ) {
+				$result = substr( bin2hex( $hash ), 0, 64 );
+			}
+			return $result;
 		} catch ( Exception $e ) {
 			return new WP_Error( 'wpjobmanager-com-token-not-generated', __( 'Token could not be generated', 'wp-job-manager' ) );
 		}
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -119,7 +119,7 @@ class WP_Job_Manager_Com_Auth_Token {
 	 * @param array $value The value persisted in the database.
 	 * @return bool True if the token is valid, false otherwise.
 	 */
-	public function is_valid( $value ) {
+	private function is_valid( $value ) {
 		return is_array( $value ) &&
 			array_key_exists( 'token', $value ) &&
 			array_key_exists( 'ts', $value ) &&
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -53,6 +53,7 @@ class WP_Job_Manager_Com_Auth_Token {
 		}
 		return $token;
 	}
+
 	/**
 	 * Generates a new random token and return it
 	 *
@@ -79,8 +80,6 @@ class WP_Job_Manager_Com_Auth_Token {
 			'ts'    => time(),
 		];
 	}
-
-
 
 	/**
 	 * Validate if a token is valid or not, and might remove it if it is valid or expired.
--- a/includes/helper/class-wp-job-manager-com-auth-token.php
+++ b/includes/helper/class-wp-job-manager-com-auth-token.php
@@ -1,0 +1,74 @@
+<?php
+/**
+ * File containing the class WP_Job_Manager_Com_Auth_Token.
+ *
+ * @package wp-job-manager
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Helper functions used for creating and validating tokens used for authenticating with WPJobManager.com.
+ *
+ * @package wp-job-manager
+ * @since   $$next-version$$
+ */
+class WP_Job_Manager_Com_Auth_Token {
+	const META_KEY = 'wpjmcom_site_auth_token';
+
+	const ACCEPTED_OBJECT_TYPES = [ 'post', 'user ' ];
+
+	const EXPIRATION_TIME = MINUTE_IN_SECONDS;
+
+	/**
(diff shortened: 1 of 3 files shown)


tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,168 @@
+<?php
+
+require_once JOB_MANAGER_PLUGIN_DIR . '/includes/helper/class-wp-job-manager-com-auth-token.php';
+
+/**
+ * @group helper
+ * @group helper-base
+ */
+class WP_Test_WP_Job_Manager_Com_Auth_Token extends WPJM_BaseTest {
```
fjorgemota, line 9: I'm not sure what the best way to test expired tokens here..if anyone has any suggestions, please feel free to mention it. :)

includes/helper/class-wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,177 @@
...
+		}
+		return $token;
+	}
+
+	/**
+	 * Generates a new random token and return it
+	 *
+	 * @return string|WP_Error The random token generated, or a WP_Error if the token could not be generated.
+	 */
+	private function generate_new_token() {
```
fjorgemota, line 86: This method will generate a string encoded as base64, but without some characters to make it URL-Safe. Because of this, it generates a string with varied size, with up to 64 characters.

includes/helper/class-wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,179 @@
...
+	public function validate( $object_type, $object_id, $token ) {
+		if ( ! in_array( $object_type, self::ACCEPTED_OBJECT_TYPES, true ) ) {
+			return false;
+		}
+		$metadatas = get_metadata( $object_type, $object_id, self::META_KEY );
+		if ( false === $metadatas ) {
+			return false;
+		}
+		foreach ( $metadatas as $metadata ) {
+			if ( ! $this->is_valid( $metadata ) ) {
```
renatho: WDYT to rename it to `is_valid_format` or something similar? It seems the main validation method.

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+	public function testInstance_WhenCalled_ReturnSameInstance() {
```
renatho, line 11: Not sure if I misunderstood something, but aren't these 2 tests with the name exchanged? (`testInstance_WhenCalled_ReturnSameInstance` and `testInstance_WhenCalled_ReturnCorrectType`)

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+		$this->assertEquals( 'wpjobmanager-com-invalid-type' , $result->get_error_code() );
+	}
+
+
+	public function testGenerate_WhenPassedInvalidObjectID_ShouldReturnError() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+
+		// Act.
+		$result = $instance->generate( 'user', "test");
```
renatho:
```suggestion
		$result = $instance->generate( 'user', 'invalid_id' );
```

I suggested a format fix and I took the opportunity to just suggest another string to make the test more explicit.

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+		$this->assertEquals( 'wpjobmanager-com-token-not-saved' , $result->get_error_code() );
+	}
+
+	public function testGenerate_WhenPassedUser_ShouldPersistMeta() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$user = $this->factory->user->create_and_get();
+
+		// Act.
+		$result = $instance->generate( 'user', $user->ID);
```
renatho:
```suggestion
		$result = $instance->generate( 'user', $user->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+		$this->assertNotEmpty( get_user_meta( $user->ID, WP_Job_Manager_Com_Auth_Token::META_KEY ) );
+	}
+
+	public function testGenerate_WhenPassedPost_ShouldPersistMeta() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$post = $this->factory->post->create_and_get();
+
+		// Act.
+		$result = $instance->generate( 'post', $post->ID);
```
renatho:
```suggestion
		$result = $instance->generate( 'post', $post->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	public function testValidate_WhenPassedUserWithMetaButInvalidToken_ShouldReturnFalse() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$user = $this->factory->user->create_and_get();
+		$token = $instance->generate('user', $user->ID);
```
renatho:
```suggestion
		$token = $instance->generate('user', $user->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	public function testValidate_WhenPassedPostWithMetaButInvalidToken_ShouldReturnFalse() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$post = $this->factory->post->create_and_get();
+		$token = $instance->generate('post', $post->ID);
```
renatho:
```suggestion
		$token = $instance->generate( 'post', $post->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	public function testValidate_WhenPassedUserWithMetaAndValidTokenTwice_ShouldReturnFalse() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$user = $this->factory->user->create_and_get();
+		$token = $instance->generate('user', $user->ID);
```
renatho:
```suggestion
		$token = $instance->generate( 'user', $user->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	public function testValidate_WhenPassedPostWithMetaAndValidTokenTwice_ShouldReturnFalse() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$post = $this->factory->post->create_and_get();
+		$token = $instance->generate('post', $post->ID);
```
renatho:
```suggestion
		$token = $instance->generate( 'post', $post->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	public function testValidate_WhenCalledWithUser_ShouldDeleteExpiredTokens() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$user = $this->factory->user->create_and_get();
+		$instance->generate('user', $user->ID);
```
renatho:
```suggestion
		$instance->generate( 'user', $user->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertEmpty( get_metadata( 'user', $user->ID, WP_Job_Manager_Com_Auth_Token::META_KEY ) );
+	}
+
+	public function testValidate_WhenCalledWithPost_ShouldDeleteExpiredTokens() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$post = $this->factory->post->create_and_get();
+		$instance->generate('post', $post->ID);
```
renatho:
```suggestion
		$instance->generate( 'post', $post->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertEmpty( get_metadata( 'user', $post->ID, WP_Job_Manager_Com_Auth_Token::META_KEY ) );
+	}
+
+	public function testValidate_WhenPassedValidUserButTokenIsExpired_ShouldReturnFalse() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$user = $this->factory->user->create_and_get();
+		$token = $instance->generate('user', $user->ID);
```
renatho:
```suggestion
		$token = $instance->generate( 'user', $user->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+		$this->expire_tokens( 'post', $post->ID );
+
+		// Act.
+		$result = $instance->validate( 'post', $post->ID, $token );
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	private function expire_tokens ( $object_type, $object_id ) {
```
renatho:
```suggestion
	private function expire_tokens( $object_type, $object_id ) {
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Assert.
+		$this->assertTrue( $result );
+	}
+
+	public function testValidate_WhenPassedPostWithMetaAndValidToken_ShouldReturnTrue() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$post = $this->factory->post->create_and_get();
+		$token = $instance->generate('post', $post->ID);
```
renatho:
```suggestion
		$token = $instance->generate( 'post', $post->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
```diff
@@ -0,0 +1,263 @@
...
+			];
+			update_metadata( $object_type, $object_id, WP_Job_Manager_Com_Auth_Token::META_KEY, $new_metadata, $metadata);
+		}
+	}
+
+	public function testValidate_WhenPassedUserWithMetaAndValidToken_ShouldReturnTrue() {
+		// Arrange.
+		$instance = WP_Job_Manager_Com_Auth_Token::instance();
+		$user = $this->factory->user->create_and_get();
+		$token = $instance->generate('user', $user->ID);
```
renatho:
```suggestion
		$token = $instance->generate( 'user', $user->ID );
```

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
renatho: IMO, the way you implemented it by updating the meta is good enough.

renatho (commented):
Sorry that you will see a lot of comments, but almost all of them are related to PHPCS, which seems to not be working on GH Actions.

I liked the code. It seems very organized and clear.
Just to understand, I'd like just to confirm your idea about user and post for the Promoted Jobs project. Are you planning to generate a token associated with the job? Or with the user?

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Yeah, I implemented that this morning. Thanks for the feedback!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: One tests if the instance returned is the same, another tests if it returns the correct type.

IMHO it is kinda a boilerplate test, as it is testing..the singleton (which isn't technically even needed by this class), but oh well

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

tests/php/tests/includes/helper/test_class.wp-job-manager-com-auth-token.php
fjorgemota: Fixed in 3e75085c236778ceb0fe8e4f055aca2b6a1323e3. Thanks!

includes/helper/class-wp-job-manager-com-auth-token.php
fjorgemota: Fair, renamed in 5f0ea4b519579ce82e49ddb2c9abb50dee44cfb4. Thanks!

fjorgemota:
>  Just to understand, I'd like just to confirm your idea about user and post for the Promoted Jobs project. Are you planning to generate a token associated with the job? Or with the user?

The idea of supporting `post` is to allow us to easily work with this issue: 433-gh-Automattic/wpjobmanager.com - and based on two assumptions:

- We will not use a `WP_User` in WPJMCOM to represent each site
- Most of this class will be shared with WPJMCOM, just with a few changes, like namespace, name of the class, etc.

Hope this answers your question. :)

aaronfc:
> Most of this class will be shared with WPJMCOM, just with a few changes, like namespace, name of the class, etc.

I am not sure I understand what is the purpose of having this code in WPJMCOM. AFAIK we will generate and validate the token on WPJM site only, right? On WPJMCOM I would expect only to "receive" a token and do a query to some public endpoint on the WPJM to assert that the result is "valid". Maybe I am missing something else?

aaronfc (commented):
> Most of this class will be shared with WPJMCOM, just with a few changes, like namespace, name of the class, etc.

I am not sure I understand what is the purpose of having this code in WPJMCOM. AFAIK we will generate and validate the token on WPJM site only, right? On WPJMCOM I would expect only to "receive" a token and do a query to some public endpoint on the WPJM to assert that the result is "valid". Maybe I am missing something else?

A part from that, I am also a little confused about how we will use this for the basic promote job flow. Would we use the 'user' type there or also the 'post' for the Job we want to promote? It's a bit confusing to me that this is "site authentication" but there is no way to authenticate the "site" as a whole (only users or posts from the site).

fjorgemota:
Hey there!

> I am not sure I understand what is the purpose of having this code in WPJMCOM. AFAIK we will generate and validate the token on WPJM site only, right? On WPJMCOM I would expect only to "receive" a token and do a query to some public endpoint on the WPJM to assert that the result is "valid". Maybe I am missing something else?

No, we will need this on WPJMCOM too, to work with this issue:  433-gh-Automattic/wpjobmanager.com, mentioned above, which will implement the reverse flow with the purpose of protecting #2442:

1. WPJMCOM receives the confirmation that the job was published by JobTarget
2. WPJMCOM generates a token (using this same system defined here), and associates it with the Promoted Site CPT (hence why this also supports `post`, more on this later);
3. It contacts the WPJM Site, more specifically the API defined in #2442, to update the job status, passing the token generated above, and probably the Promoted Site ID too (if the WPJM Site doesn't know it already)
4. WPJM Site validates the request is indeed coming from WPJMCOM (otherwise, anyone can update the job status, which would be annoying)
5. WPJMCOM verifies if the token is valid or not
6. WPJM Site proceeds with updating or not the job status

Hope this answer makes it clear why we need to have a very similar code on WPJMCOM, and also why we're supporting `post`.

> A part from that, I am also a little confused about how we will use this for the basic promote job flow. Would we use the 'user' type there or also the 'post' for the Job we want to promote?

The way I see it on the WPJM side we can either associate with the user OR with the post for the job we want to promote. . IMHO, it would make more sense to associate it with the user (e.g. so a token generated for me can't be validated for another user of the website, although that won't make a real difference in the end, to be honest), but keep in mind that, we'll need to share the ID of the thing with WPJMCOM. Are we associating the token with the user? We'll need to share the User ID. Are we associating the token with the listing being promoted? We'll need to share that listing's ID.

>  It's a bit confusing to me that this is "site authentication" but there is no way to authenticate the "site" as a whole (only users or posts from the site).

Uh..we're not authenticating the user or post from the site, but instead associating the token with that user or post, mostly as a storage per se, and by doing that we're also validating that the request and the site it is coming from is valid as well.

If we had the tokens persisted in a global way (let's say, saving as a option in the database), we would have the risk of making the token slightly less secure. When using the tokens from the WPJMCOM side, for instance, one token generated for each site would be valid for another site as well, which I don't think would be expected. This might be an edge-case in the case of the Promoted Jobs project, yeah, but oh well..

Also, I couldn't think an use case where it would be useful to have it persisted in a more..."global" way. Have you thought of any? I'd love to hear more of your thoughts about this.

Thanks for all these questions, by the way!

renatho (approved):
Thank you, Fernando!

I'm approving the PR on both sides (here and on wpjobmanager.com). Are you planning to come with another PR with the requests that connect them, allowing them to create/edit the jobs on behalf of the specific site?

fjorgemota:
Hey there, thanks for your review!

> I'm approving the PR on both sides (here and on wpjobmanager.com). Are you planning to come with another PR with the requests that connect them, allowing them to create/edit the jobs on behalf of the specific site?

As discussed on Slack, yes. 

Thanks!

PR #2467, opened 2023-06-21
Update/promoted jobs schema

mikeyarce:

* `job_type` is now an array since it's possible to have multiple job types set
* `salary` is now an array with the properties of `currency`, `value`, and `unit`.


docs/api/internal.json
```diff
@@ -82,19 +82,38 @@
                             "example": false
                           },
                           "job_type":  {
```
renatho, line 84: I think it should be something like this:

```
"job_type":  {
  "type": "array",
  "items" : {
    "type": "string",
    "enum": [
      "freelance",
      "full-time",
      "internship",
      "part-time",
      "temporary",
      "other"
    ]
  }
},
```

docs/api/internal.json
```diff
@@ -82,19 +82,38 @@
...
+                                  "type": "string",
+                                  "example": "USD"
+                                },
+                                "value": {
+                                  "type": "string",
+                                  "example": "$50,000 - $100,000"
+                                },
+                                "unit": {
+                                  "type": "string",
+                                  "example": "YEAR"
```
renatho: This could also be an `enum`, right?

docs/api/internal.json
```diff
@@ -82,19 +82,38 @@
...
+                            ],
+                            "items" : {
+                              "type": "string",
+                              "example": "Full Time"
+                            }
                           },
                           "salary":  {
-                            "type": "string",
-                            "example": "$50,000 - $100,000"
+                            "type": "array",
```
renatho: I think these 2 lines we need to remove, right?

```
"type": "array",
"items": {
```

docs/api/internal.json
renatho: @aaronfc, I was talking with @mikeyarce about this one. It seems there's a setting where the users can define multiple job types.
But JobTarget only accepts one. So since you're integrating the form, I'd suggest you consider receiving an array, but get only the first item to be used to pre-fill the job type field.

mikeyarce: Fixed in [513b102](https://github.com/Automattic/WP-Job-Manager/pull/2467/commits/513b1027ba481fe00f73f47d8ab300df43358408)

docs/api/internal.json
mikeyarce: Fixed in c7dc9536e2935b9c52f42781fc9f23f99d60e14f

docs/api/internal.json
mikeyarce: Fixed in c7dc9536e2935b9c52f42781fc9f23f99d60e14f

renatho: I think now we removed an extra curly brace.
You can use this tool to validate the document: https://editor.swagger.io/ 😉

mikeyarce: Nice tool!  Fixed now in 0c48b308d226b71423baeee481f4a061d844ada5

aaronfc (commented):
Hey, thanks for the updates!

About the salary, initially I don't think we are going to use it (we are asking for the salary on wpjobmanager.com).
That's why I left it as a free string, that we can "try" to parse on wpjobmanager.com in a "best effort" approach.

Anyway, if we want to go ahead and make it more structured. I am not very convinced on the proposal. What do you think about something like this?
```
salary: {
  low: {
    currency: "EUR",
    value: 10000,
  },
  high: {
    currency: "EUR",
    value: 100000
  },
  period: "YEAR"
}
```

- Having low and high values per separate.
- Each quantity having it's currency (even though normally we would want to use the same for both).
- Having a common "period" to set expectations on the period for the mentioned salary range.
- Both low and high would be optional. As well as the whole salary attribute. 

**Edit**:
> About the salary, initially I don't think we are going to use it (we are asking for the salary on wpjobmanager.com).
That's why I left it as a free string, that we can "try" to parse on wpjobmanager.com in a "best effort" approach.

Same for the job type. AFAIK we are going to ask for it in the Job Form on wpjobmanager.com. @renatho in the work that @yscik made this selector is a dropdown. Should we change it to support multiple choices?

renatho:
> Having low and high values per separate.

Not sure if we should try to do it on the WP Job Manager side because we don't even suggest the format of low/high for the users, currently. See screenshots:

Frontend:
Screenshot 2023-06-26 at 11 17 19

WP Admin:
Screenshot 2023-06-26 at 11 18 58

If we consider using that, we could try to parse it in wpjobmanager.com. If it has some separator, we could extract the low and high, otherwise, we could add the same value to the lower and to the higher. WDYT?

> Same for the job type. AFAIK we are going to ask for it in the Job Form on wpjobmanager.com. @renatho in the work that @yscik made this selector is a dropdown. Should we change it to support multiple choices?

I think it's similar to the previous case. We have the option to ask for multiple items (it needs to be enabled through a Job Manager setting), but in JobTarget we support only 1 value. So I'd make this API aligned with WP Job Manager, and in wpjobmanager.com we parse it and get the first one. So the user can edit it to another option if they want. Does it make sense to you?

aaronfc:
> Not sure if we should try to do it on the WP Job Manager side because we don't even suggest the format of low/high for the users, currently. See screenshots

Oh! I didn't know we had support for salary, currency and salary unit. This makes much more sense now.
Same for job type.  I am fine with the initial proposal then :)

docs/api/internal.json
```diff
@@ -53,7 +53,7 @@
                         "type": "object",
                         "properties": {
                           "id": {
-                            "type": "string",
+                            "type": "integer",
```
renatho: @aaronfc, I [suggested updating it to int](https://github.com/Automattic/WP-Job-Manager/pull/2469#discussion_r1238917056), but given this conversation (p1687532087992389-slack-C053397HN30), should we revert it to `string` and update it in the API to be parsed to `string`?

Sorry for the probably wrong suggestion here, @mikeyarce! 🤭

mikeyarce: No problem!  Done in c645824841fb08978a41b4f9c6984790b32b55fe

renatho (approved):
Thank you, @mikeyarce!

Issue #2462, opened 2023-06-19
Add deactivate Modal


mikeyarce:
Add Deactivate Modal

![Image](https://github.com/Automattic/WP-Job-Manager/assets/3220162/0039f286-7609-4c29-a269-2cb48f5c758c)

renatho:
Hey @mikeyarce! Isn't this issue duplicated of part of https://github.com/Automattic/WP-Job-Manager/issues/2438?

mikeyarce:
Closing!

PR #2468, opened 2023-06-21
Add deactivate modal

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2462

* Adds Modal for Deactivate options
* Simplifies CSS for both modals (share styles between two modals)

### Testing instructions

* Go to Job Listings Page
* Make a job Promoted by adding the post meta to a job listing `_promoted => 1`
* Click on "Deactivate" on the Promote column for that job
* Try clicking Cancel, pressing Escape, and clicking the `X` to leave the modal

### Screenshot / Video

deactivate-dialog

### DIFF
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -27,9 +27,7 @@ promote-job-modal {
 	}
 	h2.promote-jobs-heading {
 		font-size: 36px;
-		font-weight: 300;
 		line-height: 105%;
-		margin-top: 0;
 		width: 80%;
 		margin-bottom: 0px;
 	}
@@ -59,15 +57,6 @@ promote-job-modal {
 			font-weight: 700;
 		}
 	}
-	.promote-buttons-group {
-		.button {
-			padding: 5px 16px;
-			margin-right: 18px;
-		}
-		.button-secondary {
-			background: #ffffff;
-		}
-	}
 }
 @media screen and ( max-width: 1000px ) {
 	.promote-job-modal-column-right {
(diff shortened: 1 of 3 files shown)


assets/js/admin.js
```diff
@@ -190,8 +190,19 @@ function wpjmModal( selector, dialogSelector ) {
 		});
 	});
 }
+function wpjmCloseModal( selector, dialog ) {
```
renatho, line 193: Since it's part of the modal logic, what do you think about adding this logic as part of the `wpjmModal` function too? The button, instead of `cancel-promotion` could have some generic class for any type of modal.

assets/css/admin.scss
```diff
@@ -82,6 +71,10 @@ promote-job-modal {
 .wpjm-dialog {
 	border: 0;
 	border-radius: 8px;
+	h2.promote-jobs-heading {
```
renatho, line 74: It seems a little mixed because we have the `wpjm-dialog` generic class for any modal, and inside it we have classes specific for promote jobs modals.

I think your idea here was to re-use the same styles for both modals, right? Since you're considering it generic, maybe we could just change to generic names instead of related to promote jobs?

assets/css/admin.scss
```diff
@@ -99,6 +92,44 @@ promote-job-modal {
...
+			padding: 5px 16px;
+			margin-right: 18px;
+		}
+		.button-secondary {
+			background: #ffffff;
+		}
+	}
+}
+/** Deactivate Modal **/
+#deactivate-dialog {
```
renatho, line 106: I'd recommend as a good practice to use class for styles instead of id. You can keep the id to use in JS, but you could add an extra class to the element in order to add specific styles there.

renatho (commented):
Thank you @mikeyarce! I just added a few comments suggesting an organization about the generic vs specific parts.

mikeyarce:
Since so much changed to the modal, I made a new PR but took into account your feedback from this PR.  
New PR: https://github.com/Automattic/WP-Job-Manager/pull/2473

Issue #2437, opened 2023-05-31
Implement endpoint to expose promoted jobs feed


renatho:
Jobs marked as promoted (a post meta) should be exposed in a feed.
See schema created in https://github.com/Automattic/WP-Job-Manager/issues/2436.

### Success Criteria

- [ ] We can access an endpoint to see a list with all promoted Jobs.
- [ ] Jobs not promoted don't show up in the list.

mikeyarce:
A couple questions:
- Is the `description` meant to be the `post_content`
- Should `is_remote` output `1` or `0` or do we want it to say `true` or `false`? 
- I'm assuming this is a public api endpoint, so we don't need any permissions callbacks?
- Should we also add `_job_salary_currency` and `_job_salary_unit`?

I've got this as a PoC right now, does this look like what we want?
![igNkCn.png](https://github.com/Automattic/WP-Job-Manager/assets/3220162/502b9da0-96b4-4efd-875a-f1afa7cd4cb8)

renatho:
Hey @mikeyarce!

> Is the description meant to be the post_content

Yep!

> Should is_remote output 1 or 0 or do we want it to say true or false?

It's a boolean `true` or `false`.

> I'm assuming this is a public api endpoint, so we don't need any permissions callbacks?

Correct!

> Should we also add _job_salary_currency and _job_salary_unit?

Since we have the information, I think it's fine to include it. We won't use it for now, we're including it for future use only, so I think it makes sense to add it. Maybe transform the `salary` in an object with the 3 fields. Just remember to update the schema if you change it. 😉

> I've got this as a PoC right now, does this look like what we want?

The only other thing I noticed in your screenshot is the `job_type` that should be one of these options:

```
"freelance",
"full-time",
"internship",
"part-time",
"temporary",
"other"
```

mikeyarce:
Thanks @renatho !

> Since we have the information, I think it's fine to include it. We won't use it for now, we're including it for future use only, so I think it makes sense to add it. Maybe transform the salary in an object with the 3 fields. Just remember to update the schema if you change it. 😉

👍🏽 

> The only other thing I noticed in your screenshot is the job_type that should be one of these options:

Ok, will use the term`slug` for the `job_type`.   New question though:  if a job listing has more than one job type, is a string separated with comma's fine?

renatho:
> if a job listing has more than one job type, is a string separated with comma's fine?

[...]

PR #2469, opened 2023-06-21
Add promote jobs API endpoint

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2437

* Add a REST API endpoint to show all promoted jobs at `/wp-json/wpjm-internal/v1/promoted-jobs`

### Testing instructions

* Check out this PR
* Make a few job listings promoted by adding the `_promoted => 1` post meta
* Go to the REST API endpoint to ensure they show up there.  Example:`http://wpjm.local/wp-json/wpjm-internal/v1/promoted-jobs`

### Screenshot / Video
Screenshot 2023-06-21 at 3 27 19 PM


includes/class-wp-job-manager.php
```diff
@@ -175,6 +175,8 @@ public function include_template_functions() {
 	 */
 	public function rest_init() {
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/class-wp-job-manager-rest-api.php';
+		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/class-wp-job-manager-promoted-jobs-api.php';
```
renatho: I wonder if it would be better to initialize it as part of the `WP_Job_Manager_Promoted_Jobs` class. Not sure what is the standard we're using for other endpoints, if we have something defined already though.

I also just noticed that `WP_Job_Manager_Promoted_Jobs` is being included in the `includes/admin/class-wp-job-manager-admin.php`. Maybe we should include it in a more generic class since we'll have things not related to the admin?

includes/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -0,0 +1,126 @@
...
+	 */
+	public function get_items( $request ) {
+		$args = [
+			'post_type'           => 'job_listing',
+			'post_status'         => 'publish',
+			'no_found_rows'       => true,
+			'ignore_sticky_posts' => true,
+			'meta_query'          => [
+				[
+					'key'     => '_promoted',
```
renatho: Maybe we could add the key as a constant in the `WP_Job_Manager_Promoted_Jobs` to make it simpler to update in the future if we want?

includes/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -0,0 +1,126 @@
...
+	public function prepare_item_for_response( $item, $request ) {
+		$terms = get_the_terms( $item->ID, 'job_listing_type' );
+
+		$terms_array = [];
+		foreach ( $terms as $term ) {
+			$terms_array[] = $term->slug;
+		}
+
+		$data = [
+			'id'           => $item->ID,
```
renatho: I noticed we defined it as `string` in the schema and we're sending an `int` here. Maybe we should update the schema to match it?

renatho (commented):
In general, it's looking good! :)

Just adding some questions and suggestions.

includes/class-wp-job-manager-promoted-jobs-api.php
mikeyarce: Fixed in 3da81bbc1bf2950f68cad1a02b1ecae542bf2afe

includes/class-wp-job-manager-promoted-jobs-api.php
mikeyarce: Done in 58eba79ed4fd14e02ea37a234d6e67bfd4430e05

includes/class-wp-job-manager.php
mikeyarce: What if we include `WP_Job_Manager_Promoted_Jobs` in `class-wp-job-manager.php` instead, then we can load the rest API from `WP_Job_Manager_Promoted_Jobs`?  

For the plugin side, what will it have that's not related to admin?

includes/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -0,0 +1,134 @@
...
+		if ( empty( $items ) ) {
+			return rest_ensure_response( $data );
+		}
+
+		foreach ( $items as $item ) {
+			$itemdata = $this->prepare_item_for_response( $item, $request );
+			$data[]   = $this->prepare_response_for_collection( $itemdata );
+		}
+
+		return new WP_REST_Response( $data, 200 );
```
renatho: @mikeyarce, sorry for this late suggestion here, but I noticed something while working on wpjobmanager.com pointing to this PR.

It would be nice if we add a property here to have the array, instead of returning it directly. Maybe `jobs`?

So it gets easier if we need to expand this endpoint in the future adding some extra information.

mikeyarce: Good idea!  Done in 86f948e7d8002352740f0fb718533af78075e1a9

includes/class-wp-job-manager.php
renatho: > What if we include `WP_Job_Manager_Promoted_Jobs` in `class-wp-job-manager.php`.

I think it makes sense!

> For the plugin side, what will it have that's not related to admin?

It would be the admin part and the REST APIs, for this first iteration of the project.

includes/class-wp-job-manager-promoted-jobs-api.php
renatho: I think it would be better to have it as a constant in the main class related to Promoted Jobs, so we could use it in the different classes we'll have.

But I think it also depends on the decision about the architecture of the files and folders. p1687804255327959-slack-C036R3Y5ET0

mikeyarce:
There is a PHP Lint error about duplicate classes of `WP_Job_Manager_Promoted_Jobs` - but that's on purpose!  After this is merged I'll move that class into the `promoted_jobs` folder and rename it to `WP_Job_Manager_Promoted_Jobs_Admin`.

includes/class-wp-job-manager-promoted-jobs-api.php
mikeyarce: Good idea, done in 806aaf1631f9237f43fe6bd47513fdcf32384353

includes/class-wp-job-manager.php
mikeyarce: Done in https://github.com/Automattic/WP-Job-Manager/pull/2469/commits/dcbf48a0a3346bdadbf31095574b7f2e3fa3c76b

renatho (approved):
Sorry for the delay in checking your answers! Looks good to me! =)

About the admin class, maybe we could also move it to the `promoted-jobs` when we do that re-organization in another PR. Do you think it makes sense?

PR #2471, opened 2023-06-23
Add/notify listing change

### LINKED ISSUE #2441: Notify wpjobmanager.com when a promoted job is changed
The sites should notify wpjobmanager.com when a job is changed (See [schema](https://github.com/Automattic/wpjobmanager.com/issues/345)).

- The trigger should be when a promoted job is removed, deactivated or updated (fields used in the feed only).
- It should call an endpoint in wpjobmanager.com notifying that a job was updated.
- In case of an error in this request, it should retry in some hours with a scheduled job (maximum 3 times).

### Success Criteria

- [ ] When a field used in job feed is updated in a promoted job, it should notify wpjobmanager.com.
- [ ] When a field used in job feed is updated in a **not** promoted job, it should **not** notify wpjobmanager.com.
- [ ] When a field **not** used in job feed is updated in a promoted job, it should **not** notify wpjobmanager.com.
- [ ] In case of an error, the request retries in some hours.

onubrooks:
Fixes #2441 

* Adds hook for updated jobs.
* ~Hooks into `wp_trash_post`~ Now covered in `post_updated` hook.
* Sends notification to `wpjobmanager.com` when job listing is changed or updated.
* Schedules job to retry 3 times if api call fails.

### Testing instructions

This is pretty manual for now.
* Set the post meta for a job as `_promoted` and set it to 1
* Update the job listing in one of these fields: `'job_title', 'job_description', 'company_name', 'application', 'job_location'`
* The API call should fail for now [ testing with wpjobmanager.com]
* Check the new option on the database: `job_manager_should_notify_promoted_jobs` to see the values set and the error message, `last_run`, `should_notify` etc
* Use a cron job plugin to check that a new job is scheduled to run: `job_manager_promoted_jobs_notification`
* If the API call succeeds, the cron job should be removed.

### New/Updated Hooks

* ~`job_manager_edit_job_listing_before_save`: hook that fires just before a job edit is persisted to the database`.

### Some Notes

- I added a new hook `job_manager_edit_job_listing_before_save` because the hook `job_manager_user_edit_job_listing` fires **after** save and by this time we don't know the exact fields that changed.
- I have set the cron schedule to `twicedaily` just as an arbitrary value. Not sure how often we would want the job to run.
- As an alternative idea to retrying 3 times, this job is not compute intensive. Also, since a notification triggers a whole fetch of the job feeds, if one update fails and the next one succeeds, there's no need to retry the first. Therefore, this implementation determines whether to retry by setting the `should_notify_jobs` option to a boolean. If one job fails, it is set to true, if another succeeds, it is set to false so there's no need to retry the first.
- There's a `get_site_id` method that calls `get_current_blog_id`. Is this okay or there's something else that we call the site ID.
- A comment in the issue says: **The trigger should be when a promoted job is removed, deactivated or updated**, I have interpreted it to mean delete and update. Anything else? What happens when the post status is changed from say 'publish' to 'pending' or 'expired'? Some sites change a post status to pending if it was edited for moderation, this is done in the settings.


includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,335 @@
...
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->watched_fields = [ 'job_title', 'job_description', 'company_name', 'application', 'job_location' ];
+		$this->init_options();
+		add_action( 'wp_trash_post', [ $this, 'promoted_job_trashed' ] );
+		add_action( 'job_manager_edit_job_listing_before_save', [ $this, 'promoted_job_updated' ], 10, 2 );
+		add_action( 'job_manager_save_job_listing', [ $this, 'promoted_job_updated_admin' ], 10, 2 );
+		add_action( 'job_manager_promoted_jobs_notification', [ $this, 'run_scheduled_promoted_jobs_notification' ] );
```
renatho: Checking your PR, I realized I missed one extra scenario that is changing the job to a status that is not published. Could you also handle it here? Or we could open a new issue for this too, if you prefer.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,335 @@
...
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->watched_fields = [ 'job_title', 'job_description', 'company_name', 'application', 'job_location' ];
```
renatho: My fault while specifying the issue. But could you update it to be only check title, description and job permalink?

The other ones we'll just use to help prefill some fields because they don't match exactly JobTarget API. Company name would match, but since it's not required in WP Admin, I think we'll update the company name only if they edit it in wpjobmanager.com form. Does it make sense to you too @aaronfc?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,335 @@
...
+	 * @param int     $post_id Post ID.
+	 * @param WP_Post $post Post object.
+	 * @return void
+	 */
+	public function promoted_job_updated_admin( $post_id, $post ) {
+		if ( ! $this->should_notify( $post_id ) ) {
+			return;
+		}
+
+		if ( $this->post_has_changed( $this->get_watched_fields_meta( $post_id ), $this->get_post_data( $post ) ) ) {
```
renatho: For some reason, it's always returning with no changes for me. In the `post_has_changed` it receives the new values in both arguments.

Let me know if you can't reproduce the issue here, so I can help you to debug. 😉

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,335 @@
...
+		}
+	}
+
+	/**
+	 * Notify wpjobmanager.com when a Job Listing changes.
+	 *
+	 * @return void
+	 */
+	private function send_notification() {
+		$response = wp_remote_post( $this->get_notification_url() );
```
renatho: I imagine it'd be safer to use the `wp_safe_remote_post` instead.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,335 @@
...
+		$this->maybe_schedule_rerun( $response );
+	}
+
+	/**
+	 * Schedule a cron job to run the notification task if API call failed, unschedule if it succeeds.
+	 *
+	 * @param WP_Error|array $response Response from wp_remote_post.
+	 * @return void
+	 */
+	public function maybe_schedule_rerun( $response ) {
```
renatho: Do you think the retry could be simplified a little using an approach similar to [this one](https://github.com/Automattic/wpjobmanager.com/pull/444/files#diff-205ccc80cb5c8b42278efe5172eee92cf972cd0476825c0cb6b013c2a49f13c3R112-R119) (basically scheduling a single job after each fail instead a cron job and control the number of retries through its arguments)?

renatho (commented):
Hi Onuh! In general, it's looking good! But for some reason, it didn't work for me on the admin side. I added a few more details in the comments. Feel free to reach out if you can't reproduce the issue that is happening for me. 😉

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	/**
+	 * Notify wpjobmanager.com when a Job Listing is updated.
+	 *
+	 * @param int   $post_id Post ID.
+	 * @param array $values Values.
+	 * @return void
+	 */
+	public function promoted_job_updated( $post_id, $values ) {
```
renatho: I'm wondering if controlling it through the `save_post` for both options would be better to control the change in a single method. Also, it could catch the job listing being updated through other ways too (some other code calling `wp_update_post`, for example) which I think these 2 methods wouldn't catch, right?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Thanks @renatho 
I missed that `update_postmeta` was also a hook. Simplified it a lot in 7ebc3e1f265fa7c6b7f4995c47483d896fcdb845

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: 7ebc3e1f265fa7c6b7f4995c47483d896fcdb845

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Never mind about this, should work now in 7ebc3e1f265fa7c6b7f4995c47483d896fcdb845

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Yep thanks.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Yeah, much simpler in baf019ff677d12aedc7a154efdce9ed17f194328

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: 7ebc3e1f265fa7c6b7f4995c47483d896fcdb845
`post_updated` and `update_postmeta` hooks cover everything including `status`.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -0,0 +1,231 @@
...
+			if ( $current_value !== $meta_value ) {
+				$this->send_notification();
+			}
+		}
+	}
+
+	/**
+	 * Initialize options.
+	 */
+	public function init_options() {
```
renatho: I think it's something we're not using anymore, right?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
renatho: Should we also add `post_name` to the watched_fields in order to also get the job permalink change?

Also, I think `_application` that is there would be to get the "Application email/URL", right? This one we don't need to monitor.

But there's another one that I realized that we're not monitoring. We also need a notification when a promotion is deactivated. Maybe we could just monitor the `_promoted` meta changes using your logic here?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
renatho: Happy that it could be simplified! =)

While reviewing your PR, I just realized a case that we need to handle and I also fixed it in my PR.

If it fails and already has a retry scheduled, we don't need to schedule another one. And if it works, we can cancel existing retries.

For reference, it's what I did in my PR: https://github.com/Automattic/wpjobmanager.com/pull/444/commits/fa2af0ace56b89b8d006218457f2edb7d06a37c4, but I think you can make it simpler here, not needing to check the arguments.

In my PR it was more complicated because I couldn't clear all the hooks because we have arguments for different sites. In your PR, I think you can handle it more simply, checking and canceling all hooks with the retry name `wp_next_scheduled( self::RETRY_JOB_NAME )` / `wp_clear_scheduled_hook( self::RETRY_JOB_NAME )`. WDYT?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
renatho: Nice! Code looks great now!

I just noticed that after changing post and post meta (tested in WP Admin), it's calling the notification 2 times. It seems one is coming from an ajax request and the other not.
Do you have any idea how to trigger it a single time only after both were changed?

onubrooks: No idea but since we now have only `_promoted` as the meta we are interested in and it doesn't get updated in admin, it won't call the notification twice anymore.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Yep, updated in 66d1ab5d853f6e5fc1d433563817c0ab8eccf3cf

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Added in a6b619b2b8718671e7c4b2dc133f96f399c47dd2

Note however that I'm using `wp_unschedule_hook` to cancel. 
Both `wp_clear_scheduled_hook` and `wp_next_scheduled` actually require the exact arguments(ie we'll need to loop through all retry values as in your PR). 
With `wp_unschedule_hook`, it can all be cleared on one line.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Yeah thanks, removed!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
renatho: Oh! Good point!
I was a little worried about how complicated would be this. That was easy! 🙌 haha

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
renatho: Ah! TIL! I didn't know it was a different function when using it without arguments. Do you think we also have a similar one for the `wp_next_scheduled` to avoid the other loop? I tried to Google quickly, but I didn't find anything.

renatho (approved):
Looks great and works well (tested on the frontend and on the backend)!

I just left a last comment about something with the loop to get the scheduled event, but if there's not a function for that, feel free to skip it.

And answering your question that I missed previously:

> There's a `get_site_id` method that calls `get_current_blog_id`. Is this okay or there's something else that we call the site ID.

That's good for now! I created an issue to handle this "connection" later when all the necessary parts are done: https://github.com/Automattic/wpjobmanager.com/issues/463

I think the rest of the questions are things we already discussed, right? If I missed anything else, let me know.

Thanks!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
onubrooks: Yeah I checked too check too but didn't find any.

PR #2472, opened 2023-06-23
Promoted jobs modal improvements

mikeyarce:

* Don't output the template twice
* Use JS to just render the buttons
* Remove all extra slots
* Simplify the template
* Move styles to `<style>` inside template
* Move dialog close button to JS because it won't work if it's inside template

### Testing instructions

* Go to Job Listings
* Click "Promote" 
* Make sure you see the modal

### DIFF
--- a/includes/admin/class-wp-job-manager-promoted-jobs.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs.php
@@ -115,6 +115,98 @@ class WP_Job_Manager_Promoted_Jobs {
 						<div class="price-text">Starting From</div>
 						<span>$83.00</span>
 					</div>
+		<style>
+			promote-job-modal {
+				display: grid;
+				grid-template-columns: repeat(auto-fit, minmax(240px, 1fr));
+				padding: 30px 80px 50px 80px;
+			}
+			promote-job-modal img.promote-jobs-image {
+				width: 100%;
+			}
+			promote-job-modal h2.promote-jobs-heading {
+				font-size: 36px;
+				font-weight: 300;
+				line-height: 105%;
+				margin-top: 0;
+				width: 80%;
+				margin-bottom: 0px;
+			}
+			promote-job-modal .promote-job-modal-column-left {
+				display: flex;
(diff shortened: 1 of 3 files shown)


includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -101,39 +101,124 @@ public function promoted_jobs_custom_columns( $column ) {
...
+				background: none;
+				border: 0;
+				font-size: 0;
+			}
+			form.dialog button.dialog-close:after {
+				content: "\2715";
+				font-size: 20px;
+			}
+		</style>
+		<promote-job-modal>
```
yscik: Since this is not a custom element anymore, should it be just a div, like `<div class="promote-job-modal">`? (With the CSS selectors above also updated)

yscik (approved):
Looks really nice!

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -101,39 +101,124 @@ public function promoted_jobs_custom_columns( $column ) {
...
+				}
+			}
+			.wpjm-dialog {
+				border: 0;
+				border-radius: 8px;
+			}
+			form.dialog {
+				display: flex;
+			}
+			form.dialog button.dialog-close {
```
yscik, line 184: Tiny thing, adding a `cursor: pointer` for the close button (and maybe a color change on hover) would ensure it looks clickable.

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Done in 9821f24315062a7b5fc4d0ee58dcda550f655daf

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Done in 4365970ac85fc95ef23f278995de335d23872510

Issue #2440, opened 2023-05-31
Integrate deactivate button on Job Listing


renatho:
- When a job is promoted, we should display the deactivate button instead of Promote button.
- When clicking on the button, it should display a [modal](https://github.com/Automattic/WP-Job-Manager/issues/2438) asking for confirmation.
- After user confirmation in the modal, it should mark the job as not promoted (remove post meta that represents that it's promoted).
- Restrict this feature with a capability.

### Success Criteria

- [ ] Promote button doesn't show up in the Job Listing when a Job is promoted.
- [ ] Deactivate button shows up in the Job Listing when a Job is promoted.
- [ ] A modal with confirmation shows up when clicking on the deactivate button and we can interact with it.
- [ ] When users confirm the prompt, job is not promoted anymore.
- [x] #2494

Hastibe:
Could you clarify what you mean by "promote" / "promoted"? WP Job Manager allows a job to featured, is that what you mean by promoted?

renatho:
Hi @Hastibe!

Thank you for asking! We are working on a feature to allow promoting jobs from WP Job Manager on external job boards. We hope it's helpful for you and for many other users! 🙂

Hastibe:
Ah, thanks for clarifying--I was really confused! Looking forward to learning more about this.

PR #2473, opened 2023-06-23
Add deactivate modal

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2440 (partly)
Based on the changes to the promoted jobs modal: https://github.com/Automattic/WP-Job-Manager/pull/2472
It's missing the part that actually makes the job not promoted any more - this is just the scaffolding for that.

* Adds Modal for Deactivate options

### Testing instructions

* Go to Job Listings Page
* Make a job Promoted by adding the post meta to a job listing `_promoted => 1`
* Click on "Deactivate" on the Promote column for that job
* Try clicking Cancel, pressing Escape, and clicking the `X` to leave the modal

### Screenshot / Video

deactivate-dialog

### DIFF
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -50,6 +50,38 @@ a.wpjm-activate-licence-link:active {
 	}
 }
 
+/** Deactivate Modal **/
+.deactivate-dialog {
+	h2.deactivate-modal-heading {
+		font-size: 20px;
+		line-height: 28px;
+		margin-top: 0;
+		font-weight: 300;
+	}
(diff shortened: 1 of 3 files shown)


yscik (approved):
LGTM. It could be useful in the long run to build this dialog to be more generic so we can use it in other places. Maybe the `action` parameter for `wpjmModal` can be `content` instead, a callback function setting up the modal content? 
I'm working on the integration part for the deactivate action, I'll experiment a bit with a more generic helper.

PR #2475, opened 2023-06-27
Change "Site Authentication" feature to "Site Trust", and make small changes to the constants

fjorgemota:
This PR is sibling to the one in  458-gh-Automattic/wpjobmanager.com

This PR was sent to change some small things in the previously called "Site Authentication" feature:

* Rename class from `WP_Job_Manager_Com_Auth_TOKEN` to `WP_Job_Manager_Site_Trust_Token`, to align with the name used in WPJMCOM
* Change meta key to have a `_` prefix (so it is considered "private" and can't be edited by the user), and refer to the token in plural, as the meta key might save more than one token
* Improve error codes/messages used on the `WP_Job_Manager_Site_Trust_Token` class
* Fix a rare case where the token might be empty, which could generate weird results

### Testing instructions

The changes are fairly small, but, if you want to test in practice, you can run the unit tests and experiment creating tokens using WP-CLI Shell:

```php
$instance = WP_Job_Manager_Site_Trust_Token::instance(); // To get the instance of the class
$token = $instance->generate( 'user', 1 ); // Create a token associated with the User ID 1
// Maybe await a minute using sleep(60)?
var_dump($instance->validate( 'user', 1, $token));
// Make the same test with a valid post instead of user
```

### DIFF
--- a/includes/helper/class-wp-job-manager-site-trust-token.php
+++ b/includes/helper/class-wp-job-manager-site-trust-token.php
@@ -34,9 +34,9 @@ class WP_Job_Manager_Site_Trust_Token {
 	/**
 	 * The singleton instance of the class.
 	 *
-	 * @var self
+	 * @var ?self
 	 */
-	private static $instance;
+	private static $instance = null;
 
 	/**
 	 * WP_Job_Manager_Site_Trust_Token constructor.
--- a/includes/helper/class-wp-job-manager-site-trust-token.php
+++ b/includes/helper/class-wp-job-manager-site-trust-token.php
@@ -127,6 +127,7 @@ class WP_Job_Manager_Site_Trust_Token {
 		if ( false === $metadatas ) {
 			return false;
 		}
+		$found = false;
 		foreach ( $metadatas as $metadata ) {
 			if ( ! $this->is_valid_format( $metadata ) ) {
 				// If the metadata structure isn't valid, just ignore it.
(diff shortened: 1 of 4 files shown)


includes/helper/class-wp-job-manager-site-trust-token.php
```diff
@@ -127,6 +123,7 @@ public function validate( $object_type, $object_id, $token ) {
 		if ( false === $metadatas ) {
 			return false;
 		}
+		$found = false;
```
renatho, line 126: Out of curiosity, any special reason for this change?

renatho (approved):
Looks good to me!

includes/helper/class-wp-job-manager-site-trust-token.php
renatho: Ah, now that I saw it in the PR description. Probably related to this, right?

> Fix a rare case where the token might be empty, which could generate weird results

fjorgemota: Nope, sorry, this change was introduced in 878b90e38f0f6c3c150a2a5d6d34f606389452b6.

The main idea is to allow us to clean the database while NOT stopping the search early. So if for instance for some reason I have the following sequence of metas:

1. token that I want, which is not expired
2. some expired token 2 
3. some expired token 3 
4. some expired token 4 

And it is returned in that exact order...in the previous code, the three expired would never be deleted (well, unless we called `validate` with a token that simply doesn't match anything, or is expired by itself). Now, even if we have a successful result, all the expired tokens will still be deleted as expected.

Hopefully this explains the change well. But if doesn't, please let me know. :)

renatho: Ah! Gotcha! It expains well! Thanks!

Issue #2464, opened 2023-06-19
Consume promote dialog from wpjobmanager.com


renatho:
https://github.com/Automattic/WP-Job-Manager/pull/2460 introduced the promote modal, and it needs to consume the content from wpjobmanager.com, which is being added in https://github.com/Automattic/wpjobmanager.com/pull/431.

This issue also includes to extract the modal content styles (introduced in https://github.com/Automattic/WP-Job-Manager/pull/2460), and add it to wpjobmanager.com endpoint.

### Success Criteria

- [x] Promote modal content comes directly from wpjobmanager.com, so if it's changed there, it updates in the plugin without needing a plugin update.
- [x] Styles also come directly from wpjobmanager.com
- [x] It sends the site lang as argument while fetching, so we can [translate it](https://github.com/Automattic/wpjobmanager.com/issues/410).

renatho:
https://github.com/Automattic/WP-Job-Manager/pull/2478

PR #2478, opened 2023-06-27
Fetch remote promoted template

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2464

* Fetch template from wpjobmanager.com
* Store in cache for 1 day
* Add function to check if template is stored
* Hide `Promote` column if there is no template to show (if our site is down, then none of these actions will work)

### Testing instructions

* You need to change the URL from `wpjobmanager.com` to whatever your local test site is
* What I did was install `wpjmcom-plugins` on my test site.  Make sure to have PHP >= 8.1 or else your site will explode
* `wpjmcom-plugins` needs to be using the `feature/promoted-jobs` branch which has the right API
* Go to the Jobs Listings page, make sure the modal works as expected

## Notes

Right now I'm hiding the entire promote column if the promote template is not available.  My thinking is that if the site is not accessible, likely none of the actions will work, is that right?


includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -90,135 +93,34 @@ public function promoted_jobs_custom_columns( $column ) {
...
+		$promote_template = wp_cache_get( 'promote-jobs-template', 'promote-jobs', false, $found );
 
-				<ul class="promote-list">
-					<li class="promote-list-item">Your ad will get shared on our Partner Network</li>
-					<li class="promote-list-item">Featured on jobs.blog for 7 days</li>
-					<li class="promote-list-item">Featured on our weekly email blast</li>
-				</ul>
+		if ( ! $found ) {
+			$response         = wp_remote_get( 'http://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . get_locale() );
+			$promote_template = json_decode( $response['body'], true );
```
yscik: In case `wp_remote_get`returns a WP_Error, could this line fail already? The `is_wp_error` check probably needs to be before an attempt to parse the response. 
Or could be safest to wrap this all in a try-catch :)

mikeyarce: Done in 126f1c1768c3e9b467a01691ff31806d962c44f7
Also, what do you think about cache storage time?  I orignally was thinking maybe a day, but if the response isn't there it might be broken for a long time so I changed it to 1 hour.

yscik: I think we only need to update the cache if everything was successful, putting `wp_cache_set` after the 
```
$promote_template = json_decode( wp_remote_retrieve_body( $response ), true );
``` 
line. (And also caching the already parsed template instead of the response). 

That'd take care of this, falling back to the version we already cached. But I realized we'd need to change a bit more, de-coupling how often we check for new content from the cached content itself. We could have an option, something like `_promote-jobs-template_next_check`, that updates whenever we make a request, and the `promote-jobs-template` cache that never gets cleared unless we have a new version. (So maybe that too can be an option, not a cache.)

We'd make a new request if either time is up for the next check or the local template cache is empty.
For failed requests, we could make `_promote-jobs-template_next_check` to be a few minutes or hour from now, and for success it can be set longer.

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -90,135 +95,42 @@ public function promoted_jobs_custom_columns( $column ) {
...
-				</h2>
 
-				<div class="promote-job-modal-price">
-					<div class="price-text">Starting From</div>
-					<span>$83.00</span>
-				</div>
+		$response = wp_cache_get( 'promote-jobs-template', 'promote-jobs', false, $found );
+
+		if ( ! $found ) {
+			$response = wp_remote_get( 'https://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . get_locale() );
```
renatho: I think `wp_safe_remote_get` would probably be safer here.

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -54,6 +56,9 @@ public function __construct() {
 	 * @return array
 	 */
 	public function promoted_jobs_columns( $columns ) {
+		if ( ! $this->promote_jobs_template_exists() ) {
```
renatho: The idea is nice, but I think we could keep this even in case of error because it could be something specific to the template.

If we implement something like this, I think it could be in some status endpoint in wpjobmanager.com, and turn off the whole Promoted Jobs feature.

WDYT?

includes/admin/class-wp-job-manager-promoted-jobs.php
```diff
@@ -227,6 +139,10 @@ public function get_promote_jobs_template() {
 	 * @return void
 	 */
 	public function promoted_jobs_admin_footer() {
+		$screen = get_current_screen();
```
renatho: Good catch!

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Done in https://github.com/Automattic/WP-Job-Manager/pull/2478/commits/36879401a1c933f12584d4ce740414802e0c0a1d

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Removed the check for now in https://github.com/Automattic/WP-Job-Manager/pull/2478/commits/52593c0b3bf8d2bf0eb843fc1cd5e7ae188be583

Are you thinking something like... if response is not 200, we check a status endpoint on wpjobmanager.com?  And if the status check fails, we turn off promoted jobs entirely?   Would that run every time we try to get the template?

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Great idea @yscik !

I think I covered most of this in f8e9194868bdfd1f9d1dbad22232c11f2a929873 but let me know if I missed something.

includes/admin/class-wp-job-manager-promoted-jobs.php
yscik: Related to this, we should also have a filter to turn off the whole promoted jobs feature, in case a dev wants to turn it off for a client site.

onubrooks: We did discuss having a single point of entry so I'll also suggest moving this file to the `promoted-jobs` folder and including it from [here](https://github.com/Automattic/WP-Job-Manager/blob/594b31da03443d96a1f6527e4d1f55934aa6b90a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php#L62-L65) maybe using `is_admin()` or alternatively using the `admin_init` hook.

includes/admin/class-wp-job-manager-promoted-jobs.php
yscik: Looks pretty good! But in case the update fails, I'd still fall back to the version saved in the `promote-jobs-template` option and return it in the `get_promote_jobs_template()` function — a slightly outdated version would still be good enough to show for this.

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Once this is merged I'll do another PR to move the whole file over to the `promoted-jobs` folder and instantiate it properly.

includes/admin/class-wp-job-manager-promoted-jobs.php
mikeyarce: Done in https://github.com/Automattic/WP-Job-Manager/pull/2478/commits/88e5c77b88005873e24d3e515a4ded2c2dbd0cec

yscik (approved):
Looks good 🔥

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -179,134 +177,38 @@ public function promoted_jobs_custom_columns( $column ) {
...
-				content: "\2715";
-				font-size: 20px;
-			}
-		</style>
-		<div class="promote-job-modal">
-			<div class="promote-job-modal-column-left">
-				<h2 class="promote-jobs-heading">
-					Promote your job on our partner network.
-				</h2>
+		$promote_template                 = get_option( 'promote-jobs-template', false );
```
fjorgemota: Maybe we could make this dependent on the locale?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -179,134 +177,38 @@ public function promoted_jobs_custom_columns( $column ) {
...
+			$check_for_updated_template = true;
+		}
 
-				<ul class="promote-list">
-					<li class="promote-list-item">Your ad will get shared on our Partner Network</li>
-					<li class="promote-list-item">Featured on jobs.blog for 7 days</li>
-					<li class="promote-list-item">Featured on our weekly email blast</li>
-				</ul>
+		if ( $check_for_updated_template ) {
+			$response = wp_safe_remote_get( 'https://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . get_locale() );
```
fjorgemota: Maybe we could use `get_user_locale()` instead of `get_locale()`? So we get the locale that the user is effectively seeing?

Related slack thread: p1688599499810659-slack-C053397HN30

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -179,134 +177,38 @@ public function promoted_jobs_custom_columns( $column ) {
...
+			if (
+				is_wp_error( $response )
+				|| 200 !== wp_remote_retrieve_response_code( $response )
+				|| empty( wp_remote_retrieve_body( $response ) )
+			) {
+				update_option( '_promote-jobs-template_next_check', time() + MINUTE_IN_SECONDS * 5, false );
+				return $promote_template;
```
fjorgemota, line 280: What if the very first request to WPJMCOM for some reason fails? Shouldn't we maybe have a fallback here in case $promote_template isn't defined yet?

fjorgemota (commented):
Hey there, based on a discussion on Slack, I left some comments, I hope you don't mind. ;) 

Thanks!

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
mikeyarce: Done in 1d7504f0486c74fc3dff10dda1a1c0dcbb5a9e91

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
mikeyarce: done in 1d7504f0486c74fc3dff10dda1a1c0dcbb5a9e91

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
mikeyarce: We talked about this in another thread, but I think if WPJMCOM fails, then the whole promoted jobs flow would fail too, so maybe it's ok?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: Hi @mikeyarce, I think the whole logic should depend on the `get_user_locale`, so I think we should also add it to the other option. To simplify, I'd suggest creating variables with the name of the options to re-use, so you don't need to duplicate it. 😉

mikeyarce: I think I know what you mean, done in https://github.com/Automattic/WP-Job-Manager/pull/2478/commits/961a0b9ee3d6a433a6b3ae572b838ac84d7fff0c

renatho: About the variable, it looks good!
I think we should have a similar logic in the other option too, since it's also tied to the template of the specific locale.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: yeah, fair enough. Good reasoning. :)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Yeah, the `_promote-jobs-template_next_check` probably should be dependent on the locale too. :thinking:

mikeyarce: Done in e0eb0156d774629d3985de14b917356f15c778ce

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -179,134 +177,40 @@ public function promoted_jobs_custom_columns( $column ) {
...
-					<li class="promote-list-item">Featured on our weekly email blast</li>
-				</ul>
+		if ( $check_for_updated_template ) {
+			$response = wp_safe_remote_get( 'https://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . get_user_locale() );
+			if (
+				is_wp_error( $response )
+				|| 200 !== wp_remote_retrieve_response_code( $response )
+				|| empty( wp_remote_retrieve_body( $response ) )
+			) {
+				update_option( '_promote-jobs-template_next_check', time() + MINUTE_IN_SECONDS * 5, false );
```
yscik: This should use the new name too, right?

```suggestion
				update_option( $promote_jobs_template_next_check_name, time() + MINUTE_IN_SECONDS * 5, false );
```

mikeyarce: Good catch!  Fixed in c24aa522b46a1bab8fd76ab242f97d1c8cd1797f

PR #2481, opened 2023-06-30
Add redirect to promote job form  on WPJMCOM

### LINKED ISSUE #2439: Integrate promote button on Job Listing
- Display a promote button when the job is not promoted.
- On click, this should display a [modal](https://github.com/Automattic/WP-Job-Manager/issues/2438) with more details and then redirect to wpjobmanager.com authenticated as the site.
    - This should also send the job reference somehow to be used to create the job in JobTarget. I imagine a link to the WP core endpoint with the job details.
    - If [authentication](https://github.com/Automattic/wpjobmanager.com/issues/301) is not implemented yet, extract this part to a separate issue.
- Restrict this feature with a capability.

### Success Criteria

- [ ] ~Only users with a given capability related to promoted jobs can see the button in the Job Listings.~ (extracted to #2494)
- [x] The [authentication](https://github.com/Automattic/WP-Job-Manager/issues/2438) code also only works for users with the capability.
- [x] The click on the promote button navigates to wpjobmanager.com, sending all [parameters needed for authentication](https://github.com/Automattic/WP-Job-Manager/issues/2438).
- [x] It also sends the site lang, so we are able to translate the form and JobTarget page in the future.

fjorgemota:
Part of #2439 

- Make `get_api_base_url` public, so we can use it elsewhere in the code to get a dynamic link do WPJMCOM
- Make Promote Job button work, redirecting to WPJMCOM (whether it is a dev version or not) with the appropriate parameters
- Add endpoint to generate token dynamically and redirect the user appropriately
- Refactor the action to deactivate a promotion to use the `admin_action_` hook.

### Testing instructions

1. Register a job on WPJM
2. On the admin page listing all the jobs (`/wp-admin/edit.php?post_type=job_listing`), click on the "Promote" button
3. On the modal that appears, click on the "Promote your jobs" button
4. Check that the plugin validates that you're using HTTPS (if you're not, feel free to comment [this line for testing](https://github.com/Automattic/WP-Job-Manager/pull/2481/files#diff-48ee1884d5a044341d0f6de2f55553f95039416b06b775a572afb6ea907a10a3R90))
5. Make sure you're directed to a page like `https://wpjobmanager.com/promote-job?user_id=YOUR_USER_ID&job_id=YOUR_JOB_ID&token=A_RANDOM_TOKEN&site_url=THE_HTTPS_VERSION_OF_YOUR_SITE`

Also, try setting your WPJM version to include the `-dev` suffix, and add the constant `JOB_MANAGER_DEV_API_BASE_URL` in your `wp-config.php` setting the URL for your local WPJMCOM instance. Make sure the above link is changed appropriately. 

Finally, check out if the option to deactivate a promotion still works, by setting a job listing with `_promoted` set to `1` and verifying if the deactivate flow still works.


includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -45,10 +59,11 @@ public static function instance() {
 	public function __construct() {
 		add_filter( 'manage_edit-job_listing_columns', [ $this, 'promoted_jobs_columns' ] );
 		add_action( 'manage_job_listing_posts_custom_column', [ $this, 'promoted_jobs_custom_columns' ], 2 );
+		add_action( 'admin_action_' . self::PROMOTE_JOB_ACTION, [ $this, 'handle_promote_job' ] );
+		add_action( 'admin_action_' . self::DEACTIVATE_PROMOTION_ACTION, [ $this, 'handle_deactivate_promotion' ] );
 		add_action( 'admin_footer', [ $this, 'promoted_jobs_admin_footer' ] );
-		add_action( 'load-edit.php', [ $this, 'handle_deactivate_promotion' ] );
 		add_action( 'wpjm_job_listing_bulk_actions', [ $this, 'add_action_notice' ] );
-
+		add_filter( 'allowed_redirect_hosts', [ $this, 'add_to_allowed_redirect_hosts' ] );
```
fjorgemota, line 66: I'm always adding `wpjobmanager.com` to the list of `allowed_redirect_hosts` here. I...don't think this is a problem, but this logic MIGHT be better suited in another place. WDYT?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -68,19 +83,15 @@ public function promoted_jobs_columns( $columns ) {
...
-		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized,WordPress.Security.ValidatedSanitizedInput.MissingUnslash -- Nonce should not be modified.
-		if ( ! $post_id || empty( $_GET['nonce'] ) || ! wp_verify_nonce( $_GET['nonce'], 'deactivate_promotion_' . $_GET['post_id'] ) ) {
-			return;
+		if ( ! $post_id ) {
+			wp_die( esc_html__( 'No job listing ID provided for deactivation of the promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
 		}
 
 		if ( ! current_user_can( 'manage_job_listings', $post_id ) || 'job_listing' !== get_post_type( $post_id ) ) {
-			return;
+			wp_die( esc_html__( 'You do not have permission to deactivate the promotion for this job listing.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
```
fjorgemota, line 94: I added some error handling so the system never returns a blank page

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -114,6 +128,54 @@ public function add_action_notice( $actions_handled ) {
...
+			[
+				'user_id'  => $current_user,
+				'job_id'   => $post_id,
+				'token'    => $token,
+				'site_url' => site_url( '', 'https' ),
+				'locale'   => get_user_locale(),
+			],
+			WP_Job_Manager_Helper_API::get_api_base_url() . self::PROMOTE_JOB_FORM_PATH
+		);
+		wp_safe_redirect( $url );
```
fjorgemota, line 176: Should we maybe set something in the job listing to mark that this is in the process of being promoted, maybe? Not sure

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -114,6 +128,54 @@ public function add_action_notice( $actions_handled ) {
...
+	public function handle_promote_job() {
+		$post_id = absint( $_GET['post_id'] ?? 0 );
+		check_admin_referer( self::PROMOTE_JOB_ACTION . '-' . $post_id );
+		if ( ! $post_id ) {
+			wp_die( esc_html__( 'No job listing ID provided for promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
+		}
+		if ( ! is_ssl() ) {
+			wp_die( esc_html__( 'You must be using SSL to promote a job listing.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
+		}
+		if ( ! current_user_can( 'manage_job_listings', $post_id ) || 'job_listing' !== get_post_type( $post_id ) ) {
```
fjorgemota, line 153: I thought about using `job_manager_user_can_edit_job`, but given that the deactivate action was using this logic, I decided to use the same here.

WDYT? Is there any options you prefer?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -114,6 +128,54 @@ public function add_action_notice( $actions_handled ) {
...
+	 *
+	 * @return void
+	 */
+	public function handle_promote_job() {
+		$post_id = absint( $_GET['post_id'] ?? 0 );
+		check_admin_referer( self::PROMOTE_JOB_ACTION . '-' . $post_id );
+		if ( ! $post_id ) {
+			wp_die( esc_html__( 'No job listing ID provided for promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
+		}
+		if ( ! is_ssl() ) {
```
renatho: I commented it as you suggested to test. But could we add a filter here to help with development?

~~And also if a user doesn't have SSL, and consent with the risk.~~ (thinking more, it wouldn't work for this case because we wouldn't be able to fetch from wpjobmanager.com, right?)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -114,6 +128,54 @@ public function add_action_notice( $actions_handled ) {
...
+		}
+		$url = add_query_arg(
+			[
+				'user_id'  => $current_user,
+				'job_id'   => $post_id,
+				'token'    => $token,
+				'site_url' => site_url( '', 'https' ),
+				'locale'   => get_user_locale( $current_user ),
+			],
+			WP_Job_Manager_Helper_API::get_api_base_url() . self::PROMOTE_JOB_FORM_PATH
```
renatho: WDYT to add the full URL to a constant, instead of concating it with the `get_api_base_url()`. What I'm finding a little weird is that it's not an API, so maybe it doesn't make sense to use this method here.

For our dev env, if we don't want to use a constant or a filter, similar to the `get_api_base_url`, we can use something like this:

```
	add_filter( 'wp_redirect', function( $location ) {
		if ( str_starts_with( $location, 'https://wpjobmanager.com' ) ) {
			return str_replace(
				'https://wpjobmanager.com',
				'http://wpjobmanagercom.vipdev.lndo.site',
				$location
			);
		}

		return $location;
	} );
```

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: IMO, we could ~~suggest~~ wait the answer from JobTarget on how we'll receive this update. If it's immediate after it's published there, I think we wouldn't need to worry about this.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -45,10 +59,11 @@ public static function instance() {
 	public function __construct() {
 		add_filter( 'manage_edit-job_listing_columns', [ $this, 'promoted_jobs_columns' ] );
 		add_action( 'manage_job_listing_posts_custom_column', [ $this, 'promoted_jobs_custom_columns' ], 2 );
+		add_action( 'admin_action_' . self::PROMOTE_JOB_ACTION, [ $this, 'handle_promote_job' ] );
+		add_action( 'admin_action_' . self::DEACTIVATE_PROMOTION_ACTION, [ $this, 'handle_deactivate_promotion' ] );
```
renatho, line 63: Nice refactor here!

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: > I...don't think this is a problem

I agree that it's probably fine.

> this logic MIGHT be better suited in another place

I also agree with this. Since it's something very generic, maybe we could add to the `WP_Job_Manager` class? Checking the codebase, I couldn't find another place where it could fit.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: Just as a reminder, if we do it, it would also demand a change in the `add_to_allowed_redirect_hosts` method to get from the constant.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: I had thought of a specific capability, but I think it's fine for now, and we could iterate in the future if needed.

renatho: @jom, I noticed you got the issue https://github.com/Automattic/WP-Job-Manager/issues/2494. Since @fjorgemota is already implementing it here, we could just hide the buttons as part of that issue?

renatho (commented):
This is looking great! Just added some suggestions. ;)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: The requirement to use SSL is mostly to guarantee that there's no "man-in-the-middle" between WPJMCOM and WPJM.

I think I can definitely use `wp_get_environment_type()` to check if it is a production site to enforce this, but, honestly, I'm not sure. Adding a filter is also an option.  But, again, I'm a bit worried about the possible security implications of allowing the user to do this. :thinking:

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: The reason I didn't do this (used a constant) and instead preferred to make it dynamic based on `get_api_base_url()` is just to have...that "single point of change" for development.

I know that WP does provide us with the flexibility to change this easily, but then I already see three filters we'll need to change just to support a local WPJM environment:

1. wp_redirect (for this specific use case)
2. allowed_redirect_hosts (to allow the redirect above)
3. probably the HTTP request methods, to allow communicating with WPJM

In Sensei Pro, I know we have a single place/filter to change most this, so my idea was to bring some of that to here.

Maybe we could instead have a filter inside `get_api_base_url()` to make it easier to change instead of using the logic based on `-dev` versions?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: That's a good call! Should I ping them or do you want to send the question? :thinking:

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Yeah, I think the only missing piece is hidding the buttons, probably using a similar logic. :D

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: That might be a good idea! That, or maybe on `WP_Job_Manager_Com_API`, too.

I'll await on your input related [to this ](https://github.com/Automattic/WP-Job-Manager/pull/2481#discussion_r1256769424) to think the better way to handle this, too.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: Ah, good idea about the `wp_get_environment_type()`!

> worried about the possible security implications of allowing the user to do this

In the worst case, I think it just would not work, right? Or do you think of any other implication?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: My main point was that we're getting the base API URL as the site URL. Currently, it's the same and works, but it doesn't seem semantic to me.

I just realized that we also have the private method `WP_Job_Manager_Com_API::get_api_base_url` (not related to this PR), which also seems a little weird. But just mentioning it because the path is to `wp-json` there, for example.

Anyway, if you're not comfortable to change that, I'm okay to keep it as it is.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: It's something they're working on. In the next few days, they'll send us a plan for that.

BTW, my comment had a wrong word that I just fixed. Sorry if that made the message confusing to understand.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: I'm mainly worried about the user seeing the filter/option to disable the check and just...going in and disabling the check, you know?

I think I can solve that by adding a check directly on WPJMCOM, instead of adding it here. Maybe it would be better than adding here? (at least, we would have a bigger chance to document the whole issue)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Ah, good catch! I see what you mean now. Thanks for the explanation!

I think I will unify the `get_api_base_url` in a way where it returns the `/wp-json` path for WPJobManager.com, while creating another `get_site_base_url()` where it returns only the base for WPJobManager.com.

I noticed that indeed we have two methods with the same name, and one of them has a filter and another doesn't. So...this will help improve the consistency too.

I'll share the commit changing this later. Thanks!

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Hey there! I decided to remove this check in 247e46c and moved the logic to WPJMCOM here: [Automattic/wpjobmanager.com@`f9e7a0c` (#479)](https://github.com/Automattic/wpjobmanager.com/pull/479/commits/f9e7a0c3dc923c4107375b70eadfab0098fc8d51)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -114,6 +128,55 @@ public function add_action_notice( $actions_handled ) {
...
+		$job_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/' . $post_id, 'https' );
+		$job_endpoint_url = substr( $job_endpoint_url, strlen( $site_url ) );
+
+		$url = add_query_arg(
+			[
+				'user_id'  => $current_user,
+				'job_endpoint_url' => $job_endpoint_url,
+				'token'    => $token,
+				'site_url' => $site_url,
+				'locale'   => get_user_locale( $current_user ),
```
jom: I think there are some PHPCS issues here and on line 162.

fjorgemota: yep, it is on my TODO list to fix. Gonna fix it now. Thanks for the ping!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -147,4 +172,31 @@ public function update_job_status( $request ) {
...
+	public function get_job_data( $request ) {
+		$token    = $request->get_param( 'token' );
+		$user_id  = $request->get_param( 'user_id' );
+		$job_id   = $request->get_param( 'job_id' );
+		$verified = WP_Job_Manager_Site_Trust_Token::instance()->validate( 'user', $user_id, $token );
+		$result   = [
+			'verified' => $verified,
+		];
+		if ( $verified ) {
+			// We only want to return the job data if the specified user has permission to manage the job.
```
renatho, line 191: Sorry for identifying it later, but I think it would make more sense as part of the `permission_callback`, no? Anyway, I think it would be just a different way to handle it, so feel free to skip it for now if you don't consider it important.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -114,6 +128,55 @@ public function add_action_notice( $actions_handled ) {
...
 	}
 
+	/**
+	 * Add the host of the Promote Job form to the array of allowed redirect hosts.
+	 *
+	 * @param array $hosts Allowed redirect hosts.
+	 * @return array Updated array of allowed redirect hosts.
+	 */
+	public function add_to_allowed_redirect_hosts( $hosts ) {
+		$hosts[] = wp_parse_url( WP_Job_Manager_Helper_API::get_wpjmcom_url(), PHP_URL_HOST );
```
renatho, line 138: I imagine you're using `wp_parse_url ` for the case that the user is using the filter `wpjm_wpjmcom_api_url`, right? I wonder if it would make more sense in the `get_wpjmcom_url` method.

renatho (approved):
I sent 2 last questions, but not critical, so I'm already approving it in case you don't agree or decide to work on them later. 😉

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Fixed in 73f1f2c and send PR #2507  to fix the `lint-staged` behavior so it catches this sort of mistake. Thanks!

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Sorry, I forgot to document here, but the commits 4d16369, 28e2d23e2c958d8db1311db0328ff043b661df95 and 366cb88ff4ab04114e5d4353bf7be649d3d039c0 make this change.

Hopefully it is better and it solves what you mentioned. ;)

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
fjorgemota: I think this enters the same discussion [that we had on my other PR](https://github.com/Automattic/WP-Job-Manager/pull/2489#discussion_r1256648213): If we return 403 (which I think will happen if `permission_callback` returns `false`), it will be harder (maybe impossible?) to identify whether the request failed because the token verification failed, or because something else failed. WDYT?

That other PR was merged back into this one, so this code is appearing again here. Sorry for that confusion too.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Actually, the idea is just to return the host involved, because that's what I want to get here. `get_wpjmcom_url` returns a full URL, with protocol and even a path if needed (one can technically run a dev version of WPJMCOM on `http://localhost/wpjobmanager.com`, maybe?), meanwhile here I just want to get the host, and nothing else, hence why I'm using `wp_parse_url`.

Does that make sense?

renatho: Oh! Sorry! I misunderstood it while reading previously! That makes sense!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
renatho: Ah, I'm sorry we had discussed this one before!
I remember that one, but I thought it was another part!

PR #2484, opened 2023-06-30
Endpoint to update promoted job status

### LINKED ISSUE #2442: Update job status after job is promoted or expired in JobTarget

onubrooks:
Fixes #2442

* Exposes an endpoint to update the status of a promoted job when it changes from Job Target.
* Documents schema in `/docs/api/internal.json` 

### Testing instructions

* Make a `PUT` request to: `/wp-json/wpjm-internal/v1/promoted-jobs/{id}` with the body `status: 0 or 1`.
* Ensure the post is a promoted job.
* See that it's status has changed
* Note: this endpoint is not authenticated, there's another [issue](https://github.com/Automattic/wpjobmanager.com/issues/433) for it.

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -50,6 +50,30 @@ class WP_Job_Manager_Promoted_Jobs_API extends WP_REST_Controller {
 				],
 			]
 		);
+		register_rest_route(
+			$this->namespace,
+			'/' .
+			$this->rest_base . '/(?P<id>[\d]+)',
+			[
+				[
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => [ $this, 'update_job_status' ],
+					'permission_callback' => '__return_true',
+					'args'                => [
+						'id'     => [
+							'validate_callback' => function( $param ) {
+								return is_numeric( $param );
+							},
+						],
+						'status' => [
+							'validate_callback' => function( $param ) {
+								return is_numeric( $param );
+							},
+						],
+					],
+				],
+			]
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


mikeyarce:
Works good, @onubrooks ! 

My only question is - do we care if it's a `0` or `1`?  Right now there's no validation to make sure it's boolean but just that it's numeric, so technically it could be `_promoted => 29847592876542987549082745` and it would still work.

onubrooks:
> 

Good call @mikeyarce, update in https://github.com/Automattic/WP-Job-Manager/pull/2484/commits/963939573dc5331828dd4beaf9b4585925ab2af9

mikeyarce (approved):
Looks good!  🚢

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -120,6 +144,31 @@ public function prepare_item_for_response( $item, $request ) {
...
+	public function update_job_status( $request ) {
+		$post_id = $request->get_param( 'id' );
+		$status  = $request->get_param( 'status' );
+		$post    = get_post( $post_id );
+
+		if ( empty( $post ) ) {
+			return new WP_Error( 'not_found', __( 'The promoted job was not found', 'wp-job-manager' ), [ 'status' => 404 ] );
+		}
+
+		$result = update_post_meta( $post_id, WP_Job_Manager_Promoted_Jobs::META_KEY, $status );
```
renatho, line 163: > My only question is - do we care if it's a 0 or 1? Right now there's no validation to make sure it's boolean but just that it's numeric, so technically it could be _promoted => 29847592876542987549082745 and it would still work.

Good point here! Actually, the feed endpoint is getting the posts with `_promoted` as `"1"`. So maybe it should  be something like this?

```php
$result = update_post_meta( $post_id, WP_Job_Manager_Promoted_Jobs::META_KEY, $status ? '1' : '0' );
```

renatho (commented):
Sorry for the late review here, but I just noticed something about the type, and that the documentation has some structural notices.

You could use https://editor.swagger.io/ to test the documentation file. 😉

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
fjorgemota: Just FYI, I fixed this on #2488. :)

PR #2488, opened 2023-07-04
Fix some issues on the Promoted Jobs API and refactor it to be more consistent with WPJMCOM

fjorgemota:
This PR fixes a few issues that I found on #2469 and improves the validation of arguments in #2484

* Do not extend on `WP_REST_Controller` anymore, as we do on other classes in WPJMCOM
* Use constants instead of properties with set values - As these values won't change anyway
* Use `type` instead of `validate_callback` to validate if an argument has the correct type
* Use the `=` as meta query's compare operator, as `==` isn't recognized officially by WP
* Use `get_posts` instead of `WP_Query` to query for jobs
* **FIX**: Remove pagination for the API endpoint to query all the promoted jobs, otherwise we would get only 10 jobs and only that;
* **FIX**: Return the same format of response if there's no promoted jobs listed
* **FIX**: Persist meta as '1' or '0' instead of using a boolean there

### Testing instructions

Check if the behavior described in #2484 and on #2469 is still working.

Also check:

-  if all the jobs are returned when the promoted when there are more than 10 promoted jobs registered
- If the response is `{"jobs": [] }` when there are no promoted jobs registered

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -51,14 +51,12 @@ class WP_Job_Manager_Promoted_Jobs_API {
 					'permission_callback' => '__return_true',
 					'args'                => [
 						'id'     => [
-							'validate_callback' => function( $param ) {
-								return is_numeric( $param );
-							},
+							'type' => 'integer',
+							'required' => true,
 						],
 						'status' => [
-							'validate_callback' => function( $param ) {
-								return is_bool( $param );
-							},
+							'type' => 'boolean',
+							'required' => true,
 						],
 					],
 				],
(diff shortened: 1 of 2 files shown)


includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -79,55 +67,46 @@ public function register_routes() {
...
+	 * @return WP_Error|WP_REST_Response The response, or WP_Error on failure.
 	 */
-	public function get_items( $request ) {
+	public function get_items() {
 		$args = [
 			'post_type'           => 'job_listing',
 			'post_status'         => 'publish',
 			'no_found_rows'       => true,
 			'ignore_sticky_posts' => true,
+			'posts_per_page'      => -1,
```
renatho, line 78: Good catch!

renatho (approved):
Looks good and works well!

PR #2489, opened 2023-07-05
Add endpoint to get job data with Site Trust verification

### LINKED ISSUE #2439: Integrate promote button on Job Listing

fjorgemota:
Part of #2439 
Based on #2481 

- Implement a REST endpoint which validates the token+user_id+job_id combination and only returns the job data if everything is OK and verified

### Testing instructions

1. Register a job on WPJM
2. On the admin page listing all the jobs (`/wp-admin/edit.php?post_type=job_listing`), click on the "Promote" button
3. On the modal that appears, click on the "Promote your jobs" button
4. Check that the plugin validates that you're using HTTPS (if you're not, feel free to comment the condition with `is_ssl()` for checking)
5. Make sure you're directed to a page like `https://wpjobmanager.com/promote-job?user_id=YOUR_USER_ID&job_id=YOUR_JOB_ID&token=A_RANDOM_TOKEN&site_url=THE_HTTPS_VERSION_OF_YOUR_SITE`
6. Get the Job ID,  user ID, token and site URL specified and make sure you can access `http://YOUR_SITE/wp-json/wpjm-internal/v1/promoted-jobs/YOUR_JOB_ID?user_id=YOUR_USER_ID&token=A_RANDOM_TOKEN`, and that it returns `verified` set to `true` with all the job data of the specified Job ID in the `job_data` key
7. Also make sure that, if you make the same request again, it will return `verified` set to `false`, and no job data

Also, make sure that the API Doc for that endpoint is correct


includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -62,6 +62,31 @@ public function register_routes() {
...
+					'args'                => [
+						'job_id'  => [
+							'required' => true,
+							'type'     => 'integer',
+						],
+						'user_id' => [
+							'required' => true,
+							'type'     => 'integer',
+						],
+						'token'   => [
```
renatho, line 82: I'm thinking if it would make more sense to include the token in the header instead of in the body. But honestly, I don't see any benefit in changing that. I was just thinking that it could make more sense, so feel free to ignore it if you disagree.

docs/api/internal.json
```diff
@@ -130,6 +62,90 @@
...
+                          ]
+                        },
+                        "job_data": {
+                          "$ref": "#/components/schemas/JobData"
+                        }
+                      }
+                    },
+                    {
+                      "type": "object",
+                      "description": "The token/user ID is expired or invalid",
```
renatho, line 122: Shouldn't it be returned as `403` instead of `200`?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -147,4 +172,28 @@ public function update_job_status( $request ) {
...
+		$user_id  = $request->get_param( 'user_id' );
+		$job_id   = $request->get_param( 'job_id' );
+		$verified = WP_Job_Manager_Site_Trust_Token::instance()->validate( 'user', $user_id, $token );
+		$result   = [
+			'verified' => $verified,
+		];
+		if ( $verified ) {
+			// We only want to return the job data if the specified user has permission to manage the job.
+			if ( ! user_can( $user_id, 'manage_job_listings', $job_id ) || 'job_listing' !== get_post_type( $job_id ) ) {
+				return new WP_Error( __( 'Job not found or does not have enough permissions.', 'wp-job-manager' ) );
```
renatho: WDYT to add the status `401` here? Or separate and make a `404` too if the job is not found.

renatho (commented):
Thank you for breaking all this feature into smaller PRs, making it simpler to review!
I added a few suggestions in the code.

docs/api/internal.json
fjorgemota: My idea of keeping it as `200` is to "reply" to the client that the answer is "valid", but not verified.

For instance, depending on the settings, the server could already reply with 403 (if the user has some plugin to enforce API authentication for some reason, for instance). By replying with `200`, we can kinda pass the message that the request itself is OK (so it is not a server configuration error), BUT the token isn't verified.

Does that make sense?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
fjorgemota: I think the main issue with that is that, by making it a header, we might cause issues with reverse proxy (nginx/varnish/cloudflare, maybe?), and WP wouldn't enforce it as well as it does right now.

If this was on WPJMCOM side, I think it would be ok, though. :thinking:

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
fjorgemota: Good call! Fixed in 4f7eca7 and 4529c2b. Thanks!

docs/api/internal.json
renatho: Yep! Sounds good! ;)

renatho (approved):
🚀🚀🚀

fjorgemota:
Thanks for the review, Renatho!

@aaronfc Hey there! I'm going to merge this, but if you have any comments or thoughts, please, feel free to share and I'll fix them in another PR, ok?

Thanks!

PR #2506, opened 2023-07-11
Check if user can promote job before presenting button

### LINKED ISSUE #2494: Only display and allow the promote/deactivate based on a capability check
Issue extracted from https://github.com/Automattic/WP-Job-Manager/issues/2440 and https://github.com/Automattic/WP-Job-Manager/issues/2439

It's being partially implemented in https://github.com/Automattic/WP-Job-Manager/issues/2439, restricting the feature, but not hiding the button.

### Success Criteria

- [ ] Users without the Promote Jobs capability can't promote or deactivate a promotion.
- [ ] The promote column is not displayed in the job listing for users without the capability.

### LINKED ISSUE #2497: Don't offer a button to promote a not published job
Don't offer the promote button in the job listing if a job is not published.

### Success Criteria

- [x] If a job is not published, the Promote button is not displayed in the job listing.

jom:
Fixes #2494 
Fixes #2497 

Based off of #2481 

* Require correct permissions to promote or deactivate job
* Require job be published.

### Testing instructions

Use this snippet to play with the permissions. As is, it will just give the capability to jobs with `job` in the title.

```php
add_filter(
  'user_has_cap',
  function( $allcaps, $caps, $args ) {
	if ( ! in_array( 'manage_job_listings', $caps, true ) ) {
	  return $allcaps;
	}
	
	$job_id = $args[2] ?? null;
	if ( ! $job_id ) {
	  return $allcaps;
	}
	
	$job = get_post( $job_id );
	if ( 'job_listing' !== $job->post_type ) {
	  return $allcaps;
	}
	
	if ( str_contains( strtolower( $job->post_title ), 'job' ) ) {
	  $allcaps['manage_job_listings'] = true;

	  return $allcaps;
	}
	
	unset( $allcaps['manage_job_listings'] );
	
	return $allcaps;
  },
  10,
  3
);
```

Screenshot 2023-07-11 at 6 05 22 pm

### DIFF
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -90,7 +90,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 			wp_die( esc_html__( 'No job listing ID provided for deactivation of the promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
 		}
 
-		if ( ! current_user_can( 'manage_job_listings', $post_id ) || 'job_listing' !== get_post_type( $post_id ) ) {
+		if ( ! $this->can_promote_job( $post_id ) ) {
 			wp_die( esc_html__( 'You do not have permission to deactivate the promotion for this job listing.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
 		}
 
@@ -110,6 +110,21 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 			)
 		);
 		exit;
+	}
+
+	/**
+	 * Check if a user can promote a job. They must have permission to manage job listings and the post type must be a published job_listing.
+	 *
+	 * @param int $post_id Post ID.
+	 *
+	 * @return bool Returns true if they can promote a job.
+	 */
+	private function can_promote_job( int $post_id ) {
+		if ( 'job_listing' !== get_post_type( $post_id ) || 'publish' !== get_post_status( $post_id ) ) {
+			return false;
+		}
+
+		return current_user_can( 'manage_job_listings', $post_id );
 	}
 
 	/**
@@ -212,6 +227,11 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		if ( 'promoted_jobs' !== $column ) {
 			return;
 		}
+
+		if ( ! $this->can_promote_job( $post->ID ) ) {
+			return;
+		}
+
 		$base_url    = admin_url( 'admin.php' );
 		$promote_url = add_query_arg(
 			[
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -165,7 +165,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		if ( ! $post_id ) {
 			wp_die( esc_html__( 'No job listing ID provided for promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
 		}
-		if ( ! current_user_can( 'manage_job_listings', $post_id ) || 'job_listing' !== get_post_type( $post_id ) ) {
+		if ( ! $this->can_promote_job( $post_id ) ) {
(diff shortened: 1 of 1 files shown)


includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -90,7 +90,7 @@ public function handle_deactivate_promotion() {
 			wp_die( esc_html__( 'No job listing ID provided for deactivation of the promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
 		}
 
-		if ( ! current_user_can( 'manage_job_listings', $post_id ) || 'job_listing' !== get_post_type( $post_id ) ) {
+		if ( ! $this->can_promote_job( $post_id ) ) {
```
renatho, line 93: There's just one more case to replace in the method `handle_promote_job`.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -112,6 +112,21 @@ public function handle_deactivate_promotion() {
...
 
+	/**
+	 * Check if a user can promote a job. They must have permission to manage job listings and the post type must be a published job_listing.
+	 *
+	 * @param int $post_id Post ID.
+	 *
+	 * @return bool Returns true if they can promote a job.
+	 */
+	private function can_promote_job( int $post_id ) {
+		if ( 'job_listing' !== get_post_type( $post_id ) || 'publish' !== get_post_status( $post_id ) ) {
```
fjorgemota, line 123: I'm not sure if we should do that here, but maybe we can have another restriction to make sure that the job has a non-empty description (post_content)?

I just noticed that JobTarget fails invisibly when a job has no post content/description.. :thinking:

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -212,6 +227,11 @@ public function promoted_jobs_custom_columns( $column ) {
 		if ( 'promoted_jobs' !== $column ) {
 			return;
 		}
+
+		if ( ! $this->can_promote_job( $post->ID ) ) {
+			return;
```
fjorgemota, line 232: I'm not sure but...maybe we can return here *why* the user can't promote that job? Just an idea. :thinking:

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
jom: Good catch! Added in 821c8580a5d0c71a476582dbc97d1c5fee75b449.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
jom: I think it is outside of this PR's scope, but we could check with Andrei and add later. I feel like it will get quite messy repeated on the listing, but he might have a solution.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
jom: That feels like more of a validation thing we should probably do in `handle_promote_job` and show the proper error there. We could do it here, but it might be a tiny bit confusing and not 100% (stripping tags would be a bit of a heavy operation here). I'm guessing there are other validation things we should do, too, such as empty post title.

renatho: Hm! Good point! I'm thinking that we could validate it and display an error in wpjobmanager.com. Maybe it's a very late check, but I don't think it will be common users trying to promote jobs without titles and descriptions. WDYT?

If it makes sense, I can create an issue for that.

renatho (approved):
🚀

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
jom: That makes sense to me.

renatho: Since it's failing silently, I added the issue with high priority: https://github.com/Automattic/wpjobmanager.com/issues/503

PR #2508, opened 2023-07-11
Close modal when clicking on promote

renatho:

* It closes the modal when clicking on Promote, so when the user returns to the tab the modal is not open anymore.

### Testing instructions

* Go to the job listing on WP Admin.
* Click on "Promote" on a job.
* Click on "Promote your job".
* Return to the original tab, and make sure the modal is closed.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -198,8 +198,8 @@ function populateTemplate( item, dialog, action ) {
 					</form>
 					<promote-job-template>
 						<div slot="buttons" class="promote-buttons-group">
-								<a class="promote-button button button-primary" target="_blank" href="${ this.getAttribute( 'data-href' ) }">${ job_manager_admin_params.job_listing_promote_strings.promote_job }</a>
-								<a class="promote-button button button-secondary" target="_blank" href="#">${ job_manager_admin_params.job_listing_promote_strings.learn_more }</a>
+							<a class="promote-button button button-primary" target="_blank" rel="noopener noreferrer" href="${ this.getAttribute( 'data-href' ) }">${ job_manager_admin_params.job_listing_promote_strings.promote_job }</a>
+							<a class="promote-button button button-secondary" target="_blank" rel="noopener noreferrer" href="#">${ job_manager_admin_params.job_listing_promote_strings.learn_more }</a>
 						</div>
 					<promote-job-template>`;
 				}
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -202,17 +202,15 @@ function populateTemplate( item, dialog, action ) {
 							<a class="promote-button button button-secondary" target="_blank" rel="noopener noreferrer" href="#">${ job_manager_admin_params.job_listing_promote_strings.learn_more }</a>
 						</div>
 					<promote-job-template>`;
+
+					dialog.querySelector( '#wpjm-promote-button' ).addEventListener( 'click', function() {
+						dialog.close();
+					} );
 				}
 
 				if ( 'deactivate' === action ) {
 					let deactivateButton = dialog.querySelector( '.deactivate-promotion' );
 					deactivateButton.setAttribute( 'href', this.getAttribute( 'data-href' ) );
-				}
-
-				if ( 'promote' === action ) {
-					dialog.querySelector( '#wpjm-promote-button' ).addEventListener( 'click', function() {
-						dialog.close();
-					} );
 				}
 			} );
 		} );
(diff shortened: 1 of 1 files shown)


assets/js/admin.js
```diff
@@ -198,15 +198,20 @@ function populateTemplate( item, dialog, action ) {
...
 						</div>
 					<promote-job-template>`;
 				}
+
 				if ( 'deactivate' === action ) {
 					let deactivateButton = dialog.querySelector( '.deactivate-promotion' );
 					deactivateButton.setAttribute( 'href', this.getAttribute( 'data-href' ) );
 				}
+
+				document.querySelector( '#wpjm-promote-button' ).addEventListener( 'click', function() {
```
fjorgemota: This causes an error if we click to deactivate a button, because then `#wpjm-promote-button` isn't defined. Maybe we should have this only if `'promote' === action`?

Also, wouldn't it be better to have the event listener on `dialog` directly, maybe? Or maybe just use `dialog.querySelector`? I'm thinking about maybe reducing the scope of the event listener.

assets/js/admin.js
```diff
@@ -198,15 +198,20 @@ function populateTemplate( item, dialog, action ) {
 					</form>
 					<promote-job-template>
 						<div slot="buttons" class="promote-buttons-group">
-								<a class="promote-button button button-primary" target="_blank" href="${ this.getAttribute( 'data-href' ) }">${ job_manager_admin_params.job_listing_promote_strings.promote_job }</a>
-								<a class="promote-button button button-secondary" target="_blank" href="#">${ job_manager_admin_params.job_listing_promote_strings.learn_more }</a>
+							<a id="wpjm-promote-button" class="promote-button button button-primary" target="_blank" rel="noopener noreferrer" href="${ this.getAttribute( 'data-href' ) }">${ job_manager_admin_params.job_listing_promote_strings.promote_job }</a>
+							<a class="promote-button button button-secondary" target="_blank" rel="noopener noreferrer" href="#">${ job_manager_admin_params.job_listing_promote_strings.learn_more }</a>
```
fjorgemota, line 202: Won't the Learn More button have the same requirement to close the modal? :thinking:

fjorgemota (commented):
This is looking good! I found an issue and a suggestion with the current version while testing. It should be simple to solve, though. :)

assets/js/admin.js
renatho: Good catch! Updated here: https://github.com/Automattic/WP-Job-Manager/pull/2508/commits/4337c533033ae0f9334d7e02fb69dab5be9ab999

assets/js/admin.js
renatho: Good question! I don't think so. My thought is that the user will try to promote a job, but before that, they just want to understand how it works, so they go to a different tab, and return continuing from where they stopped.

This is different for the promote button because it changes the state of the job. It wouldn't make sense to click again on "Promote". Ideally, it should also update the job status on the listing, but I think only closing the modal is fine for now.

Does it make sense to you? Or do you have any other thoughts on that?

assets/js/admin.js
```diff
@@ -198,15 +198,22 @@ function populateTemplate( item, dialog, action ) {
...
 				if ( 'deactivate' === action ) {
 					let deactivateButton = dialog.querySelector( '.deactivate-promotion' );
 					deactivateButton.setAttribute( 'href', this.getAttribute( 'data-href' ) );
 				}
+
+				if ( 'promote' === action ) {
+					dialog.querySelector( '#wpjm-promote-button' ).addEventListener( 'click', function() {
+						dialog.close();
+					} );
+				}
```
fjorgemota: Could we maybe move this code to the `if ( 'promote' === action ) {` that we already have above? But still after the `innerHTML` of course? I don't see why we would have this duplicated condition based on `'promote' === action` here.. :thinking:

fjorgemota (approved):
Looks good and works well!

I left a small suggestion on the code. Thanks for the changes!

assets/js/admin.js
renatho: Ah, good catch! I didn't notice that.

Updated in https://github.com/Automattic/WP-Job-Manager/pull/2508/commits/dfe229e979179974e8837fb481edad9cd04ff373. I'm going ahead with the merge since it was a small change, but feel free to comment anything else that I can fix in another PR.

fjorgemota: Good one! It is perfect now. Thanks!

Issue #2490, opened 2023-07-05
Review promote modal content and pricing 


renatho:
We want to add a warning in the modal saying that the job promotion will stop if the job becomes not published (even if it returns to published later). It's related to https://github.com/Automattic/WP-Job-Manager/issues/2477, but a simpler approach.

Reference of the behavior on JobTarget: p1687906664898939-slack-C056EMC2ABG

We also need to review the content of the modal, and the pricing to match features and pricing that we'll offer with JobTarget.

See discussion: p1689343773188169/1689267861.599449-slack-C053397HN30

### Success Criteria

- [x] ~~Pricing matches the minimum price found on JobTarget.~~ Remove price.
- [x] The text matches the features we'll be able to offer through JobTarget.
- [ ] We have a warning saying that the job can't be removed from publish status, otherwise, the promotion will stop.
- [x] The "Learn More" points to the correct link.
- [ ] As a plus, if we can think of a good text, we could also say that some fields are synced with the plugin (on every change), while others will be completed in the next steps. And it would also be interesting to say that it can take a few hours to sync with the external job boards.

renatho:
Closing this issue since the pending part was handled in different ways and respective issues were opened. p6r3EZ-1GJ-p2

PR #2513, opened 2023-07-12
Separate promoted job endpoint into verify token and job data retrieval.

### LINKED ISSUE #2511: Split token verification from job data retrieval for Promoted Jobs feature.
We decided to separate the token verification logic from the promoted job data retrieval.

This is done in combination with 500-gh-Automattic/wpjobmanager.com

aaronfc:
Fixes #2511 

* Added new `/promoted-jobs/verify-token` endpoint.
* Removed token verification logic and relaxed checks on `/promoted-jobs/<job_id>` to return job data.

### Testing instructions
* The easiest way is to test this together with 501-gh-Automattic/wpjobmanager.com
* As an alternative you can play around with the curls below.

#### Get job data
```
$ curl --request GET \
  --url https://wpjm.local/wp-json/wpjm-internal/v1/promoted-jobs/32
```

#### Verify token
```
curl --request POST \
  --url 'https://wpjm.local/wp-json/wpjm-internal/v1/promoted-jobs/verify-token?job_id=32&token=TOKEN&user_id=1'
```

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -117,7 +117,7 @@ class WP_Job_Manager_Promoted_Jobs_API {
 			'post_status'         => 'publish',
 			'no_found_rows'       => true,
 			'ignore_sticky_posts' => true,
-			'posts_per_page'      => - 1,
+			'posts_per_page'      => -1,
 			'meta_query'          => [
 				[
 					'key'     => WP_Job_Manager_Promoted_Jobs::META_KEY,
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -71,10 +71,23 @@ class WP_Job_Manager_Promoted_Jobs_API {
 					'callback'            => [ $this, 'get_job_data' ],
 					'permission_callback' => '__return_true',
 					'args'                => [
-						'job_id'  => [
+						'job_id' => [
 							'required' => true,
 							'type'     => 'integer',
 						],
+					],
+				],
+			]
+		);
+		register_rest_route(
+			self::NAMESPACE,
+			self::REST_BASE . '/verify-token',
+			[
+				[
+					'methods'             => WP_REST_Server::CREATABLE,
+					'callback'            => [ $this, 'verify_token' ],
+					'permission_callback' => '__return_true',
+					'args'                => [
 						'user_id' => [
 							'required' => true,
 							'type'     => 'integer',
@@ -82,6 +95,10 @@ class WP_Job_Manager_Promoted_Jobs_API {
 						'token'   => [
 							'required' => true,
 							'type'     => 'string',
+						],
+						'job_id'  => [
+							'required' => true,
+							'type'     => 'integer',
 						],
 					],
 				],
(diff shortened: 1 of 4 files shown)


includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -174,29 +194,49 @@ public function update_job_status( $request ) {
...
-			'verified' => $verified,
-		];
-		if ( $verified ) {
-			// We only want to return the job data if the specified user has permission to manage the job.
-			if ( 'job_listing' !== get_post_type( $job_id ) ) {
-				return new WP_Error( __( 'Job not found.', 'wp-job-manager' ), [ 'status' => 404 ] );
-			}
-			if ( ! user_can( $user_id, 'manage_job_listings', $job_id ) ) {
-				return new WP_Error( __( 'User does not have enough permissions to get data for this job.', 'wp-job-manager' ), [ 'status' => 401 ] );
+		$job_id = $request->get_param( 'job_id' );
```
renatho, line 204: I'm thinking a little more about this, and I'm not sure if it's fine to keep it open for any job. Maybe the site can have some private jobs somehow?

To solve this we could add an extra check similar to what we had previously, but also checking the logged-in user. WDYT?

Not sure, but maybe [this method](https://github.com/Automattic/WP-Job-Manager/pull/2506/files#diff-48ee1884d5a044341d0f6de2f55553f95039416b06b775a572afb6ea907a10a3R122) could be useful here. We'd just need to think if it makes sense to move it to some other part of the code to use in the 2 parts.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -100,7 +117,7 @@ public function get_items() {
 			'post_status'         => 'publish',
 			'no_found_rows'       => true,
 			'ignore_sticky_posts' => true,
-			'posts_per_page'      => -1,
+			'posts_per_page'      => - 1,
```
renatho: This seems a little weird. Is it something you could customize in your editor to use without space?

jom (approved):
Looks good and works well!

renatho (commented):
In general, it looks good to me. I have only 1 concern that I added to the comments (it could be in another PR though).

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
renatho: Oops! The logged-in user check wouldn't work because it's a server request. Another thing I'm thinking of is maybe to mark the job as "open" with a meta key when they click on Promote (before redirecting to wpjobmanager.com). And add a warning (the warning could be part of [this issue](https://github.com/Automattic/WP-Job-Manager/issues/2490)) saying that after proceeding, their job will be exposed through APIs, so other sites can consume it.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
aaronfc: Yes, it is VERY weird. But I don't know how to disable it. I have undone the change manually: 7f1eb648

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
aaronfc: You are raising a valid point. But I would like to challenge a bit the first assumption: is this a known problem? I mean, are there any use cases we know about that would make the jobs private?

It makes me think about what happens with the default rest API. How does that work for them?

If this is indeed a problem then I would open a separate ticket to work on this, since I can't think of any easy solution to that. (A part from the warning you are suggesting).

aaronfc: @renatho I have created [this issue](https://github.com/Automattic/WP-Job-Manager/issues/2514) to continue the discussion, since I am going to merge this into the feature branch.

PR #2515, opened 2023-07-12
Fix terms handling on the promoted jobs API

fjorgemota:

* Fixes an issue where, if there were no job listing types associated with a job, the Promoted Jobs API would return an error

### Testing instructions

Create a job without selecting any job listing types, promote it and make sure it still works (without any failures to get the token)


renatho (approved):
Thank you for the quick solution!!

PR #2516, opened 2023-07-13
Open Edit Promotion link for Promoted Jobs in a new tab

fjorgemota:

* Makes "Edit" link to edit the job listing open in a new tab
* Also removes "type='submit'" attribute from link on the Promoted Jobs modal fallback

### Testing instructions

On WPJM running this branch of the code:

1. On a job listing with the `_promoted` meta set to `1`, verify if the Edit link now opens in a new tab
2. Also in a job listing that is not promoted yet, verify if the modal works correctly.

### DIFF
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -260,7 +260,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 			echo '
 			<span class="jm-promoted__status-promoted">' . esc_html__( 'Promoted', 'wp-job-manager' ) . '</span>
 			<div class="row-actions">
-				<a href="' . esc_url( $promote_url ) . '" class="jm-promoted__edit">' . esc_html__( 'Edit', 'wp-job-manager' ) . '</a>
+				<a href="' . esc_url( $promote_url ) . '" class="jm-promoted__edit" target="_blank">' . esc_html__( 'Edit', 'wp-job-manager' ) . '</a>
 				|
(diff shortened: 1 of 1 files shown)


includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -260,7 +260,7 @@ public function promoted_jobs_custom_columns( $column ) {
 			echo '
 			<span class="jm-promoted__status-promoted">' . esc_html__( 'Promoted', 'wp-job-manager' ) . '</span>
 			<div class="row-actions">
-				<a href="' . esc_url( $promote_url ) . '" class="jm-promoted__edit">' . esc_html__( 'Edit', 'wp-job-manager' ) . '</a>
+				<a href="' . esc_url( $promote_url ) . '" class="jm-promoted__edit" target="_blank">' . esc_html__( 'Edit', 'wp-job-manager' ) . '</a>
```
renatho:
```suggestion
				<a href="' . esc_url( $promote_url ) . '" class="jm-promoted__edit" target="_blank" rel="noopener noreferrer">' . esc_html__( 'Edit', 'wp-job-manager' ) . '</a>
```

renatho (approved):
Looks good and works well!
Just added one small improvement suggestion, so feel free to merge after that. ;)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: Ah, good one. I thought about adding it, but I wasn't 100% sure. Thanks!

PR #2517, opened 2023-07-13
Encode url arguments

renatho:

* Encode URL arguments in order to avoid unexpected issues.
  * One issue I noticed is on wpjobmanager.com side. If it fails before the redirect with the token (still with `/wp-json/` in the URL) it doesn't include the WooCommerce styles.

### Testing instructions

* A way to reproduce an early error when promoting is changing the verification endpoint to something else. Search for `self::REST_BASE . '/verify-token'` and replace it with another string.
* Go to WP Admin > Job Listing.
* Try to promote a job.
* In wpjobmanager.com see that it displays the error with the WooCommerce styles.
* Revert your test change, try to promote a job, and make sure wpjobmanager.com can load the job data from the plugin.


aaronfc (approved):
Looks good!

PR #2519, opened 2023-07-14
Do not throw error on Promoted Jobs API if Job Listing Types are disabled

fjorgemota:
Fix issue reported on Slack p1689351424769739-slack-C053397HN30 

* Only call `get_the_terms` in Promoted Jobs API if Job Listing Types are enabled on WPJM Settings
* Also handle a case where, if the job data couldn't be formatted, it caused the `http://YOUR.WPJM.INSTANCE/wp-json/wpjm-internal/v1/promoted-jobs/JOB_ID` endpoint to return an error in the structure `{ "job_data": WP_ERROR }`, which wasn't handled well by WPJMCOM.

### Testing instructions

1. Create a job, get the post ID of the created job
2. Set the `_promoted` meta field to `1` (`wp post meta set JOB_ID _promoted 1`)
3. Visit `http://YOUR.WPJM.INSTANCE/wp-json/wpjm-internal/v1/promoted-jobs` and make sure all the data appears there
4. Now, go to Job Listings -> Settings -> Job Listings and uncheck the field "Enable listing types". Save the modified setting.
5. Visit `http://YOUR.WPJM.INSTANCE/wp-json/wpjm-internal/v1/promoted-jobs` again and make sure all the data appears there, without any errors or anything like that

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -148,14 +148,17 @@ class WP_Job_Manager_Promoted_Jobs_API {
 	 * @return array|\WP_Error The response, or WP_Error on failure.
 	 */
 	private function prepare_item_for_response( WP_Post $item ) {
-		$terms = get_the_terms( $item->ID, 'job_listing_type' );
+		$terms = [];
+		if ( get_option( 'job_manager_enable_types' ) ) {
+			// Only query for terms if the listing types are enabled.
+			$terms = get_the_terms( $item->ID, 'job_listing_type' );
+		}
 		if ( false === $terms ) {
 			$terms = [];
 		}
 		if ( is_wp_error( $terms ) ) {
 			return $terms;
 		}
-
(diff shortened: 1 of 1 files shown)


renatho (approved):
Looks good and works well!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -148,14 +148,17 @@ public function get_items() {
 	 * @return array|\WP_Error The response, or WP_Error on failure.
 	 */
 	private function prepare_item_for_response( WP_Post $item ) {
-		$terms = get_the_terms( $item->ID, 'job_listing_type' );
+		$terms = [];
+		if ( get_option( 'job_manager_enable_types' ) ) {
+			// Only query for terms if the listing types are enabled.
+			$terms = get_the_terms( $item->ID, 'job_listing_type' );
```
renatho, line 154: @fjorgemota, sorry for the late comment here, but after approving, I noticed something in the code. I think we probably should use the `wpjm_get_the_job_types` here because there's a filter there to tweak the types.

PR #2522, opened 2023-07-17
Add promote modal on job save

### LINKED ISSUE #2518: Add modal to the editor after publishing a new job
We should display the promote modal also on the editor, when a new job is published.

More context: p1689343825159629/1689267861.599449-slack-C053397HN30

```[tasklist]
### Success Criteria
- [x] The promote modal is displayed when a user published a new job through the editor.
- [x] The modal continues working as previously in the jobs listing.
- [ ] https://github.com/Automattic/WP-Job-Manager/issues/2523
```

renatho:
Fixes #2518

* Extract some logic related to the modals, so we can use it in the editor.
* Display promote modal when a post is changed from any status to publish.
  * It doesn't work with Classic Editor.

### Testing instructions

* Make sure the promote and deactivate modals continue working properly in the job listing.
* Create a new job, and publish.
* Make sure it displays a modal to promote the job and it works properly.

### Screenshot / Video

https://github.com/Automattic/WP-Job-Manager/assets/876340/16ba5448-1a1e-45c5-acfc-5357dbbd6f97

### DIFF
--- a/assets/js/admin/editor-lifecycle.js
+++ b/assets/js/admin/editor-lifecycle.js
@@ -1,0 +1,60 @@
+/**
+ * WordPress dependencies
+ */
+import { subscribe, select } from '@wordpress/data';
+
+/**
+ * Helper function to fire callbacks on editor lifecycles.
+ *
+ * @param {Object}   options
+ * @param {Function} options.subscribeListener Callback called everytime the subscribe listener is called.
+ * @param {Function} options.onSetDirty        Callback called when the editor becomes dirty.
+ * @param {Function} options.onSaveStart       Callback called when editor starts saving.
+ * @param {Function} options.onSave            Callback called when a save is completed.
+ *
+ * @return {Function} Unsubscribe function.
+ */
+const editorLifecycle = ( {
+	subscribeListener = () => {},
+	onSetDirty = () => {},
+	onSaveStart = () => {},
+	onSave = () => {},
+} ) => {
+	const coreEditorSelector = select( 'core/editor' );
+	let wasSaving = false;
+	let wasDirty = false;
+
+	const unsubscribe = subscribe( () => {
+		subscribeListener();
+
+		const isDirty = coreEditorSelector.isEditedPostDirty();
+
+		const isSaving =
+			coreEditorSelector.isSavingPost() &&
+			! coreEditorSelector.isAutosavingPost();
+
+		if ( ! wasDirty && isDirty ) {
+			// If editor becomes dirty.
+			wasDirty = true;
+			onSetDirty();
+		} else {
+			wasDirty = isDirty;
+		}
+
+		if ( wasSaving && ! isSaving ) {
+			// If it completed a saving.
+			wasSaving = isSaving;
+			onSave();
+		} else if ( ! wasSaving && isSaving ) {
+			// If it started saving.
+			wasSaving = isSaving;
+			onSaveStart();
+		} else {
+			wasSaving = isSaving;
+		}
+	} );
+
+	return unsubscribe;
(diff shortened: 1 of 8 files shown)


assets/js/admin/editor-lifecycle.js
```diff
@@ -0,0 +1,66 @@
...
+const editorLifecycle = ( {
+	subscribeListener = () => {},
+	onSetDirty = () => {},
+	onSaveStart = () => {},
+	onSave = () => {},
+} ) => {
+	const coreEditorSelector = select( editorStore );
+	let wasSaving = false;
+	let wasDirty = false;
+	let isNew    = false;
```
renatho: The `editorLifecycle` was borrowed from Sensei LMS, just adding the extra `isNew` variable to check if it's saving a new post or updating an existing one.

With this logic, we continue considering it as new even after an autosave.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -220,37 +220,43 @@ public function get_promote_jobs_template() {
 	 */
 	public function promoted_jobs_admin_footer() {
 		$screen = get_current_screen();
-		if ( 'edit-job_listing' !== $screen->id ) {
-			return;
+
+		if ( in_array( $screen->id, [ 'edit-job_listing', 'job_listing' ], true ) ) { // Job listing and job editor.
```
renatho, line 322: This change is to also load the promote modal on the job editor page.

assets/js/admin/editor-lifecycle.js
renatho: Approach changed.

assets/js/admin/editor-lifecycle.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {Object}   options
+ * @param {Function} options.subscribeListener Callback called everytime the subscribe listener is called.
+ * @param {Function} options.onSetDirty        Callback called when the editor becomes dirty.
+ * @param {Function} options.onSaveStart       Callback called when editor starts saving.
+ * @param {Function} options.onSave            Callback called when a save is completed.
+ *
+ * @return {Function} Unsubscribe function.
+ */
+const editorLifecycle = ( {
```
renatho, line 18: Borrowed from Sensei LMS.

assets/js/admin/promote-job-modals.js
```diff
@@ -0,0 +1,49 @@
+/**
```
renatho, line 1: This logic was extracted from the `admin.js`, so we can use it properly in the editor.

assets/js/admin/wpjm-modal.js
```diff
@@ -0,0 +1,21 @@
+/**
```
renatho, line 1: It was also extracted, just in order to make it a module instead of a global function.

includes/admin/class-wp-job-manager-admin.php
```diff
@@ -128,6 +128,21 @@ public function admin_enqueue_scripts() {
 			);
 		}
 
+		if ( 'job_listing' === $screen->id && $screen->is_block_editor() ) { // Check if it's block editor in job post.
```
renatho, line 131: It seems the standard to include the JS files is here, so I included a JS for the job editor here, including a global variable for the promote URL.

jom (approved):
This looks good and works well. 

I also tested some strange WPJM scenarios where it is in draft and I publish with a old expiration date, but it looks like that moves the expiration date forward and continues to publish the job. Deactivate still works well.

PR #2525, opened 2023-07-17
Refactor promote jobs template in order to clean the data

renatho:

* Refactor promote modal template fetch.
  * @jom realized in another review that we weren't cleaning the options, so I raised this PR to fix this. Since the options in WP Job Manager are cleared with `delete_option` it wasn't possible to clear options with dynamic names. Instead of changing it and risk an issue, I changed the approach to a transient, returning a fallback in case of failure. @mikeyarce and @yscik, please let me know if I missed any important point here.
  * Notice that the transient uses a `jm_` prefix, following the plugin standard, so it already removes these transients as part of the `WP_Job_Manager_Data_Cleaner` when the plugin is uninstalled.

### Testing instructions

* Make sure the promote modal continues working on the job listing.
* Now clear the transients and turn off your wpjobmanager.com test site.
* Open it again, and check that it returns a fallback modal only with the buttons to promote or learn more.

### DIFF
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -260,35 +260,43 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 	 * @return string
 	 */
 	public function get_promote_jobs_template() {
-		$promote_template_option_name          = 'promote-jobs-template_' . get_user_locale();
-		$promote_jobs_template_next_check_name = '_promote-jobs-template_next_check_' . get_user_locale();
-		$promote_template                      = get_option( $promote_template_option_name, false );
-		$promote_jobs_template_next_check      = get_option( $promote_jobs_template_next_check_name );
-		$check_for_updated_template            = false;
+		$locale                          = get_user_locale();
+		$promote_template_transient_name = 'jm_promote-jobs-template_' . $locale;
+		$promote_template                = get_transient( $promote_template_transient_name );
 
-		if ( ! $promote_jobs_template_next_check || $promote_jobs_template_next_check < time() ) {
-			$check_for_updated_template = true;
-		}
-
-		if ( $check_for_updated_template ) {
-			$response = wp_safe_remote_get( 'https://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . get_user_locale() );
-			if (
-				is_wp_error( $response )
-				|| 200 !== wp_remote_retrieve_response_code( $response )
-				|| empty( wp_remote_retrieve_body( $response ) )
-			) {
-				update_option( $promote_jobs_template_next_check_name, time() + MINUTE_IN_SECONDS * 5, false );
-				return $promote_template;
-			} else {
-				$assets           = json_decode( wp_remote_retrieve_body( $response ), true );
-				$promote_template = $assets['assets'][0]['content'];
-				update_option( $promote_template_option_name, $promote_template, false );
-				update_option( $promote_jobs_template_next_check_name, time() + HOUR_IN_SECONDS, false );
-			}
-		}
-		if ( ! is_wp_error( $promote_template ) ) {
+		if ( false !== $promote_template ) {
 			return $promote_template;
 		}
+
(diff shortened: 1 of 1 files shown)


includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -260,35 +260,43 @@ public function promoted_jobs_custom_columns( $column ) {
...
-				$assets           = json_decode( wp_remote_retrieve_body( $response ), true );
-				$promote_template = $assets['assets'][0]['content'];
-				update_option( $promote_template_option_name, $promote_template, false );
-				update_option( $promote_jobs_template_next_check_name, time() + HOUR_IN_SECONDS, false );
-			}
+		$url      = 'https://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . $locale;
+		$response = wp_safe_remote_get( $url );
+		$fallback = '
+			<div>
+				<br />
```
fjorgemota, line 275: Maybe we could have a simple error message here?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -260,35 +260,43 @@ public function promoted_jobs_custom_columns( $column ) {
...
-			) {
-				update_option( $promote_jobs_template_next_check_name, time() + MINUTE_IN_SECONDS * 5, false );
-				return $promote_template;
-			} else {
-				$assets           = json_decode( wp_remote_retrieve_body( $response ), true );
-				$promote_template = $assets['assets'][0]['content'];
-				update_option( $promote_template_option_name, $promote_template, false );
-				update_option( $promote_jobs_template_next_check_name, time() + HOUR_IN_SECONDS, false );
-			}
+		$url      = 'https://wpjobmanager.com/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . $locale;
```
fjorgemota: WDYT about using `WP_Job_Manager_Helper_API::get_wpjmcom_url()` to build the URL here?

```suggestion
		$url      = WP_Job_Manager_Helper_API::get_wpjmcom_url() . '/wp-json/promoted-jobs/v1/assets/promote-dialog/?lang=' . $locale;
```

fjorgemota (commented):
This is looking good! Left two small suggestions. :)

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: I started trying with a message, but it was a little ugly. So I realized that maybe it wasn't very important because we're presenting in the modal what the user really needs. The information they are missing in the modal we will probably also have on the "Learn more" page.

But if you have a strong opinion on it, I could write some CSS to customize it and make it not so ugly with a message.

yscik:
Our reasoning was to use any previously saved template as the fallback. That's why the current implementation saves it as an option instead of a transient that's cleared after a day. In that case a request failure means the user sees a slightly outdated content, instead of an error message or anything hard-coded into the plugin.

jom (approved):
This looks good and works well, pending @fjorgemota's review and discussion.

> Our reasoning was to use any previously saved template as the fallback.

@yscik That makes sense. I wonder though if most of the errors fetching will be momentary and a typical user might just refresh if it didn't work the first time. If we go back to something like before, I don't think we need the 5 minute retry block. 

I think the transient solution is fine for now (especially with the buttons), but if we want to go back to an option, we could switch to a prefixed/statically named option that we can remove and is formatted something like:
```
[
   `en_CA` => '....',
   `en_CA-fetched' => '2023-07-18 12:22:04.000`,
   'es_MX' => '...',
   'es_MX-fetched' => '2023-07-18 12:22:05.000`
]
```

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
fjorgemota: By "presenting in the modal what the user really needs" you mean having the learn more and promote button working, right?

If yes, it is fine, although a little weird. :thinking: I'll consider this a non-issue.

renatho: Correct! I'll keep it as it is for now then, and anything we could iterate later. ;)

renatho:
I'm going ahead and merging this, but we could iterate later if needed then. 😉

PR #2529, opened 2023-07-19
Add system for syncing status for Promoted Jobs from WPJMCOM to WPJM

fjorgemota:
Fixes 433-gh-Automattic/wpjobmanager.com
Sibling PR to 527-gh-Automattic/wpjobmanager.com

- Add `no-cache` headers to the REST API endpoints in Promoted Jobs namespace
- Create class to handle updating Promoted Jobs Status
  - Create endpoint to trigger update of the Promoted Jobs Status
  - Create cron that runs twice a day to trigger update of the Promoted Jobs Status
- Refactor initialization of the Promoted Jobs feature to match WPJMCOM implementation 

### Testing instructions

Check TBD of the 527-gh-Automattic/wpjobmanager.com PR

On this PR, check if the `no-cache` headers are added correctly for the endpoints in the Promoted Jobs REST API namespace. For instance, create a job, set the title and description, visit `/wp-json/wpjm-internal/v1/promoted-jobs/JOB_ID`, and check if that response returns the no-cache headers. 

You can also run the cron manually. Run `wp option delete wpjm_promoted_jobs_status_update_last_execution
` and then `wp cron event run wpjm_promoted_jobs_status_update` to run the cron job, verify if the request for WPJMCOM is being made correctly.

### DIFF
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -89,7 +89,7 @@ class WP_Job_Manager {
 		$this->post_types = WP_Job_Manager_Post_Types::instance();
 
 		// Schedule cron jobs.
-		add_action( 'init', [ $this, 'maybe_schedule_cron_jobs' ] );
+		add_action( 'init', [ __CLASS__, 'maybe_schedule_cron_jobs' ] );
 
 		// Switch theme.
 		add_action( 'after_switch_theme', [ 'WP_Job_Manager_Ajax', 'add_endpoint' ], 10 );
@@ -240,7 +240,7 @@ class WP_Job_Manager {
 	/**
 	 * Schedule cron jobs for WPJM events.
 	 */
-	public function maybe_schedule_cron_jobs() {
+	public static function maybe_schedule_cron_jobs() {
 		if ( ! wp_next_scheduled( 'job_manager_check_for_expired_jobs' ) ) {
 			wp_schedule_event( time(), 'hourly', 'job_manager_check_for_expired_jobs' );
 		}
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -89,7 +89,7 @@ class WP_Job_Manager {
 		$this->post_types = WP_Job_Manager_Post_Types::instance();
 
 		// Schedule cron jobs.
-		self::maybe_schedule_cron_jobs();
+		add_action( 'init', [ $this, 'maybe_schedule_cron_jobs' ] );
 
 		// Switch theme.
 		add_action( 'after_switch_theme', [ 'WP_Job_Manager_Ajax', 'add_endpoint' ], 10 );
@@ -240,7 +240,7 @@ class WP_Job_Manager {
 	/**
 	 * Schedule cron jobs for WPJM events.
 	 */
-	public static function maybe_schedule_cron_jobs() {
+	public function maybe_schedule_cron_jobs() {
 		if ( ! wp_next_scheduled( 'job_manager_check_for_expired_jobs' ) ) {
 			wp_schedule_event( time(), 'hourly', 'job_manager_check_for_expired_jobs' );
 		}
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -250,6 +250,9 @@ class WP_Job_Manager {
 		if ( ! wp_next_scheduled( 'job_manager_email_daily_notices' ) ) {
 			wp_schedule_event( time(), 'daily', 'job_manager_email_daily_notices' );
 		}
+		if ( ! wp_next_scheduled( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_HOOK ) ) {
+			wp_schedule_event( time(), 'twicedaily', WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_HOOK );
+		}
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


includes/class-wp-job-manager.php
```diff
@@ -240,7 +240,7 @@ public function usage_tracking_cleanup() {
 	/**
 	 * Schedule cron jobs for WPJM events.
 	 */
-	public static function maybe_schedule_cron_jobs() {
+	public function maybe_schedule_cron_jobs() {
```
renatho: Does this change fix something? Asking just because maybe we could revert to avoid backward compatibility issues.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php
```diff
@@ -56,27 +63,42 @@ public static function instance() {
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_action( 'init', [ $this, 'include_dependencies' ] );
-		add_action( 'init', [ $this, 'register_post_metas' ] );
-		add_action( 'rest_api_init', [ $this, 'rest_init' ] );
+		add_action( 'init', [ $this, 'init' ] );
```
renatho, line 66: Nice refactor!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -104,6 +132,45 @@ public function register_routes() {
...
+	 *
+	 * @param WP_REST_Response $response The response data.
+	 * @param WP_REST_Server   $server The REST server instance.
+	 * @param WP_REST_Request  $request The request used to generate the response.
+	 *
+	 * @return WP_REST_Response The response data.
+	 */
+	public function add_nocache_headers( $response, $server, $request ) {
+		// Check if the request belongs to the specified namespace and the response is successful.
+		if ( str_starts_with( $request->get_route(), '/' . self::NAMESPACE ) && $response->get_status() >= 200 && $response->get_status() < 300 ) {
```
renatho: I think we should also concat the `self::REST_BASE` in the check, right?

renatho (commented):
Also looks great! Left 2 small comments. 😉

includes/class-wp-job-manager.php
fjorgemota: The idea of this change was to allow it to be called from the `init` hook, so it runs only after all the files are included (including the one that defines `WP_Job_Manager_Promoted_Jobs_Status_Handler`). I changed the approach in 6ce063d to make this method static again and, instead, make the `init` hook call the static method. Thanks!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
fjorgemota: Good catch! Fixed in 25e30e5. Thanks!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
```diff
@@ -0,0 +1,96 @@
...
+		return $json['jobs'];
+	}
+
+	/**
+	 * Gets the URL for the site feed of promoted jobs in WPJMCOM.
+	 *
+	 * @return string The site feed URL in WPJMCOM.
+	 */
+	private function get_site_feed_url() {
+		return add_query_arg( 'site_url', home_url( '', 'https' ), WP_Job_Manager_Helper_API::get_wpjmcom_url() . '/wp-json/promoted-jobs/v1/site/jobs' );
```
renatho: I think the argument here should be `site` given the last changes in wpjobmanager.com, right?

fjorgemota: Great catch! Yes! Fixed in cfde5f580a3e696f6402f38328b136d505c56552. Thanks!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
```diff
@@ -0,0 +1,96 @@
...
+
+	/**
+	 * The name of the option that stores the last time the cron job was executed.
+	 */
+	const OPTION_KEY = self::CRON_HOOK . '_last_execution';
+
+	/**
+	 * The frequency at which the cron job should be executed.
+	 */
+	private const FREQUENCY_UPDATE = HOUR_IN_SECONDS;
```
renatho: I'm thinking it it would make more sense to use 5 minutes here in order to match the lower interval we have [here](https://github.com/Automattic/wpjobmanager.com/pull/561/files#diff-0a086bb7ed1691365420085e907d905bb7fdb67319650a46ef08bbcb44ff72faR59). WDYT?

Otherwise, if we have 2 promotions in a row, we would see the status update for the second one only after the cron job running `twicedaily`, right?

renatho (approved):
Looks good and works well! Just added one last comment to think about.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
fjorgemota: @renatho Good point. As discussed on Slack, I pushed the commits 6b315a5, ee1fa71 and 8169dab to only run `fetch_updates()` if the user had redirected to the Promote Job form at least once (by clicking on the button in the modal to promote a job).

I also created issue #2558 to improve this logic.

renatho: I'd suggest only change the name to something like "used promoted jobs" or "touched promoted jobs" instead of "has promoted jobs". Otherwise, it could be a little confusing since it doesn't represent exactly having a promoted job. WDYT?

renatho (approved):
I added a last comment suggesting just a name change. The rest looks good to me! So feel free to merge after checking that.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
fjorgemota: Sure! Renamed in ec759b8, and adjusted frequency update (which I also renamed to something friendlier) in c9fb860. Thanks!

PR #2534, opened 2023-07-20
Fix deactivate promotion URL

renatho:

* If fixes and undefined variable.

### Testing instructions

* Change a job to promoted adding the post meta `_promoted` as `1`.
* In the job listing click on "Deactivate" and make sure it works properly.


fjorgemota (approved):
Good catch! Looks good and works well!

PR #2536, opened 2023-07-24
Display a notice in the job listings when at least one job was promoted

### LINKED ISSUE #2531: Display a warning in the job listings when at least one job was promoted
If we have at least one promoted job, we should display a "congratulations" warning on the top of the WP Admin > Job listings page. And it should also be dismissible.

Image

Notice that we also want to add a link to the end of the process in the JobTarget marketplace, redirecting the user to the job listings, so if they click they would see it. If not, they will see it when they access the job listings page.

```[tasklist]
### Success Criteria
- [x] A congratulations warning is displayed in the job listings when we have at least one promoted job. https://github.com/Automattic/WP-Job-Manager/pull/2536
- [x] The warning is dismissible. https://github.com/Automattic/WP-Job-Manager/pull/2536
- [ ] Language is updated to reflect our warning about drafts
```

jom:
Advances #2531 

* Adds a notice when at least one job has been promoted.
* Caching the total count of active promoted posts so that we can easily use it in data collection stats later.
* **Note**: I've update the language for now. This PR might not totally be necessary without the issue of drafts removing promoted jobs. We could add something different if we'd think it would be helpful.

### Testing instructions

* In `wp shell` run `WP_Job_Manager_Promoted_Jobs::update_promotion( 775, true );` updating the post ID to one of your job posts.
* Go to WP Admin > Job Listings > All Jobs.
* Verify the notice appears.
* Dismiss the notice.
* Verify it doesn't reappear on page refresh.

### Screenshot
Screenshot 2023-07-24 at 6 42 25 pm

### DIFF
--- a/includes/admin/class-wp-job-manager-admin-notices.php
+++ b/includes/admin/class-wp-job-manager-admin-notices.php
@@ -31,14 +31,16 @@ class WP_Job_Manager_Admin_Notices {
 	const DISMISSED_NOTICES_USER_META   = 'wp_job_manager_dismissed_notices';
 
 	const ALLOWED_HTML = [
-		'div' => [
+		'div'    => [
 			'class' => [],
 		],
-		'a'   => [
+		'a'      => [
 			'target' => [],
 			'href'   => [],
 			'rel'    => [],
 		],
+		'em'     => [],
+		'strong' => [],
 	];
 
 	/**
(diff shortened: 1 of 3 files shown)


renatho (approved):
> Note: I've update the language for now. This PR might not totally be necessary without the issue of drafts removing promoted jobs. We could add something different if we'd think it would be helpful.

Hm! Good point!

Would you mind to keep the issue open when merging this PR? And maybe move to Blocked with a comment that we should tweak the text or close the issue depending on our final decision.

Other than that, it looks good and works well!

PR #2540, opened 2023-07-25
Fix notification site url

renatho:

* It forces the `site_url` to be used with `https`, like it's being used when we promote the job.

### Testing instructions

* In an env without `https` create a job, and make sure the `site_url` received in wpjobmanage.com is with `https://` (not changed, just to compare with the change).
* Update a job, and make sure the received notification comes in the same way (with `https://`).

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
@@ -152,7 +152,7 @@ class WP_Job_Manager_Promoted_Jobs_Notifications {
 	 */
 	private function get_notification_data() {
 		$site_url = home_url( '', 'https' );
-		$feed_url = rest_url( '/wpjm-internal/v1/promoted-jobs' );
+		$feed_url = rest_url( '/wpjm-internal/v1/promoted-jobs', 'https' );
 		$feed_url = substr( $feed_url, strlen( $site_url ) );
 		return [
 			'site_url' => $site_url,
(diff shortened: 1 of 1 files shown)


aaronfc (approved):
Looks good!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -151,7 +151,7 @@ private function get_notification_url() {
 	 * @return array The data to send.
 	 */
 	private function get_notification_data() {
-		$site_url = home_url();
+		$site_url = home_url( '', 'https' );
 		$feed_url = rest_url( '/wpjm-internal/v1/promoted-jobs' );
```
fjorgemota: I think we need to have the `rest_url` in the `https` scheme as well, no?

```suggestion
		$feed_url = rest_url( '/wpjm-internal/v1/promoted-jobs', 'https' );
```

aaronfc: Probably it would be a good idea to do that for all the endpoints we are sending to WPJMCOM (I am thinking on the get job data and verify endpoints).

fjorgemota: **I think** we're already doing that on other `home_url` and `rest_url` that we send to WPJMCOM...but yeah, definitely worth checking.

renatho: Ah! Good catch!

renatho: About the other cases, they're correct already.

```php
		$site_url            = home_url( '', 'https' );
		$job_endpoint_url    = rest_url( '/wpjm-internal/v1/promoted-jobs/' . $post_id, 'https' );
		$job_endpoint_url    = substr( $job_endpoint_url, strlen( $site_url ) );
```

fjorgemota (approved):
Looks good and works well!

PR #2542, opened 2023-07-25
Fix notification before meta change

### LINKED ISSUE #2541: Job deactivation not working when it runs fast
When deactivating a job, and it runs fast, it's not deactivating the job in wpjobmanager.com.

It seems to be related to the notification running on the hook `update_post_meta` (before saving the post meta).

renatho:
Fixes #2541

* It fixes an issue that happened when the wpjobmanager.com requested the feed from the site before the meta was updated. The reason is that [`update_postmeta` is fired before the update](https://github.com/WordPress/wordpress-develop/blob/6.2/src/wp-includes/meta.php#L302).
* It also enhances a little the logic, in case of code changes in the future identify more than one change in the same request (maybe a bulk change?). In this case, it will send only a single notification.

### Testing instructions

* Promote a job.
* Set it to promoted (`_promote` meta as `1` in the Job Listing post).
* Edit post title and make sure it continues updating wpjobmanager.com properly.
* Deactivate the promotion, and make sure it updates wpjobmanager.com.

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
@@ -52,6 +52,13 @@ class WP_Job_Manager_Promoted_Jobs_Notifications {
 	private $watched_fields;
 
 	/**
+	 * Whether watched fields were changed.
+	 *
+	 * @var bool
+	 */
+	private $watched_fields_changed;
+
+	/**
 	 * The single instance of the class.
 	 *
 	 * @var self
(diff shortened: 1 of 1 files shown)


fjorgemota (commented):
Hey there, thanks for the PR.

Two points: 

Looking at the code, it seems like we probably should use `updated_postmeta` instead of `update_postmeta`, right? This would prevent us from sending the notification if the post meta failed to update for some reason.

Also, I'm not sure if running the notification on the `shutdown` filter is a good idea, from what I see it is triggered just before PHP shuts down execution, even if there's an error in the request, so we might have issues on calling the Cron API if needed, for instance.

WDYT about maybe making it NOT run the `send_notification` more than once in the same request? I think it technically should have the same behavior than using the shutdown hook, except that we're not gonna make the request more than once per site per change.

fjorgemota (approved):
Hey there

As discussed on Slack:

1. We can't use `updated_postmeta` because we need to compare if the value actually changed before triggering the notification
2. `shutdown` is way more used than I expected, so it shouldn't be a problem to have it here.

I tested it here and everything still works well, so, I'm approving this change.

Thanks!

PR #2543, opened 2023-07-25
Display promote dialog only after the save is complete

### LINKED ISSUE #2538: Error when clicking on Promote in the editor before the job save is completed
When we publish a job and the modal is displayed, we can click on "Promote" before the job save is completed. It causes an error on the wpjobmanager.com side. We need to wait until the save is complete to open the modal.

renatho:
Fixes #2538

* Display promote dialog only after the save is complete. Otherwise, it could fetch the job data before it's created, returning an error in wpjobmanager.com.

### Testing instructions

* If you have some way to make your WPJM site slower, do it as a plus for this test (in my case, running the PHP debugger I could reproduce the issue easily).
* Create a new job.
* When publishing, make sure you see the dialog only after the save is complete.
* Click as soon as possible on the promote button.
* Make sure wpbjobmanager.com is able to fetch the job data properly.

### DIFF
--- a/assets/js/admin/job-editor.js
+++ b/assets/js/admin/job-editor.js
@@ -11,6 +11,7 @@ import domReady from '@wordpress/dom-ready';
 import editorLifecycle from './editor-lifecycle';
 import { postOpenPromoteModal } from './promote-job-modals';
 
+// Open promote dialog when it's publishing a job.
 domReady( () => {
 	const coreEditorSelector = select( editorStore );
 	const promoteDialog = document.querySelector( '#promote-dialog' );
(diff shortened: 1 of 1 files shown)


fjorgemota (approved):
Looks good and works well!

I wasn't able to promote a job before it was officially published because apparently we check if a job is already published before redirecting to WPJMCOM. I used the following snippet to test:

```php
function delay_job_listing($data, $postarr) {
    if ($data['post_type'] === 'job_listing') {
		sleep(20);
    }

    return $data;
}
add_filter('wp_insert_post_data', 'delay_job_listing', 10, 2);
```

However, using the above snippet I could verify that the behavior of the modal with your PR looks way better, hence, I'm already approving it!

renatho:
Thank you @fjorgemota! As discussed in Slack, I wrote the issue wrong. The error really happened on WPJM and not on WPJMCOM side.

PR #2544, opened 2023-07-25
Consider all promoted jobs as promoted even if it's not published

### LINKED ISSUE #2477: Avoid deactivating JobTarget promotion on accidental post status change
~~I'm creating this issue in order to track a case that is not very easy to understand by the users. We don't have a solution yet, so we should think on a solution and implement it.~~

~~Basically, if a user changes a job to draft or move it to trash, it will be removed from the feed, deactivating the promotion on JobTarget. What if the user change a job to draft or move to the trash accidentally?~~

~~Should we try to resolve it with warnings to the user? Should we try an implementation solution to avoid this type of situation? Should we just ignore it for now and see if it's really a problem when the users start using it?~~

In order to avoid deactivating a promotion based on an accidental change to draft or trash:

- Send all jobs with `_promoted` meta to the feed on the plugin side. And include an extra field that represents the post status.
- On wpjobmanager.com side, in order to remove a job from the feed, we need to schedule a job (probably 1 hour?).
- If the job was reactivated before the job runs, it shouldn't remove that job.

@fjorgemota also send an alternative proposal here to consider: p1688569546447249/1688567475.443379-slack-C8QJAM52B

Reference of the behavior on JobTarget: p1687906664898939-slack-C056EMC2ABG

renatho:
Fixes #2477

I'm anticipating it while it's still in discussion in p6r3EZ-1Hf-p2, in order to try to accelerate it a little. We can continue the discussion iterating this PR.

* Allow jobs of any post status on the promoted feed.
* Add the post_status to the job object in the feed and in the job data, in case we want to iterate handling it differently in the future without depending on a plugin update.
* Allow edit/deactivate promoted jobs that aren't published.

### Testing instructions

* Create and publish a job.
* Promote the job.
* Change the job status to draft, or any other status (even moving to trash).
* Make sure you can still deactivate the promotion.
* Create a new job as draft, and make sure you can't promote it.

### DIFF
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -256,6 +256,11 @@
             "example": "9788656",
             "description": "The unique identifier for the job in the site (post ID)."
           },
+          "post_status": {
+            "type": "string",
+            "example": "publish",
+            "description": "The job post status."
+          },
           "title": {
             "type": "string",
             "example": "Senior Software Engineer"
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -256,10 +256,10 @@
             "example": "9788656",
             "description": "The unique identifier for the job in the site (post ID)."
           },
-          "post_status": {
+          "job_status": {
             "type": "string",
             "example": "publish",
-            "description": "The job post status."
+            "description": "The job status."
           },
           "title": {
             "type": "string",
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -256,7 +256,7 @@
             "example": "9788656",
             "description": "The unique identifier for the job in the site (post ID)."
           },
-          "job_status": {
+          "status": {
             "type": "string",
             "example": "publish",
             "description": "The job status."
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -246,7 +246,8 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		} elseif ( 'publish' === $post->post_status ) {
 			echo '<button class="promote_job button button-primary" data-href=' . esc_url( $promote_url ) . '>' . esc_html__( 'Promote', 'wp-job-manager' ) . '</button>';
 		} else {
-			echo '<button class="button button-primary" disabled="disabled" title="' . esc_attr__( 'The job needs to be published in order to be promoted.', 'wp-job-manager' ) . '">' . esc_html__( 'Promote', 'wp-job-manager' ) . '</button>';
+			$title = __( 'The job needs to be published in order to be promoted.', 'wp-job-manager' );
+			echo '<button type="button" class="button button-primary tips disabled" aria-disabled="true" title="' . esc_attr( $title ) . '" data-tip="' . esc_attr( $title ) . '">' . esc_html__( 'Promote', 'wp-job-manager' ) . '</button>';
 		}
 	}
 
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -138,7 +138,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 	 * @return bool Returns true if they can promote a job.
 	 */
 	private function can_promote_job( int $post_id ) {
-		if ( 'job_listing' !== get_post_type( $post_id ) || 'publish' !== get_post_status( $post_id ) ) {
+		if ( 'job_listing' !== get_post_type( $post_id ) ) {
 			return false;
 		}
 
@@ -231,7 +231,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 				<a class="jm-promoted__deactivate delete" href="#" data-href="' . esc_url( $deactivate_action_link ) . '">' . esc_html__( 'Deactivate', 'wp-job-manager' ) . '</a>
 			</div>
 			';
-		} else {
+		} elseif ( 'publish' === $post->post_status ) {
 			echo '<button class="promote_job button button-primary" data-href=' . esc_url( $promote_url ) . '>' . esc_html__( 'Promote', 'wp-job-manager' ) . '</button>';
 		}
 	}
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -127,7 +127,17 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 			return;
 		}
 
-		echo '<span title="' . esc_attr__( 'This job has been promoted to external job boards.', 'wp-job-manager' ) . '" class="job_manager_admin_badge job_manager_admin_badge--promoted">' . esc_html__( 'Promoted', 'wp-job-manager' ) . '</span>';
+		$tooltip      = '';
+		$status_class = '';
+
+		if ( 'publish' !== $post->post_status ) {
+			$tooltip      = __( 'Your job is promoted and being exposed through API but it\'s not published in your site. You can fix it by publishing it again or deactivating the promotion.', 'wp-job-manager' );
+			$status_class = 'job_manager_admin_badge--not_published';
+		} else {
+			$tooltip = __( 'This job has been promoted to external job boards.', 'wp-job-manager' );
+		}
+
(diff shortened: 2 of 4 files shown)


includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -231,7 +233,7 @@ public function promoted_jobs_custom_columns( $column ) {
 				<a class="jm-promoted__deactivate delete" href="#" data-href="' . esc_url( $deactivate_action_link ) . '">' . esc_html__( 'Deactivate', 'wp-job-manager' ) . '</a>
 			</div>
 			';
-		} else {
+		} elseif ( 'publish' === $post->post_status ) {
 			echo '<button class="promote_job button button-primary" data-href=' . esc_url( $promote_url ) . '>' . esc_html__( 'Promote', 'wp-job-manager' ) . '</button>';
 		}
```
aaronfc, line 259: Should we add a disabled promote button with maybe a tooltip explaining that the job needs to be published before they can promote it?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -172,7 +172,9 @@ public function handle_promote_job() {
 		if ( ! $post_id ) {
 			wp_die( esc_html__( 'No job listing ID provided for promotion.', 'wp-job-manager' ), '', [ 'back_link' => true ] );
 		}
-		if ( ! $this->can_promote_job( $post_id ) ) {
+
+		$is_promoted_or_published = WP_Job_Manager_Promoted_Jobs::is_promoted( $post_id ) || 'publish' === get_post_status( $post_id );
```
aaronfc: I am wondering if we should have a `WP_Jov_Manager_Promoted_Jobs::can_be_promoted( $post_id )` method encapsulating the `'publish' === get_post_status(...)` check. Maybe even including the `is_promoted` part too? Not sure about that second part, though.

docs/api/internal.json
```diff
@@ -256,6 +256,11 @@
             "example": "9788656",
             "description": "The unique identifier for the job in the site (post ID)."
           },
+          "post_status": {
```
aaronfc: I am a bit concerned about mentioning here the "post". But it might make sense.

The alternative I thought of is just calling this 'status'. But then we are coupling the "job" status value, with the implementation's (post_status). Which I don't like either. Maybe that's the reason you used `post_status`?

Rambling a bit... so feel free to let it as it is :)

aaronfc (commented):
Looking good! Left a couple of comments and found some strange behaviours:

- If a job is moved to trash and restored there is no visual cue on the Job Listings view to identify it as being in "draft". Haven't confirmed it, but maybe this is something from WPJM itself? That basically draft jobs are not shown as "draft"?

image

- The "Do you want to promote the job?" modal appears when publishing a job (even if it's already promoted). In this case it's the same job I trashed/restored and is now in draft.

image

(Not sure why the ugly design, probably I have something wrong in in my local env).

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: Good idea! WDYT?

Screenshot 2023-07-26 at 15 52 50

docs/api/internal.json
renatho: I think it could make sense as `job_status`, mainly because it's special, having values like `"expired"`, for example.

renatho:
Thank you for the review, Aarón!

> If a job is moved to trash and restored there is no visual cue on the Job Listings view to identify it as being in "draft". Haven't confirmed it, but maybe this is something from WPJM itself? That basically draft jobs are not shown as "draft"?

About the draft, it's not very highlighted as you said. It's represented with an icon in the 4th column. But I took the opportunity to highlight the badge and also add a tooltip explaining what is happening there. WDYT?

Screenshot 2023-07-26 at 15 58 06


> The "Do you want to promote the job?" modal appears when publishing a job (even if it's already promoted). In this case it's the same job I trashed/restored and is now in draft.

I fixed it in a separate PR: https://github.com/Automattic/WP-Job-Manager/pull/2547

> (Not sure why the ugly design, probably I have something wrong in in my local env).

Your env probably wasn't able to fetch the modal content from wpjobmanager.com.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: It was like this before. I extracted the `publish` check because it depends on the context now, with the changes of this PR. But if you have a good idea of how to refactor it, feel free to comment here so that I can change it. 😉

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
aaronfc: Looks awesome 😲

docs/api/internal.json
aaronfc: Didn't think about the expired value. Yeah, then I like the job_status more than post_status.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
aaronfc: Still I think that having this 'publish' status checks everywhere might not be a good idea.

What about:
- Rename the `$this->can_promote_job( $post_id )` to `$this->can_current_user_promote_job( $post_id )`.
- Add a method `WP_Job_Manager_Promoted_Jobs::can_be_promoted( $post_id )` encapsulating the 'publish' status check? This method description must be something like "Check if a post_id can be applied the promote action or not". And would be used here and in the elseif to show the promote button.

If you think it's a bit confusing or that it might not be worth the effort, feel free to continue with what you have :)

docs/api/internal.json
```diff
@@ -256,6 +256,11 @@
             "example": "9788656",
             "description": "The unique identifier for the job in the site (post ID)."
           },
+          "job_status": {
```
aaronfc: Risking to be repetitive here but: what about just status? All the other fields (title, etc) are "job title". So I think the status field is no different?

If you chose job_status for some special reason feel free ignore this comment :)

aaronfc (commented):
Hey Renatho this looks great in your screenshots!

Unfortunately I couldn't get it to work. The tooltips (title attributes) are not appearing for me. I am using Chrome Version 115.0.5790.102.
I can't see it in either of the badge nor the disabled promote button.

I was going to say that "if this needs extra/complicated JS to work" maybe it's not worthy it... but then I found that we already support some kind of tooltips in that same screen:
image

So maybe it's as easy as adding the `class="tips" data-tip="bla bla bla` attributes?

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
renatho: Have you considered the edit when you thought about this?
I tried to refactor it as you said, but I couldn't achieve good results. The reason is that if we had a separate `can_be_promoted`, it would duplicate the check of `can_current_user_promote_job`, which we also want to check previously.

Anyway, I changed this part a little to improve the readability. It made this part a little bigger, but I think it was worth it. https://github.com/Automattic/WP-Job-Manager/pull/2544/commits/a68e011fe551846407d101fd218d58174c8d4248

WDYT?

docs/api/internal.json
renatho: Makes sense! Sorry that I misunderstood it previously. https://github.com/Automattic/WP-Job-Manager/pull/2544/commits/b1e2e52b352811320d77a97a91550f087dc0a0f6

docs/api/internal.json
renatho: As discussed in the [other comment](https://github.com/Automattic/WP-Job-Manager/pull/2544#discussion_r1275845477), updated to `status` only.

renatho:
> So maybe it's as easy as adding the class="tips" data-tip="bla bla bla attributes?

@aaronfc, thank you for the suggestion! Applied in https://github.com/Automattic/WP-Job-Manager/pull/2544/commits/6a3639fbe7ebc210ae2dbfb4e1638c2849ba2519 and https://github.com/Automattic/WP-Job-Manager/pull/2544/commits/254dd26d6b702bc781986cf954e999f6be035f73.

renatho:
@aaronfc, I'm merging it in order to update our CfT with this. But if you could complete the review when you can I would appreciate it, so I'll open other PRs, if needed. 😉

aaronfc:
Checked and everything looks very nice. Congrats! 🔥

PR #2545, opened 2023-07-25
Protect Get Promoted Job API with same condition as used by WP

### LINKED ISSUE #2514: Enhance the promoted job feature to also check the jobs REST API capability
_Originally reported by @renatho in https://github.com/Automattic/WP-Job-Manager/pull/2513#discussion_r1261601931_

Renatho raised some valid concerns during #2513, about the `promoted-job/<job_id>` endpoint being open.

I created this ticket to discuss about it.

fjorgemota:
Fixes #2514

* Use same permission check as WP REST Controller to validate whether we should return data or not

### Testing instructions

1. Create a job listing 
2. Publish it 
3. Visit `http://your.wpjm.site/wp-json/wpjm-internal/v1/promoted-jobs/JOB_ID` replacing JOB_ID with the post id of the job listing and make sure you can view the post data as appropriate
4. Move it to draft (or just move to trash)
5. Visit the same endpoint again and make sure you see the error "Sorry, you are not allowed to view this job"

### DIFF
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -222,7 +222,7 @@ class WP_Job_Manager_Promoted_Jobs_API {
 			return new WP_Error( 'not_found', __( 'The promoted job was not found', 'wp-job-manager' ), [ 'status' => 404 ] );
 		}
 		$controller = get_post_type_object( 'job_listing' )->get_rest_controller();
-		if ( $controller instanceof WP_REST_Posts_Controller && ! $controller->check_read_permission( $post ) ) {
+		if ( ! ( $controller instanceof WP_REST_Posts_Controller ) || ! $controller->check_read_permission( $post ) ) {
 			return new WP_Error( 'rest_forbidden', __( 'Sorry, you are not allowed to view this job.', 'wp-job-manager' ), [ 'status' => rest_authorization_required_code() ] );
 		}
 		$job_data = $this->prepare_item_for_response( get_post( $job_id ) );
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -217,7 +217,8 @@ class WP_Job_Manager_Promoted_Jobs_API {
 	 */
 	public function get_job_data( $request ) {
 		$job_id = $request->get_param( 'job_id' );
-		if ( 'job_listing' !== get_post_type( $job_id ) ) {
+		$post   = get_post( $job_id );
+		if ( 'job_listing' !== get_post_type( $post ) ) {
 			return new WP_Error( 'not_found', __( 'The promoted job was not found', 'wp-job-manager' ), [ 'status' => 404 ] );
 		}
 		$job_data = $this->prepare_item_for_response( get_post( $job_id ) );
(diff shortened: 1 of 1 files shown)


includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -217,9 +217,19 @@ public function update_job_status( $request ) {
...
+		if ( 'job_listing' !== get_post_type( $post ) ) {
 			return new WP_Error( 'not_found', __( 'The promoted job was not found', 'wp-job-manager' ), [ 'status' => 404 ] );
 		}
+		/**
+		 * The REST controller for job listings.
+		 *
+		 * @var WP_REST_Posts_Controller $controller
+		 */
+		$controller = get_post_type_object( 'job_listing' )->get_rest_controller();
+		if ( ! $controller->check_read_permission( $post ) ) {
```
fjorgemota: Officially `get_rest_controller()` returns a `WP_REST_Controller`, which doesn't have the `check_read_permission` method.

However, in `register_post_type` for the `job_listing` post type we explicitly define the `WP_REST_Posts_Controller` as the rest controller for this post type, so...i think we're safe here? Anyway I added the phpdoc just to avoid complaints from IDEs..

aaronfc: I think we are pretty safe, but what do you think about checking the class type? Or maybe the existence of the `check_read_permission` method?

In case it isn't we can decide forbidding access (in order to not be less restrictive ever).

aaronfc: After a second check, can't we use `get_item_permissions_check` ? Which is part of the `WP_REST_Controller` signature and seems like ends up checking `check_read_permission` in the WP_REST_Posts_Controller implementation?

fjorgemota: @aaronfc Hey there, thanks for your review! I thought about using `get_item_permissions_check`, but I opted to not use it because it receives a `WP_REST_Request`..pretty tailored to the WP REST API specific for posts, where `job_id` is just `id` and there's also a `context` argument. Also, it is meant to be used as a permission callback, which I'm not sure is what we want here. In the end, it seemed worse than just using `check_read_permission`.

I re-added the check for the class type  in fe3195352bd4f3aefdae30b1238a349d52df709d and ce0dd957868d098868592ef5efd96edc4cb358ad  before calling the method, but, I'm not sure about this part you mentioned:

> In case it isn't we can decide forbidding access (in order to not be less restrictive ever).

Should we just block all the access in case the REST controller doesn't match what we expected?

aaronfc: Yeah, sorry for the wording. It doesn't look very easy to understand now that I read it a second time haha

My thought was that, since this is to not ever leak more information than what REST API does. In the case we can't do the check, I would assume we want to fail.
But it was just a suggestion because you would have to make that decision. Feel free to go with the other option if you want. I don't think it will ever happen anyway.

fjorgemota: @aaronfc Oh, I understand you now, and yeah, it sounds fair. I adapted the check to block requests if the REST Controller isn't an instance of `WP_REST_Posts_Controller` in 5356390. Thanks!

aaronfc (commented):
This is looking good!

But the test I did was returning me `rest_forbidden` error.

```
curl --request GET \
  --url https://wpjm.local/wp-json/wpjm-internal/v1/promoted-jobs/32 \
  --header 'Content-Type: application/json'
```

Is it working for you? Maybe my environment is acting weird?

aaronfc (approved):
Okay. The post I was testing with was in draft 🗡️ 

Approving! 🚀

PR #2547, opened 2023-07-26
Don't open the promote modal on the editor when the job is already promoted

renatho:
Fixes issue reported by @aaronfc https://github.com/Automattic/WP-Job-Manager/pull/2544#pullrequestreview-1547036865

> The "Do you want to promote the job?" modal appears when publishing a job (even if it's already promoted). In this case it's the same job I trashed/restored and is now in draft.

* Avoid opening the promote modal on the editor when the job is already promoted.

### Testing instructions

* Create a new job and publish it.
* Make sure you see the promote modal.
* Promote the job.
* Return to the job editor.
* Switch to draft.
* Publish again, and make sure you don't see the promote modal again.

### DIFF
--- a/assets/js/admin/job-editor.js
+++ b/assets/js/admin/job-editor.js
@@ -30,8 +30,10 @@ domReady( () => {
 				'publish' !== coreEditorSelector.getCurrentPostAttribute( 'status' );
 		},
 		onSave: () => {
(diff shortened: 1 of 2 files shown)


aaronfc (approved):
🚀 Works great!

Issue #2417, opened 2023-05-11
Use new endpoints for license management


jom:
### Success Criteria
- [x] ~Use new endpoint for single plugin activation~ We just use the bulk activation endpoint now.
- [x] Use new endpoint for single plugin deactivation
- [x] Use new endpoint to fetch plugin information
- [x] Use new endpoint to bulk check plugin information https://github.com/Automattic/WP-Job-Manager/pull/2552

PR #2552, opened 2023-07-31
Add bulk update check

jom:
Part of #2417 

* Switch over to the bulk update endpoint.
* Remove old tests and rewrite based on new techniques. I'll add additional tests as I migrate things to the new endpoints.
* Fixed issue that prevented plugin installed multiple times from getting updates.

### Testing instructions

* Make sure your local WPJMCOM is on the `add/icon-info` branch.
* Use the following snippet to point WPJM to your local env:
```

		$new_url = str_replace(
			$scheme . '://' . $host,
			'http://wpjobmanagercom.vipdev.lndo.site',
			$url
		);

	}
	return $preempt;
}, 10, 3 );
	return true;
});
```
- Install older versions of some WP Job Manager paid add-ons. Clear your site transient cache (`wp transient delete --all --network`).
- Go to Job Listings > All Jobs and make sure the updates available notice shows up.
- With licenses deactivated, ensure plugin updates show up for those plugins in WP Admin > Dashboard > Updates. No downloads should appear.
- Ensure notice still shows with plugin update available from WP Admin > Job Listings.
* Activate a license.
* Ensure plugin updates show up as available. 
* Install plugin update.
* Make sure it removes the plugin from the notice at the top of WP Admin > Job Listings. 

- 

### Deprecated Code

* `WP_Job_Manager_Helper_API::plugin_update_check`
* `WP_Job_Manager_Helper::get_installed_plugins` signature changed. Soft requirement for new second argument `$keyed_by_filename`.


includes/helper/class-wp-job-manager-helper.php
```diff
@@ -155,7 +155,7 @@ private function handle_admin_request() {
 		$product_slug = isset( $_GET['dismiss-wpjm-licence-notice'] ) ? sanitize_text_field( wp_unslash( $_GET['dismiss-wpjm-licence-notice'] ) ) : false;
 
 		if ( ! empty( $product_slug ) ) {
-			$product_plugins = $this->get_installed_plugins();
+			$product_plugins = $this->get_installed_plugins( false, false );
```
jom, line 158: This is a change in the current behavior. It seems like there was a bug that only allowed people to dismiss notices for _active_ plugins.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -194,65 +194,148 @@ public function addon_core_version_check( $do_check, $minimum_required_core_vers
...
+
+			$this->deactivate_licence( $plugin_slug, true );
+		} elseif ( in_array( 'expired_key', $error_keys, true ) ) {
+			// translators: First placeholder is the plugin name, second placeholder is the My Account URL.
+			$errors['expired_key'] = sprintf( __( '<strong>Error:</strong> The license for <strong>%1$s</strong> has expired. You must <a href="%2$s" rel="noopener noreferrer" target="_blank">renew your license</a> to receive updates and support.', 'wp-job-manager' ), $plugin_name, $my_account_url );
+		} elseif ( in_array( 'expiring_soon', $error_keys, true ) ) {
+			// translators: First placeholder is the plugin name, second placeholder is the My Account URL.
+			$errors['expiring_soon'] = sprintf( __( '<strong>Error:</strong> The license for <strong>%1$s</strong> is expiring soon. Please <a href="%2$s" rel="noopener noreferrer" target="_blank">renew your license</a> to continue receiving updates and support.', 'wp-job-manager' ), $plugin_name, $my_account_url );
```
jom, line 345: This isn't implemented yet on WPJobManager.com but thought I'd throw it in as a nice-to-have eventually.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -384,7 +466,7 @@ public function is_product_installed( $product_slug ) {
 	 * @return bool
 	 */
 	public function has_licenced_products() {
-		$product_plugins = $this->get_installed_plugins();
+		$product_plugins = $this->get_installed_plugins( false, false );
```
jom, line 479: This is another change in implementation. Beforehand this would only care about activated licensed plugins. This is only used to decide whether to force showing the license page, which we want for inactive plugins.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -449,39 +531,55 @@ public function extra_headers( $headers ) {
...
 		 *
 		 * @param bool $clear_plugin_cache True if we should clear the plugin cache.
 		 */
-		if ( ! self::$cleared_plugin_cache && apply_filters( 'job_manager_clear_plugin_cache', true ) ) {
+		if ( ! self::$cleared_plugin_cache && apply_filters( 'job_manager_clear_plugin_cache', ! isset( $plugins[ JOB_MANAGER_PLUGIN_BASENAME ] ) ) ) {
 			// Reset the plugin cache on the first call. Some plugins prematurely hydrate the cache.
 			wp_clean_plugins_cache( false );
 			self::$cleared_plugin_cache = true;
+			$plugins                    = get_plugins();
 		}
```
jom, line 576: We were being too eager on this before for an edge case. Some plugins were polluting the cache. I think it is (a) Rare and (b) Sniffable by seeing if the (obviously active) WP Job Manager plugin is listed. Before we were flushing the plugin update cache _way too often._

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -499,7 +597,7 @@ public function licence_output() {
 		if ( ! empty( $_POST ) ) {
 			$this->handle_request();
 		}
-		$licenced_plugins = $this->get_installed_plugins();
+		$licenced_plugins = $this->get_installed_plugins( false, false );
```
jom, line 610: I don't understand why we were limiting this to active only before.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -571,12 +668,11 @@ public function licence_error_notices() {
 		$this->maybe_add_license_error_notices();
 	}
 
-
 	/**
 	 * Outputs unset licence key notices.
 	 */
 	public function maybe_add_license_error_notices() {
-		foreach ( $this->get_installed_plugins() as $product_slug => $plugin_data ) {
+		foreach ( $this->get_installed_plugins( false, false ) as $product_slug => $plugin_data ) {
```
jom, line 685: Again, we should show notices for inactive plugins as well. They should be receiving updates.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -624,7 +720,7 @@ private function handle_request() {
 	 * @return void
 	 */
 	private function handle_single_request() {
-		$licenced_plugins = $this->get_installed_plugins();
+		$licenced_plugins = $this->get_installed_plugins( false, false );
```
jom, line 733: Likewise, we should be allowing folks to active/deactivate licenses for inactive plugins.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -194,65 +194,148 @@ public function addon_core_version_check( $do_check, $minimum_required_core_vers
...
-				&& version_compare( $response['new_version'], $plugin_data['Version'], '>' )
+				! empty( $response->new_version )
+				&& version_compare( $response->new_version, $plugin_data['Version'], '>' )
 			) {
-				$available_addon_updates[ $product_slug ]                      = $response;
-				$check_for_updates_data->response[ $plugin_data['_filename'] ] = (object) $response;
+				$check_for_updates_data->response[ $plugin_data['_filename'] ] = $response;
 			}
 		}
-		set_site_transient( 'wpjm_addon_updates_available', $available_addon_updates ); // No expiration set.
```
jom, line 217: We were caching this and never using it (as far as I could see).

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -194,65 +194,158 @@ public function addon_core_version_check( $do_check, $minimum_required_core_vers
...
+	 */
+	private function get_plugin_update_package( $installed_plugins ) {
+		$plugin_package = [];
+		foreach ( $installed_plugins as $plugin ) {
+			$plugin_slug = $plugin['_product_slug'];
+			$license_key = $this->get_plugin_licence( $plugin_slug );
+
+			$plugin_package[ $plugin_slug ] = [
+				'installed_version' => $plugin['Version'],
+				'license_key'       => $license_key['licence_key'] ?? '',
```
aaronfc, line 252: Given that the default return value for `WP_Job_Manager_Helper_Options::get( $product_slug, 'licence_key' );` is `false`, I think that what we want here is: `$license_key['licence_key'] ? $license_key['licence_key'] : ''`

In my testings we were sending through the API: `..."wp-job-manager-embeddable-job-widget":{"installed_version":"1.1.3","license_key":false}...` which is returning a 400 because `license_key` is expected to be a string.

includes/helper/class-wp-job-manager-helper.php
aaronfc: Maybe we should create a ticket referencing this part then?

includes/helper/class-wp-job-manager-helper.php
aaronfc: No idea either. Multisite scenario came to my mind, but not really sure there is any issue with showing this for the inactive too there.

includes/helper/class-wp-job-manager-helper.php
```diff
@@ -908,6 +992,9 @@ private function handle_product_activation_response( $response, $product_slug, $
 			self::log_event( 'license_activation_error', $event_properties );
 		} else {
 			self::log_event( 'license_activated', $event_properties );
+
+			// Clear the update cache so we can get the packages.
+			delete_site_transient( 'wpjm_helper_updates' );
```
aaronfc, line 997: Should we also delete the `update_plugins` transient?
In my test I had to flush the transients after activating the license for a plugin in order to see the "update" action in the plugins page.

aaronfc (commented):
Looking good! Left some comments :)

includes/helper/class-wp-job-manager-helper.php
jom: Cleared in c84f82e199ef0d01581daa5771243566baad27fa.

includes/helper/class-wp-job-manager-helper.php
jom: Good catch! I changed downstream in the licence PR just so we can cleanly merge these. https://github.com/Automattic/WP-Job-Manager/pull/2554/commits/2f458247c094fcec74cbe203b596b29c535bb43a

That PR is up-to-date with this branch.

includes/helper/class-wp-job-manager-helper.php
jom: Good idea: https://github.com/Automattic/wpjobmanager.com/issues/570

aaronfc (approved):
Looks good and works well! 🚀 
Tested with #2554

PR #2561, opened 2023-08-14
Change check if promoted jobs feature was used or not

renatho:

* We have a check if the user used the promoted jobs feature to start running the CRON to request job status changes.
* Currently, we're marking it as used when the user clicks on "Promote".
* Reflecting on this, my proposal with this PR is to have fewer sites running the CRON and making requests to wpjobmanager.com, so I moved the option creation to the `refresh-status` endpoint, running the CRON only when the user completed a job promotion. Notice that if it fails the first time the user uses this feature, the site will never get a status update from wpjobmanager.com until it changes status again, requesting `refresh-status`. IMO, we could try this approach, and if we have users complaining about that, we can iterate to a better approach.

### Testing instructions

* Make sure you don't have the option `job_manager_used_promoted_jobs` on your site.
* Click on "Promote" for a job.
* Make sure the option `job_manager_used_promoted_jobs` wasn't created yet.
* Complete the promotion, submitting the promote form, and going through the JobTarget checkout.
* Run `vip dev-env -- exec wp cron event run wpjobmanager_com_sync_promoted_job_status` in order to sync the status.
* Make sure the option `job_manager_used_promoted_jobs` was created.

### DIFF
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -209,8 +209,6 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
-		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::USED_PROMOTED_JOBS_OPTION_KEY, true );
-
 		$url = add_query_arg(
 			[
 				'user_id'             => $current_user,
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -339,6 +339,7 @@ class WP_Job_Manager_Promoted_Jobs_API {
 	 * @return WP_REST_Response The response.
 	 */
 	public function refresh_status( $request ) {
+		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::USED_PROMOTED_JOBS_OPTION_KEY, true );
 		$this->status_handler->fetch_updates();
 		return new WP_REST_Response( [ 'success' => true ] );
 	}
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
@@ -23,7 +23,7 @@ class WP_Job_Manager_Promoted_Jobs_Status_Handler {
 	const LAST_EXECUTION_OPTION_KEY = self::CRON_HOOK . '_last_execution';
 
 	/**
-	 * The name of the option that stores whether the site has promoted jobs or not.
+	 * The name of the option that stores whether the site used promoted jobs or not.
 	 */
 	const USED_PROMOTED_JOBS_OPTION_KEY = 'job_manager_used_promoted_jobs';
 


fjorgemota:
Hey there :wave: 

I started thinking about this issue and I think it might be a little risky to NOT have the cron running unless the site receives a webhook.

So, I started thinking here and I'd like to suggest a different thing (based on the original feature branch, by the way): what if instead of having a flag to set whether the cron should run or not, we change it to save the value of the `UPDATE_INTERVAL` constant? (which wouldn't be a constant anymore, it would be the value of the option)

Based on this change, we could apply a few changes:

1. Change the cron to `hourly`, to run more often
2. If the option just doesn't exist, the cron doesn't run, as usual
3. If the user is redirected to WPJMCOM, we set the value of that option to run the cron once per hour
4. On WPJMCOM, we define a new header (or we just use `Cache-Control` and [let VIP cache it on Edge Servers](https://docs.wpvip.com/technical-references/caching/page-cache/)?) on the WPJMCOM feed to determine when the next update should happen 
5. When fetching the WPJMCOM feed from WPJM, we read the header (or parse the `Cache-Control`?), get the value and set the option to whatever new value WPJMCOM defines (maybe with support for a special value to delete the option again?)

The main logic here is that, by doing this, we will be able to remote-control how often we fetch feeds from WPJMCOM. For instance, if a site doesn't even exist on WPJMCOM (we don't have a Promoted Site for it), it doesn't make sense to make WPJMCOM keep pinging the site. In the same way, we can adjust how often we will let WPJM update the frequency depending on whether the site has jobs currently active, or pending, or whatever.

WDYT?

cc @aaronfc for reference too

aaronfc:
Hey @renatho, I think that the change, as you already identified is a bit risky. I agree that we can go with this and just change the logic if there is any problem, but given that this is on WPJM side it might become trickier to solve AND get people updated on a reasonable time.

I haven't tested the feature but the change looks reasonable and if you prefer to go with this approach I am fine.

---

Hey @fjorgemota, thanks for the ping! I think this is actually a smart solution. But I am wondering if it might be too much for now. It involves changing both WPJM and WPJMCOM. It indeed would give us the greatest flexibility. But again, not sure if maybe it's too much given the stage of the project? (We don't even know how many people will be using this feature at all!)

---

Both, @renatho and @fjorgemota. I bring you yet another idea haha.
It's not a well-thought solution, just an idea: what if we keep track for "when" a user clicked the promote button and also "when" the site received a sync request?

If we kept track of both events, (persisting the timestamps as an option?), we could then tweak the logic of the cronjob to only run at most X days after the last time that the user clicked the "promote" button OR the site received a sync request.
This is just a simple logic, but we can make it as elaborated as we want.

fjorgemota:
>  I think that the change, as you already identified is a bit risky. I agree that we can go with this and just change the logic if there is any problem, but given that this is on WPJM side it might become trickier to solve AND get people updated on a reasonable time.
> I haven't tested the feature but the change looks reasonable and if you prefer to go with this approach I am fine.

Yeah, I was talking with @renatho yesterday that the biggest issue IMHO is that, given that this is on WPJM side, the only two options for fixing an eventual issue of "the promoted job status isn't being updated on WPJM" would be by updating the plugin (assuming that we release a fix for it) OR asking the user to set an option to make the cron job run. IMHO, both aren't ideal.

>  thanks for the ping! I think this is actually a smart solution. But I am wondering if it might be too much for now. It involves changing both WPJM and WPJMCOM. It indeed would give us the greatest flexibility. But again, not sure if maybe it's too much given the stage of the project? (We don't even know how many people will be using this feature at all!)

I'm not sure. To be clear, I think the part about returning different expiration times based on the status of the promoted job is just a nice to have, it is not a requirement for this to work at all. So, based on this, I believe the solution wouldn't be too complicated given what we already have here AND given the problem we're trying to solve.

Given that people apparently [don't update plugins very often](https://wordpress.org/plugins/wp-job-manager/advanced/) (note that most people AREN'T using the latest WPJM version), I think it would be nice to control this from WPJMCOM side, even if it is a constant there too.

> It's not a well-thought solution, just an idea: what if we keep track for "when" a user clicked the promote button and also "when" the site received a sync request?
> If we kept track of both events, (persisting the timestamps as an option?), we could then tweak the logic of the cronjob to only run at most X days after the last time that the user clicked the "promote" button OR the site received a sync request.
> This is just a simple logic, but we can make it as elaborated as we want.

I think the logic that's already implemented is quite similar to what you're suggesting here. The `fetch_updates()` method is already rate-limited and, right now, has this flag to either disable it completely AND the constant to make sure it doesn't run very often. The rate limit looks quite related to what you suggested here, IMHO, in special because it works especially when the site receives a sync request. I'm not completely sure how tracking when the user clicked the "promote" button helps here. :thinking:

fjorgemota:
FYI, I sent an alternative proposal to this implementing my idea on #2563. And the WPJMCOM side on 581-gh-Automattic/wpjobmanager.com (that side only implements the logic if the promoted site doesn't exist, it doesn't implement the whole "dynamic smart update interval" thing)

renatho:
Closed in favor of https://github.com/Automattic/WP-Job-Manager/pull/2563

renatho:
Closed in favor of https://github.com/Automattic/WP-Job-Manager/pull/2563

PR #2562, opened 2023-08-16
Change deactivation approach

renatho:

* We identified an issue that could happen in intervals between syncs (WPJM <> wpjobmanager.com <> JobTarget). It could cause a promotion stop accidentally. The discussion: p1692125471733679-slack-C053397HN30.
* In order to fix it, we're implementing @fjorgemota's idea to keep all the jobs with `_promoted` meta key on the feed, even if it's not promoted. To represent it in the feed, we added the `promote` attribute, which is not currently used but it's useful if we want to iterate in the future, and it makes the endpoint more concise.

#### The previous issue:

1. User promote a job, completing the JobTarget checkout.
2. Before the 5 minutes polling interval (or more if they delay to complete the checkout), the user change something in another promoted jobs triggering a sync (WPJM -> [wpjobmanager.com](http://wpjobmanager.com/)).
3. Their job would still have `_promoted` `0` because it didn't sync with the new status yet ([wpjobmanager.com](http://wpjobmanager.com/) -> WPJM). So in the sync WPJM -> [wpjobmanager.com](http://wpjobmanager.com/), it would remove the job from [wpjobmanager.com](http://wpjobmanager.com/), removing from the feed that JobTarget will consume, which would deactivate their promotion.

### Testing instructions

* Make sure your env has at least one promote job.
* Promote a new job, completing the JobTarget checkout.
* Before the status is updated in WPJM, update another promoted job, and make sure the new job you created wasn't removed from wpjobmanager.com (Check it through WP Admin because the feed has cache, or just clear the cache).
* On wpjobmanager.com, run the job to update the statuses `vip dev-env -- exec wp cron event run wpjobmanager_com_sync_promoted_job_status`.
* Check that the job is marked as promoted on WPJM.
* Deactivate the job on WPJM job listings, and make sure it was removed from wpjobmanager.com.
* For another promoted job of your site, update the status to expired (see instructions on https://github.com/Automattic/wpjobmanager.com/pull/579).
* Make sure the job is marked as not promoted on job listings but continues on the feed of WPJM and wpjobmanager.com.

### DIFF
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -300,8 +300,13 @@
           },
           "status": {
             "type": "string",
-            "example": "publish",
+            "example": true,
             "description": "The job status."
+          },
+          "promoted": {
+            "type": "boolean",
+            "example": false,
+            "description": "Whether the job has promoted status on the plugin."
           },
           "title": {
             "type": "string",
--- a/docs/api/internal.json
+++ b/docs/api/internal.json
@@ -300,7 +300,7 @@
           },
           "status": {
             "type": "string",
-            "example": true,
+            "example": "publish",
             "description": "The job status."
           },
           "promoted": {
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -209,6 +209,11 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
+		// Make sure the job contains the promoted job meta to be listed in the feed.
+		if ( ! $is_editing ) {
+			WP_Job_Manager_Promoted_Jobs::update_promotion( $post_id, false );
+		}
+
 		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::USED_PROMOTED_JOBS_OPTION_KEY, true );
 
 		$url = add_query_arg(
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -190,8 +190,7 @@ class WP_Job_Manager_Promoted_Jobs_API {
 			'meta_query'          => [
 				[
 					'key'     => WP_Job_Manager_Promoted_Jobs::PROMOTED_META_KEY,
-					'value'   => '1',
-					'compare' => '=',
+					'compare' => 'EXISTS',
 				],
 			],
 		];
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -234,6 +234,7 @@ class WP_Job_Manager_Promoted_Jobs_API {
 		return [
 			'id'           => (string) $item->ID,
 			'status'       => $item->post_status,
+			'promoted'     => WP_Job_Manager_Promoted_Jobs::is_promoted( $item->ID ),
 			'title'        => $item->post_title,
 			'description'  => $item->post_content,
 			'permalink'    => get_permalink( $item ),
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
@@ -125,6 +125,7 @@ class WP_Job_Manager_Promoted_Jobs_Notifications {
 
 	/**
 	 * Checks if we should send a notification to wpjobmanager.com after promoted meta key is deleted.
+	 * It happens when a job promotion is deactivated.
 	 *
 	 * @param string[] $meta_ids
 	 * @param int      $post_id  Post ID.
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
@@ -85,10 +85,9 @@ class WP_Job_Manager_Promoted_Jobs_Notifications {
 	public function __construct() {
 		$this->watched_fields = [
 			'post' => [ 'post_name', 'post_title', 'post_content', 'post_status' ],
-			'meta' => [ WP_Job_Manager_Promoted_Jobs::PROMOTED_META_KEY ],
 		];
 		add_action( 'post_updated', [ $this, 'post_updated' ], 10, 3 );
-		add_action( 'update_postmeta', [ $this, 'meta_updated' ], 10, 4 );
+		add_action( 'deleted_post_meta', [ $this, 'deleted_meta' ], 10, 3 );
 		add_action( 'shutdown', [ $this, 'maybe_trigger_notification' ] );
 		add_action( 'job_manager_promoted_jobs_notification', [ $this, 'send_notification' ] );
 	}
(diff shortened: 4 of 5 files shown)


includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
```diff
@@ -125,22 +124,18 @@ function( $key ) use ( $keys ) {
 	}
 
 	/**
-	 * Checks if we should send a notification to wpjobmanager.com after a post meta is updated.
+	 * Checks if we should send a notification to wpjobmanager.com after promoted meta key is deleted.
```
fjorgemota, line 127: Maybe it is worth documenting that the idea for this method is to detect when the job was unpromoted?

I initially found it confusing because I thought it would be cool to track whether `_promoted` changes from `0` to `1`, but..I'm not sure that's the goal of this method after reading these changes, right?

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
```diff
@@ -234,6 +233,7 @@ private function prepare_item_for_response( WP_Post $item ) {
 		return [
 			'id'           => (string) $item->ID,
 			'status'       => $item->post_status,
+			'promoted'     => WP_Job_Manager_Promoted_Jobs::is_promoted( $item->ID ),
```
fjorgemota, line 236: Do you foresee any changes we need to make on WPJMCOM to use this field?

I am just thinking about the use-case here. :thinking:

docs/api/internal.json
```diff
@@ -300,9 +300,14 @@
           },
           "status": {
             "type": "string",
-            "example": "publish",
+            "example": true,
```
fjorgemota: I think the job status isn't modified by your code changes, right? It's still effectively the post status, no?

```suggestion
            "example": "publish",
```

includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php
```diff
@@ -161,6 +162,10 @@ public static function update_promotion( $post_id, $promoted ) {
 
 		delete_option( self::PROMOTED_JOB_TRACK_OPTION );
 
+		if ( 'force_delete' === $promoted ) {
```
fjorgemota: I'm thinking here about this. It is a bit weird for me to have a parameter that might be a boolean, but also might be a string, and a string with a special value (otherwise, it is interpreted as `true`).

I'm not sure what the best approach here:

1. Add a third parameter in the spirit of `$force_delete` on `wp_delete_post`, that, when `$promoted` is `false` and that parameter is `true` (the default value is `false`), it deletes the post meta as appropriate
2. Maybe copy some of the logic present here (to validate if the post type is correct and delete the promoted job track option) to a private method , and implement the change directly on `deactivate_promotion`.

Also, now that I'm thinking here. This change that I suggested basically implies that, after a job is expired on WPJMCOM, it will never delete the post meta here on WPJM, right? It will just set `_promoted` to `0` and call it a day. :thinking:

fjorgemota (commented):
Hey there, this is looking good! 

I left a couple of comments/suggestions, and I'll test the PR in the meantime. :)

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
renatho: Currently, I don't think so. But I added it because it can be useful in the future and make the feed more concise (we don't have promoted and not promoted jobs in the feed without a differentiation).
We could use that, in case we want to list all the promoted jobs as a type of extra promotion, for example.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
renatho: Sure! WDYT? https://github.com/Automattic/WP-Job-Manager/pull/2562/commits/63db340d060cfcfe88bdbd69665306607c9bc9d7

docs/api/internal.json
renatho: Oops! It was an accidental change.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php
renatho: > I'm thinking here about this. It is a bit weird for me to have a parameter that might be a boolean, but also might be a string, and a string with a special value (otherwise, it is interpreted as true).

Implemented your suggestion 2 here: https://github.com/Automattic/WP-Job-Manager/pull/2562/commits/99c974bb5e3a663a3701874971b576ff16c4ef62

> Also, now that I'm thinking here. This change that I suggested basically implies that, after a job is expired on WPJMCOM, it will never delete the post meta here on WPJM, right? It will just set _promoted to 0 and call it a day. 🤔

Yep! This was the way to differentiate the cases where the user is deactivating manually (that we really want to remove from the feed in order to deactivate the promotions). Otherwise, the edge case of accidentally stopping a promotion because of the interval sync could happen. Pasting here the case, just for context:

> 1. User promote a job, completing the JobTarget checkout.
> 2. Before the 5 minutes polling interval (or more if they delay to complete the checkout), the user change something in a promoted jobs triggering a sync (WPJM -> [wpjobmanager.com](http://wpjobmanager.com/)).
> 3. Their job would still have _promoted 0 because it didn't sync with the new status yet ([wpjobmanager.com](http://wpjobmanager.com/) -> WPJM). So in the sync WPJM -> [wpjobmanager.com](http://wpjobmanager.com/), it would remove the job from [wpjobmanager.com](http://wpjobmanager.com/), removing from the feed that JobTarget will consume, which would deactivate their promotion.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-notifications.php
fjorgemota: Much better. Thanks :)

includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php
fjorgemota: Much better, thanks! And thanks for the clarification!

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -209,6 +209,11 @@ public function handle_promote_job() {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
+		// Make sure the job contains the promoted job meta to be listed in the feed.
+		if ( ! $is_editing ) {
+			WP_Job_Manager_Promoted_Jobs::update_promotion( $post_id, false );
```
fjorgemota, line 214: Good catch on setting the flag only if we're creating a new job

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
fjorgemota: Sounds good, thanks!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php
```diff
@@ -149,18 +149,17 @@ public static function is_promoted( $post_id ) {
 	/**
 	 * Update promotion.
 	 *
-	 * @param int  $post_id
-	 * @param bool $promoted
+	 * @param int         $post_id
+	 * @param bool|string $promoted `true` to promoted, `false` to not promoted, `force_delete` to delete.
+	 *                              The deletion is used to force a removal from the feed, deactivating the promotion while syncing.
```
fjorgemota: I think we should probably update this PHPDoc, right?

fjorgemota (approved):
Looks good and works well! I left a comment for updating a PHPDoc, but everything else seems to be working just fine. Thanks!

includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php
renatho: Yep! Good catch! I updated the previous commit just changing this. ;)
https://github.com/Automattic/WP-Job-Manager/pull/2562/commits/51ebc4b1032916694a0271080d4d8afcf95b8335

PR #2563, opened 2023-08-17
Add support for setting update interval dynamically from WPJMCOM

fjorgemota:
Alternative to #2561

* Changes the Promoted Jobs Status Handle to run the check hourly  (but it WILL NOT run hourly unless the server specifies to do so)
* Split the update interval setting into two different options: CRON and WEBHOOK. The first is considered when the WP-CRON is triggering the request, and the other is considering in every other call (usually coming out from the webhook).
* Every time the status handler fetches an update from the server, it will look for the `X-WPJM-Webhook-Interval` and `X-WPJM-Cron-Interval` headers and will update the value of the respective option accordingly (or delete it, if the header's value is `false`)
* The system considers always the last execution time (whether it comes from the cron or the webhook) when running the update, to not run very often
* From WPJM side, this PR is effectively identical to what's already implemented on the feature branch. Except that now we have the power to change the options at will from WPJM. :D 

### Testing instructions

**NOTE**: Consider that "running the webhook" is the same as running the command ` curl -X POST "http://your.wpjm.instance/?rest_route=/wpjm-internal/v1/promoted-jobs/refresh-status"`

**TIP**: You can use `wp option list --search="job_manager_promoted_jobs_*"` to see the options related to Promoted Jobs. :) 

On a clean WP installation running this branch and running a WPJMCOM instance based on 581-gh-Automattic/wpjobmanager.com.

1. Make sure the options `job_manager_promoted_jobs_cron_interval` and `job_manager_promoted_jobs_webhook_interval` aren't set
2. Try running the webhook using  and verify that it doesn't make any requests (you can check that by verifying if the option `job_manager_promoted_jobs_status_update_last_execution` doesn't exist)
3. Promote a job
4. Make sure the options `job_manager_promoted_jobs_cron_interval` and `job_manager_promoted_jobs_webhook_interval` are set and are numbers (the cron_interval should be set to 43200, and the webhook_interval to 300). Make sure the `job_manager_promoted_jobs_status_update_last_execution` option still doesn't exist (if it exists, delete it)
5. Try running the webhook now and make sure the option `job_manager_promoted_jobs_status_update_last_execution` is set
7. Delete the Promoted Site (and their Promoted Jobs) from WPJMCOM side
8. Delete the `job_manager_promoted_jobs_status_update_last_execution` (or set it to `0`, it is effectively the same)
9. Try running the webhook again and verify if the options  `job_manager_promoted_jobs_cron_interval` and `job_manager_promoted_jobs_webhook_interval` are deleted now

Sorry for the complicated testing instructions, by the way. :)

### DIFF
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -209,6 +209,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
+		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::LAST_CHECK_OPTION_KEY, time() - ( 5 * MINUTE_IN_SECONDS ), '', false );
 		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_INTERVAL_OPTION_KEY, 12 * HOUR_IN_SECONDS, '', false );
 		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::WEBHOOK_INTERVAL_OPTION_KEY, 5 * MINUTE_IN_SECONDS, '', false );
 
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -209,7 +209,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
-		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_INTERVAL_OPTION_KEY, 5 * MINUTE_IN_SECONDS );
+		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_INTERVAL_OPTION_KEY, 12 * HOUR_IN_SECONDS );
 		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::WEBHOOK_INTERVAL_OPTION_KEY, 5 * MINUTE_IN_SECONDS );
 
 		$url = add_query_arg(
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -209,9 +209,7 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
-		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::LAST_CHECK_OPTION_KEY, time() - ( 5 * MINUTE_IN_SECONDS ), '', false );
-		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_INTERVAL_OPTION_KEY, 12 * HOUR_IN_SECONDS, '', false );
-		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::WEBHOOK_INTERVAL_OPTION_KEY, 5 * MINUTE_IN_SECONDS, '', false );
+		WP_Job_Manager_Promoted_Jobs_Status_Handler::initialize_defaults();
 
 		$url = add_query_arg(
 			[
--- a/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
+++ b/includes/admin/class-wp-job-manager-promoted-jobs-admin.php
@@ -209,8 +209,8 @@ class WP_Job_Manager_Promoted_Jobs_Admin {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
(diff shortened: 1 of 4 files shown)


includes/class-wp-job-manager.php
```diff
@@ -251,7 +251,7 @@ public static function maybe_schedule_cron_jobs() {
 			wp_schedule_event( time(), 'daily', 'job_manager_email_daily_notices' );
 		}
 		if ( ! wp_next_scheduled( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_HOOK ) ) {
-			wp_schedule_event( time(), 'twicedaily', WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_HOOK );
+			wp_schedule_event( time(), 'hourly', WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_HOOK );
```
fjorgemota, line 254: It is important to note that, although this change is a bit weird at first, it is only done to increase the number of checks on the fallback per day.

We're still limited to running only twice a day on the internal check that the Promoted Jobs Status Handler verify. BUT we now run this check much more often, which allows us to increase the number of checks we do if we want on WPJMCOM side.

includes/admin/class-wp-job-manager-promoted-jobs-admin.php
```diff
@@ -209,7 +209,8 @@ public function handle_promote_job() {
 		$verify_endpoint_url = rest_url( '/wpjm-internal/v1/promoted-jobs/verify-token', 'https' );
 		$verify_endpoint_url = substr( $verify_endpoint_url, strlen( $site_url ) );
 
-		update_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::USED_PROMOTED_JOBS_OPTION_KEY, true );
+		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::CRON_INTERVAL_OPTION_KEY, 12 * HOUR_IN_SECONDS, '', false );
+		add_option( WP_Job_Manager_Promoted_Jobs_Status_Handler::WEBHOOK_INTERVAL_OPTION_KEY, 5 * MINUTE_IN_SECONDS, '', false );
```
fjorgemota: These values are literally the default that we had set on the feature branch. Except that now they're options and are controllable via HTTP Response Headers. Also note that, if the options already exist on the database, we won't overwrite them here, too. :) (the priority comes from the data on the server, always)

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
```diff
@@ -102,4 +126,24 @@ private function get_site_feed_url() {
...
+		if ( isset( $headers[ $header_name ] ) ) {
+			$header = $headers [ $header_name ];
+			if ( 'false' === $header ) {
+				delete_option( $option_name );
+			} elseif ( rest_is_integer( $header ) ) {
+				update_option( $option_name, (int) $header, false );
+			}
+		}
+	}
+
```
fjorgemota, line 159: This is the logic to update the option according to the response headers. If the value is the magic `false`, we delete the option, otherwise we verify if it is an integer and, if it is, we update the value on the database

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
```diff
@@ -73,6 +68,33 @@ public function fetch_updates() {
...
+	/**
+	 * Get the current interval for fetching updates. This is either the interval for the cron or the webhook, depending
+	 * on the context of the current request.
+	 *
+	 * @return int The current interval for fetching updates.
+	 */
+	private function get_current_interval() {
+		$option_name = wp_doing_cron() ? self::CRON_INTERVAL_OPTION_KEY : self::WEBHOOK_INTERVAL_OPTION_KEY;
```
fjorgemota, line 94: I'm using `wp_doing_cron()` here to detect whether this is a CRON request or not and use the appropriate option key.

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
```diff
@@ -73,6 +68,33 @@ public function fetch_updates() {
...
+	 * @param int $current_time The current time.
+	 * @return bool True if the update logic should be executed, false otherwise.
+	 */
+	private function should_update( $current_time ) {
+		$interval = $this->get_current_interval();
+		if ( ! $interval ) {
+			// If the interval is not set or is zero, we don't update.
+			return false;
+		}
+		$last_execution = get_option( self::LAST_EXECUTION_OPTION_KEY, 0 );
```
renatho: Not a big deal, but I think it's worth also casting it to `int`, similar to the `get_current_interval`.

renatho (approved):
Looks good and works well! Thank you for implementing it!

fjorgemota:
Hey @renatho, thanks for your review!

As discussed on Slack, I changed a few additional commits to fix some issues that I discovered, and also to improve the code a bit more. Hence, I'm requesting your review again.

For documentation purposes, the issue that I discovered is that this approach had two issues:

1. If the cron for updating the status ran while the user was still filling the form, it would delete the options of intervals for webhooks and crons, causing the code to effectively block the webhook/cron from fetching the feed
2. If WPJMCOM returned `false` or `0` for the interval of the webhook, it would effectively block the webhook from running (which isn't ideal)

The 1º issue made this PR similar to #2561 (in the sense that the job status might never be updated) and was fixed in 3039e61. The 2º issue was fixed in aa6ef7f.

Sorry for the trouble! :)

includes/promoted-jobs/class-wp-job-manager-promoted-jobs-status-handler.php
```diff
@@ -102,4 +126,47 @@ private function get_site_feed_url() {
...
+	}
+
+	/**
+	 * Initialize the default options for the Promoted Jobs Status Handler.
+	 *
+	 * @return void
+	 */
+	public static function initialize_defaults() {
+		$webhook_interval = 5 * MINUTE_IN_SECONDS;
+		update_option( self::LAST_CHECK_OPTION_KEY, time() - $webhook_interval, '', false );
```
renatho: Thank you, Fernando! Just one final thing I noticed:
Since we're handling the webhook interval dynamically, I think we should get this interval from the option and use this only if the option doesn't exist, right?

The rest is looking great to me! I'll just wait it and the conflict fix before running a final test to approve. ;)

fjorgemota: I implemented your suggestion in 4e0b5a1. Thanks!

renatho (approved):
Looks good and works well! 🚀🚀🚀

PR #2565, opened 2023-08-21
Fix tests

renatho:
The tests were broken because of the license -> license renaming. The reason is that in the tests, the `get_base_args` is used for some functions, and also to create the URL, but [we temporarily didn't change the URL to keep backward compatibility](https://github.com/Automattic/WP-Job-Manager/commit/50787702ab2b68900ea55487fe90946109eba472).

So my temporary approach here was to replace the key only when creating the URL.

### DIFF
--- a/tests/php/tests/includes/helper/test_class.wp-job-manager-helper-api.php
+++ b/tests/php/tests/includes/helper/test_class.wp-job-manager-helper-api.php
@@ -225,6 +225,24 @@ class WP_Test_WP_Job_Manager_Helper_API extends WPJM_BaseTest {
 	}
 
 	protected function build_url( $args ) {
+		// The legacy endpoints are temporary. For now, translate `license_key` => `licence_key` at this point.
+		$args = $this->updateArrayKey( $args, 'license_key', 'licence_key' );
+
 		return 'https://wpjobmanager.com/?' . http_build_query( $args, '', '&' );
+	}
+
+	private function updateArrayKey( $array, $oldKey, $newKey ) {
+		if ( array_key_exists( $oldKey, $array ) ) {
+			$newArray = array();
+			foreach ( $array as $key => $value ) {
+				if ( $key === $oldKey ) {
+					$newArray[ $newKey ] = $value;
+				} else {
+					$newArray[ $key ] = $value;
+				}
+			}
+			return $newArray;
+		}
+		return $array;
 	}
 }


tests/php/tests/includes/helper/test_class.wp-job-manager-helper-api.php
```diff
@@ -225,6 +225,24 @@ protected function default_invalid_response() {
...
 
 	protected function build_url( $args ) {
+		// The legacy endpoints are temporary. For now, translate `license_key` => `licence_key` at this point.
+		$args = $this->updateArrayKey( $args, 'license_key', 'licence_key' );
+
 		return 'https://wpjobmanager.com/?' . http_build_query( $args, '', '&' );
 	}
+
+	private function updateArrayKey( $array, $oldKey, $newKey ) {
+		if ( array_key_exists( $oldKey, $array ) ) {
```
renatho, line 235: This method is used to keep the arguments in the same order to match the transport faker hash (See `Requests_Transport_Faker::make_request_signature`).

tests/php/tests/includes/helper/test_class.wp-job-manager-helper-api.php
```diff
@@ -225,6 +225,24 @@ protected function default_invalid_response() {
...
+			$newArray = array();
+			foreach ( $array as $key => $value ) {
+				if ( $key === $oldKey ) {
+					$newArray[ $newKey ] = $value;
+				} else {
+					$newArray[ $key ] = $value;
+				}
+			}
+			return $newArray;
```
aaronfc, line 244: You could have done something like:
```
if ( array_key_exists( $oldKey, $array ) ) {
  $array[$newKey] = $array[$oldKey];
  unset($array[$oldKey]);
}
return $array;
```
So we don't need to iterate over all the keys.

aaronfc (approved):
Looks good!

Minor optimisation in the comments, but feel free to use yours. I don't think it makes any difference for this specific case.

tests/php/tests/includes/helper/test_class.wp-job-manager-helper-api.php
renatho: Hey @aaronfc! Thank you for your review.
This implementation is because of the order of the keys. More details in this comment: https://github.com/Automattic/WP-Job-Manager/pull/2565#discussion_r1300776663

So I'm going ahead and merging it.

PR #2573, opened 2023-09-08
Add event to datepicker method

gikaragia:
This issue came up while I was having a look at this ticket: 6733381-zd-a8c

It seems that although we support a type of 'date' in WPJM forms, we don't support this type in fields that are generated dynamically. This happens in resumes.

* Adds a wpJobManagerFieldAdded event to datepicker method

### Testing instructions

* Tested as part of this PR: 115-gh-Automattic/wpjm-addons

### Release Notes

* Support dynamically added date form fields

### DIFF
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -9,19 +9,21 @@ jQuery(document).on( 'ready', function() {
 	}
 
 	var initializeDatepicker = function ( targetInput ) {
-		var $hidden_input = jQuery( '<input />', { type: 'hidden', name: jQuery(targetInput).attr( 'name' ) } ).insertAfter( jQuery( targetInput ) );
-		jQuery(targetInput).attr( 'name', jQuery(targetInput).attr( 'name' ) + '-datepicker' );
-		jQuery(targetInput).keyup( function() {
-			if ( '' === jQuery(targetInput).val() ) {
+		var $target = jQuery( targetInput );
+		var $hidden_input = jQuery( '<input />', { type: 'hidden', name: $target.attr( 'name' ) } ).insertAfter( $target );
+
+		$target.attr( 'name', $target.attr( 'name' ) + '-datepicker' );
+		$target.keyup( function() {
+			if ( '' === $target.val() ) {
 				$hidden_input.val( '' );
 			}
 		} );
-		jQuery(targetInput).datepicker( jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
-		if ( jQuery(targetInput).val() ) {
-			var dateParts = jQuery(targetInput).val().split('-');
+		$target.datepicker( jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
+		if ( $target.val() ) {
+			var dateParts = $target.val().split('-');
 			if ( 3 === dateParts.length ) {
 				var selectedDate = new Date(parseInt(dateParts[0], 10), (parseInt(dateParts[1], 10) - 1), parseInt(dateParts[2], 10));
-				jQuery(targetInput).datepicker('setDate', selectedDate);
+				$target.datepicker('setDate', selectedDate);
 			}
 		}
 	};
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -1,5 +1,5 @@
 /* global job_manager_datepicker */
-jQuery(document).on( 'ready', function() {
+jQuery(document).ready( function() {
 	var datePickerOptions = {
 		altFormat  : 'yy-mm-dd',
 	};
@@ -13,7 +13,7 @@ jQuery(document).on( 'ready', function() {
 		var $hidden_input = jQuery( '<input />', { type: 'hidden', name: $target.attr( 'name' ) } ).insertAfter( $target );
 
 		$target.attr( 'name', $target.attr( 'name' ) + '-datepicker' );
-		$target.keyup( function() {
+		$target.on( 'keyup', function() {
 			if ( '' === $target.val() ) {
 				$hidden_input.val( '' );
 			}
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -1,5 +1,5 @@
 /* global job_manager_datepicker */
-jQuery(document).on( 'ready wpJobManagerFieldAdded', function() {
+jQuery(document).on( 'ready', function() {
 	var datePickerOptions = {
 		altFormat  : 'yy-mm-dd',
 	};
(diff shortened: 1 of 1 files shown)


assets/js/datepicker.js
```diff
@@ -1,26 +1,27 @@
 /* global job_manager_datepicker */
-jQuery(document).ready(function($) {
+jQuery(document).on( 'ready wpJobManagerFieldAdded', function() {
```
yscik: Using  the same callback for these means all the date fields will be initialized again when the new event is triggered. Is that safe, could it lead to problems? Should we extract the callback that initializes a single input, and run only that for new fields?

Meaning something like this, though the new event would need to reference the newly added input field:

`jQuery( 'input.job-manager-datepicker, input#_job_expires' ).each( initializeDatepicker )`)
`jQuery(document).on( 'wpJobManagerFieldAdded' , initializeDatepicker )`

yscik (approved):
Looks good. I'm seeing one potential issue but if in testing it didn't cause trouble then it should be fine?

assets/js/datepicker.js
gikaragia: Great point @yscik! Missed that! So I am not sure if this actually caused an issue (although I am pretty sure that it did) as I found out another issue where if you added dynamically more than one dates and tried to select the date it would break.

The problem is that datepicker relies on the HTML id and unfortunately the ids that were generated for this form were not unique. I have updated the code to have unique ids, can you have another look please?

I tried another approach initially, to rename the actual fields but it wasn't possible as these are stored in the db and we would break backward compatibility. There are updates in this PR too: https://github.com/Automattic/wpjm-addons/pull/115

assets/js/datepicker.js
```diff
@@ -1,27 +1,37 @@
 /* global job_manager_datepicker */
-jQuery(document).ready(function($) {
+jQuery(document).on( 'ready', function() {
```
fjorgemota: Is there any specific reason for this change?

I'm asking because using on the PR in `wpjm-addons`, we use the first original syntax and use `$` in the code. Meanwhile, here now we're using the `jQuery` identifier directly instead.

assets/js/datepicker.js
```diff
@@ -1,27 +1,37 @@
...
 	}
 
-	$( 'input.job-manager-datepicker, input#_job_expires' ).each( function(){
-		var $hidden_input = $( '<input />', { type: 'hidden', name: $(this).attr( 'name' ) } ).insertAfter( $( this ) );
-		$(this).attr( 'name', $(this).attr( 'name' ) + '-datepicker' );
-		$(this).keyup( function() {
-			if ( '' === $(this).val() ) {
+	var initializeDatepicker = function ( targetInput ) {
+		var $hidden_input = jQuery( '<input />', { type: 'hidden', name: jQuery(targetInput).attr( 'name' ) } ).insertAfter( jQuery( targetInput ) );
+		jQuery(targetInput).attr( 'name', jQuery(targetInput).attr( 'name' ) + '-datepicker' );
```
fjorgemota: Maybe this is a nit-pick (probably it is), but WDYT about saving `jQuery(targetInput)` to a variable (maybe `$target_input`?) or something and use that instead?

fjorgemota (commented):
Hey there! This  is looking good. I just left a code review with a few suggestions. I'm going to test it to make sure it works in the meantime. :)

assets/js/datepicker.js
```diff
@@ -1,27 +1,37 @@
...
+			if ( '' === jQuery(targetInput).val() ) {
 				$hidden_input.val( '' );
 			}
 		} );
-		$(this).datepicker( $.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
-		if ( $(this).val() ) {
-			var dateParts = $(this).val().split("-");
+		jQuery(targetInput).datepicker( jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
+		if ( jQuery(targetInput).val() ) {
+			var dateParts = jQuery(targetInput).val().split('-');
```
fjorgemota: hmm, I think maybe we should use `.attr("value")` to read the value instead of using `.val()`?

If I understand this correctly, the idea is to read the value that PHP added as the `value` of the field. However, apparently, the `.datepicker` above is changing the `.value` of the element (which is returned by jQuery's `.val()` ) to a formatted date, which doesn't use `-` anymore (at least not here on my PC).

If we use `.attr("value")`, it will run a classic `.getAttribute("value")`, which returns the original value in the HTML, which I think is what we want.

This solves the issue I found on the review here: 115-gh-Automattic/wpjm-addons#pullrequestreview-1634134670 

WDYT?

fjorgemota (commented):
I tested this and it looks generally good. 

I'm adding another review just to add a comment about the issue I found while testing WPJM Resumes. Might be worth looking into. :)

Thanks!

assets/js/datepicker.js
gikaragia: Good point implemented on 26185de999d630be21941f5e3b9da3e083f421aa

assets/js/datepicker.js
gikaragia: So regarding $ vs JQuery, I don't think it makes much difference. I have a slight preferance for JQuery as it avoids potential conflicts (although conflicts were not possible in this case). So the change was done because I wanted to subscribe to many events in a previous commit.

The problem however was the usage of `on('ready')` which although it runs, it causes a weird initialization issue. Although `wp_localize_jquery_ui_datepicker` was running before `datepicker.js`, it seemed like the datepicker wasn't initialized during the initial page load. This caused the issue that you mentioned in your other comment which should be fixed now. 59fc67dfa704f737e6488c277cd94b15b0d05675

assets/js/datepicker.js
gikaragia: See my comment above.

assets/js/datepicker.js
fjorgemota: > So regarding $ vs JQuery, I don't think it makes much difference. I have a slight preferance for JQuery as it avoids potential conflicts (although conflicts were not possible in this case). So the change was done because I wanted to subscribe to many events in a previous commit.

I see. For me it is pretty indifferent, too. However, I do like to maintain a pattern in the code, as it is easier to me to look at a code and know what to expect according to the rest of the project, so maintaining the usage of `jQuery` vs. `$` in a consistent way helps me a lot.

> The problem however was the usage of on('ready') which although it runs, it causes a weird initialization issue. Although wp_localize_jquery_ui_datepicker was running before datepicker.js, it seemed like the datepicker wasn't initialized during the initial page load. This caused the issue that you mentioned in your other comment which should be fixed now. https://github.com/Automattic/WP-Job-Manager/commit/59fc67dfa704f737e6488c277cd94b15b0d05675

I can confirm that this fixes the issue. Thanks!

fjorgemota (approved):
Looks good and works well. =)

PR #2576, opened 2023-09-13
Add more plugin statistics

gikaragia:

* Adds a few more plugins to track to our stats
* Starts tracking plugins that have `wp-job-manager` anywhere in their slug, not just the start
* Specifically tracks which plugins have an active license
* Fixes an issue where WPJM plugins weren't reported. The issue was caused in installations where a plugin calls `get_plugins` before WPJM has a chance to register the `WPJM-Product` custom header. In that case the header wasn't populated and the plugins weren't included in the results of `get_installed_plugins`

### Testing instructions

* As the changes are in data that are submitted to Tracks, it is easier if you simply add some logging to see the reported data.
* Install some plugins like classic editor, Woo etc. Also install a few of our paid plugins.
* Run the `job_manager_usage_tracking_send_usage_data` cronjob. You can use WP Control plugin for that.
* Observe that the data reported are correctly.
* Activate/deactivate licenses to our plugins and observe that the data are still correct.

### DIFF
--- a/includes/class-wp-job-manager-usage-tracking-data.php
+++ b/includes/class-wp-job-manager-usage-tracking-data.php
@@ -30,7 +30,7 @@ class WP_Job_Manager_Usage_Tracking_Data {
 			$categories = wp_count_terms( 'job_listing_category', [ 'hide_empty' => false ] );
 		}
 
-		return [
+		$usage_data = [
 			'employers'                   => self::get_employer_count(),
 			'job_categories'              => $categories,
 			'job_categories_desc'         => self::get_job_category_has_description_count(),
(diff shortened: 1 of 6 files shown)


yscik (approved):
Looks good! Asked in p1694699949666949-slack-C056Q2JS5B5 if we should enable tracking for all plugins.

gikaragia:
Good idea @yscik, the more data, the better. I imagine the reason that a small number of plugins were tracked initially was to not overload the stats endpoint.

I ran some tests and it seems that there is a limit imposed by the tracks endpoint. It is hit when the site has ~480 plugins installed so I think it's pretty safe to track every plugin.

yscik (approved):

PR #2577, opened 2023-09-13
Add plugin upload and playground setup to PR builds

yscik:

* Change the plugin build step action to upload the plugin zip file to wpjobmanager.com's media library, instead of as an artifact of the action
* Post the plugin zip link and a link to open it in a testing site ~as a comment~ to the PR
* Delete previous zips when a PR is updated or merged

See p6r3EZ-1PL-p2 for more details. 

### Testing instructions

* See GH action logs for this PR: https://github.com/Automattic/WP-Job-Manager/actions/runs/6176999314/job/16767164534?pr=2577
* See generated comment below

----

| Plugin build for d7b3e6dafe75c6d63749a71b7df2f249205ebbad <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/09/wp-job-manager-zip-2577-d7b3e6da.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/09/2577-d7b3e6da)             |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,49 +1,51 @@
 name: Plugin Build
 
-on: pull_request
+on:
+  - pull_request
 
 jobs:
-    build:
-        name: Plugin Build
-        runs-on: ubuntu-latest
-        steps:
-            - uses: actions/checkout@v2
-            - name: Get npm cache directory
-              id: npm-cache
-              run: |
-                  echo "::set-output name=dir::$(npm config get cache)"
-            - uses: actions/cache@v2
-              with:
-                  path: ${{ steps.npm-cache.outputs.dir }}
-                  key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
-                  restore-keys: |
-                      ${{ runner.os }}-node-
-            - uses: shivammathur/setup-php@v2
-              with:
-                php-version: '7.4'
-                tools: composer
-                coverage: none
-            - name: Get composer cache directory
-              id: composer-cache
-              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-            - uses: actions/cache@v2
-              with:
-                  path: ${{ steps.composer-cache.outputs.dir }}
-                  key: ${{ runner.os }}-composer-${{ hashFiles('composer.lock') }}
-                  restore-keys: |
-                      ${{ runner.os }}-composer-
-            - name: Install JS dependencies
-              run: npm ci
-            - name: Install PHP dependencies
-              run: composer install --no-ansi --no-interaction --prefer-dist --no-progress
-            - name: Build Plugin
-              run: npm run build
-            - name: Decompress plugin
-              run: unzip wp-job-manager.zip -d wp-job-manager
-            - name: Store Artifact
-              uses: actions/upload-artifact@v2
-              with:
-                name: wp-job-manager-${{ github.event.pull_request.head.sha }}
-                path: ${{ github.workspace }}/wp-job-manager/
-                retention-days: 7
-                  
+  build:
+    name: Plugin Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get npm cache directory
+        id: npm-cache
+        run: |
+          echo "::set-output name=dir::$(npm config get cache)"
+      - uses: actions/cache@v2
+        with:
+          path: ${{ steps.npm-cache.outputs.dir }}
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          tools: composer
+          coverage: none
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+      - uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-
+      - name: Install JS dependencies
+        run: npm ci
+      - name: Install PHP dependencies
+        run: composer install --no-ansi --no-interaction --prefer-dist --no-progress
+      - name: Build Plugin
+        run: npm run build
+      - name: Decompress plugin
+        run: unzip wp-job-manager.zip -d wp-job-manager
+      - name: Store Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: wp-job-manager-${{ github.event.pull_request.head.sha }}
+          path: ${{ github.workspace }}/wp-job-manager/
+          retention-days: 7
+      - name: Save plugin zip and add link to it in a comment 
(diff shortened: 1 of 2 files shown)


yscik:
Updated to put the links in the PR description instead of a comment, to prevent getting notifications each time the build is run

fjorgemota:
FYI, one approach that IMHO works well is to just update the comment previously posted instead of posting a new comment each time.

I used that approach with WordPress Dependencies Report (p6rkRX-5Aw-p2), and it works especially well because it also avoids an additional notification each time the build is run. :eyes: I think Jetpack also does something similar to this, btw.

yscik:
There is still an initial notification for the PR author for the first comment which I didn't like :) It seems right to put it close to 'Testing instructions', but we can change it to whatever is everyone's preference.

scripts/upload-and-link-plugin-zip.mjs
```diff
@@ -0,0 +1,97 @@
...
+
+	const oldZips = JSON.parse( execSync( `curl -s -u "${ login }" "${ MEDIA_LIBRARY_ENDPOINT }?mime_type=application/zip&_fields=id,date,title,source_url"` ).toString() );
+	if ( oldZips?.code || ! Array.isArray( oldZips ) ) {
+		console.log( `[${ oldZips.code }]`, oldZips.message );
+		return;
+	}
+	oldZips?.forEach( ( zip ) => {
+		const
+			title = zip.title.rendered;
+		if ( title.match( new RegExp( `wp-job-manager-zip-${pr}` ) ) ) {
```
fjorgemota: Can't this be replaced by a simpler `.startsWith`?

```suggestion
		if ( title.startsWith( `wp-job-manager-zip-${pr}-`  ) ) {
```

scripts/upload-and-link-plugin-zip.mjs
```diff
@@ -0,0 +1,97 @@
...
+	console.log( chalk.green( '✓' ), `Plugin file uploaded to ${ zip }` )
+
+	return zip;
+}
+
+/**
+ * Post a comment on the PR with a link to the plugin zip and playground.
+ * @param {string} zip URL to plugin zip file.
+ */
+function addComment( zip ) {
```
fjorgemota: Probably we will want to update the function name/jsdoc to match the new intent?

scripts/upload-and-link-plugin-zip.mjs
```diff
@@ -0,0 +1,97 @@
...
+( function main() {
+
+	// deleteOldZips();
+	//
+	// if ( merged ) {
+	// 	return;
+	// }
+	//
+	// const zip = uploadZip();
+	const zip = "https://wpjobmanager.com/wp-content/uploads/2023/09/wp-job-manager-zip-2577-26255dc9.zip";
```
fjorgemota: Did you change this just for the sake of the demo?

scripts/upload-and-link-plugin-zip.mjs
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Upload plugin zip to media library.
+ *
+ * @returns {string} URL to plugin zip file.
+ */
+function uploadZip() {
+
+	const id = `${ pr }-${ commit.substring( 0, 8 ) }`;
+
+	const zip = execSync( `curl -u "${ login }" --http1.1 --data-binary @wp-job-manager.zip -H "Content-Disposition: attachment; filename=\"wp-job-manager-zip-${ id }.zip\"" ${ MEDIA_LIBRARY_ENDPOINT }?title=wp-job-manager-zip-${ id } | jq .source_url` ).toString().replaceAll( '"', '' ).trim();
```
fjorgemota: If a non-automattician forks this repo, and then send a PR to it, then `login` will be empty, as [the secret isn't available in these scenarios](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). Maybe we shold avoid trying to run the `curl` command if `login` is empty? (and then avoid modifying the description too, naturally?)

Also, what happens if this command fails to run properly?

scripts/upload-and-link-plugin-zip.mjs
```diff
@@ -0,0 +1,97 @@
...
+function addComment( zip ) {
+
+	const [ , path, id ] = zip.match( 'wp-content/uploads/(.*)/wp-job-manager-zip-(.*).zip' );
+	const playgroundLink = `https://wpjobmanager.com/playground/?core=${ path }/${ id }`
+
+	const links = `
+<!-- wpjm:plugin-zip -->
+----
+
+| Plugin build for ${ commit } <a href="#"><img width=600></a> |
```
fjorgemota, line 106: Do we need this link? 

```suggestion
| Plugin build for ${ commit } |
```

fjorgemota (commented):
Hey there, this is looking generally good.

IMHO, I feel like having a separate comment is better than just changing the description, especially because the description is something that the user can edit to point anywhere else, but a comment from `github-actions` cannot.

With that said, I found a few issues that I think are worth mentioning. =)

Thanks for the PR!

scripts/upload-and-link-plugin-zip.mjs
yscik: 🤦  Yup, restored in 138a689

scripts/upload-and-link-plugin-zip.mjs
yscik: Good catch, I added an early exit. Generally if anything throws an error it should make it to the workflow's logs, but made it a bit more graceful ([`commits`](https://github.com/Automattic/WP-Job-Manager/pull/2577/files/64e63fba8aa437c5e6be6606af534ea97df47282..265b709cac8676e43712f1818d30f552b8bd8876))

scripts/upload-and-link-plugin-zip.mjs
yscik: The image is a hack for making the table full-width, and the link is a hack to override the default link generated for images 👨‍💻

fjorgemota: Oh, that's neat. Nice! Thanks for explaining what it is. :)

scripts/upload-and-link-plugin-zip.mjs
yscik: Updated in [859fe9e](https://github.com/Automattic/WP-Job-Manager/pull/2577/commits/859fe9e9cc62677e10113374680768ae18004cb9)

.github/workflows/build.yml
```diff
@@ -1,49 +1,54 @@
...
+        run: npm run build
+      - name: Decompress plugin
+        run: unzip wp-job-manager.zip -d wp-job-manager
+      - name: Store Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: wp-job-manager-${{ github.event.pull_request.head.sha }}
+          path: ${{ github.workspace }}/wp-job-manager/
+          retention-days: 7
+      - name: Save plugin zip and add link to it in a comment
```
fjorgemota: Just a reminder to update this:

```suggestion
      - name: Save plugin zip and add link to the PR description
```

fjorgemota (approved):
Looks good and works well!

I see two possible improvements for this task. One being a must, another being more optional:

1. I created an issue on WPJMCOM to handle the first issue: 654-gh-Automattic/wpjobmanager.com, just to avoid talking about the issue publicly
2. The second is moving most of the logic to WPJMCOM (so e.g. it downloads files from Github instead of uploading files from Github to WPJMCOM, as well as handle the ZIP files appropriately). This isn't a requirement to make this work.

With that said, I'm already approving the issue. I also left a suggestion to fix the name of the action in the `build.yml` workflow, too.

Thanks!

PR #2585, opened 2023-09-20
Use new "Deactivate License" endpoint for deactivating licenses on WPJMCOM

fjorgemota:
Part of #2417

### Changes proposed in this Pull Request

- Change endpoint used to deactivate licenses in `WP_Job_Manager_Helper_API`

### Testing instructions

On a WPJM installation based on this version of the plugin:

1. Install a WPJM addon;
2. Activate the license for that addon
3. Make sure the license is correctly associated with the host of your WP installation on wpjobmanager.com
4. Deactivate the license for that addon via WPJM UI (just click on the "Deactivate License" button)
5. Make sure the license is correctly disassociated with the host of your WP installation on wpjobmanager.com

PR #2588, opened 2023-09-22
Change endpoint used for getting plugin information from WPJMCOM

### LINKED ISSUE #2417: Use new endpoints for license management

fjorgemota:
Based on #2585
Fixes #2417 

- Use new endpoint on WPJMCOM for getting plugin information from WPJM

### Testing instructions

Using a WP installation running this branch of the WPJM plugin:

1. Install a WPJM addon, let's say, the `wp-job-manager-application-deadline` addon
2. Activate it with a license from WPJMCOM
3. Run the following code on `wp shell`:

```php
require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
var_dump(plugins_api('plugin_information', ['slug'=>'wp-job-manager-application-deadline']));
```
5. Make sure the plugin information printed makes sense (has the correct name, `plugin`, slug, version, etc.). Also, make sure the `download_link` isn't empty.
6. Now, deactivate the license for that plugin on the WPJM panel
7. Run the above snippet again, make sure the `download_link` is now empty.

### DIFF
--- a/includes/helper/class-wp-job-manager-helper-api.php
+++ b/includes/helper/class-wp-job-manager-helper-api.php
@@ -77,13 +77,46 @@ class WP_Job_Manager_Helper_API {
 	 * Sends and receives data related to plugin information from the WPJobManager.com API.
 	 *
 	 * @param array $args  The arguments to pass to the endpoint.
-	 * @return array|false The response, or false if the request failed.
+	 * @return \stdClass|false The response, or false if the request failed.
 	 */
 	public function plugin_information( $args ) {
-		$args            = wp_parse_args( $args );
-		$args['wc-api']  = 'wp_plugin_licencing_update_api';
-		$args['request'] = 'plugininformation';
-		return $this->request( $args );
+		$args = wp_parse_args( $args );
+		$data = $this->request_endpoint(
+			'wp-json/wpjmcom-licensing/v1/plugin-information',
+			[
+				'method' => 'POST',
+				'body'   => wp_json_encode(
+					[
+						'site_url'     => $this->get_site_url(),
+						'license_key'  => $args['license_key'],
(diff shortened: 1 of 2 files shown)


includes/helper/class-wp-job-manager-helper-api.php
```diff
@@ -77,13 +77,44 @@ public function bulk_update_check( array $plugins ) {
...
+				'body'   => [
+					'site_url'     => $this->get_site_url(),
+					'license_key'  => $args['license_key'],
+					'product_slug' => $args['api_product_id'],
+				],
+			]
+		);
+		if ( ! is_array( $data ) ) {
+			return false;
+		}
```
fjorgemota, line 97: Here we have a difference in behavior compared to the current WPJM version: Currently, when the legacy endpoint on WPJMCOM receives a `product_slug` that doesn't exist, it returns an empty `stdClass` (but still with all the correct format). In every other case, we don't return an error per se.

Here, that doesn't happen: When the `product_slug` doesn't match anything we expect, we return only `false`, which is handled by WP as if the request failed (and hence we should return what WP already knows about it: nothing). 

Technically, WPJM already has protections in place to guarantee that this method won't even be called in case we're not handling a WPJM addon (by looking at the `WPJM-Product` on the plugin header), so I don't think this is a big deal. However, it is worth to keep in mind, anyway

gikaragia (approved):
:shipit:

PR #2593, opened 2023-09-26
Add/usage tracking improvements

### LINKED ISSUE #2580: Improve usage tracking on installation
We should change the way that we currently as for users to enable usage tracking in the following ways:
- On the setup wizard, have the checkbox checked by default. This will enable usage tracking regardless if the user clicks ‘Start setup’ or ‘Skip setup’. It will not enable usage tracking only if the user unchecks the checkbox. We can look into updating the text on the buttons to make it more clear (e.g. Save and Skip)
- On activation do not show the ‘Enable usage tracking’ notice at all.
- For new users, we should display the 'Enable usage tracking' notice one week after installation, if they don't have tracking already enabled.
- For existing users, we should display the 'Enable usage tracking' notice one week after updating to the latest version.
- If the users click 'Disable Usage Tracking' we shouldn't ask them again.

Image

Image

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2580

* Usage tracking checkbox in Setup is now checked by default
* The "Usage Tracking" notice will show up 1 week after the plugin has been activated
* When the plugin is updated, we'll show the "Usage Tracking" notice again, but only one time. If it's dismissed, it won't show up on future updates
* I also added a few `target="_blank"` attributes to some links that did not have them.

### Testing instructions

* Pull this branch to a test site
* Make sure you DON'T have usage tracking enabled
* Go to run a setup on this page:  `/wp-admin/index.php?page=job-manager-setup`
* See that the box is checked by default
* Click `Save and Skip setup` and make sure the option is saved properly
---
* To mimic the 1 week wait, you can edit the time in `delay_tracking_notice` to display the notice.
* To mimic a plugin update, you can edit the `WP_Job_manager::updater()` function so that the condition is true and so that the code runs
* Make sure that the usage tracking prompt shows up after that.

### DIFF
--- a/includes/admin/views/html-admin-setup-step-1.php
+++ b/includes/admin/views/html-admin-setup-step-1.php
@@ -11,6 +11,16 @@ if ( ! defined( 'ABSPATH' ) ) {
 ?>
 <h3><?php esc_html_e( 'Welcome to the Setup Wizard!', 'wp-job-manager' ); ?></h3>
 
+<script type="text/javascript">
+	document.addEventListener('DOMContentLoaded', function() {
+		const usageTrackingEnabled = document.querySelector("input[name='job_manager_usage_tracking_enabled']");
+
+		if (usageTrackingEnabled) {
+			usageTrackingEnabled.checked = true;
+		}
+	});
+</script>
+
 <p><?php echo wp_kses_post( __( 'Thanks for installing <em>WP Job Manager</em>! Let\'s get your site ready to accept job listings.', 'wp-job-manager' ) ); ?></p>
 <p><?php echo wp_kses_post( __( 'This setup wizard will walk you through the process of creating pages for job submissions, management, and listings.', 'wp-job-manager' ) ); ?></p>
 <p>
--- a/includes/admin/views/html-admin-setup-step-1.php
+++ b/includes/admin/views/html-admin-setup-step-1.php
@@ -16,7 +16,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 <p>
 	<?php
 	// translators: Placeholder %s is the path to WPJM documentation site.
-	echo wp_kses_post( sprintf( __( 'If you\'d prefer to skip this and set up your pages manually, our <a href="%s">documentation</a> will walk you through each step.', 'wp-job-manager' ), 'https://wpjobmanager.com/documentation/' ) );
+	echo wp_kses_post( sprintf( __( 'If you\'d prefer to skip this and set up your pages manually, our <a target="_blank" href="%s">documentation</a> will walk you through each step.', 'wp-job-manager' ), 'https://wpjobmanager.com/documentation/' ) );
 	?>
 </p>
 
--- a/includes/admin/views/html-admin-setup-step-1.php
+++ b/includes/admin/views/html-admin-setup-step-1.php
@@ -11,16 +11,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 ?>
 <h3><?php esc_html_e( 'Welcome to the Setup Wizard!', 'wp-job-manager' ); ?></h3>
 
-<script type="text/javascript">
-	document.addEventListener('DOMContentLoaded', function() {
-		const usageTrackingEnabled = document.querySelector("input[name='job_manager_usage_tracking_enabled']");
-
-		if (usageTrackingEnabled) {
-			usageTrackingEnabled.checked = true;
-		}
-	});
-</script>
-
 <p><?php echo wp_kses_post( __( 'Thanks for installing <em>WP Job Manager</em>! Let\'s get your site ready to accept job listings.', 'wp-job-manager' ) ); ?></p>
 <p><?php echo wp_kses_post( __( 'This setup wizard will walk you through the process of creating pages for job submissions, management, and listings.', 'wp-job-manager' ) ); ?></p>
 <p>
--- a/includes/admin/views/html-admin-setup-step-1.php
+++ b/includes/admin/views/html-admin-setup-step-1.php
@@ -30,13 +30,15 @@ if ( ! defined( 'ABSPATH' ) ) {
 	?>
 </p>
 
-<form method="post" action="<?php echo esc_url( add_query_arg( 'step', 2 ) ); ?>">
+<form method="post">
 	<input type="hidden" name="nonce" value="<?php echo esc_attr( wp_create_nonce( 'enable-usage-tracking' ) ); ?>" />
 
 	<?php $this->maybe_output_opt_in_checkbox(); ?>
 
 	<p class="submit">
-		<input type="submit" value="<?php esc_html_e( 'Start setup', 'wp-job-manager' ); ?>" class="button button-primary" />
-		<a href="<?php echo esc_url( add_query_arg( 'skip-job-manager-setup', 1, admin_url( 'index.php?page=job-manager-setup&step=3' ) ) ); ?>" class="button"><?php esc_html_e( 'Skip setup. I will set up the plugin manually.', 'wp-job-manager' ); ?></a>
+		<input type="submit" name="start-setup" value="<?php esc_html_e( 'Start setup', 'wp-job-manager' ); ?>" class="button button-primary" formaction="<?php echo esc_url( add_query_arg( 'step', 2 ) ); ?>" />
+
+		<input type="submit" name="skip-setup" value="<?php esc_html_e( 'Save and skip setup', 'wp-job-manager' ); ?>" class="button" formaction="<?php echo esc_url( add_query_arg( 'step', 3 ) ); ?>">
+
 	</p>
 </form>
--- a/includes/class-wp-job-manager-install.php
+++ b/includes/class-wp-job-manager-install.php
@@ -34,6 +34,8 @@ class WP_Job_Manager_Install {
 			$is_new_install = true;
 		}
 
+		require_once __DIR__ . '/../lib/usage-tracking/class-wp-job-manager-usage-tracking-base.php';
+
 		// On new installs display the usage tracking notice with one week delay and for existing installs display it right away.
 		if ( false === get_option( WP_Job_Manager_Usage_Tracking_Base::DISPLAY_ONCE_OPTION ) ) {
 			$time_to_show_notice = $is_new_install ? time() + WEEK_IN_SECONDS : time() - 10;
--- a/includes/class-wp-job-manager-install.php
+++ b/includes/class-wp-job-manager-install.php
@@ -34,6 +34,12 @@ class WP_Job_Manager_Install {
 			$is_new_install = true;
 		}
 
+		// On new installs display the usage tracking notice with one week delay and for existing installs display it right away.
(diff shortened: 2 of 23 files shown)


includes/admin/views/html-admin-setup-step-1.php
```diff
@@ -11,22 +11,34 @@
 ?>
 <h3><?php esc_html_e( 'Welcome to the Setup Wizard!', 'wp-job-manager' ); ?></h3>
 
+<script type="text/javascript">
```
gikaragia: I don't understand why we don't do this in PHP directly. Changed that in 00ba0f023fd6ecbedda56cb3d2e2152514452246

@mikeyarce let me know if there is something that I'm missing here.

includes/class-wp-job-manager.php
```diff
@@ -597,4 +603,15 @@ public function include_admin_files() {
...
 	}
+
+	/**
+	 * Sets a transient to track the activation time of WP Job Manager plugin.
+	 *
+	 * @return void
+	 *
+	 * @since $$next-version$$
+	 */
+	public function set_activation_time() {
```
gikaragia: I think that creating a function that has a single statement makes code a bit harder to read, especially when this function is called only once. You can argue that sometimes when a single statement is very complicated, it makes it easier to understand it if you give it a name with a function but I don't think this is the case here.

includes/class-wp-job-manager.php
```diff
@@ -597,4 +603,15 @@ public function include_admin_files() {
...
+		set_transient( 'job_manager_activation_time', time() );
```
gikaragia: Transients should be thought of as a cache that can be deleted any time. In our case this won't be a big issue but I think that having an option for that will be better.

gikaragia (commented):
Since you are on Support Rotation this week Mikey, I'll pick up the changes for this PR. I just left the comments for documentation purposes.

lib/usage-tracking/class-usage-tracking-base.php
```diff
@@ -451,6 +451,21 @@ protected function opt_in_dialog_text_allowed_html() {
...
+	/**
+	 * Check if the plugin has been active for more than a week.
+	 *
+	 * @return bool Returns true if the plugin has been active for more than a week, false otherwise.
+	 *
+	 * @since $$next-version$$
+	 */
+	function delay_tracking_notice() {
+		$activation_time = get_transient( 'job_manager_activation_time' );
+		if ( $activation_time && ( time() - $activation_time ) > 604800 ) {
```
gikaragia: WP defines a few useful constants for these:
```
	define( 'MINUTE_IN_SECONDS', 60 );
	define( 'HOUR_IN_SECONDS', 60 * MINUTE_IN_SECONDS );
	define( 'DAY_IN_SECONDS', 24 * HOUR_IN_SECONDS );
	define( 'WEEK_IN_SECONDS', 7 * DAY_IN_SECONDS );
	define( 'MONTH_IN_SECONDS', 30 * DAY_IN_SECONDS );
	define( 'YEAR_IN_SECONDS', 365 * DAY_IN_SECONDS );
```

gikaragia:
Hey @fjorgemota since the rest of the team is in support rotation, can you take a look at these changes? The testing instructions should be updated as:
- To simulate a week passing you can update `job_manager_display_usage_tracking_once` to a small number.
- To simulate a new plugin installation, you can delete `wp_job_manager_version`.
- Any changes happen on plugin activation so to test you should: 1) Deactivate WPJM 2) Do any changes to the db 3) Re-enable the plugin.

includes/admin/views/html-admin-setup-step-1.php
```diff
@@ -16,17 +16,19 @@
...
 	<input type="hidden" name="nonce" value="<?php echo esc_attr( wp_create_nonce( 'enable-usage-tracking' ) ); ?>" />
 
 	<?php $this->maybe_output_opt_in_checkbox(); ?>
 
 	<p class="submit">
-		<input type="submit" value="<?php esc_html_e( 'Start setup', 'wp-job-manager' ); ?>" class="button button-primary" />
-		<a href="<?php echo esc_url( add_query_arg( 'skip-job-manager-setup', 1, admin_url( 'index.php?page=job-manager-setup&step=3' ) ) ); ?>" class="button"><?php esc_html_e( 'Skip setup. I will set up the plugin manually.', 'wp-job-manager' ); ?></a>
+		<input type="submit" name="start-setup" value="<?php esc_html_e( 'Start setup', 'wp-job-manager' ); ?>" class="button button-primary" formaction="<?php echo esc_url( add_query_arg( 'step', 2 ) ); ?>" />
+
+		<input type="submit" name="skip-setup" value="<?php esc_html_e( 'Save and skip setup', 'wp-job-manager' ); ?>" class="button" formaction="<?php echo esc_url( add_query_arg( 'step', 3 ) ); ?>">
```
fjorgemota, line 31: TIL about `formaction`. Nice approach!

lib/usage-tracking/class-usage-tracking-base.php
```diff
@@ -430,6 +430,14 @@ protected function hide_tracking_opt_in() {
 	 * @return bool true if the opt-in is hidden, false otherwise.
 	 **/
 	protected function is_opt_in_hidden() {
+		$delayed_notice_timestamp  = (int) get_option( 'job_manager_display_usage_tracking_once' );
```
fjorgemota: Can we maybe move this option name to a constant somewhere?

I see it being used at least 4 times just on this PR, and I think using constants can really help to avoid mistakes/typos/etc.

lib/usage-tracking/class-usage-tracking-base.php
```diff
@@ -430,6 +430,14 @@ protected function hide_tracking_opt_in() {
...
+
+		// Display only once they delayed notice regardless if the user has declined in the past.
+		if ( $delayed_notice_timestamp > 0 && $delayed_notice_timestamp < time() ) {
+			update_option('job_manager_display_usage_tracking_once', 0 );
```
fjorgemota: Usually WP complains about missing spaces on function calls like this.

I'm not sure why it didn't complain here, but I'm adding the suggestion anyway:

```suggestion
			update_option( 'job_manager_display_usage_tracking_once', 0 );
```

includes/class-wp-job-manager-install.php
```diff
@@ -34,6 +34,12 @@ public static function install() {
 			$is_new_install = true;
 		}
 
+		// On new installs display the usage tracking notice with one week delay and for existing installs display it right away.
+		if ( false === get_option( 'job_manager_display_usage_tracking_once' ) ) {
+			$time_to_show_notice = $is_new_install ? time() + WEEK_IN_SECONDS : time() - 10;
```
fjorgemota, line 41: Not really sure if this simplifies the code but..WDYT?

```suggestion
			$time_to_show_notice = time()  + ( $is_new_install ? WEEK_IN_SECONDS : - 10 );
```

lib/usage-tracking/class-usage-tracking-base.php
```diff
@@ -430,6 +430,14 @@ protected function hide_tracking_opt_in() {
 	 * @return bool true if the opt-in is hidden, false otherwise.
 	 **/
 	protected function is_opt_in_hidden() {
+		$delayed_notice_timestamp  = (int) get_option( 'job_manager_display_usage_tracking_once' );
+
+		// Display only once they delayed notice regardless if the user has declined in the past.
```
fjorgemota: Typo

```suggestion
		// Display only once the delayed notice regardless if the user has declined in the past.
```

fjorgemota (commented):
This is looking generally good. I'll test it further tomorrow.

From what I'm seeing, when I deactivate and then activate, the notice looks to be appearing even when the option `job_manager_display_usage_tracking_once` is in the future (with value `1696893421`), after removing the option `wp_job_manager_version`. I'm not sure if I'm doing something wrong, but I'll test more tomorrow.

includes/class-wp-job-manager-install.php
gikaragia: Hmm I think that this version is harder to read IMO

lib/usage-tracking/class-usage-tracking-base.php
gikaragia: It seems that we weren't running phpcs in the lib directory (maybe initially it was meant to only put libraries in there?). I added this dir, run phpcbf, added a couple of rules and updated existing code.

lib/usage-tracking/class-usage-tracking-base.php
gikaragia: c10483031a0db03348a7c81595de76cfb8d80a68

lib/usage-tracking/class-usage-tracking-base.php
gikaragia: Fixed in 5c49e009f1e9c4fae200dc5c1076409f0e9f8a0e

gikaragia:
> the notice looks to be appearing even when the option job_manager_display_usage_tracking_once is in the future

Well spotted! The logic was flawed, fixed in cc81563cd07569fadc9d178f7d5ba2aff73208c8

includes/class-wp-job-manager-post-types.php
```diff
@@ -584,8 +584,8 @@ public function job_feed() {
 			'ignore_sticky_posts' => 1,
 			'posts_per_page'      => $input_posts_per_page,
 			'paged'               => absint( get_query_var( 'paged', 1 ) ),
-			'tax_query'           => [],
-			'meta_query'          => [],
+			'tax_query'           => [], // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query -- Empty.
+			'meta_query'          => [], //// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- Empty.
```
fjorgemota: I think we need just a single `//` right?

```suggestion
			'meta_query'          => [], // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- Empty.
```

includes/class-wp-job-manager-usage-tracking.php
```diff
@@ -182,7 +182,9 @@ protected function get_event_prefix() {
 	}
 
 	/**
-	 * Get the text domain used in the plugin.
+	 * Get the text domain used in the plugin. Deprecated - use 'wp-job-manager' directly.
+	 *
+	 * @deprecated 1.41.1
```
fjorgemota: I believe we should use `$$next-version$$` here, no?

```suggestion
	 * @deprecated $$next-version$$
```

fjorgemota (approved):
Wow! Thanks a lot for making these changes. 

I reviewed them all and now the feature works properly. Also tried updating the option `job_manager_display_usage_tracking_once` to a value generated by `strtotime('+1 minute');` and I could confirm that now the feature works beautifully.

Also, the linting changes look great.

Thanks a lot! Approved! 

(I still left two very minor comments with suggestions, but I believe you can consider the PR as approved after the changes)

gikaragia:
There are some issues unrelated to this PR. I'll go ahead with merging this one and fix them as part of a separate PR

PR #2605, opened 2023-10-05
Do not check for plugin updates when there are no plugins

jom:

* Only sends plugin update requests when there are first-party plugins installed

### Testing Instructions

* Install Query Monitor. Test with no other WPJM plugins installed (activated or not).
* Clear all network/site transients.
* Ensure requests aren't made to https://wpjobmanager.com/wp-json/wpjmcom-licensing/v1/updates
* Add a first party add on.
* Observe a request is made to the `/updates` endpoint.

----

| Plugin build for 678169ca88f8bbb165d45a5f9649e488481b47d6 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/10/wp-job-manager-zip-2605-678169ca.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/10/2605-678169ca)             |

### DIFF
--- a/includes/helper/class-wp-job-manager-helper.php
+++ b/includes/helper/class-wp-job-manager-helper.php
@@ -250,7 +250,11 @@ class WP_Job_Manager_Helper {
 	 */
 	public function check_for_updates( $check_for_updates_data ) {
 		$installed_plugins = $this->get_installed_plugins( false, true );
-		$updates           = $this->get_plugin_update_info( $installed_plugins );
+		if ( empty( $installed_plugins ) ) {
+			return $check_for_updates_data;
+		}
+
+		$updates = $this->get_plugin_update_info( $installed_plugins );
 
 		$notice_data = [];
 


fjorgemota (commented):
The code looks good, but, where (e.g. on which pages) should this code be running?

I see that on the `/wp-admin/update-core.php?force-check=1` page the `update_plugins` transient is being updated, but it doesn't seem to be triggering the hook `pre_set_site_transient_update_plugins`, although the addon is correctly appearing on the list of plugins to be updated... :thinking: 

Testing on PHP 8.2.10 and WP 6.3.1

fjorgemota:
Nevermind the comment above, I just found out that `wp transient delete --all` deletes only the site transients, while the ones used by this algorithm are related to the "network", so we need to run `wp transient delete --network --all` instead.

And the weird thing is that I'm testing on a non-multisite WP installation, but oh well. I'll test further and update the review =)

fjorgemota (approved):
Looks good and works well!

jom:
Thanks, @fjorgemota!

PR #2607, opened 2023-10-05
Update readme text

yscik:

* Update copy
* Include the new promote jobs feature
* Include the jobs.blog demo site
* Remove unit testing section
* Fix changelog section headings

### Release Notes

----

| Plugin build for 1cde695c714bba617a1e11677f965ab49dff1ac8 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/11/wp-job-manager-zip-2607-1cde695c.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/11/2607-1cde695c)             |


readme.txt
```diff
@@ -131,16 +136,7 @@ If you wish to be notified of new postings on your site you can use a plugin suc
...
-
-The plugin contains all the files needed for running tests.
-Developers who would like to run the existing tests or add their tests to the test suite and execute them will have to follow these steps:
-1. `cd` into the plugin directory.
-2. Run the install script(you will need to have `wget` installed) - `bash tests/bin/install-wp-tests.sh <db-name> <db-user> <db-pass> <db-host> <wp-version>`.
-3. Run the plugin tests - `phpunit`
-
-The install script installs a copy of WordPress in the `/tmp` directory along with the WordPress unit testing tools.
-It then creates a database based on the parameters passed to it.
+== Changelog ==
```
yscik: I'm hoping this will add a Changelog header before the changelog here:

image

readme.txt
```diff
@@ -131,16 +136,7 @@ If you wish to be notified of new postings on your site you can use a plugin suc
...
-== Unit Testing ==
-
-The plugin contains all the files needed for running tests.
-Developers who would like to run the existing tests or add their tests to the test suite and execute them will have to follow these steps:
-1. `cd` into the plugin directory.
-2. Run the install script(you will need to have `wget` installed) - `bash tests/bin/install-wp-tests.sh <db-name> <db-user> <db-pass> <db-host> <wp-version>`.
-3. Run the plugin tests - `phpunit`
-
-The install script installs a copy of WordPress in the `/tmp` directory along with the WordPress unit testing tools.
-It then creates a database based on the parameters passed to it.
```
fjorgemota, line 143: WDYT about moving this to a separate file (or maybe even a page on wpjobmanager.com, but I believe a file is better for this) and link it here?

yscik: 👍 We can add it to README.md to show it on GitHub, or a CONTRIBUTING.md — this is for developers, not users.

Not sure why was there a random section about unit testing in the plugin readme, without the rest of the developer-focused docs on how to build and work with the plugin. (Though those actually doesn't seem to exist?)

gikaragia (commented):
You can use this tool if you want to try out reformating the readme file: https://wpreadme.com/

As for the current state of the file there is an issue with the changelog. It seems that for the changelog part, the `##` markdown doesn't work. This causes the changelog to move to the plugin description instead of the 'Development' tab where it should be. I think we should do the following:
1. Use a format like the following:
```
== Changelog ==

= 1.42.0 - 2023-10-04 =
* Fix: Fix my-alerts.php template HTML
* Fix: Make 'Add alert' link relative
* Fix: Fix redirection after actions on My alerts page
```
2. In the end of the changelog use something like this (stole the idea from Yoast)
```
= Earlier versions =
For the changelog of earlier versions, please refer to [the changelog on wpjobmanager.com](https://wpjobmanager.com/release-notes/?job-manager-product=wp-job-manager&job-manager-release-year=).
```

yscik:
Oh, good catch! Looks like it works with `#` too, Sensei uses those, but the versions need to be a 3rd level heading. I'll check which format works both for readme.txt and changelog.txt.

gikaragia (approved):
👌

PR #2615, opened 2023-10-10
Fix plugin zip workflow

yscik:

* Use zx to escape shell commands better

### Testing Instructions

* Add `code` and hope it doesn't break
* See Plugin Build / Plugin Build (pull_request) check

----

| Plugin build for 027ab53c0fc1c39339d28fde3b7c3e2e7e11fa5e <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/10/wp-job-manager-zip-2615-027ab53c.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/10/2615-027ab53c)             |


fjorgemota (approved):
Looks good and works well!

Note: I was going to ask if we should add `zx` to `package.json`, only to find out that it was added on #2604 :eyes:

PR #2627, opened 2023-10-28
Upsells: Check for Applications/Resumes plugin installs directly

yscik:

* Change the approach from filter to look for the `wp_job_manager_applications_version`/`wp_resume_manager_version` directly when deciding if an upsell page should be displayed for those addons
* Don't show the upsells if those options are set, meaning the addons are or were already installed
* This covers a few cases where upsells would be confusingly/annoyingly displayed:
  * Running an older version of the addons would still show an upsells screen
  * De-activating the addons would bring back the upsells screens

### Testing Instructions

* Use this branch with a test site
* Confirm the upsells are showing
* Install & activate the latest released version of the applications/resumes plugin on the test site
* Confirm the upsells are not showing

### Screenshot / Video

----

| Plugin build for fa95833ff51bb2371b56eb84e2d80b6349e32094 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/11/wp-job-manager-zip-2627-fa95833f.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/11/2627-fa95833f)             |

### DIFF
--- a/includes/admin/class-wp-job-manager-addons-landing-page.php
+++ b/includes/admin/class-wp-job-manager-addons-landing-page.php
@@ -69,8 +69,8 @@ class WP_Job_Manager_Addons_Landing_Page {
 	 * @since $$next-version$$
 	 */
 	public function add_applications_menu_item() {
-		$was_addon_installed = ! ! get_option( 'wp_job_manager_applications_version', false );
-		$was_page_dismissed  = ! ! get_option( 'job_manager_addon_upsell_applications', false );
+		$was_addon_installed = (bool) get_option( 'wp_job_manager_applications_version', false );
+		$was_page_dismissed  = (bool) get_option( 'job_manager_addon_upsell_applications', false );
 
 		/**
 		 * Filters whether the 'Applications' landing page should be added to the Job Manager menu.
--- a/includes/admin/class-wp-job-manager-addons-landing-page.php
+++ b/includes/admin/class-wp-job-manager-addons-landing-page.php
@@ -46,7 +46,8 @@ class WP_Job_Manager_Addons_Landing_Page {
 	 * @since $$next-version$$
 	 */
 	public function __construct() {
-		add_action( 'admin_menu', [ $this, 'add_menu_items' ], 12 );
+		add_action( 'admin_menu', [ $this, 'add_applications_menu_item' ], 12 );
+		add_action( 'admin_menu', [ $this, 'add_resumes_menu_item' ], 12 );
 		add_action( 'admin_enqueue_scripts', [ $this, 'register_styles' ] );
 	}
 
@@ -60,16 +61,16 @@ class WP_Job_Manager_Addons_Landing_Page {
 		WP_Job_Manager::register_style( 'job_manager_admin_landing_css', 'css/admin-landing.css', [ 'job_manager_brand' ] );
 	}
 
+
 	/**
-	 * Add add-on menu items if needed.
+	 * Add Applications add-on menu item if needed.
 	 *
 	 * @access private
 	 * @since $$next-version$$
 	 */
-	public function add_menu_items() {
-
-		$parent_page = 'edit.php?post_type=job_listing';
-		$badge_text  = __( 'Pro', 'wp-job-manager' );
+	public function add_applications_menu_item() {
+		$was_addon_installed = ! ! get_option( 'wp_job_manager_applications_version', false );
+		$was_page_dismissed  = ! ! get_option( 'job_manager_addon_upsell_applications', false );
 
 		/**
 		 * Filters whether the 'Applications' landing page should be added to the Job Manager menu.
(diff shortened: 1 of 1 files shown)


includes/admin/class-wp-job-manager-addons-landing-page.php
```diff
@@ -60,16 +61,16 @@ public function register_styles() {
...
 	 *
 	 * @access private
 	 * @since $$next-version$$
 	 */
-	public function add_menu_items() {
-
-		$parent_page = 'edit.php?post_type=job_listing';
-		$badge_text  = __( 'Pro', 'wp-job-manager' );
+	public function add_applications_menu_item() {
+		$was_addon_installed = ! ! get_option( 'wp_job_manager_applications_version', false );
```
pgk: Nothing wrong with how it is, I 've seen it tackled like this more often:
```suggestion
		$was_addon_installed = (bool) get_option( 'wp_job_manager_applications_version', false );
```

includes/admin/class-wp-job-manager-addons-landing-page.php
```diff
@@ -60,16 +61,16 @@ public function register_styles() {
...
+		$was_page_dismissed  = ! ! get_option( 'job_manager_addon_upsell_applications', false );
```
pgk: same
```suggestion
		$was_page_dismissed  = (bool) get_option( 'job_manager_addon_upsell_applications', false );
```

includes/admin/class-wp-job-manager-addons-landing-page.php
```diff
@@ -78,17 +79,32 @@ public function add_menu_items() {
...
+		if ( apply_filters( 'job_manager_addon_upsell_applications', ! $was_addon_installed && ! $was_page_dismissed ) ) {
 
 			add_submenu_page(
-				$parent_page,
+				'edit.php?post_type=job_listing',
 				__( 'WP Job Manager - Applications', 'wp-job-manager' ),
-				sprintf( '%s <span class="awaiting-mod wpjm-addon-upsell__badge">%s</span>', __( 'Applications', 'wp-job-manager' ), $badge_text ),
+				sprintf(
+					'%s <span class="awaiting-mod wpjm-addon-upsell__badge">%s</span>',
+					__( 'Applications', 'wp-job-manager' ),
```
pgk, line 89: I think that we need to escape or sanitize this, as it may be an attack vector (
```suggestion
					esc_html__( 'Applications', 'wp-job-manager' ),
```
Or:
```suggestion
					sanitize_text_field( __( 'Applications', 'wp-job-manager' ) ),
```

includes/admin/class-wp-job-manager-addons-landing-page.php
```diff
@@ -97,18 +113,21 @@ public function add_menu_items() {
...
+		if ( apply_filters( 'job_manager_addon_upsell_resumes', ! $was_addon_installed && ! $was_page_dismissed ) ) {
 
 			add_submenu_page(
-				$parent_page,
+				'edit.php?post_type=job_listing',
 				__( 'WP Job Manager - Resumes', 'wp-job-manager' ),
-				sprintf( '%s <span class="awaiting-mod wpjm-addon-upsell__badge">%s</span>', __( 'Resumes', 'wp-job-manager' ), $badge_text ),
+				sprintf(
+					'%s <span class="awaiting-mod wpjm-addon-upsell__badge">%s</span>',
+					__( 'Resumes', 'wp-job-manager' ),
```
pgk, line 123: same as the above translations I think.

pgk (changes requested):
Worked great! I can see the "Pro" features gone if plugins are present!

I have some thoughts on escaping translated strings, which can be addressed several ways, leaving it up to you.

Screenshot 2023-11-01 at 11 08 27 AM

includes/admin/class-wp-job-manager-addons-landing-page.php
yscik: Hm looks like it's not usually escaped in our plugin (or [WC](https://github.com/search?q=repo%3AAutomattic%2Fwoocommerce%20add_submenu_page&type=code)), but wouldn't hurt. It goes through `wptexturize` but doesn't seem to be escaped anywhere before rendering.

I'll open a dedicated PR to update these in in the whole plugin.

includes/admin/class-wp-job-manager-addons-landing-page.php
yscik: Yeah that's probably more readable.

includes/admin/class-wp-job-manager-addons-landing-page.php
```diff
@@ -46,7 +46,8 @@ public static function instance() {
 	 * @since $$next-version$$
 	 */
 	public function __construct() {
-		add_action( 'admin_menu', [ $this, 'add_menu_items' ], 12 );
+		add_action( 'admin_menu', [ $this, 'add_applications_menu_item' ], 12 );
+		add_action( 'admin_menu', [ $this, 'add_resumes_menu_item' ], 12 );
```
thedebian, line 50:
```suggestion
		add_action( 'admin_menu', [ $this, 'add_menu_items' ], 12 );
```
Not a blocker: Instead of having two actions there, could we have just one, calling add_menu_items, then add_menu_items calls both methods ?
Both methods could also be set as private then.
It might feel a bit clearer and easier to maintain

yscik: I started out with that, but eliminated that method since it wasn't adding much value, just calling the two other ones each doing their own thing. Seemed a bit better to have two more granular hooks, but I don't have strong opinions about it.

For these cases where the method has to be `public` for hooks but otherwise is not really a public method, we add the `@access private` annotation to indicate that it shouldn't be used directly.

pgk (approved):
Since we are going to be addressing the escaping separately, the rest looks good!

PR #2633, opened 2023-11-11
Add Salary Range Field & JobSchema to WPJM

### LINKED ISSUE #2629: Add Salary Range support & add to Google Schema
As i saw on #2263 the salary schema was added. For now it only supports one salary amount but no range.
This was added in the PR as further enhancement but no task was created or followed up as far as i see, right @onubrooks?

Ranges should also be supported asap. I would suppose the following. For all current installations we use the _job_salary Field same as it is right now. It will generate a schema like this:

```
"baseSalary": {
  "@type": "MonetaryAmount",
  "currency": "USD",
  "value": {
    "@type": "QuantitativeValue",
    "value": 40.00,
    "unitText": "HOUR"
  }
}
```

Additionally we'll add a second field for _job_salary_max. if this field is also filled out we'll generate schema like this:

```
"baseSalary": {
  "@type": "MonetaryAmount",
  "currency": "USD",
  "value": {
    "@type": "QuantitativeValue",
    "minValue": 40.00,
    "maxValue": 50.00,
    "unitText": "HOUR"
  }
}
```

It respects the Structure of google requested here: https://developers.google.com/search/docs/appearance/structured-data/job-posting

-

masteradhoc:
Fixes #2629

* Adds a max salary field to the frontend and admin forms.
* Adds a max salary information to Google's Job Search Structured Data

### Testing Instructions

  * 1) add only one salary number (in min field)
  * 2) add both min salary and max salary
* Copy the output HTML and paste here - https://search.google.com/test/rich-results

### Release Notes

* Add: WPJM now supportes Salary Ranges and includes them into the job schema

### DIFF
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -1377,7 +1377,7 @@ function the_job_salary( $before = '', $after = '', $echo = true, $post = null )
 	 * @param string  $unit
 	 * @param string  $after
 	 */
-	$job_salary = apply_filters( 'the_job_salary_message', $job_salary, $post, $before, $salary, $currency, $salarymax, $unit, $after );
+	$job_salary = apply_filters( 'the_job_salary_message', $job_salary, $post, $before, $salary, $currency, $unit, $after, $salarymax );
 
 	if ( $echo ) {
 		echo esc_html( $job_salary );
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -418,10 +418,10 @@ function wpjm_get_job_listing_structured_data( $post = null ) {
 		}
 	}
 
-	$salary       = get_the_job_salary( $post );
-	$salary_max   = get_the_job_salary_max( $post );
-	$currency     = get_the_job_salary_currency( $post );
-	$unit         = get_the_job_salary_unit( $post );
+	$salary     = get_the_job_salary( $post );
+	$salary_max = get_the_job_salary_max( $post );
+	$currency   = get_the_job_salary_currency( $post );
+	$unit       = get_the_job_salary_unit( $post );
 	if ( ! empty( $salary_max ) ) {
 		$data['baseSalary']                      = [];
 		$data['baseSalary']['@type']             = 'MonetaryAmount';
@@ -1345,11 +1345,11 @@ function get_the_job_salary_max( $post = null ) {
  * @return string|void
  */
 function the_job_salary( $before = '', $after = '', $echo = true, $post = null ) {
-	$post     = get_post( $post );
-	$salary   = get_the_job_salary( $post );
+	$post      = get_post( $post );
+	$salary    = get_the_job_salary( $post );
 	$salarymax = get_the_job_salary_max( $post );
-	$currency = get_the_job_salary_currency( $post );
-	$unit     = get_the_job_salary_unit_display_text( $post );
+	$currency  = get_the_job_salary_currency( $post );
+	$unit      = get_the_job_salary_unit_display_text( $post );
 
 	if ( strlen( $salary ) === 0 ) {
 		return;
--- a/wp-job-manager-template.php
+++ b/wp-job-manager-template.php
@@ -1323,11 +1323,11 @@ function get_the_job_salary_max( $post = null ) {
 	$job_salary_max = $post->_job_salary_max;
 
 	/**
-	 * Filter the returned job salary.
+	 * Filter the returned max job salary.
 	 *
 	 * @since 1.36.0
 	 *
-	 * @param string  $job_salary
+	 * @param string  $job_salary_max
 	 * @param WP_Post $post
 	 */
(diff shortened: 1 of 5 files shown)


masteradhoc:
@fjorgemota @mikeyarce @onubrooks asking for a review on this as you worked last on the salary area :)

mikeyarce:
Thanks for the PR @masteradhoc !
@yscik what do you think about this one?

The code implementation looks good - but my concern is with the labelling of the fields:
Screenshot 2023-11-13 at 2 44 13 PM

`Salary / Salary Min` and `Salary Max` is confusing to me.  I would either just want `Salary` OR `Salary Min` and `Salary Max`.   What if we made it an option so that you can decide if you want a simple Salary or a Salary Range?

Overall, I think it would be great to have the option for Min/Max salary.

wp-job-manager-template.php
```diff
@@ -1332,11 +1372,12 @@ function the_job_salary( $before = '', $after = '', $echo = true, $post = null )
 	 * @param WP_Post $post
 	 * @param string  $before
 	 * @param string  $salary
+	 * @param string  $salarymax
 	 * @param string  $currency
 	 * @param string  $unit
 	 * @param string  $after
 	 */
-	$job_salary = apply_filters( 'the_job_salary_message', $job_salary, $post, $before, $salary, $currency, $unit, $after );
+	$job_salary = apply_filters( 'the_job_salary_message', $job_salary, $post, $before, $salary, $currency, $salarymax, $unit, $after );
```
thedebian: This might break existing integrations

masteradhoc: @thedebian what do you suggest?

masteradhoc:
> Thanks for the PR @masteradhoc ! @yscik what do you think about this one?
> 
> The code implementation looks good - but my concern is with the labelling of the fields: Screenshot 2023-11-13 at 2 44 13 PM
> 
> `Salary / Salary Min` and `Salary Max` is confusing to me. I would either just want `Salary` OR `Salary Min` and `Salary Max`. What if we made it an option so that you can decide if you want a simple Salary or a Salary Range?
> 
> Overall, I think it would be great to have the option for Min/Max salary.

Quite open to that. Just didnt wanted to add another setting for this as its quite common to have some jobs with no range (f.e. intern/trainee) and some with range. How would you like this to be changed?

i think also this could make sense:
- Salary Min.
- Salary Max.

wp-job-manager-template.php
thedebian: Good question, I don't have an empirical solution to that, it can be anything from adding the param after $after, to changing $salary to be a concatenation of $min . '-' . $max, if it's a salary range .
I'd prefer the second option, as $salary is documented as string, it wouldn't break the contract, @gkaragia do you have an idea ?

gikaragia: I think appending it as an extra argument in the end is the safest. 

If we don't add it as an extra argument and concatenate instead, this will break existing sites that use the `$after` function argument or specify a `$unit`. In these cases the job salary mind end up with weird values in the filter.

masteradhoc:
@gkaragia @thedebian any option to get this reviewed again? :)

gikaragia:
Hey @masteradhoc, as @mikeyarce suggested above I think that we should have separate fields for the salary range and actual salary. 

I think that the most versatile approach would be to have all three fields available, Salary, Salary Min, and Salary Max. Then we don't show the Salary field if both Salary Min and Salary Max are defined. @mikeyarce and @yscik what do you think about that?

masteradhoc:
@gkaragia sounds like an option for me.
From User Experience View im not sure if its though a good idea. What if you started with a regular salary number as WPJM didnt had the range feature until now and would like to upgrade to ranges now? 

You would have to turn it on and loose all of your existing values or copy it upfront and then reenter it again. The solution like this is designed that it works even for users that are upgrading to ranges or only using them sometimes..

gikaragia:
Hey @masteradhoc, I am not suggesting adding an option for this. After having some second thoughts we should do the following:
- Add a checkbox called 'Enable Salary Range' in the job itself, not WPJM settings.
- This checkbox when clicked, will hide the existing 'Salary' field and show 'Salary Min' and 'Salary Max'.
- When the checkbox is enabled and the listing is updated, salary range will be used for this job instead.

If you don't have the capacity to implement the above, let us know and we will consider implementing it in a future release.

gikaragia:
Just as an update @masteradhoc we will look into implementing this in the future.

gikaragia:
We have decided that we will rework this in the near future so switching this to draft.

masteradhoc:
@thedebian Giannis said i should contact you about this PR. Any movement? Please feel free to make the edits needed to get this into the next version.

thedebian:
> @thedebian Giannis said i should contact you about this PR. Any movement? Please feel free to make the edits needed to get this into the next version.

Hey @masteradhoc, it's in our backlog, for now we can't provide an ETA for it though, we'll let you know when we have an acceptable solution for this feature.

PR #2637, opened 2023-11-13
Align actions in notices in the center

gikaragia:
Before
Screenshot 2023-11-13 at 15 32 39

After
Screenshot 2023-11-13 at 15 31 37

* A small fix which aligns in the center the actions in admin notices

### Testing Instructions

* Load a notice and observe that the actions in the notice are now center-aligned.

----

| Plugin build for b3e0505f8532f3d11bc9743d9c2c8f86c69480a0 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/11/wp-job-manager-zip-2637-b3e0505f.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/11/2637-b3e0505f)             |

### DIFF
--- a/assets/css/admin-notices.scss
+++ b/assets/css/admin-notices.scss
@@ -39,7 +39,7 @@ $notice-success: #43af99;
 	&__top {
 		display: flex;
 		flex-direction: row;
-		align-items: flex-start;
+		align-items: center;
 		flex-wrap: wrap;
 		gap: 20px;
 	}
--- a/assets/css/admin-notices.scss
+++ b/assets/css/admin-notices.scss
@@ -25,6 +25,7 @@ $notice-success: #43af99;
 
 	gap: 10px;
 	padding: 20px 24px;
+	padding-right: calc(48px + 24px);
 	transform-origin: top center;
 	overflow: hidden;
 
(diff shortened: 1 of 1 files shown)


yscik:
Hm that's better, but we should align the dismiss icon too. 
Played around and we can replace its absolute positioning with this:

```
.wpjm-admin-notice button.wpjm-notice-dismiss--icon {
  padding: 6px;
  color: inherit;
  position: static;
  margin-right: -48px;
}
```

(One aspect of this is buttons from different notices look better aligned, and some of them might not have the dismiss action: )

image

gikaragia:
Good point @yscik! I applied your fix in a slightly different way to avoid the negative margin.

yscik:
Yup that's cleaner :) But we need to set extra padding for non-dismissible notices, and the notice content too. This is the alignment I was talking about, when having multiple notices and some is not dismissible:

image

You can try it by setting some notice to `'dismissible' => false` (eg `includes/admin/class-wp-job-manager-admin-notices.php:715`)

I'll drop in a commit with the extra padding.

yscik (approved):
Looks good on my end :)

gikaragia:
Yep I tried that I just thought that it looks nicer if the buttons between dismissable and non-dismissable aren't aligned as now non-dismissable notices have a big gap on the right. Anyways, I think that it looks nice both ways.

PR #2646, opened 2023-11-20
Small fixes to actions and release script

gikaragia:

* Updates actions/checkout to v4 to avoid some warnings
* Adds a default release note 'Updated plugin headers' when the release notes are empty.

### Testing Instructions

* You can run the script locally to test
* Or alternatively you can check https://github.com/Automattic/WP-Job-Manager/pull/2648 to see for the result of running the script.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     name: Plugin Build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Get npm cache directory
         id: npm-cache
         run: |
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     name: WPJM Release
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Comment on PR
         env:
           GITHUB_TOKEN: ${{ github.token }}
--- a/.github/workflows/gardening.yml
+++ b/.github/workflows/gardening.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
      - name: Checkout
-       uses: actions/checkout@v3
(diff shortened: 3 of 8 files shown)


jgcaruso (commented):
Changelog generation code looks good 👍 

I'm not entirely sure what the `actions/checkout@v4` is so I'm reluctant to approve the PR. Hoping someone else familiar with the release script can look at it.

Also, when running the release script I got the following error, but it also happened on Trunk so I think it might just be my environment and not this PR.
SCR-20231122-npqn

gikaragia:
Thank you for taking a look @jgcaruso!

> I got the following error,

Tried it again and it didn't happen. I would start debugging by running the pot build command individually: `npm run i18n:build `

> I'm not entirely sure what the actions/checkout@v4 is 

This action checks out the code. It's a common first step before doing the rest of the stuff like building, running tests etc.
The version upgrade should be backwards compatible. I did the update because v2 uses node 12 IIRC and this version is deprecated by Github. So it just uses node 16 anyway and prints a warning. I did a similar change for [wpjm-addons ](https://github.com/Automattic/wpjm-addons/pull/180 )here.

jgcaruso:
> I would start debugging by running the pot build command individually: `npm run i18n:build`

So running that gives me `wp: command not found`

This is probably just because I'm a fresh install (and maybe didn't real all of the documentation 😅 ), but I wonder if some dependencies should be added, and maybe the `i18n:build` should be part of the release? Or is it mentioned as a pre-release step in some release documentation, so it doesn't need to be automated?

> This action checks out the code. It's a common first step before doing the rest of the stuff like building, running tests etc.

Oh so its a github actions specific thing? Cool!

gikaragia:
Ah yes @jgcaruso! You need [WP-cli](https://make.wordpress.org/cli/handbook/guides/installing/) to be installed and globally available for this command to run. I'll add this to the release documentation.

jgcaruso (approved):
Tested well! 

Once I installed `wp-cli` globally and ran `composer install` the release script worked and created the release PR https://github.com/Automattic/WP-Job-Manager/pull/2656

PR #2657, opened 2023-11-23
Add filter to allow Polylang to translate our job listing categories

mikeyarce:

* Polylang by default won't translate custom taxonomies - this adds a filter to make them translatable

### Testing Instructions

* Install Polylang
* Set a few languages
* Apply this patch!
* Go to Job Manager -> Categories, you should see this:

### Release Notes

*Set our Job Listings Category to be translatable by Polylang by default

----

| Plugin build for 48b85c4a7eeb5b3d0bce52872acd980f8bd224de <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2657-48b85c4a.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2657-48b85c4a)             |

### DIFF
--- a/includes/3rd-party/polylang.php
+++ b/includes/3rd-party/polylang.php
@@ -83,3 +83,23 @@ function polylang_wpjm_doing_ajax( $is_ajax ) {
 	return isset( $_SERVER['REQUEST_URI'] ) && false === strpos( sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ), '/jm-ajax/' ) ? $is_ajax : true;
 }
 add_filter( 'pll_is_ajax_on_front', 'polylang_wpjm_doing_ajax' );
+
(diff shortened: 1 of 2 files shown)


gikaragia (commented):
Seems like there is a wpml-config.xml file which both Polylang and WPML respect. Maybe we should add this in there so we can allow translations in both plugins?

mikeyarce:
Thanks @gkaragia!  I had no idea about `wpml-config.xml` but that is a way better idea.  Done in https://github.com/Automattic/WP-Job-Manager/pull/2657/commits/48b85c4a7eeb5b3d0bce52872acd980f8bd224de

gikaragia (approved):
👌

PR #2659, opened 2023-11-28
Add a new UI system for notices

yscik:
Basis for 145-gh-Automattic/wpjm-addons and similar notices. 

The new notice class has a low-level API for now, I expect we'll add a more semantic layer on top if it, like `success_notice()` etc that has most of the arguments pre-configured.

I marked the classes as internal and private with a note that it's not yet stable.

Also basing it on a feature branch to build it out a bit more in additional PRs before it's ready for release.

* Add new notice styles
* Add PHP classes to generate notices and load styles
* Add a shortcode to preview new UI elements for development and future customizations

### Testing Instructions

* Run/restart `npm run start` or `npm run build` to compile new stylesheets
* Add the `[job_manager_ui_test]` shortcode to a page and view it

### Release Notes

* Add new notice and message styling (frontend)

### Screenshot / Video

image

----

| Plugin build for daf2b7761815c8cf155b66be3e8266f1aaaeffd4 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2659-daf2b776.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2659-daf2b776)             |


includes/class-wp-job-manager-ui-notice.php
```diff
@@ -0,0 +1,213 @@
...
+	 * @var self
+	 */
+	private static $instance = null;
+
+	/**
+	 * Singleton instance getter.
+	 *
+	 * @return self
+	 */
+	public static function instance() {
```
gikaragia: Why does this class need to be a singleton? I would expect to be able to create multiple instances for multiple notices if I wanted to display many notices. This means that the options would go to the constructor and be stored in a private field.

includes/class-wp-job-manager-ui-notice.php
```diff
@@ -0,0 +1,213 @@
...
+	 * @type string $message_icon Alternative to icon, if it should be placed next to the message, not the title.
+	 * @type string $details Additional details.
+	 * @type array  $buttons Action links styled as buttons.
+	 * @type array  $links Action links.
+	 * @type array  $classes Additional classes for the notice.
+	 * }
+	 *
+	 * @return string HTML.
+	 */
+	public static function notice( $options ) {
```
gikaragia: Having a method which has the same name with the class is a bit confusing. Maybe `get_notice_html` would be better?

Also, there is no reason to have this function as static.

gikaragia: Why don't we pass each option as an argument? Plain arguments are easier to document. 

Also, they are easier to maintain as you need to track down how the options array is populated if you want to understand how a method is called.

includes/class-wp-job-manager-ui-notice.php
```diff
@@ -0,0 +1,213 @@
...
+	 * Notice element.
+	 *
+	 * @param array $options {
+	 *    Options for the notice.
+	 *
+	 * @type string $title Optional title.
+	 * @type string $icon Notice icon. Should be a supported icon name, or safe SVG code.
+	 * @type string $message Main notice message.
+	 * @type string $message_icon Alternative to icon, if it should be placed next to the message, not the title.
+	 * @type string $details Additional details.
```
gikaragia: `details` is a bit too generic. Maybe `message` or `additional_html`?

gikaragia (commented):
I've paused my review here to clarify something first. Do we intend to allow users to change the notices? If yes, how do we intend to do that?

If we allow it (which I think we should), maybe it would make more sense to have a normal template for each of the notices? I can see the value of implementing an API however WP and WPJM users are used to make similar changes by overriding templates. I can also imagine a part of our users don't have the technical expertise to extend and use a programming API.

yscik:
> If we allow it (which I think we should), maybe it would make more sense to have a normal template for each of the notices?

Yeah, good idea to create templates for this. It'd be one generic notice template though, I really don't want the API to be "create a new template to add notice". That's a lot of friction and I'm not sure what the benefit would be.

Additionally, this notice is designed to be adaptive from one quick feedback message to a full dialog with actions, and makes sure everything is consistent and has good usability. We'd quickly lose that if we have to build each notice separately. We can include a filter and a notice ID though to customize the options that'll allow this to be extended but still remain consistent. I'll look into adding this.

Generally though most of the visual customization can be done via the CSS variables, they are pretty easy to tweak to match any style. (Eg colors, background, borders, rounded corners.)

Regarding the API, I'm rearranging it a bit, this is what I'm thinking:

- Singleton UI core class managing registering styles/scripts/etc
- Static Notice class, there is nothing to instance really
- To create a notice:
  - `UI\Notice::render( [ options ] )` generates the notice
- A few shortcut methods for common usages, eg:
  - `UI\Notice::success( "message" )` or `UI\Notice::success( [ custom options ] )`
  - `UI\Notice::error( "message" )`

includes/class-wp-job-manager-ui-notice.php
yscik: With this many arguments that's really hard to use and read. This is a pretty common pattern in WP too for complex methods, eg `get_posts`. But maybe in a few years we can use the [PHP 8 named arguments](https://stitcher.io/blog/php-8-named-arguments) :)

includes/class-wp-job-manager-ui-notice.php
yscik: Clarified this in the description.

yscik:
Moved the main HTML to a template and added filters that allow modifying or disabling notices.
Also moved the new code to a `ui/` folder with namespaces and PSR-4 naming, and iterated on the API a bit.

templates/notice.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Notice.
+ *
+ * This template can be overridden by copying it to yourtheme/job_manager/
+ *
+ * @see         https://wpjobmanager.com/document/template-overrides/
+ * @author      Automattic
+ * @package     wp-job-manager
+ * @category    Template
+ * @version     $$next-version
```
gikaragia:
```suggestion
 * @version     $$next-version$$
```

includes/ui/Notice.php
```diff
@@ -0,0 +1,226 @@
...
+	 * Notice element.
+	 *
+	 * @param array $options {
+	 *    Options for the notice.
+	 *
+	 * @type string $id Unique ID of the notice.
+	 * @type string $title Optional title.
+	 * @type string $icon Notice icon. Should be a supported icon name, or safe SVG code.
+	 * @type string $message Main notice message.
+	 * @type string $details Additional HTML content for the notice. Displayed below the message.
```
gikaragia: Since you came up with a much better description, I don't get why shouldn't name it as $additional_html but I don't have strong feelings about it 🙂

assets/css/ui.notice.scss
```diff
@@ -0,0 +1,244 @@
+.jm-notice,
```
gikaragia, line 1: Maybe prefix it with wpjm to be a bit more specific and the same with other classes?

assets/css/ui.neutral.scss
```diff
@@ -0,0 +1,55 @@
+
+/**
+ * Fade current text color to the given opacity.
+ */
+@function currentColor( $opacity ) {
```
gikaragia: Maybe name this `fadeColor` or smth similar? I can see this being confused with the `currentcolor` value

includes/ui/Notice.php
```diff
@@ -0,0 +1,226 @@
...
+ * @access private
+ */
+class Notice {
+
+	/**
+	 * A success message.
+	 *
+	 * @param string|array $args A message, or an array of options accepted by Notice::render.
+	 */
+	public static function success( $args ) {
```
gikaragia: As discussed in my own PR, using static methods everywhere is the same with using a procedural style. I think that we should use normal methods and instances if we want the benefits of OOP. 

There are also plenty of discussions on the matter online, for example [here](https://stackoverflow.com/questions/33705976/when-should-i-use-static-methods) and [here](https://stackoverflow.com/questions/47118926/should-i-avoid-using-too-many-static-methods-in-php).

includes/ui/Settings.php
```diff
@@ -0,0 +1,268 @@
...
+ */
+class Settings {
+
+	use Singleton;
+
+	/**
+	 * Instance constructor
+	 */
+	private function __construct() {
+		add_shortcode( 'job_manager_ui_test', [ $this, 'preview_ui_elements' ] );
```
gikaragia: This makes testing a lot easier ❤️

includes/class-wp-job-manager-ui-notice.php
gikaragia: > With this many arguments that's really hard to use and read

I see your point but encapsulating them in an array doesn't change anything, it just adds another layer. 

Named arguments is another reason not to have an array IMO as the way the function is now will never be able to benefit from it.

gikaragia (commented):
The changes look great! Left some comments

includes/ui/Notice.php
yscik: What would be the benefits of an OOP approach here? The downside is it's more boilerplate code and a harder to use API, replacing `echo Notice::success( $args )` with something like
```php
$notice = new Success_Notice( $args );
echo $notice->render();
```

assets/css/ui.notice.scss
yscik: Looks like we have a mix of prefixes: `wpjm`, `jm`, `job_manager`, `job-manager`, `wp-job-manager` (plus no prefix 😄). 
Since we are starting to brand the plugin just as 'Job Manager', especially when already inside WordPress, I'd go with the `jm` one for the new styles, it's short and easy to remember and say :)

includes/ui/Notice.php
yscik: Just trying to keep this a bit semantic and hint at the hierarchy between the elements. I'll think a bit on a better name

includes/class-wp-job-manager-ui-notice.php
yscik: It's what's possible right now for named arguments. This function has a lot of optional arguments, so calls could look like this:

`Notice:render( null, null, "Alert created", null, null, [[ 'url' => $manage_url, 'label' => 'Manage Alerts' ]])`

That's extremely hard to figure out, missing context for the what each argument is, and very easy to make mistakes with. 

There are some clean code literature on this: https://refactoring.guru/smells/long-parameter-list, https://dev.to/thinkster/code-smell-too-many-parameters-435e

includes/ui/Notice.php
gikaragia: Sorry didn't clarify that but I was not suggesting that an OOP approach is better than procedural here, what I was trying to say was that if we use a class it becomes OOP otherwise we should use functions.

There are benefits of using instance methods instead of static ones though, as static methods cannot be mocked AFAIK and they are not as easily overriden. You can always do `(new Success_Notice()->render() ) if you want to call render in a single expression.

includes/ui/Notice.php
```diff
@@ -0,0 +1,226 @@
...
+				[
+					'label' => '',
+					'url'   => '',
+				]
+			);
+
+			$html = '<a href="' . esc_url( $args['url'] ) . '" class="' . esc_attr( $class ) . '"><span>' . esc_html( $args['text'] ) . '</span></a>';
+		}
+
+		return $html;
```
thedebian: Can you reverse the logic here ?
if empty( $args ) return $html

includes/ui/Settings.php
```diff
@@ -0,0 +1,268 @@
...
+					[
+						'text'    => 'Outlined Button',
+						'href'    => '/',
+						'primary' => false,
+					],
+				],
+			]
+		);
+
+		return implode( '', $elements );
```
thedebian: Instead of using an intermediary array, can you just use a string and append ?
$elements .= Notice::render(blablabla)

includes/ui/Notice.php
```diff
@@ -0,0 +1,226 @@
...
+			$html .= '>';
+
+			if ( ! $is_classname ) {
+				$html .= ( $icon );
+			}
+
+			$html .= '</div>';
+		}
+
+		return $html;
```
thedebian: Can you reverse the logic here ?
if empty( $icon ) return $html

includes/ui/Notice.php
```diff
@@ -0,0 +1,226 @@
...
+			$html = '<div class="jm-notice__icon"';
+
+			if ( $is_classname ) {
+				$html .= ' data-icon="' . esc_attr( $icon ) . '"';
+			}
+
+			$html .= '>';
+
+			if ( ! $is_classname ) {
+				$html .= ( $icon );
```
thedebian: Does it need to be escaped ?

yscik: This is for using SVG code for the icon, which makes it a bit tricky in WP. Added a wp_kses call now covering the basics.

yscik:
Fixed some more issues and renamed the files to use the WP coding standard.

includes/ui/Settings.php
yscik: Kept this one as array, just to allow tweaking the rendering later, like adding some separator between the items.

yscik:
Merging this and #2667 + 393-gh-Automattic/wpjm-addons to keep things moving. We can address any further fixes needed in a follow-up PR.

PR #2661, opened 2023-11-28
Add custom post type for storing guest user data

### LINKED ISSUE #2660: Create custom post type for guest accounts

fjorgemota:
Fixes #2660 

* Add custom post type for storing guest user data

### Testing Instructions

Just check if the code makes sense, considering the purpose for this CPT and also the parameters of the `register_post_type` function. 

----

| Plugin build for 72adc4783d5912aad2990a3ea98e540f1f66ee20 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/11/wp-job-manager-zip-2661-72adc478.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/11/2661-72adc478)             |

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -417,6 +417,34 @@ class WP_Job_Manager_Post_Types {
 				'label_count'               => _n_noop( 'Preview <span class="count">(%s)</span>', 'Preview <span class="count">(%s)</span>', 'wp-job-manager' ),
 			]
 		);
+
+		/**
+		 * Custom post type used to store guest user data.
+		 */
+		register_post_type(
+			'job_guest_user',
+			[
+				apply_filters(
+					'register_post_type_job_guest_user',
+					[
+						'description'         => esc_html__( 'This is where guest user data is stored.', 'wp-job-manager' ),
+						'public'              => false,
+						'show_ui'             => false,
+						'capability_type'     => 'job_guest_user',
+						'map_meta_cap'        => false,
+						'publicly_queryable'  => false,
+						'exclude_from_search' => true,
+						'hierarchical'        => false,
+						'rewrite'             => false,
+						'query_var'           => false,
+						'supports'            => [ 'title', 'custom-fields', 'thumbnail' ],
+						'has_archive'         => false,
+						'show_in_nav_menus'   => false,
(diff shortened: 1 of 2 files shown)


includes/class-wp-job-manager-post-types.php
```diff
@@ -417,6 +417,34 @@ public function register_post_types() {
...
+						'public'              => false,
+						'show_ui'             => false,
+						'capability_type'     => 'job_guest_user',
+						'map_meta_cap'        => false,
+						'publicly_queryable'  => false,
+						'exclude_from_search' => true,
+						'hierarchical'        => false,
+						'rewrite'             => false,
+						'query_var'           => false,
+						'supports'            => [ 'title', 'custom-fields', 'thumbnail' ],
```
mikeyarce: What are your thoughts behind `thumbnail` here?

mikeyarce (approved):
Looks good!  Just left one comment about `thumbnail`.

includes/class-wp-job-manager-post-types.php
fjorgemota: hmmm I was originally thinking about avatars, but I guess those doesn't make a ton of sense for guest users, right? :sweat_smile: 

Removed in a80c0634201fd7176996eb30db862d33c5bddfd9. Thanks!

thedebian (approved):
👍 
Could you replace "job_guest_user" with a public constant PT_GUEST_USER in WP_Job_Manager_Post_Types, and use it in $custom_post_types ?
Same for the capability_type, it can be CAP_GUEST_USER.
I'll create a card to do the same for all the custom post types/taxonomies there, it'll make the code cleaner and less error-prone

includes/class-wp-job-manager-post-types.php
yscik: We'll use Gravatar! :)

fjorgemota:
Hey, thanks for the review! 

> Could you replace "job_guest_user" with a public constant PT_GUEST_USER in WP_Job_Manager_Post_Types, and use it in $custom_post_types ?
> Same for the capability_type, it can be CAP_GUEST_USER.

Done in 72adc4783d5912aad2990a3ea98e540f1f66ee20, could you take a quick last look? :eyes:

PR #2662, opened 2023-11-28
Add option to determine how to handle shortcodes in job listings

### LINKED ISSUE #1146: Disable shortcode use in job submission

fjorgemota:
Fixes #1146 

* Add option to determine how to handle shortcodes in job listings

### Testing Instructions

Using this branch of WPJM:

1. Create a new job listing
2. Use a valid shortcode on the description of that job listing, you can for instance use this as post content:

```
Test

[job_apply id=”X”]

Another Test
```
3. Check the job listing on the frontend (the individual job listing page), make sure the shortcode runs as usual
4. Now, on wp-admin, go to "Job Manager" -> "Settings" -> "Job Listings" and check the option "Strip Shortcodes" (that option should be disabled by default)
5. Go back to the individual job listing page for the job listing you just created and make sure the shortcode DOESN'T RUN at all, like, make sure it is actually stripped from the page.

### Release Notes

- Add option to strip shortcodes from the job description

### Screenshot / Video

New option to strip shortcodes

### Notes

Just for the sake of documentation, initially I was going to add an option to Allow Shortcodes (the current default), Keep them (by just not running them at all) or Strip them (which is what I'm doing here). This was done until commit 61ca9ae8cfb11faac1334713af38484bb6fbfe48, so if you want to check out that approach too, you can. I decided to simplify things because the option to Keep the shortcodes was a bit...potentially troublesome, in special because it might not work because of third-party plugins/themes. So just allowing to strip them sounded easier.

----

| Plugin build for 744a58e56eeb9ace2a79570cd88f23f0ac9e36ed <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/11/wp-job-manager-zip-2662-744a58e5.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/11/2662-744a58e5)             |


gikaragia (approved):
I was thinking that this might be troublesome for descriptions that want to include a string with the shortcode format (i.e. [a string]) but it seems that this only removes the shortcodes that exist. So [jobs] will be removed but [a job] will not.

LGTM!

yscik:
Hm I'm not sure that removing text from the content because they match a shortcode is a good idea, hard to justify to job posters why [jobs] disappears from a description but [something] doesn't. 

We could escape the shortcodes instead, or better yet, if possible, not run the shortcode processing for the description if the option is off.

fjorgemota:
Hey.

It is a bit weird, but it is basically what the [current snippet that we published on our docs](https://wpjobmanager.com/document/developer-reference/code-snippets/wpjm-core-snippets/#strip-shortcodes-from-the-job-description) does, anyway. I'm just adding an option so the user doesn't need to add a custom snippet on WPJM just to handle thuis.

> We could escape the shortcodes instead, or better yet, if possible, not run the shortcode processing for the description if the option is off.

That said, while implementing this PR, I tried to take this approach but quickly found out that it is hard to implement this in a way that other third-party plugins/themes don't break the functionality. See the notes and if possible checkout https://github.com/Automattic/WP-Job-Manager/commit/61ca9ae8cfb11faac1334713af38484bb6fbfe48 to see what I did here. I also published a P2 about this on pbFulr-XQ-p2, too.

PR #2666, opened 2023-12-01
Add access tokens

gikaragia:
Closes https://github.com/Automattic/wpjm-addons/issues/161, https://github.com/Automattic/wpjm-addons/issues/160

This meant to be be an initial approach to get some early feedback. Main points:
- The tokens support expiry. I think that we should use this for guest accounts.
- The code will be in WPJM core to be able to be used by all plugins.
- The current implementation is meant to be a replacement for the unsusbscribe links. The next iteration will save data in post meta for guest accounts.
- I based my solution in an [existing approach ](https://github.com/Automattic/wpjm-addons/blob/e46bef58903934ef6fb91873e9823e5c97334440/src/alerts/wp-job-manager-alerts.php#L412C2-L412C2)that we had in alerts for unsusbscribe links. I think that this approach is a bit more secure as:
	* It doesn't use DES
	* It stores only hashed values.
- If we agree on the approach I will also request a review from our security team.
- I am thinking that we should replace [these methods](https://github.com/Automattic/wpjm-addons/blob/e46bef58903934ef6fb91873e9823e5c97334440/src/alerts/wp-job-manager-alerts.php#L412C2-L412C2) with this class. We could potentially use both methods for some time to not break backward compatibility.

* Adds a token class which can be used to create and verify tokens.
* It's possible for tokens to have an expiry date.

### Testing Instructions

* Just check the approach and make sure that the tests pass.

----

| Plugin build for 622323363e98e89303675956a6524c01f640887b <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2666-62232336.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2666-62232336)             |

### DIFF
--- a/includes/class-access-token.php
+++ b/includes/class-access-token.php
@@ -31,6 +31,7 @@ class Access_Token {
 	 */
 	public function __construct( array $metadata ) {
 		$this->metadata = $metadata;
+		ksort( $this->metadata );
 	}
 
 	/**
--- a/includes/class-access-token.php
+++ b/includes/class-access-token.php
@@ -74,7 +74,7 @@ class Access_Token {
 	 * @return bool True if the token is correct.
 	 */
 	public function verify( string $token ) : bool {
-		$token_data = $this->lood_token_data( $this->payload );
+		$token_data = $this->load_token_data( $this->payload );
 
 		if ( true === $this->check_token_expiry( $token_data['payload'] ) ) {
 			return false;
@@ -89,7 +89,7 @@ class Access_Token {
 	 * @return bool
 	 */
 	public function is_expired() : bool {
-		$token_data = $this->lood_token_data( $this->payload );
+		$token_data = $this->load_token_data( $this->payload );
 
 		return $this->check_token_expiry( $token_data['payload'] );
 	}
@@ -116,7 +116,7 @@ class Access_Token {
 	 * @type string $payload    The token payload.
 	 * }
 	 */
-	protected function lood_token_data( array $payload ): array {
+	protected function load_token_data( array $payload ): array {
 		$token_hash = get_user_meta( $payload['user_id'], 'job_manager_alerts_secret_key', true );
 		$payload    = get_user_meta( $payload['user_id'], 'job_manager_alerts_token_payload', true );
 
--- a/includes/class-access-token.php
+++ b/includes/class-access-token.php
@@ -17,14 +17,6 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since $$next-version$$
  */
 class Access_Token {
-
-	/**
-	 * An expiry timestamp.
-	 *
-	 * @var int|null
-	 */
-	private ?int $expiry;
-
 	/**
 	 * Token's metadata. They are hashed together with the token.
 	 *
@@ -35,12 +27,10 @@ class Access_Token {
 	/**
 	 * Constructor.
 	 *
-	 * @param array    $metadata Metadata to be hashed together with the token.
-	 * @param int|null $expiry Expiry timestamp.
+	 * @param array $metadata Metadata to be hashed together with the token.
 	 */
-	public function __construct( array $metadata, int $expiry = null ) {
+	public function __construct( array $metadata ) {
 		$this->metadata = $metadata;
-		$this->expiry   = $expiry;
 	}
 
 	/**
@@ -49,94 +39,43 @@ class Access_Token {
 	 * @return string The token.
 	 */
 	public function create() : string {
-		// Create a random token, rtrim and strtr converts the base64 to a url-friendly format.
-		// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode -- Method used safely.
-		$token = rtrim( strtr( base64_encode( openssl_random_pseudo_bytes( 18 ) ), '+/', '-_' ), '=' );
+		$tick = wp_nonce_tick();
 
+		return $this->generate_token( $tick );
+	}
+
+	/**
+	 * Generates a token for a tick.
+	 *
+	 * @param float $tick
+	 *
+	 * @return false|string
+	 */
+	private function generate_token( float $tick ) {
 		$metadata_json = wp_json_encode( $this->metadata );
-		$token_hash    = password_hash( $metadata_json . $token, PASSWORD_DEFAULT );
 
-		$this->store_token_data( $this->metadata, $token_hash, $this->expiry );
-
-		return $token;
+		return substr( wp_hash( $tick . '|' . $metadata_json, 'nonce' ), -12, 10 );
 	}
 
 	/**
 	 * Verifies that a token is correct.
 	 *
 	 * @param string $token The token to verify.
+	 * @param int    $duration_days Days that the token should be valid.
 	 *
 	 * @return bool True if the token is correct.
 	 */
-	public function verify( string $token ) : bool {
-		$stored_data = $this->load_token_data( $this->metadata );
+	public function verify( string $token, int $duration_days ) : bool {
+		$duration_ticks = $duration_days * 2;
+		$tick           = wp_nonce_tick();
 
-		if ( true === $this->check_token_expiry( $stored_data['expiry'] ) ) {
-			return false;
+		for ( $i = 0; $i < $duration_ticks; $i++ ) {
+			if ( hash_equals( $this->generate_token( $tick - $i ), $token ) ) {
+				return true;
+			}
 		}
 
-		return password_verify( wp_json_encode( $this->metadata ) . $token, $stored_data['token_hash'] );
-	}
-
-	/**
-	 * Whether the token is expired.
-	 *
-	 * @return bool
-	 */
-	public function is_expired() : bool {
-		$stored_data = $this->load_token_data( $this->metadata );
-
-		return $this->check_token_expiry( $stored_data['expiry'] );
-	}
-
-	/**
-	 * Checks if a token is expired.
-	 *
-	 * @param int $expiry The token expiry.
-	 *
-	 * @return bool
-	 */
-	private function check_token_expiry( int $expiry ) : bool {
-		return ! empty( $expiry ) && time() > $expiry;
-	}
-
-	/**
-	 * Load the token data from storage.
-	 *
-	 * @param array $metadata The metadata of the token.
-	 *
-	 * @return array $stored_data {
-	 *
-	 * @type string $token_hash The token hash. Used for verification.
-	 * @type string $metadata The stored token metadata.
-	 * @type int    $expiry The expiry timestamp.
-	 * }
-	 */
-	protected function load_token_data( array $metadata ): array {
-		$token_hash      = get_user_meta( $metadata['user_id'], 'job_manager_alerts_secret_key', true );
-		$stored_metadata = get_user_meta( $metadata['user_id'], 'job_manager_alerts_token_metadata', true );
-		$expiry          = (int) get_user_meta( $metadata['user_id'], 'job_manager_alerts_token_expiry', true );
-
-		return [
-			'token_hash' => $token_hash,
-			'metadata'   => $stored_metadata ?: [],
-			'expiry'     => $expiry,
-		];
-	}
-
-	/**
-	 * Store the token data.
-	 *
-	 * @param array    $metadata   The token metadata.
-	 * @param string   $token_hash The token hash that will be used for the verification.
-	 * @param int|null $expiry The token expiry.
-	 *
-	 * @return void
-	 */
-	protected function store_token_data( array $metadata, string $token_hash, ?int $expiry ): void {
-		update_user_meta( $metadata['user_id'], 'job_manager_alerts_secret_key', $token_hash );
-		update_user_meta( $metadata['user_id'], 'job_manager_alerts_token_metadata', $metadata );
-		update_user_meta( $metadata['user_id'], 'job_manager_alerts_token_expiry', $expiry );
+		return false;
 	}
 }
 
--- a/includes/class-access-token.php
+++ b/includes/class-access-token.php
@@ -55,7 +55,7 @@ class Access_Token {
 	private function generate_token( float $tick ) {
 		$metadata_json = wp_json_encode( $this->metadata );
 
-		return substr( wp_hash( $tick . '|' . $metadata_json, 'nonce' ), -12, 10 );
+		return substr( wp_hash( $tick . '|' . $metadata_json, 'nonce' ), -18, 16 );
 	}
 
 	/**
--- a/includes/class-access-token.php
+++ b/includes/class-access-token.php
@@ -40,7 +40,7 @@ class Access_Token {
 	 * @return string The token.
 	 */
 	public function create() : string {
-		$tick = wp_nonce_tick();
+		$tick = $this->token_tick();
 
 		return $this->generate_token( $tick );
 	}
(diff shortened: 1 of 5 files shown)


includes/class-access-token.php
```diff
@@ -0,0 +1,143 @@
...
+	 * @var int|null
+	 */
+	private ?int $expiry;
+
+	/**
+	 * Token's payload. It is hashed together with the token.
+	 *
+	 * @var array
+	 */
+	private array $payload;
```
yscik: It's confusing how 'payload' is used for multiple things, being an instance property and then loaded from the database as something else. Maybe it's simpler to make this into just `$user_id`?

includes/class-access-token.php
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * Store the token data.
+	 *
+	 * @param array  $payload    The token payload.
+	 * @param string $token_hash The token hash that will be used for the verification.
+	 *
+	 * @return void
+	 */
+	protected function store_token_data( array $payload, string $token_hash ): void {
+		update_user_meta( $payload['user_id'], 'job_manager_alerts_secret_key', $token_hash );
```
yscik: We need to rename the branch again, but I think this should be in `feature/ui-update`, and can use the new guest user post type right away from https://github.com/Automattic/WP-Job-Manager/pull/2661

tests/php/tests/includes/test_class.access-token.php
```diff
@@ -0,0 +1,33 @@
...
+
+use WP_UnitTestCase;
+
+class WP_Test_Access_Token extends WP_UnitTestCase {
+
+	public function test_correct_token_verified_succesfully() {
+		$access_token = new Access_Token( [ 'user_id' => 10 ] );
+		$token = $access_token->create();
+
+		self::assertTrue( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token ) );
```
yscik: WDYT of a static API, `Access_Token::create( $user_id, $expiry )` and `Access_Token::verify( $user_id, $token )`? Is there a benefit in creating instances of the class?

includes/class-access-token.php
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	private array $payload;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param array    $payload A payload to be hashed together with the token.
+	 * @param int|null $expiry Expiry timestamp.
+	 */
+	public function __construct( array $payload, int $expiry = null ) {
```
yscik: So if I understand correctly, the secret URL would look like `/alerts?user_id=1232&token=XXXXXXXX`? Could we encode the user ID in the token?

includes/class-access-token.php
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * Store the token data.
+	 *
+	 * @param array  $payload    The token payload.
+	 * @param string $token_hash The token hash that will be used for the verification.
+	 *
+	 * @return void
+	 */
+	protected function store_token_data( array $payload, string $token_hash ): void {
```
yscik: One big problem is that because the database needs to change for every generated token, only the last one is valid. I think that's not a good fit for our use case: with the Alert e-mails, only one would contain a working link. (Including when the user has multiple alerts and they are sent a few seconds apart.)

Could something like this work?
- A persistent secret key is stored in the DB to sign the tokens
- Tokens have payload (including expiration) encoded, and signed with the secret key
- Verification verifies signature, then unpacks payload and checks for expiration

includes/class-access-token.php
gikaragia: So the idea is that the payload can be many different things that relate to what this token authorize you to do. As we store the hashed payload in the db, it is a good idea to have it as generic as it can get for the API to be backwards compatible in the future in case we want to hash more stuff.

I was thinking also that maybe we should migrate the existing unsubscribe tokens to use this new API. These links already use alert_id and user_id as part of the token so just having a $user_id isn't enough. Also, sometimes this is going to be the $user_id but for accountless alerts it is going to be a $post_id.

I agree with your point though that it's confusing and even payload isn't a good enough name. I'll look into moving around things to make it more clear.

includes/class-access-token.php
gikaragia: Yep! So the plan is to create two different implementations of store_token_data and load_token data, one for accountless alerts and one for the unsubscribe emails.

includes/class-access-token.php
gikaragia: Yep! So we hash the whole payload to the token if this is what you are asking.

tests/php/tests/includes/test_class.access-token.php
gikaragia: So this depends on what you suggest `create` and `verify` are going to be.

If the suggestion is to avoid instances entirely and define static methods everywhere then this is practically the same with adapting a procedural style. If this is the case I think we should switch to using namespaces and simple functions instead of using classes as classes are meant to be used for OOP. On the other hand if we want adopt an OOP style then using contructors and instances is the way to go. There are many new features (e.g. constructor promotion, named arguments etc.) in PHP that suggest it moves towards an OOP style and the sooner we adapt our code the easier it's going to be to use them in the future.

I know that existing code does not follow that and classes are just a collection of staff the relate to something but I think this is a remnant of the early days where namespaces didn't exist and defining classes was the only option to avoid global conflicts.

If on the other hand the suggestion is to use a static method to create the instances themselves, I think it would make sense to do it if we had many and complicated usages of the constructor where in that case defining named constructors would make things easier to read.

includes/class-access-token.php
gikaragia: > only the last one is valid. I think that's not a good fit for our use case: with the Alert e-mails, only one would contain a working link. (Including when the user has multiple alerts and they are sent a few seconds apart.

Hmm I am not sure if I understand this point correctly. So the `store_token_data` and `load_token_data` are meant to be different for the accountless alerts. I just added these now as an example and an alternative approach to the unsusbscribe links to be able to write some tests. As storing and loading is going to be done by these two functions we can support anything in future extensions of the API, i.e. storing a single token per user or multiple ones.

Maybe I am missing something here. Can you explain a bit the use case an arbitrary store/load method can't achieve?

> Could something like this work?

There are two benefits of the current approach.
- The unhashed secret key is not saved in the db which makes it a bit more secure
- The implementation is a bit simpler as I find encrypt/decrypt methods a bit more cumbersome to use.

We can use normal arguments for whatever we store in the payload so there isn't really a need to get these values from the token. By comparing the hash we also verify that the token was issued for these specific arguments.

gikaragia:
Sorry for clicking 'request for review' @yscik this isn't ready yet. I'll notify you when it is.

includes/class-access-token.php
gikaragia: Ah ok so I had another look at this and I think you're right. Although we could potentially create a token for each link, it's not a great approach. I'll switch to storing the secret in the db

includes/class-access-token.php
yscik: Sorry the question is, do we need to include the `user_id` as a parameter too in the link?

includes/class-access-token.php
yscik: Thinking about this, would it greatly simplify things if this token was only for accountless alerts, and we keep the existing implementation for the current unsubscribe when using an account part?

tests/php/tests/includes/test_class.access-token.php
yscik: Yeah looking broadly at code in these plugins in a lot of cases global singletons or just procedural functions seem to be the best fit. For me OOP and creating objects makes sense mainly when we have some state that's kept around and used for multiple method calls or used in multiple ways. For one-off functions that just take some parameters and return a value, we shouldn't complicate things.

And you are right, that's pretty much just namespaced functions. Static classes are still pretty useful for those, having private helper methods and public functions as a clear API surface.

includes/class-access-token.php
gikaragia: With the latest approach it will be simple enough I think. I'll come up with a quick PR and we can discuss there.

includes/class-access-token.php
gikaragia: Yes. So if we want the user id to be included in the token, then we will need to use a symmetric key cipher which I think adds extra complexity without gaining much.

tests/php/tests/includes/test_class.access-token.php
gikaragia: In my view, global singletons are never the best fit and procedural functions would not be my choice in most cases but this I think is part of a bigger discussion.

What I am trying to say is that we should use the appropriate language constructs depending on the style of the approach, namespaces and functions for procedural and instances for OOP.

gikaragia:
Hey @yscik this should be ready to review again. I discussed a bit with @alexsanford and he came up with a much better idea. With usage of WP's already provided secret keys and the `wp_nonce` methods we don't really need to store anothing to provide tokens. So with this approach, we do something similar with what WP does for nonces and expiries. 

As a next step I will look into migrating alerts into using this token class. This should be ready for review now.

tests/php/tests/includes/test_class.access-token.php
gikaragia: > For one-off functions that just take some parameters and return a value, we shouldn't complicate things.

In the beginning I was thinking that these might become more complex, with state that needs to be managed in the db, not the two small functions that ended up being. I could petontially move these outside of a class if you think it adds any value.

includes/class-access-token.php
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Generates a token for a tick.
+	 *
+	 * @param float $tick
+	 *
+	 * @return false|string
+	 */
+	private function generate_token( float $tick ) {
+		$metadata_json = wp_json_encode( $this->metadata );
```
alexsanford, line 43: This will create a different output for `[ 'a' => 1, 'b' => 2 ]` and `[ 'b' => 2, 'a' => 1 ]`, which is probably not what you want. Maybe consider using `ksort` before encoding?

includes/class-access-token.php
```diff
@@ -0,0 +1,82 @@
...
+	/**
+	 * Verifies that a token is correct.
+	 *
+	 * @param string $token The token to verify.
+	 * @param int    $duration_days Days that the token should be valid.
+	 *
+	 * @return bool True if the token is correct.
+	 */
+	public function verify( string $token, int $duration_days ) : bool {
+		$duration_ticks = $duration_days * 2;
```
alexsanford: This assumes that the lifetime of a nonce is 1 day, with two ticks over that lifetime. Examining the `wp_nonce_tick` code, there is always 2 ticks over the lifetime, but the lifetime itself can be filtered using `nonce_life`. Any site using this filter will affect your expiration timeline here.

tests/php/tests/includes/test_class.access-token.php
```diff
@@ -0,0 +1,47 @@
...
+		self::assertTrue( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token, 2 ) );
+		self::assertTrue( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token_12_hours, 2 ) );
+		self::assertTrue( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token_24_hours, 2 ) );
+		self::assertTrue( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token_36_hours, 2 ) );
+		self::assertFalse( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token_48_hours, 2 ) );
+		self::assertFalse( ( new Access_Token( [ 'user_id' => 10] ))->verify( $token_60_hours, 2 ) );
+	}
+
+	public function test_incorrect_payload_not_verified() {
+		$access_token = new Access_Token( [ 'user_id' => 10 ], time() + 1000 );
```
alexsanford: I haven't run the tests to verify, but does this line actually work? Looks like it's handing in 2 args to the contructor, when it should only accept one.

alexsanford (commented):
Added a few comments, but in general this seems like a good solution to me! The use of `wp_hash` will use the server-side secret key to generate a hash for the metadata and the tick. If an attacker knows the metadata and the tick, they still will not be able to determine the correct hash without the server-side secret key.

A couple of considerations:

- If the server-side secret key ever needs to be rotated, all existing tokens will be invalidated. This is also true of nonces and session tokens, I believe.
- Depending on how often the alert emails are received, there is a chance that a user will, at some point, not have any link with a valid token (i.e. if the latest alert came more than `$duration_days` ago). It's something I thought of after we chatted yesterday. Not sure if this is something you are concerned about?

tests/php/tests/includes/test_class.access-token.php
gikaragia: Yeah PHP ignores extra args. Fixed in 463a1e5be96967bfac0883e59b988974f28dd114

includes/class-access-token.php
gikaragia: Good point fixed in 463a1e5be96967bfac0883e59b988974f28dd114

includes/class-access-token.php
gikaragia: Hmm so it seems that the only way to make sure that this will work always is to avoid the filter. If for example for some reason a site increases the life of a token, then our tokens can't event work.

 Added method `token_tick` in febd1d09b71d529e8fdff9ba45294848b5e6e883

gikaragia:
Thank you for looking into this @alexsanford, good points!

> If the server-side secret key ever needs to be rotated, all existing tokens will be invalidated. This is also true of nonces and session tokens, I believe.

We could potentially store the token somewhere but I am thinking that if the key is compromised our tokens should be expired anyway.

> not have any link with a valid token

As we know how often these emails are sent we can adjust the expiry accordingly. We also have a flow to send another email if the token is expired or if the user just visits the alert management screen so I think we should be ok.

includes/class-access-token.php
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * @param string $token The token to verify.
+	 * @param int    $duration_days Days that the token should be valid.
+	 *
+	 * @return bool True if the token is correct.
+	 */
+	public function verify( string $token, int $duration_days ) : bool {
+		$tick = $this->token_tick();
+
+		for ( $i = 0; $i < $duration_days; $i++ ) {
```
yscik: We have a 'monthly' option for alerts e-mails, so for that use case at the minimum this duration would be 30 days, but preferably more like 60. Is that feasible with this approach?

includes/class-access-token.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
yscik, line 1: If we use the standard [PSR-4 scheme](https://www.php-fig.org/psr/psr-4/) for new namespaced classes, the filename should be `Access_Token.php`. And we haven't really talked about keeping underscores, but with the camelcase I'd slight prefer 'AccessToken' over 'Access_Token', except for the root namespace.

includes/class-access-token.php
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * The tick controls how often the token is going to be rotated. Copied from wp_nonce_tick to avoid the nonce_life
+	 * filter.
+	 *
+	 * @return false|float
+	 */
+	private function token_tick() {
+		return ceil( time() / DAY_IN_SECONDS );
+	}
```
yscik: Maybe instead of a fixed internal property, this could be some input parameter or somehow configurable? That'd make testing easier, and unlock making non-expiring tokens too. I guess it could also be done with a subclass overriding this method.

yscik (approved):
Looks good if the long expiration period is not going to be a problem.

includes/class-access-token.php
gikaragia: I tested the hash method on my laptop and it seems that it's very fast (something like 5 microseconds for 1000 iterations) so from a performance perspective this was okay. This won't exist in the new approach though.

includes/class-access-token.php
gikaragia: +1

gikaragia: I'll leave this as it is as we are keeping the `class` prefix in filenames for now.

yscik (approved):
Looks good, I like the new approach!

PR #2667, opened 2023-12-01
Add support to display messages after a redirect

yscik:
Based on https://github.com/Automattic/WP-Job-Manager/pull/2659

* Add a helper that lets you set a message, redirect the page, and show the message
* This serves a common pattern used when serving a POST/GET action like saving a form or changing a setting: we redirect the page after the action is done so it's not triggered again on page reload or a link share.
* Stores the message in a transient with an ID, and passes along the ID as a query string argument to the redirected URL, like `?message=234d72af2`
* Includes a small JS script to remove that query argument

### Testing Instructions

* See 393-gh-Automattic/wpjm-addons for testing

### Release Notes

* Add support to display messages after a redirect 

----

| Plugin build for fbdf0d056e2881bf2dd99f856f295c3952b13d28 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2667-fbdf0d05.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2667-fbdf0d05)             |

### DIFF
--- a/includes/ui/RedirectMessage.php
+++ b/includes/ui/RedirectMessage.php
@@ -62,15 +62,15 @@ class RedirectMessage {
 
 		$content = self::get_transient_message( $key );
 
-		if ( $content ) {
-			$content .= self::add_inline_script( $query_var );
-
-			UI::ensure_styles();
-
-			return $content;
+		if ( ! $content ) {
+			return null;
 		}
 
-		return null;
+		$content .= self::add_inline_script( $query_var );
+
+		UI::ensure_styles();
+
+		return $content;
 	}
 
 	/**
@@ -99,15 +99,15 @@ class RedirectMessage {
 	private static function get_transient_message( $key, $delete = true ) {
 		$content = get_transient( self::TRANSIENT_PREFIX . $key );
 
-		if ( is_string( $content ) ) {
-			if ( $delete ) {
-				delete_transient( self::TRANSIENT_PREFIX . $key );
-			}
-
-			return $content;
+		if ( ! is_string( $content ) ) {
+			return null;
 		}
 
-		return null;
+		if ( $delete ) {
+			delete_transient( self::TRANSIENT_PREFIX . $key );
+		}
+
+		return $content;
 	}
 
 	/**
--- a/includes/ui/RedirectMessage.php
+++ b/includes/ui/RedirectMessage.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * Frontend redirect messages.
+ *
+ * @package wp-job-manager
+ * @since $$next-version$$
+ */
+
+namespace WP_Job_Manager\UI;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Redirect after an action is finished and make sure a message can displayed on the next page load.
+ *
+ * Stores the messages in a transient with an ID, and passes along the ID as a query string to the redirected URL.
+ *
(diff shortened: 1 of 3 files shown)


includes/ui/RedirectMessage.php
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Redirect to a URL with a message to be displayed on the next page load.
+	 *
+	 * @param string $url URL to redirect to.
+	 * @param string $message Optional message to be displayed on the next page load.
+	 * @param string $query_var A query variable added to the redirect URL, referencing the message transient.
+	 */
+	public static function redirect( $url, $message = null, $query_var = '' ) {
+		if ( ! empty( $message ) ) {
+			$url = add_query_arg( [ $query_var ?? self::DEFAULT_QUERY_VAR => self::set_transient_message( $message ) ], $url );
```
thedebian: The ?? operator will still return $query_var if it's empty, for a shorthand, you can use the short-ternary operator (?:), but I'd say that a proper if/else here might be better for readability

https://www.tehplayground.com/cUYSR0Ht8SvWQTCs

includes/ui/RedirectMessage.php
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Get a message set during the previous redirect.
+	 *
+	 * @param string $query_var Optional namespace for the message.
+	 *
+	 * @return string|null Message content if there is one.
+	 */
+	public static function get_message( $query_var = '' ) {
+
+		$query_var = $query_var ?? self::DEFAULT_QUERY_VAR;
```
thedebian: Same issue as above

includes/ui/RedirectMessage.php
thedebian: Alternatively you can also set the default $query_var to null, but it might be better to use empty( $query_var ) anyway, so it covers all empty cases and never fatals

includes/ui/RedirectMessage.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ( $content ) {
+			$content .= self::add_inline_script( $query_var );
+
+			UI::ensure_styles();
+
+			return $content;
+		}
+
+		return null;
```
thedebian:
```suggestion
		if ( ! $content ) {
			return null;
		}

		$content .= self::add_inline_script( $query_var );

		UI::ensure_styles();

		return $content;
```

includes/ui/RedirectMessage.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ( is_string( $content ) ) {
+			if ( $delete ) {
+				delete_transient( self::TRANSIENT_PREFIX . $key );
+			}
+
+			return $content;
+		}
+
+		return null;
```
thedebian:
```suggestion
		if ( ! is_string( $content ) ) {
			return null;
		}

		if ( $delete ) {
			delete_transient( self::TRANSIENT_PREFIX . $key );
		}

		return $content;
```

thedebian (changes requested):

includes/ui/RedirectMessage.php
yscik: Good catch, thanks. Fixing in 7059ce8

thedebian (approved):
Code changes look good, ✅ as 393-gh-Automattic/wpjm-addons is tested and ✅ed too

PR #2669, opened 2023-12-02
Add multiple checkbox input fields

mikeyarce:
Needed for https://github.com/Automattic/wpjm-addons/issues/152

* Add a new type of setting which is `multiple checkboxes` for a single setting.  

### Testing Instructions

* Check this PR out to have a setting to test https://github.com/Automattic/wpjm-addons/pull/394
* Verify that the setting shows up properly

### Screenshot / Video

----

| Plugin build for c136d942384655076d9c6545c2e21046c7f95078 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2669-c136d942.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2669-c136d942)             |

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -812,13 +812,13 @@ class WP_Job_Manager_Settings {
 	}
 
 	/**
-	 * Radio and Checkbox Outputs.
+	 * Multiple Choice field output (radio or checkbox).
 	 *
 	 * @param array  $option Option data.
 	 * @param mixed  $value  Current value.
 	 * @param string $type   'radio' or 'checkbox'.
 	 */
-	protected function radio_checkbox_outputs( $option, $value, $type ) {
+	protected function multiple_choice_output( $option, $value, $type ) {
 		?>
 		<fieldset>
 		<legend class="screen-reader-text">
@@ -851,7 +851,7 @@ class WP_Job_Manager_Settings {
 	 * @param string $ignored_placeholder
 	 */
 	protected function input_radio( $option, $ignored_attributes, $value, $ignored_placeholder ) {
-		$this->radio_checkbox_outputs( $option, $value, 'radio' );
+		$this->multiple_choice_output( $option, $value, 'radio' );
 	}
 
 	/**
@@ -863,7 +863,7 @@ class WP_Job_Manager_Settings {
 	 * @param string $ignored_placeholder
 	 */
 	protected function input_multi_checkbox( $option, $ignored_attributes, $value, $ignored_placeholder ) {
-		$this->radio_checkbox_outputs( $option, $value, 'checkbox' );
+		$this->multiple_choice_output( $option, $value, 'checkbox' );
 	}
 	/**
 	 * Page input field.
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -828,15 +828,16 @@ class WP_Job_Manager_Settings {
 		if ( ! empty( $option['desc'] ) ) {
 			echo ' <p class="description">' . wp_kses_post( $option['desc'] ) . '</p>';
 		}
-		if ( 'checkbox' === $type ) {
-			foreach ( $option['options'] as $key => $name ) {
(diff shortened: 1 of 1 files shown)


includes/admin/class-wp-job-manager-settings.php
```diff
@@ -829,15 +828,43 @@ protected function input_radio( $option, $ignored_attributes, $value, $ignored_p
 		if ( ! empty( $option['desc'] ) ) {
 			echo ' <p class="description">' . wp_kses_post( $option['desc'] ) . '</p>';
 		}
-
-		foreach ( $option['options'] as $key => $name ) {
-			echo '<label><input name="' . esc_attr( $option['name'] ) . '" type="radio" value="' . esc_attr( $key ) . '" ' . checked( $value, $key, false ) . ' />' . esc_html( $name ) . '</label><br>';
+		if ( 'checkbox' === $type ) {
+			foreach ( $option['options'] as $key => $name ) {
+				$value = (array) $value;
+				echo '<label><input name="' . esc_attr( $option['name'] . '[fields][]' ) . '" type="' . esc_attr( $type ) . '" value="' . esc_attr( strtolower( $name ) ) . '" ' . checked( isset( $value['fields'] ) && is_array( $value['fields'] ) && in_array( strtolower( $name ), $value['fields'], true ), true, 0 ) . '/>' . esc_html( $name ) . '</label><br>';
```
pgk: Might be worth extracting the code that outputs the input+label into it's own function

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -812,14 +812,13 @@ class="regular-text"
...
-	 * @param array  $option
-	 * @param array  $ignored_attributes
-	 * @param mixed  $value
-	 * @param string $ignored_placeholder
+	 * @param array  $option Option data.
+	 * @param mixed  $value  Current value.
+	 * @param string $type   'radio' or 'checkbox'.
 	 */
-	protected function input_radio( $option, $ignored_attributes, $value, $ignored_placeholder ) {
+	protected function radio_checkbox_outputs( $option, $value, $type ) {
```
gikaragia: Should we also add support for the 'std' option? This is going to be the preselected value if present.

pgk: Maybe name it something like "multiple_choice" or similar, because it is either outputting a multiple choice field where you either select a single option (radio) or multiple options (checkboxes).

includes/admin/class-wp-job-manager-settings.php
pgk: Or alternatively, we can refactor this into one single loop and do the if checks within the loop body.

pgk (changes requested):
It works well along https://github.com/Automattic/wpjm-addons/pull/394.!

I have some thoughts on naming (always hard) and maybe squeezing a bit of reuse within the for loops. We could even squash this into one loop I think.

includes/admin/class-wp-job-manager-settings.php
mikeyarce: I'm not totally sure what you mean by this.  Right now if there is no value the `std` option will show up as the default.

mikeyarce: Done in f3785d8fcd68968f7d62cd66aa2afb1f8bdda968

includes/admin/class-wp-job-manager-settings.php
mikeyarce: Thanks Panos!  What do you think about 373c5173441463a71b804b7917833ffa6381629c ?

pgk: I like it! I would additionally probably do the checkbox check once and save it into a var but not a show stopper!

pgk (approved):
Looks good after the changes!

includes/admin/class-wp-job-manager-settings.php
gikaragia: So I thought that we needed to do some special handling for the std option to be the default. Wasn't aware that it was working already!

PR #2670, opened 2023-12-02
Add color input setting type

mikeyarce:
Part of https://github.com/Automattic/wpjm-addons/issues/154

* Add a `color` setting type

### Testing Instructions

* You need to add a setting like this to be able to see it in action:

```
array(
		'name'     => 'job_manager_alerts_brand_color',
		'label'    => __( 'Brand Color', 'wp-job-manager-alerts' ),
		'std'      => '#0453EB',
		'type'     => 'color',
		'desc'     => __( 'Set the color used for links and buttons in e-mails and shortcodes.', 'wp-job-manager-alerts' ),
),
```

### Screenshot / Video

----

| Plugin build for 2de3427d6e5e4f8db9c0e46304d2581fdf79ce1b <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2670-2de3427d.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2670-2de3427d)             |

### DIFF
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -19,6 +19,11 @@
 		margin: 12px 0;
 		font-size: 12px;
 	}
+
+	input.job-alerts-color-picker {
+		width: 30px;
+		height: 30px;
+	}
 }
 
 .job-manager-settings-header-wrap {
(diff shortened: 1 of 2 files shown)


thedebian (approved):
Worked as advertised 👍 
Is there a reason we use kses instead of esc_html in the description though ?

mikeyarce:
Thanks @thedebian !  

>  Is there a reason we use kses instead of esc_html in the description though ?

I think that's the default we use for settings in case we want to use HTML in the descriptions.

PR #2675, opened 2023-12-09
Add/featured first argument

### LINKED ISSUE #1487: Always show Featured Jobs in Listings

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/1487

* Add a new `featured_first` argument to the `[jobs]` shortcode
* This ensures that now regardless of the order, featured jobs will be first!

### Testing Instructions

* There are a lot of combinations to test for the jobs!

```
[jobs orderby="featured" featured_first="true" order="desc"]
[jobs orderby="title" featured_first="true" order="desc"]
[jobs orderby="ID" featured_first="true" order="desc"]
[jobs orderby="name" featured_first="true" order="desc"]
[jobs orderby="date" featured_first="true" order="desc"]
[jobs orderby="modified" featured_first="true" order="desc"]
[jobs orderby="rand" featured_first="true" order="desc"]
[jobs orderby="rand_featured" featured_first="true" order="desc"]
[jobs orderby="featured" featured_first="true" order="asc"]
[jobs orderby="rand_featured" featured_first="true" order="asc"]
[jobs orderby="title" featured_first="true" order="asc"]
[jobs orderby="ID" featured_first="true" order="asc"]
[jobs orderby="name" featured_first="true" order="asc"]
[jobs orderby="date" featured_first="true" order="asc"]
[jobs orderby="modified" featured_first="true" order="asc"]
[jobs orderby="rand" featured_first="true" order="asc"]
[jobs orderby="rand_featured" featured_first="true" order="asc"]
[jobs orderby="featured" featured_first="true" order="asc"]
```

I would also make sure to try at least a few `featured_first="false"` to ensure that works as well.

### Release Notes

* Add new shortcode parameter, "featured_first" so you can ensure featured listings always show up on top.

----

| Plugin build for c3b858a88193136c708f9508ab0ea779707195b0 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2675-c3b858a8.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2675-c3b858a8)             |

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -769,9 +769,6 @@ class WP_Job_Manager_Shortcodes {
 		if ( ! empty( $atts['post_status'] ) ) {
 			$data_attributes['post_status'] = implode( ',', $atts['post_status'] );
 		}
-		if ( ! empty( $atts['featured_first'] ) ) {
-			$data_attributes['featured_first'] = $atts['featured_first'] ? 'true' : 'false';
-		}
 
 		$data_attributes['post_id'] = isset( $GLOBALS['post'] ) ? $GLOBALS['post']->ID : 0;
 
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -627,9 +627,6 @@ class WP_Job_Manager_Shortcodes {
 		if ( ! is_null( $atts['remote_position'] ) ) {
 			$atts['remote_position'] = ( is_bool( $atts['remote_position'] ) && $atts['remote_position'] ) || in_array( $atts['remote_position'], [ 1, '1', 'true', 'yes' ], true );
 		}
-		if ( ! is_null( $atts['featured_first'] ) ) {
-			$atts['featured_first'] = ( is_bool( $atts['featured_first'] ) && $atts['featured_first'] ) || in_array( $atts['featured_first'], [ 1, '1', 'true', 'yes' ], true );
-		}
 
 		// By default, use client-side state to populate form fields.
 		$disable_client_state = false;
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -592,6 +592,7 @@ class WP_Job_Manager_Shortcodes {
 					'featured'                  => null, // True to show only featured, false to hide featured, leave null to show both.
 					'filled'                    => null, // True to show only filled, false to hide filled, leave null to show both/use the settings.
 					'remote_position'           => null, // True to show only remote, false to hide remote, leave null to show both.
+					'featured_first'            => false, // True to show featured first, false to show in default order.
 
 					// Default values for filters.
 					'location'                  => '',
@@ -613,6 +614,7 @@ class WP_Job_Manager_Shortcodes {
 		$atts['show_category_multiselect'] = $this->string_to_bool( $atts['show_category_multiselect'] );
 		$atts['show_more']                 = $this->string_to_bool( $atts['show_more'] );
 		$atts['show_pagination']           = $this->string_to_bool( $atts['show_pagination'] );
+		$atts['featured_first']            = $this->string_to_bool( $atts['featured_first'] );
 
 		if ( ! is_null( $atts['featured'] ) ) {
 			$atts['featured'] = ( is_bool( $atts['featured'] ) && $atts['featured'] ) || in_array( $atts['featured'], [ 1, '1', 'true', 'yes' ], true );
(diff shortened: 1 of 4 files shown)


includes/class-wp-job-manager-shortcodes.php
```diff
@@ -625,6 +627,9 @@ public function output_jobs( $atts ) {
 		if ( ! is_null( $atts['remote_position'] ) ) {
 			$atts['remote_position'] = ( is_bool( $atts['remote_position'] ) && $atts['remote_position'] ) || in_array( $atts['remote_position'], [ 1, '1', 'true', 'yes' ], true );
 		}
+		if ( ! is_null( $atts['featured_first'] ) ) {
+			$atts['featured_first'] = ( is_bool( $atts['featured_first'] ) && $atts['featured_first'] ) || in_array( $atts['featured_first'], [ 1, '1', 'true', 'yes' ], true );
+		}
```
thedebian:
```suggestion
```
$this->string_to_bool is already called on $atts['featured_first'], which does exactly that

includes/class-wp-job-manager-shortcodes.php
```diff
@@ -766,6 +772,9 @@ public function output_jobs( $atts ) {
 		if ( ! empty( $atts['post_status'] ) ) {
 			$data_attributes['post_status'] = implode( ',', $atts['post_status'] );
 		}
+		if ( ! empty( $atts['featured_first'] ) ) {
+			$data_attributes['featured_first'] = $atts['featured_first'] ? 'true' : 'false';
+		}
```
thedebian: Is it really what you wanted to do ?
If $atts['featured_first'] is false, then it's empty, which means $data_attributes won't be populated.
If it's what you wanted to do, then it's ok :)

wp-job-manager-functions.php
```diff
@@ -180,6 +181,13 @@ function get_job_listings( $args = [] ) {
 			];
 		}
 
+		if ( 'true' === $args['featured_first'] && 'featured' !== $args['orderby'] && 'rand_featured' !== $args['orderby'] ) {
```
thedebian, line 184: Does $args['featured_first'] (and all added data_attributes) have to be a string ?
If we want to test it against a bool, it's way better to test it ... well ... against a bool, but maybe there's a reason for that ?

thedebian (changes requested):
Maybe I'm missing something here, I pasted the test shortcodes, these are all supposed to be true

Capture d’écran 2023-12-11 à 14 44 52

Only one of them is featured

Capture d’écran 2023-12-11 à 14 44 14

includes/class-wp-job-manager-shortcodes.php
mikeyarce: Removed in d716a58bdd6dcb2742017797582c671869725f4b

includes/class-wp-job-manager-shortcodes.php
mikeyarce: yeah no, not what I want to do.  Removed in c3b858a88193136c708f9508ab0ea779707195b0

wp-job-manager-functions.php
mikeyarce: I'm trying to also account for people who might use `featured_first="0"` or `featured_first="1"` so checking the string worked for all cases.  What do you think about something like this:

```
if ( true === filter_var( $args['featured_first'], FILTER_VALIDATE_BOOLEAN ) && 'featured' !== $args['orderby'] && 'rand_featured' !== $args['orderby'] ) {
```

mikeyarce:
> Maybe I'm missing something here, I pasted the test shortcodes, these are all supposed to be true

What shortcode specifically wasn't working for you?

wp-job-manager-functions.php
thedebian: I thought the data here were populated by https://github.com/Automattic/WP-Job-Manager/pull/2675/files#diff-a0e82a74ed6b687d4ee6efce7bb4c78d6fa157212afa751c5c40f88306c489ddR684 , hence my question about making it boolean directly instead of string, but, if it's supposed to do more than what I thought, that's ok as-is

thedebian:
> > Maybe I'm missing something here, I pasted the test shortcodes, these are all supposed to be true
> 
> What shortcode specifically wasn't working for you?

Ok I think I forgot to npm run build, I know this might be tedious, but can it be included in test instructions when needed ? It helps a lot when reviewing a bunch of PRs

thedebian (approved):
After code changes (and npm build :D), worked like a charm

PR #2676, opened 2023-12-09
Save previous license upon deactivation

### LINKED ISSUE #2493: Persist license key

When deactivating a plugin, the license key is deleted. So when activating again, the user needs to go to wpjobmanager.com, copy the license, and activate it again.

It could deactivate the license, but it could also persist the key, so when re-activating a plugin they could just click on the activate button with the key already pre-filled.

### Additional context

We received a request (p1688568236271489-slack-C6GGX896G) related to the license_key not being persisted when deactivating a plugin. It's happening in `WP_Job_Manager_Helper` class, through the following action:

```php
add_action( 'deactivated_plugin', [ $this, 'plugin_deactivated' ] ); 
```

This hook deactivates the license, and removes the license_key. Maybe we could have another option to persist the license_key, only used to pre-fill the input when the license is not activated?

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2493

* When you deactivate a plugin, we'll now store the license in an option!
* That way, if you decide to activate the plugin again, it will show up in the license box pre-filled so you can activate it easily without needing to find it again.

### Testing Instructions

* Install a plugin separately (not wpjm-addons monorepo)
* Go to wpjobmanager.com and get a license for it
* Activate the license in Job Manager > Marketplace > Licenses
* Deactivate the plugin now!
* Go back to the Licenses page, see that your license is not active but the old license key is there!
* Profit.

### Release Notes

* Store previous license when plugin is deactivated for easier reactivation later.

----

| Plugin build for 19f909b1296b1e1b37a172dd1b1b0de796340c21 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2676-19f909b1.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2676-19f909b1)             |

### DIFF
--- a/includes/helper/views/html-licenses.php
+++ b/includes/helper/views/html-licenses.php
@@ -160,7 +160,7 @@ $plugin_section_first = 'plugin-license-section--first';
 						}
 						if ( apply_filters( 'wpjm_display_license_form_for_addon', true, $product_slug ) ) {
 							$has_error            = in_array( 'error', array_column( $notices, 'type' ), true );
-							$previous_license_key = get_option( 'job_manager_previous_license_' . $product_slug ) ? get_option( 'job_manager_previous_license_' . $product_slug ) : '';
+							$previous_license_key = get_option( 'job_manager_previous_license_' . $product_slug, null ) ?? '';
 							?>
 							<form method="post" class='plugin-license-form'>
 								<?php wp_nonce_field( 'wpjm-manage-license' ); ?>
--- a/includes/helper/views/html-licenses.php
+++ b/includes/helper/views/html-licenses.php
@@ -159,14 +159,15 @@ $plugin_section_first = 'plugin-license-section--first';
 							}
 						}
 						if ( apply_filters( 'wpjm_display_license_form_for_addon', true, $product_slug ) ) {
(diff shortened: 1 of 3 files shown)


includes/helper/views/html-licenses.php
```diff
@@ -159,14 +159,15 @@
 							}
 						}
 						if ( apply_filters( 'wpjm_display_license_form_for_addon', true, $product_slug ) ) {
-							$has_error = in_array( 'error', array_column( $notices, 'type' ), true );
+							$has_error            = in_array( 'error', array_column( $notices, 'type' ), true );
+							$previous_license_key = get_option( 'job_manager_previous_license_' . $product_slug ) ? get_option( 'job_manager_previous_license_' . $product_slug ) : '';
```
thedebian:
```suggestion
							$previous_license_key = get_option( 'job_manager_previous_license_' . $product_slug, null ) ?? '';
```

thedebian (approved):
Worked as advertised.
Left a small code change, apart from that 🚀

PR #2682, opened 2023-12-13
Expand the UI to inline CSS

gikaragia:
Needed by: https://github.com/Automattic/wpjm-addons/pull/415/files

* Adds a way to inline CSS to the WPJM UI

### Testing Instructions

* Should be tested as part of the related PR.

----

| Plugin build for 894f4b2c219d890c5fa9527f7712ccfdd12bc0fe <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2682-894f4b2c.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2682-894f4b2c)             |

### DIFF
--- a/includes/ui/class-ui.php
+++ b/includes/ui/class-ui.php
@@ -89,7 +89,7 @@ class UI {
 		$css = ':root{';
 
 		foreach ( $this->css_variables as $name => $value ) {
-			$css .= $name . ': ' . $value . ';';
+			$css .= esc_attr( $name ) . ': ' . esc_attr( $value ) . ';';
 		}
 
 		$css .= '}';
--- a/includes/ui/class-ui.php
+++ b/includes/ui/class-ui.php
@@ -76,7 +76,7 @@ class UI {
 	 */
 	public static function ensure_styles( array $css_variables = [] ) {
 		self::instance()->has_ui        = true;
-		self::instance()->css_variables = $css_variables;
+		self::instance()->css_variables = array_merge( self::instance()->css_variables, $css_variables );
 	}
 
 	/**
--- a/includes/ui/class-ui.php
+++ b/includes/ui/class-ui.php
@@ -73,7 +73,7 @@ class UI {
 	 *
 	 * @param array $css_variables An array of CSS variables to be enqueued: <variable_name> => <value>.
 	 */
-	public static function ensure_styles( $css_variables = null ) {
+	public static function ensure_styles( array $css_variables = [] ) {
 		self::instance()->has_ui        = true;
 		self::instance()->css_variables = $css_variables;
 	}
--- a/includes/ui/class-ui.php
+++ b/includes/ui/class-ui.php
@@ -34,7 +34,7 @@ class UI {
 	 *
 	 * @var bool
 	 */
-	private $has_ui;
+	private bool $has_ui;
 
 	/**
 	 * An array of css variables to be enqueued with the styles.
--- a/includes/ui/class-ui.php
+++ b/includes/ui/class-ui.php
@@ -34,12 +34,20 @@ class UI {
 	 *
 	 * @var bool
 	 */
-	private $has_ui = false;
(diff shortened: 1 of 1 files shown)


includes/ui/class-ui.php
```diff
@@ -52,14 +60,39 @@ public function enqueue_styles() {
...
+			}
 		}
 	}
 
 	/**
 	 * Request the styles to be loaded for the page.
+	 *
+	 * @param array $css_variables An array of CSS variables to be enqueued: <variable_name> => <value>.
+	 */
+	public static function ensure_styles( $css_variables = null ) {
```
fjorgemota:
```suggestion
	public static function ensure_styles( array $css_variables = [] ) {
```

If `$css_variables` becomes `null` for any reason, the `foreach` below will throw a warning, so better to fix that in the API. :)

includes/ui/class-ui.php
```diff
@@ -33,12 +33,20 @@ class UI {
...
 	 */
-	private $has_ui = false;
+	private $has_ui;
+
+	/**
+	 * An array of css variables to be enqueued with the styles.
+	 *
+	 * @var array
+	 */
+	private $css_variables;
```
fjorgemota:
```suggestion
	private array $css_variables;
```

I think we can type properties on PHP 7.4 :)

includes/ui/class-ui.php
```diff
@@ -33,12 +33,20 @@ class UI {
 	 *
 	 * @var bool
 	 */
-	private $has_ui = false;
+	private $has_ui;
```
fjorgemota:
```suggestion
	private bool $has_ui;
```

includes/ui/class-ui.php
```diff
@@ -33,12 +33,20 @@ class UI {
...
+	 *
+	 * @var array
+	 */
+	private $css_variables;
 
 	/**
 	 * Instance constructor
 	 */
 	private function __construct() {
+		$this->has_ui = false;
```
fjorgemota:
```suggestion
		$this->has_ui        = false;
		$this->css_variables = [];
```

Just to initialize everything on constructor (so I think we can avoid some potential notices/warnings here)

fjorgemota (commented):
Just a few improvements for the code. I'll test the code with the other PR soon.

includes/ui/class-ui.php
```diff
@@ -52,14 +60,39 @@ public function enqueue_styles() {
...
+	 * @return string The CSS.
 	 */
-	public static function ensure_styles() {
-		self::instance()->has_ui = true;
+	private function generate_inline_css() {
+
+		$css = ':root{';
+
+		foreach ( $this->css_variables as $name => $value ) {
+			$css .= $name . ': ' . $value . ';';
```
yscik: Best to escape:

```suggestion
			$css .= esc_attr( $name ) . ': ' . esc_attr( $value ) . ';';
```

yscik (commented):
Nice addition! 
I'd separate it from `ensure_styles` and add a new method, something like `set_variables`, to keep it clear what does what. To safely use this in plugins we should also make sure that multiple calls don't overwrite each other, but merge the variables with the existing set.

gikaragia:
> To safely use this in plugins we should also make sure that multiple calls don't overwrite each other, but merge the variables with the existing set.

Good catch! Fixed in 894f4b2c219d890c5fa9527f7712ccfdd12bc0fe

> I'd separate it from ensure_styles and add a new method, something like set_variables, to keep it clear what does what. 

Hmm the thing is that when you want to enqueue the CSS variables you also want to always call `ensure_styles` for the actual enqueue to happen. So if we add another method we will always need to call these two methods one after the other which would complicate things.

yscik (approved):
Looks good!

PR #2686, opened 2023-12-18
Add guest sessions

yscik:

* Add a `Guest_User` class to manage creation and lookup of guest users
* Add `Guest_Session` to 'log in' guest users

Works like this:

`Guest_User( $id/$email )->create_token()`
- Creates a token — this joins together the guest user id and their access token into an URL token so it's just one thing. 

*[ link with token gets sent out, user clicks it ]*

`Guest_Session::get_current_guest()`:
- Checks if there is a token in the URL or in a cookie
- Verifies the token, sets the guest user if it checks out
- Sets a cookie if it's a new session
- Returns the guest user object

Cookie is the same as the URL token, which keeps things simple.

### Testing Instructions

* See 420-gh-Automattic/wpjm-addons for the whole flow

### Release Notes

* Add support for user sessions without a full account

----

| Plugin build for c1bac6971c264d913e6f8516a0ebe013d83c3638 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2686-c1bac697.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2686-c1bac697)             |

### DIFF
--- a/assets/css/ui.notice.scss
+++ b/assets/css/ui.notice.scss
@@ -8,7 +8,7 @@
 	border: var(--jm-ui-border-size) solid var(--jm-ui-border-light);
 	background: var(--jm-ui-background-color, transparent);
 	color: var(--jm-ui-text-color, inherit);
-	padding: var(--jm-ui-space-m);
+	padding: var(--jm-ui-space-m) var(--jm-ui-space-ml);
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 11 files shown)


gikaragia (approved):
Looks great!

PR #2687, opened 2023-12-18
Replace strings with constants in WP_Job_Manager_Post_Types

### LINKED ISSUE #2663: Replace strings with constants in class-wp-job-manager-post-types.php
In order to make the code cleaner and less error-prone:
* Replace custom taxonomies and post type names strings with a public constant (prefixed PT_ for post type, TAX_ for taxonomies and CAP_ for capacities)
* Update the code to use these constants instead of strings

fjorgemota:
Fixes #2663

* Replace strings with constants in `WP_Job_Manager_Post_Types`

### Testing Instructions

Test if the plugin is generally work. Some examples:

1. Create a Job Type
2. Create job categories
3. Create job listings using the data created above
4. Update/delete job categories, types and listings
5. Try to uninstall the plugin
6. etc.

### Release Notes

- [New] Add constants to refer to IDs of post types, capabilities and taxonomies

----

| Plugin build for df7b39fdeac43846b95aed26e4e6c35ec689df17 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2687-df7b39fd.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2687-df7b39fd)             |

### DIFF
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -112,7 +112,7 @@ class WP_Job_Manager_CPT {
 		 *
 		 * @param array $actions_handled {
 		 *     Bulk actions that can be handled, indexed by a unique key name (approve_jobs, expire_jobs, etc). Handlers
-		 *     are responsible for checking abilities (`current_user_can( 'manage_job_listings', $post_id )`) before
+		 *     are responsible for checking abilities (`current_user_can( \WP_Job_Manager_Post_Types::CAP_MANAGE_LISTINGS, $post_id )`) before
 		 *     performing action.
 		 *
 		 *     @type string   $label   Label for the bulk actions dropdown. Passed through sprintf with label name of job listing post type.
@@ -206,7 +206,7 @@ class WP_Job_Manager_CPT {
 			'post_status' => 'expired',
 		];
 		if (
-			current_user_can( 'manage_job_listings', $post_id )
+			current_user_can( \WP_Job_Manager_Post_Types::CAP_MANAGE_LISTINGS, $post_id )
 			&& wp_update_post( $job_data )
 		) {
 			return true;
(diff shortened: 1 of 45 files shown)


includes/admin/class-wp-job-manager-cpt.php
```diff
@@ -334,11 +334,8 @@ public function jobs_by_category() {
 				'class'    => [],
 			],
 		];
-
-		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- No changes or data exposed based on input.
-		$selected_category = isset( $_GET['job_listing_category'] ) ? sanitize_text_field( wp_unslash( $_GET['job_listing_category'] ) ) : '';
-		echo "<select name='job_listing_category' id='dropdown_job_listing_category'>";
-		echo '<option value="" ' . selected( $selected_category, '', false ) . '>' . esc_html__( 'Select category', 'wp-job-manager' ) . '</option>';
+		echo "<select name='" . esc_attr( \WP_Job_Manager_Post_Types::TAX_LISTING_CATEGORY ) . "' id='dropdown_job_listing_category'>";
```
mikeyarce, line 337: Do we also want to use the constant in a place like this?  `id='dropdown_job_listing_category'`?

mikeyarce (approved):
Thanks for taking on this task, @fjorgemota ! 
I looked through every replacement and double checked them and they all look good!  I left a non-blocking question about if we want to use Constants in some class names.

includes/admin/class-wp-job-manager-cpt.php
fjorgemota: I thought about doing so, but I think the main issue with that is that `id` is the sort of attribute that might be used from the CSS/JS side to handle the field somehow (e.g., select2 stuff, for instance). 

And given that passing these constants to the JS side just to handle this looks a bit..too much, I decided just to leave it as it is. Right now, this whole "replacing magical strings with constants" looks already challenging enough (in the sense of risk of breaking things)

PR #2689, opened 2023-12-20
Hide the 'Add new' menu with CSS

### LINKED ISSUE #2673: Permission issue adding jobs with custom role
Permission issue adding jobs with custom role

**Describe the bug**
We are having an issue with adding a job with a custom role. We have a custom role called Job Poster and that role has all the capabilities added, that WP Job Manager creates using a role editor plugin. Now after an update, the user with that role cannot add a new job. They can still edit and delete. Here is the troubleshooting I have done:

    Disabled all plugins except WP Job Manager and Role editor plugin. Same issue.
    Tried a different role editor plugin and made sure capabilities are added. Same issue.
    Switched to the 2024 theme. Same issue.

**To Reproduce**
1. install WP Job Manager and a role editor plugin
2. set up a custom role with all the job manager capabilities
3. log in as a user with that role and try to add a job. You should get a permission issue.

**Screenshots**
screenshot of admin
screenshot of error

**Expected behavior**
To be able to add a job rather than get the permission issue warning

- [X ] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [X ] This bug happens with a default WordPress theme active.
- [X ] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
- WordPress Version:  6.4.2
- PHP Version: 8.0.28
- Other important details:

gikaragia:
Fixes #2673

* It seems that Core WP expects the 'Add New' menu to exist in order to check if the user has the capabilities to add a new post.
* This happens in `user_can_access_admin_page` if you want to take a look. For our case, the 'Add New' page should have a parent page for the appropriate capability to be checked in line 2178
* I think that the only viable option is to re-add the menu and hide it with CSS.

### Testing Instructions

* Follow the steps in the linked ticket and ensure that the user can access the 'Add new' job listing page.
* The user needs to have only the job listing related rights.
* WooCommerce should not be active as Woo redirects the user to the 'My account' page if he doesn't have 'edit_posts' rights.

----

| Plugin build for 3a72581fb398c9facbf9228286678381e9d738f3 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2023/12/wp-job-manager-zip-2689-3a72581f.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2023/12/2689-3a72581f)             |

### DIFF
--- a/assets/css/menu.scss
+++ b/assets/css/menu.scss
@@ -10,6 +10,10 @@
 			@include jm-icon();
 			content: "\e830";
 		}
+
+		a[href="post-new.php?post_type=job_listing"] {
+			display: none;
+		}
 	}
 
 	#menu-posts-resume {
--- a/includes/admin/class-wp-job-manager-admin.php
+++ b/includes/admin/class-wp-job-manager-admin.php
@@ -175,7 +175,6 @@ class WP_Job_Manager_Admin {
 	 * Adds pages to admin menu.
 	 */
 	public function admin_menu() {
-		remove_submenu_page( 'edit.php?post_type=job_listing', 'post-new.php?post_type=job_listing' );
 		$item = remove_submenu_page( 'edit.php?post_type=job_listing', 'edit.php?post_type=job_listing' );
 		if ( ! $item ) {
 			return;


fjorgemota (approved):
Looks good and works well!

Nice catch and nice fix, thanks for all the explanation. TIL about that WP behavior regarding menus pages.

PR #2695, opened 2024-01-05
Bound listing duration between 1 and 100 years, or no expiration

### LINKED ISSUE #2690: Editing Job Listing breaks if Listing Duration is set too large.
**Describe the bug**
When the listing duration in Settings > Job Manager is set too large, such as 4500000000000000000 days, the job editor will no longer work. Editing Product Package will also break, if WC Paid Listing is used.

This fatal error occurs:

```
Ein Fehler vom Typ E_ERROR wurde in der Zeile 1493 der Datei /var/www/clients/client1/web66/web/wp-content/plugins/wp-job-manager/wp-job-manager-functions.php verursacht. Fehlermeldung: Uncaught Exception: Unknown or bad format (P4500000000000000000D) in /var/www/clients/client1/web66/web/wp-content/plugins/wp-job-manager/wp-job-manager-functions.php:1493

Stack trace:

#0 /var/www/clients/client1/web66/web/wp-content/plugins/wp-job-manager/wp-job-manager-functions.php(1493): DateInterval->__construct()
```

**To Reproduce**
1. Go to Settings > Job Submission
2. Set the listing duration to 4500000000000000000 days
3. Edit a job listing

**Expected Behaviour**

There should be a maximum limit to how long the listing duration can be set.

- [ x] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [ x] This bug happens with a default WordPress theme active.
- [ x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
- WordPress Version: 6.4.2
- PHP Version: 8.1.26
- Other important details: 7430553-zen

thedebian:
Fixes #2690

This PR fixes an issue with the listing duration being a free text setting.
It is now a number, bound from 1 to 100 years.
If a number if below or above, it will be reset to "no expiration".
It's also adding a placeholder "No limit" to the field.

### Testing Instructions

* Don't apply PR
* Follow test steps in #2690 
* Should break
* Apply PR
* Do it again
* ✅ It should be set to "no limit"
* Update limit to a number between 1 and 36500
* ✅ The limit should be properly set
* Update it to a number out of these bounds (eg. 0, -1, 36501)
* ✅ It should be set to "no limit"
* The field should also only accept numbers, you can try manually setting the field to some random text (you'll have to bypass browser submit checks), it should be set to "no limit"

----

| Plugin build for 008c0fbb1105b54c53cac9be799d70300cdb4b08 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2695-008c0fbb.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2695-008c0fbb)             |

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -384,11 +384,14 @@ class WP_Job_Manager_Settings {
 							'attributes' => [],
 						],
 						[
-							'name'       => 'job_manager_submission_duration',
-							'std'        => '30',
-							'label'      => __( 'Listing Duration', 'wp-job-manager' ),
-							'desc'       => __( 'Listings will display for the set number of days, then expire. Leave this field blank if you don\'t want listings to have an expiration date.', 'wp-job-manager' ),
-							'attributes' => [],
+							'name'              => 'job_manager_submission_duration',
+							'std'               => '30',
+							'label'             => __( 'Listing Duration', 'wp-job-manager' ),
+							'desc'              => __( 'Listings will display for the set number of days, then expire. Leave this field blank if you don\'t want listings to have an expiration date.', 'wp-job-manager' ),
+							'type'              => 'number',
+							'attributes'        => [],
+							'sanitize_callback' => [ $this, 'sanitize_submission_duration' ],
+							'placeholder'       => __( 'No limit', 'wp-job-manager' ),
 						],
 						[
 							'name'       => 'job_manager_renewal_days',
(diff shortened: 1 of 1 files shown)


yscik (approved):
Looks good. Could the same issue happen with the 'Listing Limit' setting? Looks like I can even save it as a string:

image

thedebian:
> Looks good. Could the same issue happen with the 'Listing Limit' setting? Looks like I can even save it as a string:

Good catch, I'll check other settings and update them too 👍

yscik (approved):
👍

PR #2700, opened 2024-01-12
Add a method that checks if the current guest has a normal account

gikaragia:

* Adds a method that checks if the email in the guest account exists in a normal account

### Testing Instructions

* To be tested with: https://github.com/Automattic/wpjm-addons/pull/436

----

| Plugin build for 38d5ed59b2391d26135d5eb56d0095f97148545f <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2700-38d5ed59.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2700-38d5ed59)             |

### DIFF
--- a/includes/class-guest-session.php
+++ b/includes/class-guest-session.php
@@ -52,14 +52,13 @@ class Guest_Session {
 	 * @return bool
 	 */
 	public static function current_guest_has_account() : bool {
-		$session = self::instance();
-		$session->init();
+		$user = self::get_current_guest();
 
-		if ( false === $session->user ) {
+		if ( false === $user ) {
 			return false;
 		}
 
-		return (bool) get_user_by( 'email', $session->user->user_email );
+		return (bool) get_user_by( 'email', $user->user_email );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-guest-session.php
```diff
@@ -46,6 +46,22 @@ public static function get_current_guest() {
...
+	 */
+	public static function current_guest_has_account() : bool {
+		$session = self::instance();
+		$session->init();
+
+		if ( false === $session->user ) {
+			return false;
+		}
+
+		return (bool) get_user_by( 'email', $session->user->user_email );
```
fjorgemota: Just a suggestion to improve a bit the code, increasing code re-use. :) 

```suggestion
		$user = self::get_current_guest();

		if ( false === $user ) {
			return false;
		}

		return (bool) get_user_by( 'email', $user->user_email );
```

fjorgemota (approved):
Looks good and works well! I left a suggestion to improve the code and re-use just a bit. :)

PR #2706, opened 2024-01-16
Add modal dialogs

yscik:
Part of 186-gh-Automattic/wpjm-addons

Includes some related CSS fixes that came up along the way.

* Add a new modal dialog class that can open a popup modal in an overlay
* Uses the native `<dialog>` element
* Add styling and scripts to open and close the dialog
* Move styles and code around for some elements (buttons), to share it between notice, dialog and other future classes
* Fix the buttons and links not picking up theme colors
* Fix alignment issue in hint notices (438-gh-Automattic/wpjm-addons#issuecomment-1892794397), and add a new variation for 436-gh-Automattic/wpjm-addons#discussion_r1453638803

### Testing Instructions

* Add the `[job_manager_ui_test]` shortcode to a page, publish it and open it
* First element should have an 'Open Modal' button, use it to open the modal dialog
* Clicking outside, on the X or the Close button should close the modal
* Double-check that the rest of the notices keep working

### Release Notes

* Add support for modal dialogs

### Screenshot / Video

Desktop:

image

Mobile:

image

----

| Plugin build for ab543ddaa4f796fbe2019c38a2dc35dada63ac1e <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2706-ab543dda.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2706-ab543dda)             |

### DIFF
--- a/includes/ui/class-notice.php
+++ b/includes/ui/class-notice.php
@@ -105,7 +105,7 @@ class Notice {
 	 * @type string $title Optional title.
 	 * @type string $icon Notice icon. Should be a supported icon name, or safe SVG code.
 	 * @type string $message Main notice message.
-	 * @type string $details Additional HTML content for the notice. Displayed below the message.
+	 * @type string $html Raw HTML content. User input must be escaped by the caller function.
 	 * @type array  $buttons Action links styled as buttons.
 	 * @type array  $links Action links.
 	 * @type array  $classes Additional classes for the notice. Possible options:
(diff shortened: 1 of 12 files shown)


yscik:
Fixed style loading for non-block themes in ead3c17, and rebased to latest trunk.

includes/ui/class-notice.php
```diff
@@ -32,10 +32,10 @@ public static function success( $args ) {
 
 		return self::render(
 			array_merge(
+				$args,
```
pgk: If these are flipped it means that passing an icon change in $args will not take precedence. Is this somethihg we want?

includes/ui/class-ui-elements.php
```diff
@@ -0,0 +1,176 @@
+<?php
+/**
+ * Notice component for the frontend.
```
pgk: I think this "Notice" comment is not relevant to what the class does. It seems to be encapsulating various UI utils and elements.

pgk (approved):
Modal worked well, left some nitpicky comments, feel free to address as needed

includes/ui/class-notice.php
yscik: Good catch, thanks. Did this to fix the `classes` merges, tweaking it a bit more in ab543dd so the icon overwrites keep working too.

PR #2708, opened 2024-01-18
Add supporting changes for the alert modal

### LINKED ISSUE #1749: The "RSS", "Reset" and "Add Alert" links when no search result

yscik:
Part of 186-gh-Automattic/wpjm-addons

* Always show "RSS", "Add alert" links in search so feeds or alerts can be used without searching for a keyword first. (Fixes #1749 — looks like the previous fix wasn't really working)
* Allow adding `onclick` handlers to these links
* Allow adding custom styles to the modal dialog
* Add form styles. This is not used anywhere yet, and for now it'll only be for the form inside the new alert modal. Later this will be the drop-in form styles for the various WPJM and add-on forms when the theme is not styling forms. 

### Testing Instructions

* See 440-gh-Automattic/wpjm-addons
* To test the whole form stylesheet, add the `jm-form` class manually to a `<form>`, like the submit job or add alert forms. 

### Release Notes

* Fix RSS, Reset, Add Alert links not showing on search page without a keyword

### Screenshot / Video

For context these are the links in question:

image

----

| Plugin build for 0722e1d5d8c010930de5003ba1b148be1fee6873 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2708-0722e1d5.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2708-0722e1d5)             |

### DIFF
--- a/assets/css/ui.dialog.scss
+++ b/assets/css/ui.dialog.scss
@@ -4,7 +4,6 @@
 }
 
 .jm-dialog-open {
-
 	position: fixed;
 	top: 0;
 	left: 0;
@@ -21,6 +20,7 @@
 }
 
 .jm-dialog {
+	font-size: var(--jm-ui-font-size);
 	--jm-local-notice-padding: var(--jm-ui-space-l);
 }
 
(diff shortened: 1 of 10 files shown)


pgk (approved):
Tested well along with the Alerts PR. I changed the form class and it looked like this:

Screenshot 2024-01-19 at 1 49 58 PM

I also see that there are some linting failures, oncee these are fixed this seems good to go.

PR #2719, opened 2024-01-30
Add Stats Schema

### LINKED ISSUE #2715: Create schema for stats data on upgrade

fjorgemota:
Fixes #2715

* Add logic to initialize alias for stats table on wpdb 
* Add method to migrate the table for the latest version (and run it on plugin activation!)

### Testing Instructions

Check if the plugin can be activated on a new WP install just fine, without any errors on the error log. Also check if the `wp_job_manager_stats` is created as appropriate

### Release Notes

* 

*

### Deprecated Code

*

### Screenshot / Video

----

| Plugin build for dd028a7782d2800439fa3a2fb2c41248697c3bc8 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2719-dd028a77.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2719-dd028a77)             |

### DIFF
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -28,7 +28,18 @@ class Stats {
 	 * Do initialization of all the things needed for stats.
 	 */
 	public function init() {
-		// Do init stuff.
+		$this->initialize_wpdb();
+	}
+
+	/**
+	 * Initialize the alias for the stats table on the wpdb object.
+	 *
+	 * @return void
+	 */
+	private function initialize_wpdb() {
+		global $wpdb;
+		$wpdb->job_manager_stats = $wpdb->prefix . 'job_manager_stats';
+		$wpdb->tables[]          = 'job_manager_stats';
 	}
 
 	/**
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -38,6 +38,9 @@ class Stats {
 	 */
 	private function initialize_wpdb() {
 		global $wpdb;
+		if ( isset( $wpdb->job_manager_stats ) ) {
+			return;
+		}
 		$wpdb->job_manager_stats = $wpdb->prefix . 'job_manager_stats';
 		$wpdb->tables[]          = 'job_manager_stats';
 	}
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -64,7 +64,7 @@ class Stats {
 				`group` varchar(255) DEFAULT '',
 				`count` bigint(20) unsigned not null default 1,
 				PRIMARY KEY (`id`),
-				KEY `name_date_group`  (`name`, `date`, `group`)
+				INDEX `idx_wpjm_stats_name_date_group`  (`name`, `date`, `group`)
 			) {$collate}",
 			]
 		);
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -55,13 +55,13 @@ class Stats {
 			[
 				"CREATE TABLE {$wpdb->job_manager_stats} (
 				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
-				`date` DATE NOT NULL CURRENT_DATE
+				`date` DATE NOT NULL,
 				`post_id` bigint(20) DEFAULT NULL,
 				`name` varchar(255) NOT NULL,
 				`group` varchar(255) DEFAULT '',
 				`count` bigint(20) unsigned not null default 1,
-				PRIMARY KEY (id),
-				KEY name_date_group (`name`, `date`, `group`)
+				PRIMARY KEY (`id`),
+				KEY `name_date_group`  (`name`, `date`, `group`)
 			) {$collate}",
 			]
 		);
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -55,7 +55,7 @@ class Stats {
 			[
 				"CREATE TABLE {$wpdb->job_manager_stats} (
 				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
-				`date` DATE NOT NULL,
+				`date` date NOT NULL,
 				`post_id` bigint(20) DEFAULT NULL,
 				`name` varchar(255) NOT NULL,
 				`group` varchar(255) DEFAULT '',
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -43,6 +43,31 @@ class Stats {
 	}
 
 	/**
+	 * Migrate the stats table to the latest version.
+	 *
+	 * @return void
+	 */
+	public function migrate() {
+		global $wpdb;
+		$collate = $wpdb->get_charset_collate();
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		\dbDelta(
+			[
+				"CREATE TABLE {$wpdb->job_manager_stats} (
+				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+				`date` DATE NOT NULL CURRENT_DATE
+				`post_id` bigint(20) DEFAULT NULL,
+				`name` varchar(255) NOT NULL,
+				`group` varchar(255) DEFAULT '',
+				`count` bigint(20) unsigned not null default 1,
+				PRIMARY KEY (id),
+				KEY name_date_group (`name`, `date`, `group`)
+			) {$collate}",
+			]
+		);
+	}
+
+	/**
 	 * Perform plugin activation-related stats actions.
 	 *
 	 * @return void
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -38,11 +38,11 @@ class Stats {
 	 */
 	private function initialize_wpdb() {
 		global $wpdb;
-		if ( isset( $wpdb->job_manager_stats ) ) {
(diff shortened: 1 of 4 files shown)


includes/class-stats.php
```diff
@@ -28,7 +28,43 @@ private function __construct() {
...
+	 */
+	public function migrate() {
+		global $wpdb;
+		$collate = $wpdb->get_charset_collate();
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		\dbDelta(
+			[
+				"CREATE TABLE {$wpdb->job_manager_stats} (
+				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+				`date` DATE NOT NULL,
```
fjorgemota: I had to remove the `CURRENT_DATE` here because apparently it is not supported by MySQL older than the 8.0.20

includes/class-stats.php
```diff
@@ -28,7 +28,43 @@ private function __construct() {
...
+
+	/**
+	 * Initialize the alias for the stats table on the wpdb object.
+	 *
+	 * @return void
+	 */
+	private function initialize_wpdb() {
+		global $wpdb;
+		$wpdb->job_manager_stats = $wpdb->prefix . 'job_manager_stats';
+		$wpdb->tables[]          = 'job_manager_stats';
```
fjorgemota: The reason for including the table name on `tables` is so WP automatically updates the attribute with the proper name with the prefix when switching to another blog on a multisite installation.

includes/class-stats.php
```diff
@@ -28,7 +28,43 @@ private function __construct() {
...
+	/**
+	 * Migrate the stats table to the latest version.
+	 *
+	 * @return void
+	 */
+	public function migrate() {
+		global $wpdb;
+		$collate = $wpdb->get_charset_collate();
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		\dbDelta(
```
fjorgemota, line 57: This function will run automatically a diff for us on the SQL and do any changes on the database side.

includes/class-stats.php
```diff
@@ -0,0 +1,80 @@
...
+			[
+				"CREATE TABLE {$wpdb->job_manager_stats} (
+				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+				`date` date NOT NULL,
+				`post_id` bigint(20) DEFAULT NULL,
+				`name` varchar(255) NOT NULL,
+				`group` varchar(255) DEFAULT '',
+				`count` bigint(20) unsigned not null default 1,
+				PRIMARY KEY (`id`),
+				KEY `name_date_group`  (`name`, `date`, `group`)
```
pgk: Can we use index instead of `KEY`? They are [synonyms](https://dev.mysql.com/doc/refman/8.3/en/create-table.html#create-table-indexes-keys), but `INDEX` is the most iso-compatible one. (and wp now wants to play better with sqlite). Also nit,  prefixing indices with idx_{table_name}.

```suggestion
				INDEX `idx_wpjm_stats_name_date_group`  (`name`, `date`, `group`)
```

includes/class-stats.php
```diff
@@ -0,0 +1,80 @@
...
+	 *
+	 * @return void
+	 */
+	public function migrate() {
+		global $wpdb;
+		$collate = $wpdb->get_charset_collate();
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		\dbDelta(
+			[
+				"CREATE TABLE {$wpdb->job_manager_stats} (
```
pgk:
```suggestion
				"CREATE TABLE IF NOT EXISTS {$wpdb->job_manager_stats} (
```

includes/class-wp-job-manager-data-cleaner.php
```diff
@@ -28,6 +28,15 @@ class WP_Job_Manager_Data_Cleaner {
 		\WP_Job_Manager_Post_Types::PT_GUEST_USER,
 	];
 
+	/**
+	 * Custom tables to be deleted.
+	 *
+	 * @var $custom_tables
+	 */
+	private static $custom_tables = [
```
pgk: should this be a const?

includes/class-stats.php
pgk: Nice!

includes/class-wp-job-manager-data-cleaner.php
```diff
@@ -231,6 +241,29 @@ private static function cleanup_custom_post_types() {
...
+	private static function cleanup_custom_tables() {
+		global $wpdb;
+
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery -- We need to delete the custom tables.
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching -- We don't cache DROP TABLE.
+		// phpcs:disable WordPress.DB.PreparedSQLPlaceholders.UnsupportedPlaceholder -- %i is supported since WP 6.2.
+		// phpcs:disable WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare -- %i is supported since WP 6.2.
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.SchemaChange -- We really need to delete the custom tables.
+		foreach ( self::$custom_tables as $custom_table ) {
+			$wpdb->query( $wpdb->prepare( 'DROP TABLE IF EXISTS %i', $wpdb->prefix . $custom_table ) );
```
pgk, line 258: nice with the `IF EXISTS`!

includes/class-wp-job-manager-install.php
```diff
@@ -76,6 +76,8 @@ public static function install() {
 			update_option( 'job_manager_permalinks', wp_json_encode( $permalink_options ) );
 		}
 
+		\WP_Job_Manager\Stats::instance()->migrate();
```
pgk, line 79: Does it make sense to move this call to the activate() method of Stats instead? Note that this will mean that migrate will be called earlier, but I think it is ok to do so.

pgk (changes requested):
It worked well! I left some comments, mainly around compat and for moving some things around a bit. 

Edit: Forgot to mention something important: The table name we use is the same table name used by Astoundify stats. I think that this might be a problem, specially since enabling the plugin might dbDelta an existing astoundify table and possibly corrupt existing stats.

includes/class-wp-job-manager-data-cleaner.php
fjorgemota: All the other properties in the class are `private static`, hence why I put this as `private static` too.

I agree that changing everything to `private const` might make more sense, but I'd change all the other properties too. Do you think it is worth it, @pgk?

includes/class-stats.php
fjorgemota: I know this isn't clear at first, but `dbDelta` already handles that for us, and I think it doesn't support having `IF NOT EXISTS` on the table name (it will consider the `IF NOT EXISTS`  part of the table name, which I don't think we want to have). See how it handles that here: https://github.com/WordPress/WordPress/blob/27348531f3fdf048809da977fced3f40b90259de/wp-admin/includes/upgrade.php#L2836-L2840

You can [see here](https://github.com/WordPress/WordPress/blob/27348531f3fdf048809da977fced3f40b90259de/wp-admin/includes/schema.php#L56-L237) that no WP tables use `IF NOT EXISTS` when creating the schema (those are passed to `dbDelta`, too). So I'm inclined to not add them. WDYT?

includes/class-stats.php
fjorgemota: Good catch on the index name. Regarding using `INDEX`, for me it is kinda indifferent. WP apparently converts everything to `KEY` [anyway on the `dbDelta` function](https://github.com/WordPress/WordPress/blob/27348531f3fdf048809da977fced3f40b90259de/wp-admin/includes/upgrade.php#L2973-L2974), and it also uses [`KEY` everywhere on the schema](https://github.com/WordPress/WordPress/blob/27348531f3fdf048809da977fced3f40b90259de/wp-admin/includes/schema.php#L56-L237), but oh well...

includes/class-wp-job-manager-install.php
fjorgemota: I don't think that makes sense, @pgk. Because `activate()` AFAIK will not run when the plugin is updated (e.g. from an old WPJM installation), while this method will be run both on new plugin installations as well as plugin upgrades.

Does that make sense?

fjorgemota:
> Edit: Forgot to mention something important: The table name we use is the same table name used by Astoundify stats. I think that this might be a problem, specially since enabling the plugin might dbDelta an existing astoundify table and possibly corrupt existing stats.

Great catch!

I renamed it in 5f63042 . Thanks!

includes/class-wp-job-manager-data-cleaner.php
pgk: I think that if this property is always constant it makes sense for it to be const.

fjorgemota: I changed that specific property in dd028a7. As mentioned on Slack, I'll submit another PR to change the other properties too. :)

pgk (approved):
Table name change looks good, and activating/deactivating logic worked well!

PR #2724, opened 2024-01-30
Refactor WPJM Data Cleaner to use constants everywhere

fjorgemota:
Follow-up to #2719 

* Use `private const` instead of `private static` to save the properties of the WP Job Manager Data Cleaner class, for consistency

### Testing Instructions

* Check if the tests still pass
* Check if you can disable/uninstall the plugin correctly while having the option to clean your data enabled

### Release Notes

----

| Plugin build for e1c4b08d56ef828f1bc84c81f2a4fd26e52d271d <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2724-e1c4b08d.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2724-e1c4b08d)             |

### DIFF
--- a/includes/class-wp-job-manager-data-cleaner.php
+++ b/includes/class-wp-job-manager-data-cleaner.php
@@ -151,9 +151,9 @@ class WP_Job_Manager_Data_Cleaner {
 	/**
 	 * Capabilities to be deleted.
 	 *
-	 * @var $caps
+	 * @var array
 	 */
-	private static $caps = [
+	private const CAPS = [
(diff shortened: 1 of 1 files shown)


pgk (approved):
I like it! Not a stopper, but how do you feel about merging this to trunk instead of stats feature?

fjorgemota:
@pgk Thanks for your review and suggestion! I just did what you suggested, do you think it needs an additional review?

PR #2728, opened 2024-01-31
Add/stats log stat method

### LINKED ISSUE #2716: Create log_stat function to increment counters related to stat

pgk:
Fixes #2716

* Introduce a log_stat function. Also adds an example of how it could be used by incrementing listing page views.

Note:

I decided to do the stat update in one go by using `ON DUPLICATE UPDATE`. In order for duplicate to trigger, I introduced a unique index and I changed the stat table definition a bit for this. LMK if the new structure seems reasonable.

I crammed it all into the Stats class. They should probably live in separate classes eventually, but I think it is convenient for reading all the changes in one go and for not overengineering without reason.

The DB changes have testing implications. You probs need to drop the stats table entirely. dBdelta does not do a good job with handling e.g. the default field change or the unique index change.

TBD: Should we drop the other index entirely? 

### Testing Instructions

* First off, drop the existing stats table.
* deactivate/activate Job Manager.
* Check that the db schema has a unique index and post_id defaults to 0 now.
* Visit a generic page. No stats entries.
* Visit a job listing page. Stats entry.
* Refresh. Stat increments.
* Visit different listing. New stat entry,

### Release Notes

* N/A

* N/A

### Deprecated Code

* N/A

### Screenshot / Video

----

| Plugin build for 074eccb57c9466bc86b138ce7ea9207f9260f11b <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2728-074eccb5.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2728-074eccb5)             |

### DIFF
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -66,12 +66,11 @@ class Stats {
 				"CREATE TABLE {$wpdb->wpjm_stats} (
 				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
 				`date` date NOT NULL,
-				`post_id` bigint(20) DEFAULT 0,
+				`post_id` bigint(20) DEFAULT NULL,
 				`name` varchar(255) NOT NULL,
 				`group` varchar(255) DEFAULT '',
 				`count` bigint(20) unsigned not null default 1,
 				PRIMARY KEY (`id`),
-				INDEX `idx_wpjm_stats_name_date_group`  (`name`, `date`, `group`),
 				UNIQUE INDEX `idx_wpjm_stats_name_date_group_post_id`  (`name`, `date`, `group`, `post_id`)
 			) {$collate}",
 			]
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -29,6 +29,7 @@ class Stats {
 	 */
 	public function init() {
 		$this->initialize_wpdb();
+		$this->hook();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
...
+	 * Log a stat into the db.
+	 *
+	 * @param string $name         The stat name.
+	 * @param string $group        The group this stat belongs to.
+	 * @param int    $post_id      The post_id this stat belongs to.
+	 * @param int    $increment_by The amount to increment the stat by.
+	 *
+	 * @return bool
+	 */
+	public function log_stat( string $name, string $group = '', int $post_id = 0, int $increment_by = 1 ) {
```
yscik: Since there are a lot of optional parameters that we'll be using in various combinations, WDYT of using an `log_stat( string $name, array $args = [] )` signature for this?

includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
...
+				"INSERT INTO {$wpdb->wpjm_stats} " .
+				'(`name`, `date`, `group`, `post_id`, `count` ) ' .
+				'VALUES (%s, %s, %s, %d, %d) ' .
+				'ON DUPLICATE KEY UPDATE `count` = `count` + %d',
+				$name,
+				$date_today,
+				$group,
+				$post_id,
+				$increment_by,
+				$increment_by
```
yscik, line 124: Positional placeholders are supposedly supported here so we can do something like ``` `count` = `count` + %5$d' ``` and only pass the argument once.

includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
...
+			)
+		);
+
+		if ( false === $result ) {
+			return false;
+		}
+
+		$cache_key = $this->get_cache_key_for_stat( $name, $group, $post_id );
+
+		wp_cache_delete( $cache_key, 'wpjm_stats' );
```
yscik, line 134: I think we are not really better off caching individual rows — or do you imagine this being used to invalidate cache for more complex queries?

includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
...
 	public function activate() {
 	}
+
+	/**
+	 * Run any hooks related to stats.
+	 *
+	 * @return void
+	 */
+	private function hook() {
+		add_action( 'wp', [ $this, 'maybe_log_listing_view' ] );
```
yscik, line 167: We can add this to a later hook where the post is already loaded. I'd also check later if putting the db writes in the `shutdown` hook works, that'd be great for performance.

includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
...
 				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
 				`date` date NOT NULL,
-				`post_id` bigint(20) DEFAULT NULL,
+				`post_id` bigint(20) DEFAULT 0,
 				`name` varchar(255) NOT NULL,
 				`group` varchar(255) DEFAULT '',
 				`count` bigint(20) unsigned not null default 1,
 				PRIMARY KEY (`id`),
-				INDEX `idx_wpjm_stats_name_date_group`  (`name`, `date`, `group`)
+				INDEX `idx_wpjm_stats_name_date_group`  (`name`, `date`, `group`),
```
fjorgemota: > TBD: Should we drop the other index entirely?

I think this index really doesn't make a ton of sense given the new unique index.

yscik (approved):
Looks great! Added some notes but mostly for future PRs.

includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
 				"CREATE TABLE {$wpdb->wpjm_stats} (
 				`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
 				`date` date NOT NULL,
-				`post_id` bigint(20) DEFAULT NULL,
+				`post_id` bigint(20) DEFAULT 0,
```
fjorgemota: what's the reason for this change? I think `NULL` might make more sense here?

If we really want this, maybe it makes more sense to set the field as `NOT NULL`?

includes/class-stats.php
fjorgemota: WDYT about having `$name` as the main parameter and all the others are items in an array? 

Like the API that Peter suggested here: p6r3EZ-201-p2#comment-3505

That way, we can have not only `increment_by` but also a `count` property, which can be used to set the `count` column directly.

WDYT?

includes/class-stats.php
fjorgemota, line 167: I know that this is only an example but...WDYT about maybe using `job_manager_single_job_content` to handle this?

That way you can simplify a lot the code on `maybe_log_listing_view`, given that we know that the post is is definitely a job listing. :)

includes/class-stats.php
```diff
@@ -59,22 +60,115 @@ public function migrate() {
...
+	 * @return void
+	 */
+	public function maybe_log_listing_view() {
+		if ( is_admin() || ( defined( 'DOING_AJAX' ) && DOING_AJAX ) ) {
+			return;
+		}
+
+		$post_id   = absint( get_queried_object_id() );
+		$post_type = get_post_type( $post_id );
+		if ( 'job_listing' !== $post_type ) {
```
fjorgemota: Reminder to use the constants, in case you decide to NOT use the `job_manager_single_job_content` filter:

```suggestion
		if ( \WP_Job_Manager_Post_Types::PT_LISTING !== $post_type ) {
```

fjorgemota (commented):
Hey, this is looking generally good!

I left some comments and suggestions. Let me know what you think about them. :)

I'm also interested in knowing what @yscik thinks about this, so I'm asking his review too. :)

fjorgemota:
LMAO I didn't saw that Peter just shared a review. 

 

Nice :)

includes/class-stats.php
pgk: Yesh! -- this is an example. Like instead of invalidating individual rows, we would invalidate the cached result of a more complex query if the row affects it. In reality, probably we would need to replace this code with an action.

includes/class-stats.php
pgk: I got a warning when doing this, so maybe we can look into it in a different pr.

```
Notice: Function wpdb::prepare was called incorrectly. The query does not contain the correct number of placeholders (6) for the number of arguments passed (5). Please see Debugging in WordPress for more information. (This message was added in version 4.8.3.) in /var/www/html/wp-includes/functions.php on line 6031
```

includes/class-stats.php
pgk: Good idea with the other hook. I will change just that for now.

pgk: Tried it, It didn't work as expected. It broke the rendering of the listing. Leaving as is for now, and we can review leter.

pgk: Leaving as is for now, and we can tweak as needed

fjorgemota: I think it broke because it is not an action, but a filter, so you need to get the first parameter and return it unmodified.

I will submit a PR to change that later. It will be quick. :) 

Thanks!

PR #2738, opened 2024-02-05
Add stats unique listing view and ajax - based endpoint

### LINKED ISSUE #2731: Create an example of storing a unique stat view

pgk:
Fixes #2731

* Introduces an ajax endpoint for recording multiple stats at once.
* Adds ~cookie-based~ localSorage-based unique stat recording.
* Removes logging of stats in the backend. Reason is that pages can be cached.

Notes:

* Caveat: this will not work if JS is disabled. Maybe something we should think about.
* This is not a comprehensive solution for all the different stat use cases we want to cover, but contains the kernels of a solution.
* The general idea of stats being a declarative key-value thing seems appealing and close to how wpjm does things in other places like e.g. forms, or how wp rest api declares endpoints. Also it is easy to extend and refactor, without a class explosion.
* Deriving a stat's unique key should probably be a callback in the future. This is also somethihg that can be handled in this declarative stat style by having certain definition elements accept/be callbacks.
* We might want to optimize inserts, but we can do that later.
* Any better idea for the js cookie funcs? (Edit: we used localStorage instead)

### Testing Instructions

* Visit a job listing page.
* Check that one view and one unique view are added to the db.
* Refresh.
* Check that unique view is not incremented.
* Delete cookie. Refresh.
* Check that one view and one unique view are added to the db.

### Release Notes

* 

*

### Deprecated Code

*

### Screenshot / Video

----

| Plugin build for 83ba197ee2e960b1c48f5af9d788d27de8504fee <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2738-83ba197e.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2738-83ba197e)             |

### DIFF
--- a/assets/js/wpjm-stats.js
+++ b/assets/js/wpjm-stats.js
@@ -60,7 +60,6 @@
 				}
 			}
 		} );
-		// console.log( statsToRecord ); return;
 
 		var postData = new URLSearchParams();
 		postData.append( '_ajax_nonce', ajaxNonce );
--- a/assets/js/wpjm-stats.js
+++ b/assets/js/wpjm-stats.js
@@ -9,30 +9,20 @@
 		}
 	}
 
-	// Cookie funcs are copied verbatim from http://www.quirksmode.org/js/cookies.html, tweaked to include path.
-	function createCookie( name, value, days, path ) {
-		var expires = '';
-		if ( days ) {
+	function updateDailyUnique( key ) {
+		var date = new Date();
+		var expiresAtTimestamp = date.getTime() + 24 * 60 * 60 * 1000;
+		window.localStorage[key] = expiresAtTimestamp;
+	}
+
+	function getDailyUnique( name ) {
+		if ( window.localStorage[name] ) {
 			var date = new Date();
-			date.setTime( date.getTime() + days * 24 * 60 * 60 * 1000 );
-			var expires = '; expires=' + date.toGMTString();
-		} else var expires = '';
-		document.cookie = name + '=' + value + expires + '; path=' + path;
-	}
-
-	function readCookie( name ) {
-		var nameEQ = name + '=';
-		var ca = document.cookie.split( ';' );
-		for ( var i = 0; i < ca.length; i++ ) {
-			var c = ca[ i ];
-			while ( c.charAt( 0 ) == ' ' ) c = c.substring( 1, c.length );
-			if ( c.indexOf( nameEQ ) == 0 ) return c.substring( nameEQ.length, c.length );
+			var now = date.getTime();
+			var expiration = parseInt( window.localStorage[name], 10 );
+			return expiration >= now;
 		}
-		return null;
-	}
-
-	function eraseCookie( name ) {
-		createCookie( name, '', -1 );
+		return false;
 	}
 
 	ready( function () {
@@ -40,10 +30,10 @@
 		var jobStatsSettings = window.job_manager_stats;
 		var ajaxUrl = jobStatsSettings.ajax_url;
 		var ajaxNonce = jobStatsSettings.ajax_nonce;
-		var cookiesToSet = [];
-		var setCookies = function () {
-			cookiesToSet.forEach( function ( uniqueKey ) {
-				createCookie( uniqueKey, 1, 1, '' );
+		var uniquesToSet = [];
+		var setUniques = function () {
+			uniquesToSet.forEach( function ( uniqueKey ) {
+				updateDailyUnique( uniqueKey );
 			} );
 		};
 
@@ -54,24 +44,26 @@
 			if ( uniqueKey.length === 0 ) {
 				statsToRecord.push( statToRecordKey );
 			} else {
-				if ( null === readCookie( uniqueKey ) ) {
-					cookiesToSet.push( uniqueKey );
+				if ( false === getDailyUnique( uniqueKey ) ) {
+					uniquesToSet.push( uniqueKey );
 					statsToRecord.push( statToRecordKey );
 				}
 			}
 		} );
 
-		var postData = new URLSearchParams();
-		postData.append( '_ajax_nonce', ajaxNonce );
-		postData.append( 'post_id', jobStatsSettings.post_id || 0 );
-		postData.append( 'action', 'job_manager_log_stat' );
-		postData.append( 'stats', statsToRecord.join( ',' ) );
+		var postData = new URLSearchParams( {
+			_ajax_nonce: ajaxNonce,
+			post_id: jobStatsSettings.post_id || 0,
+			action: 'job_manager_log_stat',
+			stats: statsToRecord.join( ',' )
+		} );
+
 		fetch( ajaxUrl, {
 			method: 'POST',
 			credentials: 'same-origin',
 			body: postData,
 		} ).finally( function () {
-			setCookies();
+			setUniques();
 			console.log( 'sent' );
 		} );
 	} );
--- a/assets/js/wpjm-stats.js
+++ b/assets/js/wpjm-stats.js
@@ -1,4 +1,4 @@
-/* global wpjm_stats */
+/* global job_manager_stats */
 ( function () {
 	function ready(fn) {
 		if (document.readyState !== 'loading') {
--- a/assets/js/wpjm-stats.js
+++ b/assets/js/wpjm-stats.js
@@ -1,14 +1,8 @@
 /* global job_manager_stats */
-( function () {
-	// From https://youmightnotneedjquery.com/#ready
-	function ready( fn ) {
-		if ( document.readyState !== 'loading' ) {
-			fn();
-		} else {
-			document.addEventListener( 'DOMContentLoaded', fn );
-		}
-	}
 
+import domReady from '@wordpress/dom-ready';
+
+( function () {
 	function updateDailyUnique( key ) {
 		var date = new Date();
 		var expiresAtTimestamp = date.getTime() + 24 * 60 * 60 * 1000;
@@ -25,7 +19,7 @@
 		return false;
 	}
 
-	ready( function () {
+	domReady( function () {
 		var statsToRecord = [];
 		var jobStatsSettings = window.job_manager_stats;
 		var ajaxUrl = jobStatsSettings.ajax_url;
--- a/assets/js/wpjm-stats.js
+++ b/assets/js/wpjm-stats.js
@@ -1,5 +1,6 @@
 /* global job_manager_stats */
 ( function () {
+	// From https://youmightnotneedjquery.com/#ready
 	function ready(fn) {
 		if (document.readyState !== 'loading') {
 			fn();
@@ -8,11 +9,13 @@
 		}
 	}
 
-	function createCookie(name,value,days,path) {
+	// Cookie funcs are copied verbatim from http://www.quirksmode.org/js/cookies.html, tweaked to include path.
+	function createCookie( name, value, days, path ) {
+		var expires = "";
 		if (days) {
 			var date = new Date();
-			date.setTime(date.getTime()+(days*24*60*60*1000));
-			var expires = "; expires="+date.toGMTString();
+			date.setTime( date.getTime() + ( days * 24 * 60 * 60 * 1000 ) );
+			var expires = "; expires=" + date.toGMTString();
 		}
 		else var expires = "";
 		document.cookie = name+"="+value+expires+"; path=" + path;
--- a/assets/js/wpjm-stats.js
+++ b/assets/js/wpjm-stats.js
@@ -64,7 +64,6 @@
 			body: postData,
 		} ).finally( function () {
 			setUniques();
-			console.log( 'sent' );
 		} );
 	} );
 } )();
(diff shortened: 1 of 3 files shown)


assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,79 @@
...
+			if ( uniqueKey.length === 0 ) {
+				statsToRecord.push( statToRecordKey );
+			} else {
+				if ( null === readCookie( uniqueKey ) ) {
+					cookiesToSet.push( uniqueKey );
+					statsToRecord.push( statToRecordKey );
+				}
+			}
+		} );
+		// console.log( statsToRecord ); return;
```
pgk:
```suggestion
```

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
+
+		$post_data = stripslashes_deep( $_POST );
+
+		if ( ! isset( $post_data['_ajax_nonce'] ) || ! wp_verify_nonce( $post_data['_ajax_nonce'], 'ajax-nonce' ) ) {
+			return;
+		}
+
+		$post_id = isset( $post_data['post_id'] ) ? absint( $post_data['post_id'] ) : 0;
+
+		if ( empty( $post_id ) ) {
```
pgk, line 209: We might want to be more relaxed about this later.

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
+			return;
+		}
+
+		$post_type = get_post_type( $post_id );
+
+		if ( \WP_Job_Manager_Post_Types::PT_LISTING !== $post_type ) {
```
pgk, line 215: this can also be generalized later, by adding it to the stat definition.

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
+		$post_id   = absint( get_queried_object_id() );
+		$post_type = get_post_type( $post_id );
+		if ( \WP_Job_Manager_Post_Types::PT_LISTING !== $post_type ) {
+			return;
+		}
+
+		\WP_Job_Manager::register_script(
+			'wp-job-manager-stats',
+			'js/wpjm-stats.js',
+			[ 'jquery' ],
```
pgk, line 255: we don't need jq.

pgk:
```suggestion
			[],
```

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
+	 * @return array
+	 */
+	private function get_stats_for_ajax( $post_id = 0 ) {
+		$ajax_stats = [];
+		foreach ( $this->get_registered_stats() as $stat_name => $stat_data ) {
+			$stat_ajax = [
+				'name' => $stat_name,
+			];
+			if ( ! empty( $stat_data['unique'] ) ) {
+				$stat_ajax['unique_key'] = $stat_name . '_' . $post_id;
```
pgk: in the future, this could be a callback

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,78 @@
...
+					statsToRecord.push( statToRecordKey );
+				}
+			}
+		} );
+
+		var postData = new URLSearchParams();
+		postData.append( '_ajax_nonce', ajaxNonce );
+		postData.append( 'post_id', jobStatsSettings.post_id || 0 );
+		postData.append( 'action', 'job_manager_log_stat' );
+		postData.append( 'stats', statsToRecord.join( ',' ) );
```
pgk: this might become more complex later, and warrant JSON/similar

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,78 @@
...
+	function ready( fn ) {
+		if ( document.readyState !== 'loading' ) {
+			fn();
+		} else {
+			document.addEventListener( 'DOMContentLoaded', fn );
+		}
+	}
+
+	// Cookie funcs are copied verbatim from http://www.quirksmode.org/js/cookies.html, tweaked to include path.
+	function createCookie( name, value, days, path ) {
```
yscik: Since we are only using the data from JS, I think we'd better off with [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) instead of cookies.
- Cookies get sent to the server on every request
- Way more complicated API to set and read them

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,78 @@
+/* global job_manager_stats */
+( function () {
+	// From https://youmightnotneedjquery.com/#ready
+	function ready( fn ) {
+		if ( document.readyState !== 'loading' ) {
```
yscik: WordPress has a built-in package for this: [@wordpress/dom-ready](https://github.com/WordPress/gutenberg/blob/trunk/packages/dom-ready/README.md#)

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,78 @@
...
+				statsToRecord.push( statToRecordKey );
+			} else {
+				if ( null === readCookie( uniqueKey ) ) {
+					cookiesToSet.push( uniqueKey );
+					statsToRecord.push( statToRecordKey );
+				}
+			}
+		} );
+
+		var postData = new URLSearchParams();
```
yscik: Looks like passing an object here, like `new URLSearchParams({ foo: "1", bar: "2" });` [also works.](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams)

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,78 @@
...
+			var c = ca[ i ];
+			while ( c.charAt( 0 ) == ' ' ) c = c.substring( 1, c.length );
+			if ( c.indexOf( nameEQ ) == 0 ) return c.substring( nameEQ.length, c.length );
+		}
+		return null;
+	}
+
+	function eraseCookie( name ) {
+		createCookie( name, '', -1 );
+	}
```
fjorgemota: I was initially expecting to do something like this in the backend, not the frontend.

If we're going to do it in the frontend, WDYT about supporting a more modern API, like `localStorage`, instead of using cookies to save this? The main problem I see with this suggestion is that we would have different "unique views" depending on whether the visit was made using HTTP vs. HTTPS, but given that most site nowadays just redirect to HTTPS anyway, I'm not sure this is a big problem.

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,78 @@
...
+		var ajaxUrl = jobStatsSettings.ajax_url;
+		var ajaxNonce = jobStatsSettings.ajax_nonce;
+		var cookiesToSet = [];
+		var setCookies = function () {
+			cookiesToSet.forEach( function ( uniqueKey ) {
+				createCookie( uniqueKey, 1, 1, '' );
+			} );
+		};
+
+		jobStatsSettings.stats_to_log.forEach( function ( statToRecord ) {
```
yscik: I'd include the stat name in the payload instead of having a fixed list — that'd be hard to scale to other kind of events.

Thinking about how to architect this all, we could add an ajax handler to process the request, and then pass on the event to specialized handlers. Looks best to start separating some parts out to a new `Job_Listing_Stats` class to make reasoning about this easier. 

On the JS side, we'd want to expose some global public `log_stats` function that only handles sending the event to the server. Then we have various entry points that'd call this:

- After page load, a function logs the `job_listing_view` and `job_listing_view_unique` events, depending on cookies/storage. Or with the schema explored below, it calls the `job_listing_view` handler with `$unique` as a parameter, and let the server decide.
- On clicking 'Apply Now', it calls the `job_listing_apply_click` handler

Here is a sketch for the backend parts. I really like the idea of loosely coupling all this with the WordPress hooks approach:

```php

class Stats {

	// Option one, register callbacks for each
	public function get_registered_stats() {
		return apply_filters('wpjm_get_registered_stats', [
			'job_listing_view'  => [ Job_Listing_Stats::class, 'log_view' ],
			'job_listing_apply' => [ Job_Listing_Stats::class, 'log_apply_click' ],
		] );
	}

	public function maybe_log_stat_ajax() {
		
		// process request & everything

		call_user_func( $this->get_registered_stats()[ $params[ 'stat' ] ], $params );

		// The WordPress hooks option
		do_action( "job_manager_stat__{$params[ 'stat' ]}" )

	}
}

class Job_Listing_Stats {

	// Alternatively option, using WordPress hooks
		add_action("job_manager_stat__job_listing_view", [ $this, 'log_view' ] );
		add_action("job_manager_stat__job_listing_apply", [ $this, 'log_apply' ] );
	}

	public function log_view( $post_id, $unique = false ) {

		// checks and whistles

		$stats->log_stat("job_listing_view", $post_id )

		if $unique log_stat("job_listing_view_unique", $post_id )
	}
}
```

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
+		\WP_Job_Manager::register_script(
+			'wp-job-manager-stats',
+			'js/wpjm-stats.js',
+			[],
+			true
+		);
+
+		$script_data = [
+			'ajax_url'     => admin_url( 'admin-ajax.php' ),
+			'ajax_nonce'   => wp_create_nonce( 'ajax-nonce' ),
```
fjorgemota, line 265: I'm not sure if this can be dynamic but..WDYT about having some dynamic info on the nonce, like for instance the `$post_id` and maybe (not 100% sure about this) the list of possibilities of `stats_to_log`?

The  idea is to block an advanced enough user from changing the `$post_id` and spamming the site with "fake visits". 

WDYT?

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
+		}
+
+		$stats = isset( $post_data['stats'] ) ? explode( ',', sanitize_text_field( $post_data['stats'] ) ) : [];
+
+		// TODO: Maybe optimize this into a single insert?
+		foreach ( $stats as $stat_name ) {
+			$stat_name = trim( strtolower( $stat_name ) );
+			if ( ! in_array( $stat_name, $this->get_registered_stat_names(), true ) ) {
+				continue;
+			}
```
fjorgemota, line 228: Nice protection! :)

includes/class-stats.php
```diff
@@ -185,4 +187,126 @@ public function maybe_log_listing_view() {
...
 		$this->log_stat( 'job_listing_view', [ 'post_id' => $post_id ] );
 	}
+
+	/**
+	 * Log multiple stats in one go. Triggered in an ajax call.
+	 *
+	 * @return void
+	 */
+	public function maybe_log_stat_ajax() {
+		if ( ! ( defined( 'DOING_AJAX' ) || ! DOING_AJAX ) ) {
```
fjorgemota: WDYT about using the proper function for this?

```suggestion
		if ( ! wp_doing_ajax() ) {
```

assets/js/wpjm-stats.js
fjorgemota: > I was initially expecting to do something like this in the backend, not the frontend.

If we move this to the backend, then I think we could do the whole cookie management in the backend directly. And instead of a whole JS code, we just enqueue the API dynamically (maybe we can use the REST API to load one of those 1x1 images just for the purpose of tracking?)

fjorgemota (commented):
This is looking good! Left some comments with some thoughts. Thanks! :)

includes/class-stats.php
pgk: we would need to add a separate nonce on top of the ajax nonce which is required by the ajax endpoint. Let's maybe do it in a separate pr if needed.

assets/js/wpjm-stats.js
pgk: Is this available in the frontend or should it be a build dependency?

assets/js/wpjm-stats.js
```diff
@@ -0,0 +1,70 @@
...
+			stats: statsToRecord.join( ',' )
+		} );
+
+		fetch( ajaxUrl, {
+			method: 'POST',
+			credentials: 'same-origin',
+			body: postData,
+		} ).finally( function () {
+			setUniques();
+			console.log( 'sent' );
```
pgk:
```suggestion
```

assets/js/wpjm-stats.js
pgk: I think I like the first version, and steered the pr to that direction. I will leave things in the same class for now and we can update/re-architect as we implement more stats use cases.

assets/js/wpjm-stats.js
yscik: Yes no extra steps needed, webpack will generate a `wpjm-stats.asset.php` for the script with the list of WP dependencies, and we add them to the wp_register_script call in [WP_Job_Manager::register_script](https://github.com/Automattic/WP-Job-Manager/blob/2be6144c8014f988873f9bdea43a3569a7048f9e/includes/class-wp-job-manager.php#L322)

yscik (approved):
As we talked about we'll need to break apart the page-view specific parts from sending the data, but overall it works nicely. I updated the syntax to ES6, which should be used for new files — mostly the `var` -> `const` changes.

PR #2750, opened 2024-02-05
Add job listing stats queries and views column

### LINKED ISSUE #2717: Create a way to query daily stats

### LINKED ISSUE #2740: Stats: Add view column to Job Dashboard

yscik:
Fixes #2717, fixes #2740

* Add a `Job_Listing_Stats` class that handles queries for stats of a single job listing
* Add methods to sum up total stats, and to get daily breakdown
* Add a stats column with the total views in the [job_dashboard] shortcode and in the job listing admin screen

### Testing Instructions

* Enable stats in the setting
* Visit a few listings
* View the frontend job dashboard and see that the views are counted
* Also check the same thing in the WP-Admin Job Listings screen

----

| Plugin build for 2c2a7e38f188a9a4ada52155759bae33826fce96 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2750-2c2a7e38.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2750-2c2a7e38)             |

### DIFF
--- a/includes/class-stats-dashboard.php
+++ b/includes/class-stats-dashboard.php
@@ -16,6 +16,8 @@ class Stats_Dashboard {
 
 	use Singleton;
 
+	private const COLUMN_NAME = 'stats';
+
 	/**
 	 * Constructor.
 	 */
@@ -26,7 +28,7 @@ class Stats_Dashboard {
 		}
 
 		add_filter( 'job_manager_job_dashboard_columns', [ $this, 'add_stats_column' ] );
-		add_action( 'job_manager_job_dashboard_column_stats', [ $this, 'render_stats_column' ] );
+		add_action( 'job_manager_job_dashboard_column_' . self::COLUMN_NAME, [ $this, 'render_stats_column' ] );
 
 		add_filter( 'manage_edit-job_listing_columns', [ $this, 'add_stats_column' ], 20 );
 		add_action( 'manage_job_listing_posts_custom_column', [ $this, 'maybe_render_job_listing_posts_custom_column' ], 2 );
@@ -39,7 +41,7 @@ class Stats_Dashboard {
 	 * @return array
 	 */
 	public function add_stats_column( $columns ) {
-		$columns['stats'] = __( 'Views', 'wp-job-manager' );
+		$columns[ self::COLUMN_NAME ] = __( 'Views', 'wp-job-manager' );
 		return $columns;
 
 	}
@@ -52,7 +54,6 @@ class Stats_Dashboard {
 	public function render_stats_column( $job ) {
 		$stats = new Job_Listing_Stats( $job->ID );
 		$total = $stats->get_total_stats();
-		$daily = $stats->get_daily_stats();
 
 		$views = $total['view'];
 
@@ -70,7 +71,7 @@ class Stats_Dashboard {
 	public function maybe_render_job_listing_posts_custom_column( $column ) {
 		global $post;
 
-		if ( 'stats' === $column ) {
+		if ( self::COLUMN_NAME === $column ) {
 			$this->render_stats_column( $post );
 		}
 	}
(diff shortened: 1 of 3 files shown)


includes/class-stats-dashboard.php
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * Output the stats column content.
+	 *
+	 * @param \WP_Post $job
+	 */
+	public function render_stats_column( $job ) {
+		$stats = new Job_Listing_Stats( $job->ID );
+		$total = $stats->get_total_stats();
+		$daily = $stats->get_daily_stats();
```
fjorgemota: Technically we're not using `$daily` here, right?

```suggestion
```

includes/class-stats-dashboard.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	/**
+	 * Add a new column to the job dashboards.
+	 *
+	 * @param array $columns
+	 * @return array
+	 */
+	public function add_stats_column( $columns ) {
+		$columns['stats'] = __( 'Views', 'wp-job-manager' );
```
fjorgemota: WDYT about having a more specific name than `stats`? Just to avoid eventual possible conflicts with other plugins.

Maybe `wpjm_stats`?

includes/class-stats.php
```diff
@@ -17,6 +17,9 @@
 class Stats {
 	use Singleton;
 
+	const TABLE       = 'wpjm_stats';
```
fjorgemota: IMHO it is a good idea to have the `CACHE_GROUP` constant, but I'm a bit :thinking:  about having the `TABLE` constant too. It may lead to people trying to use it when the correct is to use the `$wpdb->wpjm_stats` alias.

WDYT about removing it?

includes/class-stats-dashboard.php
fjorgemota: Also maybe we could move this to a private constant in this class?

fjorgemota (commented):
Hey there, this is looking good! Just left a few comments about the code. :)

includes/class-stats.php
yscik: Yup I added it in the first place expecting to use it, good point :) I made the `TABLE` constant private now.

includes/class-stats-dashboard.php
yscik: Moved it to a constant. I think we are fine with `stats`, this is just for our own post type and shortcode, the context is already WPJM there. The other stats plugin uses `total_views` so there is no conflict there.

includes/class-stats-dashboard.php
```diff
@@ -0,0 +1,78 @@
...
+	 * @param \WP_Post $job
+	 */
+	public function render_stats_column( $job ) {
+		$stats = new Job_Listing_Stats( $job->ID );
+		$total = $stats->get_total_stats();
+
+		$views = $total['view'];
+
+		// translators: %1d is the number of views.
+		$views_str = '<span>' . sprintf( _n( '%1d view', '%1d views', $views, 'wp-job-manager' ), $views ) . '</span>';
```
fjorgemota, line 61: I just noticed this makes it print `0 views`.

I'm not sure if this is a huge problem, but maybe it can make sense to handle if `0 === $views` and then print `No Views`? Not sure if it is worth it, though.

fjorgemota (approved):
Looks good and works well! Just left a small thing I noticed about the views string, but I'm not sure it is waaaaay too important

includes/class-stats-dashboard.php
yscik: `0 views` sounds correct, I'd rather not complicate it. Usually job listings will have some views anyway after the first day.

PR #2756, opened 2024-02-09
Add class autoloader

yscik:

* Add a new class autoloader that can be used for WP Job Manager and the addon plugins
* Resolves new namespaced classes to `includes/class-{name}.php`

### Testing Instructions

Usage example in plugins:
`\WP_Job_Manager_Autoload::register( __NAMESPACE__, JOB_MANAGER_ALERTS_PLUGIN_DIR . '/includes' );`

### Release Notes

----

| Plugin build for 0b52eaa6fabb3e1c39b2159bc6746285800d2d2e <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2756-0b52eaa6.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2756-0b52eaa6)             |

### DIFF
--- a/wp-job-manager-autoload.php
+++ b/wp-job-manager-autoload.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Autoload plugin classes.
+ *
+ * @package wp-job-manager
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Autoload plugin classes.
+ *
+ * @since $$next-version$$
+ */
+class WP_Job_Manager_Autoload {
+
+	/**
+	 * Namespace -> directory mappings.
+	 *
+	 * @var array
+	 */
+	private static $autoload_map = [];
+
+	/**
+	 * Add the autoloader.
+	 */
+	public static function init() {
+		spl_autoload_register( [ self::class, 'autoload' ] );
+	}
+
+	/**
+	 * Register a new plugin with a class prefix and directory to autoload.
+	 *
+	 * @param string $namespace Root namespace. Should start with WP_Job_Manager_.
+	 * @param string $dir Directory to autoload.
+	 */
+	public static function register( $namespace, $dir ) {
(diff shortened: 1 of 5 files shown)


wp-job-manager-autoload.php
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	public static function autoload( $class_name ) {
+
+		if ( ! str_starts_with( $class_name, 'WP_Job_Manager' ) || ! str_contains( $class_name, '\\' ) ) {
+			return;
+		}
+
+		list( $namespace, $file_name ) = explode( '\\', $class_name, 2 );
+
+		$root_dir = self::$autoload_map[ $namespace ];
```
fjorgemota, line 62: Maybe it is worth checking if that `$namespace` in fact exists?

Otherwise, if someone tries to load a class from a `WP_Job_Manager` namespace and that namespace for some reason doesn't exist, we will have an "Undefined array key" error.

fjorgemota: For instance, while testing this PR I had the following notice emitted with the WPJM Alerts addon enabled:

```
Notice: Undefined index: WP_Job_Manager_Alerts in /var/www/html/wp-content/plugins/WP-Job-Manager/wp-job-manager-autoload.php on line 58
```

fjorgemota (changes requested):
There's an issue that we must solve before merging, but from my tests after fixing it everything is fine and ready to be approved/merged. :)

wp-job-manager-autoload.php
yscik: Thanks, added some checks in 4ff906f 
Also removed some includes in 0b52eaa to start using this autoloader for the recent namespaced classes.

fjorgemota: Perfect! Thanks!

fjorgemota (approved):
Looks good and works well! :rocket:

PR #2757, opened 2024-02-09
Use PHPMailer hook to set up multipart e-mails

### LINKED ISSUE #2753: Multipart e-mails not working with some SMTP providers / plugins
**Describe the bug**

When sending e-mails after 2.2.0, where multipart e-mails were introduced, the final e-mail is usually sent out with a `text/html` content type, and the plain text part and the boundary headers are output.

image

image

image

> Tried 3 different SMTP plugins
> fluent smtp: won't sent (error code below)
> WP Mail SMTP: sends - but includes the error text below
> Post SMTP Mailer: sends, not not plain text. also includes some of the broken code below. 

So far SendGrid is a confirmed provider where things break.

More context: p1707231412995879-slack-C6GGX896G

yscik:
Fixes #2753 

* Leave assembling the e-mail body and setting the boundary header to PHPMailer

### Testing Instructions

* Send out HTML emails (eg new job submitted notification, job alert) and check that they are rendered correctly

### Release Notes

* Fix issue with rich e-mails on some e-mail providers (#2753)

----

| Plugin build for 2a84507763317efaa0a6ac081fefba9711b200f5 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2757-2a845077.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2757-2a845077)             |

### DIFF
--- a/includes/class-wp-job-manager-email-notifications.php
+++ b/includes/class-wp-job-manager-email-notifications.php
@@ -18,7 +18,6 @@ final class WP_Job_Manager_Email_Notifications {
 	const EMAIL_SETTING_PREFIX     = 'job_manager_email_';
 	const EMAIL_SETTING_ENABLED    = 'enabled';
 	const EMAIL_SETTING_PLAIN_TEXT = 'plain_text';
-	const MULTIPART_BOUNDARY       = 'jm-boundary';
 
 	/**
 	 * Notifications to be scheduled.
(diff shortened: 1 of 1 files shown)


pgk (approved):
Tested well for both the plain and the html part!

PR #2766, opened 2024-02-16
Fix for the hidden and displayed datepicker fields not holding the cu…

### LINKED ISSUE #2739: Job Listing Expiry Date not working in editor
From https://wordpress.org/support/topic/job-listing-expiry-date-not-working-in-editor/
 
> When editing an existing Job Listing, the Expiry Date field is not populated with the actual value. Saving the Job Listing will set its expiry date to a month after the current date, regardless of what was previously set.
> 
> Looking at the list of job listings, the test listing has the correct expiry date:
> 
> image
> 
> But if I edit it, the expiry date field is empty:
> 
> image
> 
> Issue was noticed on a live site but was replicated on a separate WordPress 6.4.3 install using the Twenty Twenty-Four theme with no other plugins.

btsdev:
Fix for the hidden and displayed datepicker fields not holding the current values for _job_expires in datepicker.js.

Fixes #2739

* Properly updates the inputs selected with these selectors: "[name='_job_expires']" and "input[name='_job_expires-datepicker']" with the correct values on the edit page.

### Testing Instructions
* Go to a job edit page and confirm that the value in the field is not blank and is set to the correct value.
* Look in the dev console to make sure that the hidden field also has the correct value (rather than its value being empty).

### Release Notes
* Fix for the hidden and displayed datepicker fields not holding the current values for _job_expires in datepicker.js.


btsdev:
Not sure how important it was to ensure that the names I added were camelCase (as I see a bit of mixing of naming styles: snake_case and "$snake_case" in some places. However, I won't touch any of that code as that would be out of scope for the purpose of this PR).

The need to add the `dateFormatFunction` function itself does feel a bit odd as it feels a bit redundant with jquery-ui-datepicker's functionality, but getting the day of the month correct meant I needed to ensure that it was incremented by 1 and also ensure it was always two digits (e.g. "2" should be "02") for some reason (e.g. "2024-02-4" gets converted to "February 4, 2024", but "2024-02-04" gets converted to "February 3, 2024", going back a day for some reason).

The fix to the field values now occurs at the end of the `initializeDatepicker` function (I initially was applying this at the end of the ready function block, but that seemed inaccurate).

Note that at the end of `initializeDatepicker` I'm using `$target[0].getAttribute("value")` rather than `$target.val()`. I ran into issues where the former code gave me the value, but the latter gave me nothing. I've no idea why this bug was happening, but at least this work around does the job.

I've tested my changes on my client's WP install to ensure this fix is still working.

(also note the force pushes were just to amend some typos in my commit messages)

mikeyarce:
Hi @btsdev - I had a look at this issue as well and am proposing a more simple solution which you can see here: https://github.com/Automattic/WP-Job-Manager/pull/2779 

Let me know what you think and if you're able to test it at all.

mikeyarce:
Closing this as this other PR fixes it https://github.com/Automattic/WP-Job-Manager/pull/2779

PR #2763, opened 2024-02-15
Add apply clicked stat

pgk:
Fixes #2752

### Changes Proposed in this Pull Request

* Adds a unique stat for when apply button is clicked on a listing.
* Refactors JS frontend code to accommodate hooking events to arbitrary elements.
* Refactors php code to account for ^^.

### Testing Instructions

* Visit a listing page and click on apply several times.
* Check db that only one time is recorded.

### Release Notes

* 

### New or Updated Hooks and Templates

*

### Deprecated Code

*

### Screenshot / Video

----

| Plugin build for 888cfc853845799ca9c52f653fd0a9bef630eff0 <a href="#"></a> |
| ------------------------------------------------------------ |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2763-888cfc85.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2763-888cfc85)             |

PR #2771, opened 2024-02-19
Introduce stat for jobs impressions.

### LINKED ISSUE #2751: Log search impression stat for search results
Every time search results are returned for the [jobs] shortcode, we should increase the daily search impressions event for all the job listings returned. 

## Implementation

Instead of logging all the results once when they are returned, we should do this from the frontend and log an impressions when the listing is actually on the screen:

- Watch job listings with [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)
- When it becomes visible, add the job id to a queue, and mark the element as logged
- When the queue has 10 jobs or a few seconds have passed since the last job was added, log the event to the backend and clear the queue
- Maybe log the first 10 or so results right away?

pgk:
Fixes #2751

* Introduce listing impressions on job search.
* Refactors the declarative php and the frontend js side.
* Refactors stats ajax endpoint to accept more/multiple data.

### Testing Instructions

* Open a page that contains the [jobs] shortcode.
* check that on page load, the first 10 listings are logged as impressions.
* Press load more
* Check that impressons are queued up and ajax calls are periodically done as you scroll.

### Release Notes

* 

*

### Deprecated Code

*

### Screenshot / Video

----

| Plugin build for 868bd05165a48041be7568b66e8bf6c7ab7dcf57 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2771-868bd051.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2771-868bd051)             |


assets/js/wpjm-stats.js
```diff
@@ -4,7 +4,148 @@ import domReady from '@wordpress/dom-ready';
 import { createHooks } from '@wordpress/hooks';
 
 ( function () {
```
yscik: Just a heads up this top-level wrapper is not needed, with webpack/modules-based code top-level variables/functions in the file won't be globals, but local to the file.

yscik: Oh and on that note it's also easy to break out code to separate files and import them as the code is starting to get bigger, they will be packaged into one file (the entry point specified in the webpack config, `wpjm-stats.js` in this case.)

assets/js/wpjm-stats.js
```diff
@@ -4,7 +4,148 @@ import domReady from '@wordpress/dom-ready';
...
+			});
+
+			observer.observe(document.documentElement, {
+				childList: true,
+				subtree: true
+			});
+		});
+	}
+
+	function existsAndCallable( globalAsString ) {
```
yscik: Instead of defining the callbacks as `'WPJMStats.initListingImpression',` and needing this kind of check, I'd keep it fixed to one object, define it in PHP as `'initListingImpression'`, and call it like `WPJMStats[ stat.js_callback ]` in JS. 

Together with the DOM-event based stats in #2763, maybe we could generalize these different setup needs as the stat `type`. Something like:

```js
WPJMStats.types = {
  impression: ( stat ) => { /* set up observer */ }, 
  domEvent: ( stat ) => { /* addEventListener setups */ }
}

// Stat is set up like this:
WPJMStats.types[ stat.type ]?( stat )
```

assets/js/wpjm-stats.js
```diff
@@ -4,7 +4,148 @@ import domReady from '@wordpress/dom-ready';
...
+
+	function waitForNextVisibleListing( listingVisibleCallback ) {
+		const jobListingsContainer = document.querySelector( 'ul.job_listings' );
+		const config = { childList: true };
+		const alreadyViewed = {};
+
+		const observer = new MutationObserver(function ( mutations ) {
+			mutations.forEach(function ( mutation ) {
+				mutation.addedNodes.forEach( function ( node ) {
+					if ( 1 === node.nodeType && node.classList.contains( 'job_listing' ) && ! node.classList.contains( 'viewed' ) ) {
```
yscik: Would it make sense to make the selectors part of the config?

includes/class-stats.php
```diff
@@ -365,4 +485,15 @@ private function get_stats_for_ajax( $post_id = 0 ) {
...
+
+	/**
+	 * Any page containing a job shortcode is eligible.
+	 *
+	 * @param \WP_Post $post The post.
+	 *
+	 * @return bool
+	 */
+	public function page_has_jobs_shortcode( $post ) {
+		return is_page() && has_shortcode( $post->post_content, 'jobs' );
```
yscik: Does this work for AJAX calls? If it does and we really want to keep this check, we need to cache it, this sounds very slow to do for each post when logging them in batch. But I don't think we need it or anything in `can_record_stats_for_post`. Even the post type check is a potential slowdown. 

We already have a nonce check for the request, which should prevent most shenanigans. (We can make the nonce action specific to the page, it's just keyed 'ajax-nonce' now)

templates/content-job_listing.php
```diff
@@ -18,7 +18,7 @@
 
 global $post;
 ?>
-<li <?php job_listing_class(); ?> data-longitude="<?php echo esc_attr( $post->geolocation_long ); ?>" data-latitude="<?php echo esc_attr( $post->geolocation_lat ); ?>">
+<li <?php job_listing_class(); ?> data-longitude="<?php echo esc_attr( $post->geolocation_long ); ?>" data-latitude="<?php echo esc_attr( $post->geolocation_lat ); ?>" data-id="<?php echo esc_attr( $post->ID ); ?>">
```
yscik: `job_listing_class()` outputs a `post-{ID}` class — if we use that instead of `data-id`, then this stat will work for current themes and customized templates. Otherwise folks will have update their templates to include the new attribute, which could take a while and result in some reports of 'stats not working'

includes/class-stats.php
```diff
@@ -309,38 +401,66 @@ private function get_registered_stats() {
 				'job_listing_view'                 => [
 					'log_callback' => [ $this, 'log_stat' ], // Example of overriding how we log this.
 					'trigger'      => 'page-load',
+					'page'         => 'listing',
```
yscik, line 408: Should we split this list up, separate for listing pages and job search?

includes/class-stats.php
```diff
@@ -309,38 +401,66 @@ private function get_registered_stats() {
...
+	 * @param int    $post_id Optional post id.
+	 * @param string $page    The page in question.
 	 *
 	 * @return array
 	 */
-	private function get_stats_for_ajax( $post_id = 0 ) {
+	private function get_stats_for_ajax( $post_id = 0, $page = 'listing' ) {
 		$ajax_stats = [];
 		foreach ( $this->get_registered_stats() as $stat_name => $stat_data ) {
+			if ( $page !== $stat_data['page'] ) {
```
yscik, line 455: If we don't split up the stats data, let's use `array_filter` here, so we don't have to copy the fields below.

assets/js/wpjm-stats.js
pgk: Sure! I guess it is a good time to do that, I have been postponing it for a bit

assets/js/wpjm-stats.js
pgk: The reason for making this so generic is I thought it woudl be a cool idea to allow for arbitrary namespaces/objects (catering for 3rd-party stuff)

yscik: Yeah, in the example above a plugin could add its thing like `WPJMStats.types.initAlerts = ...`

templates/content-job_listing.php
pgk: done

assets/js/wpjm-stats.js
pgk: If we end up reusing this for other things then maybe yes? Let's maybe do it in a consecutive PR.

includes/class-stats.php
pgk: ajax-nonce is required, do you mean adding another one more specific to the stat?

assets/js/wpjm-stats.js
pgk: done

assets/js/wpjm-stats.js
```diff
@@ -118,11 +108,134 @@ import { createHooks } from '@wordpress/hooks';
...
+		Object.keys( statsByType ).forEach( function ( type ) {
+			console.log( 'stats by type', type );
+			WPJMStats.types[type] && WPJMStats.types[type]( statsByType[type] );
+		} );
+	},
+
+	hooks: createHooks(),
+	types: {
+		pageLoad: function ( stats ) {
+			console.log( 'pageLoad init script.' );
```
pgk: todo, remove all console logs

includes/class-stats.php
yscik: The $action of a nonce can be specific to the given action. ([example](https://github.com/Automattic/WP-Job-Manager/blob/7cac449889ee2b2e05e9b1f577209092d02f9763/includes/forms/class-wp-job-manager-form-submit-job.php#L1183))

So if when generating the nonce we are more specific, eg using `'stat_listing_' . $job_id` or `stat_listing_jobs_page`, we don't need to do an extra check that we are on the correct page — the nonce is only accepted if it was generated when rendering the correct page.

includes/class-stats.php
```diff
@@ -309,42 +401,69 @@ private function get_registered_stats() {
...
+					'type'    => 'pageLoad',
+					'page'    => 'jobs',
+				],
+				'jobs_view_unique'                 => [
+					'trigger' => 'page-load',
+					'type'    => 'pageLoad',
+					'page'    => 'jobs',
+					'unique'  => true,
+				],
+				'job_listing_impressions'          => [
```
yscik: Let's make this name singular, `job_listing_impressions` → `job_listing_impression`, like the other events. Plus should we include `search` in the name, since we are planning to have an alert impression too? `job_listing_search_impression`, `job_listing_alert_impression`?

Maybe we should make all the event names a bit shorter, just to save database space? `job_listing` → `job`? Like `job_search_impression`, job_apply_click`.

assets/js/wpjm-stats.js
```diff
@@ -118,11 +108,134 @@ import { createHooks } from '@wordpress/hooks';
...
+
+							WPJMStats.hooks.doAction( triggerName );
+						} );
+						events[statToRecord.element] = true;
+					}
+				}
+			} );
+		},
+		// New style of declaration, a stat that relies on calling a custom js func.
+		initListingImpression: function ( stats ) {
```
yscik: I'd move all the listing impression code to its own file that exports just this function, that way it's neatly organized :)

Note the `stats` argument is not actually being used here, looks like the event is hard-coded in `logThem` above.

assets/js/wpjm-stats.js
```diff
@@ -118,11 +108,134 @@ import { createHooks } from '@wordpress/hooks';
...
+		domEvent: function ( stats ) {
+			console.log( 'pageLoad init script.' );
+			const events = {};
+			stats.forEach( function ( statToRecord ) {
+				const triggerName = statToRecord.trigger;
+				if ( statToRecord.element && statToRecord.event ) {
+					const elemToAttach = document.querySelector( statToRecord.element );
+					if ( elemToAttach && ! events[statToRecord.element] ) {
+						elemToAttach.addEventListener( statToRecord.event, function ( e ) {
+							if ( checkUniqueRecordedToday( statToRecord ) ) {
```
yscik: This check is duplicated since it's also part of `wpjmLogStats` now. 

To clarify what  I suggested in the other PR, we don't really need to wait to the request finish to call `setUniques`. We could filter the stats to log in the addAction callback, and have some logic like this for uniques:
- is this the first record today? then let it through to log and **mark it as logged immediately**
- otherwise remove it from the array

It's probably better if the unique check operation is atomic, otherwise, if a user quickly clicks the apply button twice, before the first log request is finished, we'd count it as two unique events.

assets/js/wpjm-stats.js
```diff
@@ -118,11 +108,134 @@ import { createHooks } from '@wordpress/hooks';
 		}, {} );
 
 		Object.keys( statsByTrigger ).forEach( function ( triggerName) {
-			hookStatsForTrigger( statsByTrigger, triggerName );
+			WPJMStats.hookStatsForTrigger( statsByTrigger, triggerName );
+		} );
+
+		WPJMStats.initCallbacks.forEach( function ( initCallback ) {
+			initCallback.call( null );
+		} );
```
yscik: Is this used? Nothing seems to add to the `initCallbacks` array.

assets/js/wpjm-stats.js
```diff
@@ -2,106 +2,96 @@
...
+			if ( entry.isIntersecting && entry.intersectionRatio > 0.99 ) {
+				const node = entry.target;
+				if ( 1 === node.nodeType && node.classList.contains( 'job_listing' ) ) {
+					const nodeId = findIdInClassNames( node );
+					if ( nodeId > 0 && ! alreadyViewedListings[nodeId] ) {
+						alreadyViewedListings[nodeId] = true;
+						visibleCallback( node );
+					}
 				}
+				node.classList.add( 'viewed' );
```
yscik: Seems like we track what was viewed both via this class and in `alreadyViewedListings`, would only one of them be enough?

The class would be gone when paging back and forth or running a new search, so if we don't want to log an impression again then probably the variable is better.

assets/js/stats/utils.js
```diff
@@ -0,0 +1,33 @@
...
+	for ( let i = 0; i < classes.length; i++ ) {
+		const className = classes[i];
+		if ( 0 === className.indexOf( 'post-' ) ) {
+			const maybeId = parseInt( className.substring(5), 10 );
+			if ( ! isNaN( maybeId ) ) {
+				return maybeId;
+			}
+		}
+	}
+	return 0;
```
yscik: Leeting out a bit :)

```suggestion
	return + node.className.match(/\bpost-(\d+)\b/)?.[1] || 0;
```

assets/js/stats/unique.js
```diff
@@ -0,0 +1,33 @@
+export function updateDailyUnique( key ) {
+	const date = new Date();
+	const expiresAtTimestamp = date.getTime() + 24 * 60 * 60 * 1000;
+	window.localStorage[key] = expiresAtTimestamp;
```
yscik: We should add a way to clean these up eventually, `localStorage` is for forever.

assets/js/stats/unique.js
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export function getDailyUnique( name ) {
+	if ( window.localStorage[name] ) {
+		const date = new Date();
+		const now  = date.getTime();
+		const expiration = parseInt( window.localStorage[ name ], 10 );
+		return expiration >= now;
```
yscik: We should check if it parsed to a number, otherwise this seems to be always false.

includes/class-stats.php
pgk: I will let y'all decide about names, changing to singular for now.

assets/js/stats/unique.js
pgk: added a possible way to do that, we can refine later.

includes/class-stats.php
pgk: I removed is page. As for the rest, we can optimize later.

yscik (approved):
Really nice work on this one! I saw a small issue of some listings being skipped occasionally, adding it to a [new card here](https://github.com/Automattic/WP-Job-Manager/issues/2782) along with the various optimization things that came up.

PR #2773, opened 2024-02-20
Add/recaptcha v3

### LINKED ISSUE #2025: Upgrade reCAPTCHA to use v3
Our reCAPTCHA options require v2 of reCAPTCHA: https://wpjobmanager.com/document/setting-up-wp-job-manager/#section-3

Other plugins such as Contact Form 7 use v3, which can cause conflicts, and v3 is more user friendly.

### Additional context
Requested in 3033999-zen

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2025

* Move all the captcha logic to it's own class 
* Add reCAPTCHA v3 support

### Testing Instructions

* You will need both a v2 and a v3 key for this.  
* Go to https://www.google.com/recaptcha/admin/ and create a V2 and V3 site and copy the keys
* In Job Manager > Settings > CAPTCHA use v2 to start, it should look like this:

* Try submitting without clicking the captcha, you should get an error like this:

* Now Select v3, it should look like this:

* Try submitting the form and make sure it works.  If you want to test validation you can error log the validation from 'v3' to see the response.

### Release Notes

* Add Google reCAPTCHA v3 support

* New hook `job_manager_recaptcha_v3_score_tolerance` that allows you to set a custom score tolerance for reCAPTCHA v3. 
* New template used for reCAPTCHA v3 - `form-fields/recaptcha-v3-field.php`

----

| Plugin build for 7ee450a5abce202265981d6c7225822772715fa9 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/03/wp-job-manager-zip-2773-7ee450a5.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/03/2773-7ee450a5)             |


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -79,17 +79,14 @@ public function __construct() {
 		// Listing renewal support.
 		WP_Job_Manager_Helper_Renewals::instance( $this );
 
+		// Recaptcha support.
+		WP_Job_Manager_Recaptcha::instance( $this );
```
fjorgemota: As suggested by PSalm, we should remove the parameter here

```suggestion
		WP_Job_Manager_Recaptcha::instance();
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,242 @@
...
+ *
+ * @since 1.40.0
+ * @internal
+ */
+class WP_Job_Manager_Recaptcha {
+
+	/**
+	 * Stores static instance of class.
+	 *
+	 * @var WP_Job_Manager_Helper_Renewals The single instance of the class.
```
fjorgemota: As suggested by PSalm:

```suggestion
	 * @var WP_Job_Manager_Recaptcha The single instance of the class.
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,242 @@
...
+ *
+ * @package wp-job-manager
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Notices_Conditions_Checker class.
```
fjorgemota:
```suggestion
 * WP_Job_Manager_Recaptcha class.
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,242 @@
...
+ *
+ * @since 1.40.0
```
fjorgemota:
```suggestion
 * @since $$next-version$$
```

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -455,24 +455,36 @@ protected function init_settings() {
...
-							'desc'        => __( 'The label used for the reCAPTCHA field on forms.', 'wp-job-manager' ),
+							'desc'        => __( 'The label used for the CAPTCHA field on forms.', 'wp-job-manager' ),
 							'attributes'  => [],
 						],
+						[
+							'name'        => 'job_manager_recaptcha_version',
+							'std'         => '',
+							'placeholder' => '',
+							'label'       => __( 'reCAPTCHA Version', 'wp-job-manager' ),
+							'desc'        => 'Choose between reCAPTCHA v2 or v3. Note: you will need API keys for the specific version you choose.',
```
fjorgemota: We probably want to translate this, right?

```suggestion
							'desc'        => __( 'Choose between reCAPTCHA v2 or v3. Note: you will need API keys for the specific version you choose.', 'wp-job-manager' ),
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,242 @@
...
+ * @internal
+ */
+class WP_Job_Manager_Recaptcha {
```
fjorgemota, line 20: We probably want to put this class under the namespace `WP_Job_Manager`

includes/class-wp-job-manager.php
```diff
@@ -77,6 +77,7 @@ public function __construct() {
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/class-wp-job-manager-data-exporter.php';
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/class-wp-job-manager-com-api.php';
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/promoted-jobs/class-wp-job-manager-promoted-jobs.php';
+		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/class-wp-job-manager-recaptcha.php';
```
fjorgemota: If we put the class under the namespace `WP_Job_Manager`, we won't need to include it here because it can use the autoload :D

templates/form-fields/recaptcha-v3-field.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Shows the `recaptcha` form field on job listing forms.
+ *
+ * This template can be overridden by copying it to yourtheme/job_manager/form-fields/recaptcha-field.php.
+ *
+ * @see         https://wpjobmanager.com/document/template-overrides/
+ * @author      Automattic
+ * @package     wp-job-manager
+ * @category    Template
+ * @version     1.31.1
```
fjorgemota:
```suggestion
 * @version     $$next-version$$
```

templates/form-fields/recaptcha-v3-field.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * Shows the `recaptcha` form field on job listing forms.
+ *
+ * This template can be overridden by copying it to yourtheme/job_manager/form-fields/recaptcha-field.php.
```
fjorgemota:
```suggestion
 * This template can be overridden by copying it to yourtheme/job_manager/form-fields/recaptcha-v3-field.php.
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,242 @@
...
+			);
+		} elseif ( 'v3' === $this->recaptcha_version ) {
+			get_job_manager_template(
+				'form-fields/recaptcha-v3-field.php',
+				[
+					'key'   => 'recaptcha',
+					'field' => $field,
+				]
+			);
+		}
```
fjorgemota: Given that the parameters to the teemplate are the same, WDYT about changing this logic to just change the template filename?

So we would have just one call to `get_job_manager_template` and the `if` / `elseif` to decide on the filename to load..

fjorgemota (dismissed):
Hey, this is looking generally good!

I found a few points for improvement. :)

Thanks!

includes/forms/class-wp-job-manager-form-submit-job.php
mikeyarce: Done in 14ce1f574fb9b218acc10b13c92237cb73fd5ca8

includes/class-wp-job-manager-recaptcha.php
mikeyarce: Done in 7f4d9d052bc89ea9a2952811160ef6200bf0c535

includes/class-wp-job-manager-recaptcha.php
mikeyarce: Done in 7f4d9d052bc89ea9a2952811160ef6200bf0c535

includes/admin/class-wp-job-manager-settings.php
mikeyarce: Done in 475b4b11a90d40396322b0820afab0253d6d2995

includes/class-wp-job-manager-recaptcha.php
mikeyarce: done in cb4210dfe6288eb66c8c715897b1f9ec794e855f

includes/class-wp-job-manager.php
mikeyarce: done in cb4210dfe6288eb66c8c715897b1f9ec794e855f

templates/form-fields/recaptcha-v3-field.php
mikeyarce: done in 9f1c6217200aa0e65150ca94a576074a2f3c89e1

templates/form-fields/recaptcha-v3-field.php
mikeyarce: done in 9f1c6217200aa0e65150ca94a576074a2f3c89e1

includes/class-wp-job-manager-recaptcha.php
mikeyarce: Good idea!  done in f6aea0315466075b6e9d99e332460c54b33b062d

mikeyarce:
Thanks @fjorgemota - I've addressed all your feedback 🎉

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -318,106 +318,6 @@ public function clear_fields() {
 		$this->fields = [];
 	}
 
-	/**
-	 * Enqueue the scripts for the form.
-	 */
-	public function enqueue_scripts() {
```
gikaragia, line 324: As these are public functions, we will need to deprecate them for a couple of versions until we eventually remove them as they will break sites that use them. So we should basically add `deprecated_function` calls and call the function from the recaptcha class.

We use this method in applications also which causes this fatal:
```
PHP Fatal error:  Uncaught Error: Call to undefined method WP_Job_Manager_Form::enqueue_scripts() in /wp-content/plugins/wpjm-addons/src/applications/includes/class-wp-job-manager-applications-apply.php:79
Stack trace:
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,236 @@
...
+			add_filter( 'submit_draft_job_form_validate_fields', [ $this, 'validate_recaptcha_field' ] );
+		}
+	}
+
+	/**
+	 * Returns static instance of class.
+	 *
+	 * @return self Main instance.
+	 */
+	public static function instance() {
```
gikaragia: We have a Singleton trait which you can use here I think

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,236 @@
...
+		$validation_error = new WP_Error( 'validation-error', sprintf( esc_html__( '"%s" check failed. Please try again.', 'wp-job-manager' ), $recaptcha_field_label ) );
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing -- Nonce check happens earlier (when possible).
+		$input_recaptcha_response = isset( $_POST['g-recaptcha-response'] ) ? sanitize_text_field( wp_unslash( $_POST['g-recaptcha-response'] ) ) : '';
+
+		if ( empty( $input_recaptcha_response ) ) {
+			return $validation_error;
+		}
+
+		if ( 'v2' === $this->recaptcha_version ) {
```
gikaragia, line 160: Something that you can consider as an improvement. This seems like a good case of defining a base class for recaptcha and then create two subclasses for v2 and v3. You might need to define some extra methods that are overridden in subclasses. This way we can avoid the separate flows for each version.

gikaragia (commented):
Looks good overall but I found a couple of issues:
- If I create recaptcha in trunk, and then switch to this branch the recapthca stops working. In 'reCAPTCHA Version' setting no version is selected and in the frontend I see the following:
Screenshot 2024-02-28 at 11 52 11

 I would expect a mini-migration to happen and my site to work with v2 out of the box.
- recaptchas are also used in the resume submission and application forms. When applications are enabled there is a fatal and for resume submission the recaptcha isn't displayed.

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -455,24 +455,36 @@ protected function init_settings() {
...
 							'std'         => __( 'Are you human?', 'wp-job-manager' ),
 							'placeholder' => '',
 							'label'       => __( 'Field Label', 'wp-job-manager' ),
-							'desc'        => __( 'The label used for the reCAPTCHA field on forms.', 'wp-job-manager' ),
+							'desc'        => __( 'The label used for the CAPTCHA field on forms.', 'wp-job-manager' ),
 							'attributes'  => [],
 						],
+						[
+							'name'        => 'job_manager_recaptcha_version',
+							'std'         => '',
```
yscik: @mikeyarce related to what Giannis mentioned about the default version, we can put 'v2' here, but also need to add it to the `get_option( 'job_manager_recaptcha_version' )` calls. 
Important that whatever is set as the default in the `std` field in settings is only applied after the settings are saved once.

templates/form-fields/recaptcha-v3-field.php
```diff
@@ -0,0 +1,33 @@
...
+ * @version     $$next-version$$
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+$site_key = get_option( 'job_manager_recaptcha_site_key' );
+?>
+<script>
+	function onClick(e) {
```
yscik: We should add a plugin-specific prefix to this since it's a global, eg `jm_job_submit_click`

templates/job-submit.php
```diff
@@ -16,6 +16,7 @@
 }
 
 global $job_manager;
+$captcha_version = get_option( 'job_manager_recaptcha_version' );
```
yscik: Might be worth adding a `get_captcha_version()` function somewhere that puts getting the option or returning the default value in one place.

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,236 @@
...
+
+	/**
+	 * Initialize class for landing pages.
+	 *
+	 * @since 2.0.0
+	 */
+	public function __construct() {
+		$this->site_key          = get_option( 'job_manager_recaptcha_site_key' );
+		$this->secret_key        = get_option( 'job_manager_recaptcha_secret_key' );
+		$this->recaptcha_version = get_option( 'job_manager_recaptcha_version' );
```
yscik: Should we add class constants for these option names (and one for the default version)?

templates/job-submit.php
mikeyarce: Done in 260ba28f0540e2a73d48dda4b4bae2e99e3e56bf

includes/class-wp-job-manager-recaptcha.php
mikeyarce: Done in 260ba28f0540e2a73d48dda4b4bae2e99e3e56bf

templates/form-fields/recaptcha-v3-field.php
mikeyarce: Done in 260ba28f0540e2a73d48dda4b4bae2e99e3e56bf

includes/class-wp-job-manager-recaptcha.php
mikeyarce: Done in ed453f473f4f0b602a3117436c1666349d94a36f

includes/abstracts/abstract-wp-job-manager-form.php
mikeyarce: Deprecated in 74ce7d0c5322d1e6579d44e13bbf999e13e9c0b6

includes/class-wp-job-manager-recaptcha.php
mikeyarce: Thanks @gikaragia !  I had a talk with @fjorgemota  about this and after discussing this and a few other approaches we thought maybe since it's just small we should keep it as-is for now.  Appreciate the suggestion though!!

includes/admin/class-wp-job-manager-settings.php
mikeyarce: done in badfb87a77790444831b8c7225c80eab9f893ae3  and also whenever getting the option we default to `v2` if it doesn't exist.

mikeyarce:
Hi @yscik and @gikaragia - thanks for your detailed reviews!  I have a companion PR for `wpjm-addons` that you can use here: https://github.com/Automattic/wpjm-addons/pull/472

The only added testing instructions are:
- Make sure Applications and Resumes works both with V2 and V3
- Also make sure to test Applications and Resumes with this PR and trunk, as it should work with both

Note:  In Applications I was able to put the logic for when to enqueue the scripts within that plugin, but with Resumes it became quite complicated with how Resumes works, so the easiest was to add the logic to WPJM Core itself.

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -136,7 +136,17 @@ public function process() {
...
+		$resumes_enabled_recaptcha = (bool) get_option( 'resume_manager_enable_recaptcha_resume_submission', false );
+		if ( is_array( $atts ) ) {
+			if ( in_array( 'resume', $atts, true ) ) {
+				if ( $resumes_enabled_recaptcha ) {
+					WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
+				}
+			} else {
+				WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
+			}
+		}
```
gikaragia: This can be simplified with:
```suggestion
		if ( is_array( $atts ) ) {
			if ( ! in_array( 'resume', $atts, true ) || $resumes_enabled_recaptcha ) {
				WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
			}
		}
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,231 @@
...
+	const RECAPTCHA_SITE_KEY   = 'job_manager_recaptcha_site_key';
+	const RECAPTCHA_SECRET_KEY = 'job_manager_recaptcha_secret_key';
+	const RECAPTCHA_VERSION    = 'job_manager_recaptcha_version';
+
+	/**
+	 * Initialize class for landing pages.
+	 *
+	 * @since 2.0.0
+	 */
+	public function __construct() {
```
gikaragia:
```suggestion
	private function __construct() {
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -0,0 +1,231 @@
...
+		$instance = self::instance();
+
+		if (
+			$instance->use_recaptcha_field() &&
+			in_array( $instance->recaptcha_version, [ 'v2', 'v3' ], true )
+		) {
+			$recaptcha_version = $instance->recaptcha_version;
+			$recaptcha_url     = '';
+
+			if ( 'v2' === $recaptcha_version ) {
```
gikaragia, line 91: When the user upgrades, the recaptcha_version variable is not set which means that although the had a recaptcha before it will stop working.

I think we should either do a mini migration and set to v2 for such cases or assume v2 when it's empty.

gikaragia (commented):
Had another go, changes look good in general! If the users had a recaptcha set up it will stop working when they upgrade, I think we should fix that.

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -136,7 +136,17 @@ public function process() {
 	 * @param array $atts Attributes to use in the view handler.
 	 */
 	public function output( $atts = [] ) {
-		$this->enqueue_scripts();
```
yscik, line 139: If I'm understanding correctly, the current `enqueue_scripts()` overrides in the descendant classes (`WP_Job_Manager_Form_Submit_Job`, `WP_Resume_Manager_Form_Submit_Resume`) would solve the problem of detecting if it's a resume or submit job form already, why move this logic up to the parent class?

mikeyarce: I thought so, too - but in Resumes the class is never instantiated, it's just output through the parent class, so overriding the child class does nothing.

includes/abstracts/abstract-wp-job-manager-form.php
mikeyarce: Good call!  d6cb78736bef930b891d5d26b7576b1a6dc623f8

includes/class-wp-job-manager-recaptcha.php
mikeyarce: I think it should work because we set `v2` as the default here:
`$this->recaptcha_version = get_option( self::RECAPTCHA_VERSION, 'v2' );`
https://github.com/Automattic/WP-Job-Manager/pull/2773/files#diff-20f5f5d09f20927ac0c7a20add75346d466aca6f66eb2ae09daeb54c27c2b785R57

What I tested was deleting the recaptcha version option, setting recaptcha in `trunk`, then moving to the new branch, and it did work for Resumes, Applications, and Core.  

Let me know if I'm missing something else 😄

includes/abstracts/abstract-wp-job-manager-form.php
yscik: Would that be fixable in Resumes? It'd be nice to avoid having to put addon-specific exceptions in the core plugin.

includes/class-wp-job-manager-recaptcha.php
gikaragia: Ah yes indeed the option was set from a previous test and this is the reason it didn't work. 

But there is another case, and it happens only in core. If there are no attributes set in the shortcode, `atts` is not going to be an array in `abstract-wp-job-manager-form.php` so the recaptcha doesn't work.

mikeyarce: Thanks for the details!  I've removed the check for `atts` so I think this should work now :)

includes/abstracts/abstract-wp-job-manager-form.php
mikeyarce: Removed the logic here https://github.com/Automattic/WP-Job-Manager/pull/2773/commits/48835aa4054a34a3d4e54714874314582298588c    and added here https://github.com/Automattic/wpjm-addons/pull/472/commits/8926d93fe7bb9b2a4ea881a34d98190e3f714763

gikaragia (approved):
LGTM!

PR #2774, opened 2024-02-20
Fix Psalm config errors

fjorgemota:

* Add WP constants to the psalm-loader
* Disable feature for finding unused codes (which was throwing a lot of errors related to methods used by WP hooks)
* Remove templates folder from analysis (at least until https://github.com/vimeo/psalm/issues/6065 is fixed)

### Testing Instructions

Check if the errors reported in the baseline make sense

### Context

p1708463903828029-slack-C069FKQBQQM 

----

| Plugin build for 8a3705bbae71e6159c110c1a3df5abee9630eaaf <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2774-8a3705bb.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2774-8a3705bb)             |

### DIFF
--- a/.psalm/psalm-loader.php
+++ b/.psalm/psalm-loader.php
@@ -5,10 +5,63 @@
  * @package wp-job-manager
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound
-	define( 'ABSPATH', __DIR__ . '/../' );
-}
+/** WordPress Constants */
+// phpcs:disable WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound
+define( 'ABSPATH', __DIR__ . '/../' );
+
+// ./wp-includes/default-constants.php
+
+define( 'WP_DEBUG', true );
+define( 'WP_DEBUG_LOG', true );
+
+define( 'EMPTY_TRASH_DAYS', 30 );
+
+define( 'MINUTE_IN_SECONDS', 60 );
+define( 'HOUR_IN_SECONDS', 60 * MINUTE_IN_SECONDS );
+define( 'DAY_IN_SECONDS', 24 * HOUR_IN_SECONDS );
+define( 'WEEK_IN_SECONDS', 7 * DAY_IN_SECONDS );
+define( 'MONTH_IN_SECONDS', 30 * DAY_IN_SECONDS );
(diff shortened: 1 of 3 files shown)


.psalm/psalm-loader.php
```diff
@@ -5,10 +5,63 @@
...
-	define( 'ABSPATH', __DIR__ . '/../' );
-}
+/** WordPress Constants */
+// phpcs:disable WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound
+define( 'ABSPATH', __DIR__ . '/../' );
+
+// ./wp-includes/default-constants.php
+
+define( 'WP_DEBUG', true );
+define( 'WP_DEBUG_LOG', true );
```
yscik, line 15: Is there an alternative to having the copy of all the WordPress constants? It doesn't seem maintainable. Maybe https://github.com/psalm/psalm-plugin-wordpress has better stubs?

fjorgemota: This is basically what that plugin does: https://github.com/psalm/psalm-plugin-wordpress/blob/master/stubs/globals.php

~However, I'll remove the constants from the loader and just use that plugin directly. :)~

fjorgemota: I checked here and, as mentioned on Slack, that plugin tries to parse all the `add_filter` / `add_action` calls and:

- Throws parse errors when it finds a hook name using a class constant
- When the above parse errors are fixed, it throws errors about not knowing about WPJM custom hooks
- The above error can't be fixed easily because https://github.com/wp-hooks/generator - which is supposed to generate a file containing all the WPJM hooks - doesn't work (it tries to load a class that doesn't exist)

So for now we have two options:

1. Just keep the WP constants there 
2. Or, install the plugin but do not enable it on psalm. Instead, load the file referred above directly here in the loader.

WDYT?

yscik: I'd probably defer to the psalm-plugin-wordpress plugin for the constant list and include it from there, but I imagine it won't change that often so no strong opinions on it.

yscik (approved):
🕵️

.psalm/psalm-loader.php
fjorgemota: Let's go this way for now, in any case we can change in the future. :)

PR #2779, opened 2024-02-24
Set the default dateFormat if it’s empty

### LINKED ISSUE #2739: Job Listing Expiry Date not working in editor

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2739

* Previously, we were setting a datepicker option if it was empty, but it was causing some issues with showing the dateformat in the editor.  Now, this will just set the default in case the default dateFormat is empty.
* Bonus: also added a fallback for the main Jobs view in case there is no DateFormat

### Testing Instructions

* Settings > General - set the Date Format to `Custom`, clear out the custom format and save.
* Go to a Job Listing, make sure you can save an expiry date
* Refresh to make sure it's still there!
* Go to the Jobs page and make sure you see the dates in the column view as well.
* Do the same thing now but select a Date Format of your choosing!

### Release Notes

* Fix: Expiry date not showing up in backend editor.
* Fix: Add fallback to date format for Jobs view in case it's missing.

----

| Plugin build for 5b318b12806dbc22f58216ca0f4fe37f9b16b4b5 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/03/wp-job-manager-zip-2779-5b318b12.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/03/2779-5b318b12)             |

### DIFF
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -1,5 +1,8 @@
 /* global job_manager_datepicker */
 jQuery(document).ready( function() {
+	if ( jQuery.datepicker._defaults.dateFormat == '' ) {
+		jQuery.datepicker._defaults.dateFormat = 'MM d, yy';
+	}
 	var $date_today = new Date();
 	var datePickerOptions = {
 		altFormat  : 'yy-mm-dd',
@@ -21,7 +24,6 @@ jQuery(document).ready( function() {
 			}
 		} );
 		$target.datepicker( jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
-		$target.datepicker("option", "dateFormat", $target.datepicker("option", "dateFormat") || "MM d, yy");
 		if ( $target.val() ) {
 			var dateParts = $target.val().split('-');
 			if ( 3 === dateParts.length ) {
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -593,6 +593,7 @@ class WP_Job_Manager_CPT {
 	 */
 	public function custom_columns( $column ) {
 		global $post;
+		$date_format = get_option( 'date_format' ) ?: 'F j, Y';
 
 		switch ( $column ) {
 			case \WP_Job_Manager_Post_Types::TAX_LISTING_TYPE:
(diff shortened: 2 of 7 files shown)


thedebian (changes requested):
I think we might need to update somewhere all components of WPJM and addons could check, maybe ?

Here, for instance, it's not appearing in the jobs dashboard:

Capture d’écran 2024-02-28 à 09 58 44

On the jobs search page the date appears 👍 :
Capture d’écran 2024-02-28 à 10 00 02

If I update the date format in general settings the date appears properly:

Capture d’écran 2024-02-28 à 10 00 44

mikeyarce:
Thanks @thedebian - my initial PR was addressing the problem when jQuery UI was missing the format, but now I've expanded that to add a fallback for when the actual `get_option( 'date_format' )` is missing in the rest of the plugin 😄

thedebian (approved):
Looking good.
Not a blocker, but if we can put 'F j, Y' in a constant somewhere, it'd probably be better.

ibiza69:
Hello, we updated the plugin with the patch provided on github and it completely broke our website, both the front and backend, without the possibility of access to both. We have had to download the original plugin (not the one patched at github), version 2.2.2 from WordPress and upload it via FTP, overwriting the files, in order to solve the problem.

mikeyarce:
Hi @ibiza69 - did you just download the source or did you also run the build steps?  The safest bet would be to wait till we have a release and test that!

ibiza69:
Hi @mikeyarce, we downloaded thru github the fix and uploaded the files thru FTP and that crashed our site. 

We shall wait untill plugin update at wordpress :) Thank you.

PR #2780, opened 2024-02-24
Allow duplication on all post statuses

### LINKED ISSUE #2470: Duplicate a Job - Additional Status Options
Currently, users have the ability to duplicate a post in their jobs dashboard. However, this option is only possible when the listing is active. This limits the capability of this feature when job listings are in a different status.

Ideally, the Duplicate option would appear to users when jobs are pending and expired in addition to when jobs are active. We have many users requesting this option. For example, when employers join our site we set their member status to pending while we validate their information. While a pending employer, we allow employers to post a job listing with the understanding that the job will not be published until their account is approved. In many cases, employers want to publish similar jobs, yet different, and it would be of great value to allow them to duplicate the first listing when trying to create a second. The same issue applies to many employers have expired listings who want to save time by duplicating an existing job description.

Perhaps this could be an option under WPJM / Applications settings. This would allow the administrator to decide if they wanted to extend these settings or not.

We used to have access to the template that assigned these status' in the jobs dashboard but that does not seem to be the case anymore. We found this very helpful to customize the jobs dashboard.

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2470

* Adds a filter `job_manager_allowed_post_status_duplicate_jobs` that will allow people to filter which post status will be allowed to duplicate jobs in the Job Dashboard
* Default is to allow just ‘publish’ status, which won’t change any behaviours.

### Testing Instructions

* Apply this PR
* Add a filter like this to your site:

```
add_filter( 'job_manager_allowed_post_status_duplicate_jobs', function( $post_status ) {
    $post_status[] = 'preview';
    return $post_status;
});
```
* Make sure you have a "draft" job in your job dashboard
* Ensure that you can now duplicate it!

### Release Notes

* New filter, `job_manager_allowed_post_status_duplicate_jobs` will let you set specific post statuses that will have the duplicate optionin the Job Dashboard.

----

| Plugin build for 82286ded26febb5ba3b64e21f0b29c11b10cccbc <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/02/wp-job-manager-zip-2780-82286ded.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/02/2780-82286ded)             |

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -477,11 +477,6 @@ class WP_Job_Manager_Shortcodes {
 						'nonce' => $base_nonce_action_name,
 					];
 				}
-
-				$actions['duplicate'] = [
-					'label' => __( 'Duplicate', 'wp-job-manager' ),
-					'nonce' => $base_nonce_action_name,
-				];
 				break;
 			case 'expired':
 				if ( job_manager_get_permalink( 'submit_job_form' ) ) {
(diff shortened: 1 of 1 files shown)


yscik (approved):
Looks good, but I'm wondering if we need the filter at all. 'Duplicate' seems useful regardless of job status, copying old jobs or a draft one etc. Should we just always show the action? 
Actions will be in a dropdown with the dashboard updates in `feature/stats`, so we have the space: 
image

mikeyarce:
@yscik I was leaning that way originally, so now that you say that I think we should do it for all post statuses.  Done in https://github.com/Automattic/WP-Job-Manager/pull/2780/commits/82286ded26febb5ba3b64e21f0b29c11b10cccbc

PR #2776, opened 2024-02-22
Add stats section to job overlay

yscik:
Fixes #2744, fixes #2745, fixes #2746

### Changes Proposed in this Pull Request

* Add stats section and daily chart to job overlay
* Add accent color, with a script to auto-detect link color by default
* Fix/tune related styles, improve job overlay on mobile
* Add a dev cli command to generate stats
* Add search impression count to the dashboard and admin tables

### Testing Instructions

* Open the [job_dashboard], and click on various jobs to view them. 
* Generate stats to show by clicking around job search opening jobs, or by the CLI helper command: `wp jm dev stats generate <postId>`
* The daily views chart goes from publish date to expiration date, make sure those are realistic. Test data generation goes from publish date to current date.
* Check on a few themes. Chart color should pick up the site's link color.

### New or Updated Hooks and Templates

* `job_manager_job_stats_chart` - Filter the data for the daily breakdown chart
* `job_manager_job_stats_summary` - Filter the sections and data for the summaries
* `job_manager_ui_accent_color` - Set an accent color manually

### Screenshot / Video

image

----

| Plugin build for 78c905a09926b9af7d59f469d4b65bcaa16a02b9 <a href="#"></a> |
| ------------------------------------------------------------ |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/03/wp-job-manager-zip-2776-78c905a0.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/03/2776-78c905a0)             |

PR #2785, opened 2024-03-07
Optimize stats, clean up code, respect settings

### LINKED ISSUE #2782: Optimize stat logging
A few follow-up optimizations and improvements to look at

All stats:

- [Maybe make the stat names shorter](https://github.com/Automattic/WP-Job-Manager/pull/2771#discussion_r1498902609)?

Search impressions stats (#2771):

- Batch writing the events to DB
- Maybe reduce request payload size? Could compress into one event with multiple post IDs? Eg `{ job_listing_impression: [ 11, 12, 13, 14 ] }`
- Debounce: send if 10 or so events are in the queue, don't wait for scrolling to stop
- Check if the JS is blocking rendering and if anything could be moved off the main thread
- Fix: investigate why sometimes a few listings are missed — registered as viewed but not as sent
- Maybe move CSS selectors in the observers to PHP so it's filterable if a theme is using different classes?
- [Optimize page_has_jobs_shortcode check](https://github.com/Automattic/WP-Job-Manager/pull/2771#discussion_r1494958601)

### LINKED ISSUE #2784: Apply stats settings
- Only log stats when the setting is turned on.
- Only display stats when the setting is turned on
- By default, the setting should be off for existing sites and on for new sites

yscik:
Fixes #2782 #2784

Stacked on #2776

* Remove unused stats settings, only leave "Enable job statistics". Move it to the General settings tab since it's only one setting for now. Update labels.
* Make sure the setting is respected & stats are not logged or displayed when it's off
* Rename stats to make DB keys a bit shorter (`job_listing_view` -> `job_view`)
* Add constants, cleanup code
* Various tweaks for the frontend framework:
  * Organize and rename a few things for better clarity
  * Make stats triggered by a custom event/action (only `page-load` now) a dedicated stat type with it's own setup (`action`)
* And to observing search impressions:
    * Use a single IntersectionObserver to monitor the listing elements. Also use this for the initially visible listings, it'll trigger even when they are already in view
    * Clean up observer when a listing element is removed (this happens when search filters/keywords are changed)
    * Send the log request right away when 10 events are in the queue
    * Fix some timing issues causing nodes to be skipped sometimes
    * Move the job listing selectors to the PHP stat config, so theme authors can customize them

### Testing Instructions

* In Job Manager settings, turn on the Job Statistics option
* Check that stats are logged correctly when viewing a job page ( job_listing, job_listing_unique, job_apply_click)
* Check that stats are logged when viewing the job list. Test with both ajax search and static: `[jobs]` and `[jobs show_filters=false]`
* Check that the above logging is only done when the settings is turned on
* Check that the Views column in Job listings admin and the job dashboard, and the Job statistics overlay is only shown when the settings is turned on

### Screenshot / Video

image

----

| Plugin build for c418e1b8dce425adc4dadfdb6d9a34e28e00f985 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/03/wp-job-manager-zip-2785-c418e1b8.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/03/2785-c418e1b8)             |


assets/js/stats/utils.js
```diff
@@ -10,12 +10,20 @@ export function debounce( func, delay ) {
 	};
 }
 
-export function filterZeroes( list ) {
-	return list.filter( function ( i ) {
-		return i > 0;
-	} );
+export function getPostId( node ) {
+	return +node.className.match( /\bpost-(\d+)\b/ )?.[ 1 ] || 0;
```
thedebian, line 14:
```suggestion
	return node.className.match( /\bpost-(\d+)\b/ )?.[ 1 ] || 0;
```

includes/class-job-listing-stats.php
```diff
@@ -38,42 +38,21 @@ class Job_Listing_Stats {
 	/**
 	 * Stats for a single job listing.
 	 *
-	 * @param int $job_id
+	 * @since $$next-version$$
+	 *
+	 * @param int  $job_id
+	 * @param null $date_range
```
thedebian:
```suggestion
	 * @param array|null $date_range
```

includes/class-job-listing-stats.php
```diff
@@ -38,42 +38,21 @@ class Job_Listing_Stats {
...
 	 */
-	public function __construct( $job_id ) {
+	public function __construct( $job_id, $date_range = null ) {
 
 		$this->job_id     = $job_id;
-		$this->start_date = get_post_datetime( $job_id )->format( 'Y-m-d' );
-
-	}
+		$this->start_date = $date_range[0] ?? get_post_datetime( $job_id )->format( 'Y-m-d' );
+		$this->end_date   = $date_range[1] ?? ( new \DateTime( 'today' ) )->format( 'Y-m-d' );
```
thedebian: Both of these should probably use get_option( 'date_format' ) ?: 'F j, Y' as in #2779 ?

thedebian (approved):
Worked well, it needs a rebase though.
I had issues unrelated to this PR, reported on Slack.

includes/class-job-listing-stats.php
yscik: This one is just for the array keys, matching the YYYY-MM-DD date format from the DB, not user-visible.

assets/js/stats/utils.js
yscik: The `+` is a cheeky little type cast to number :)

PR #2787, opened 2024-03-13
Add stats batch logging, unit tests, small tweaks

yscik:
Collecting a few more things here that support addons adding their own stats.  

Under the hood:
* Add batch logging, insert multiple stat records in one query. This is most useful for the Alerts add-on but I also switched the ajax logger over for quicker search impressions
* Move ajax stats to dedicated class, remove some unused methods
* Add unit tests
* Lower column size to `varchar(50)` for `name` and `group` column in the stats table
* Improve UI icon helper for addons adding using custom icons

Stats overlay:
* Add repeat job views count (just the difference between total and unique for now)
* Add click through rate
* Format numbers, tweak labels

### Testing Instructions

* Enable stats, click around job search and open job pages
* Check in the job dashboard that the impressions and views are showing up

### Screenshot / Video

image

----

| Plugin build for 15e705bec87f9ac206b8822f53353b3907e71070 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/03/wp-job-manager-zip-2787-15e705be.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/03/2787-15e705be)             |

### DIFF
--- a/includes/class-stats-dashboard.php
+++ b/includes/class-stats-dashboard.php
@@ -212,6 +212,10 @@ class Stats_Dashboard {
 	private function get_stat_summaries( \WP_Post $job ) {
 		$job_stats = new Job_Listing_Stats( $job->ID );
 
+		$views        = $job_stats->get_event_total( Job_Listing_Stats::VIEW );
+		$views_unique = $job_stats->get_event_total( Job_Listing_Stats::VIEW_UNIQUE );
+		$views_repeat = $views - $views_unique;
+
 		/**
 		 * Filter the job stat summaries, displayed in the job overlay.
 		 *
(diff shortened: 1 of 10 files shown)


includes/class-stats-dashboard.php
```diff
@@ -212,6 +212,12 @@ private function get_daily_stats_chart( \WP_Post $job ): array {
 	private function get_stat_summaries( \WP_Post $job ) {
 		$job_stats = new Job_Listing_Stats( $job->ID );
 
+		$views              = $job_stats->get_event_total( Job_Listing_Stats::VIEW );
+		$views_unique       = $job_stats->get_event_total( Job_Listing_Stats::VIEW_UNIQUE );
+		$views_repeat       = $views - $views_unique;
+		$search_impressions = $job_stats->get_event_total( Job_Listing_Stats::SEARCH_IMPRESSION );
+		$search_clicks      = $views_unique / $search_impressions * 100;
```
mikeyarce: If `$search_impressions` here is `0` then this could be a fatal `Division by zero`.  Maybe we should add a defensive check here to only do search clicks if there are impressions?

yscik: Added a check in 15e705b

mikeyarce (approved):
Looks good!  👍🏽

PR #2792, opened 2024-03-19
Fix stat display date

mikeyarce:

* Show dates according to site's timezone and not UTC

### Testing Instructions

* Set site's timezone to something far from UTC, like Pacific time was what I had
* Set a site impression or view and make sure that it would be UTC for the next day
* You can even just edit the stat manually in the database if that's easier
* Now make sure that the stat shows for the right day in the overlay

----

| Plugin build for 4e4a0fab3e51e96743ac8d9d18229aec443e4763 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/03/wp-job-manager-zip-2792-4e4a0fab.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/03/2792-4e4a0fab)             |

### DIFF
--- a/includes/class-stats-dashboard.php
+++ b/includes/class-stats-dashboard.php
@@ -179,10 +179,12 @@ class Stats_Dashboard {
 		ksort( $by_day );
 
 		$date_format = apply_filters( 'job_manager_get_dashboard_date_format', 'M d, Y' );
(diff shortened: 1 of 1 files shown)


yscik (approved):
👌

PR #2798, opened 2024-03-27
Add new notice about latest release

### LINKED ISSUE #2796: Stats: add feature announcement notice
To let users who updated know about it.

image

yscik:
Fixes #2796

* Add a new release notice class
* Add new styles and options to support this notice layout

### Testing Instructions

* Notice should show up on the Job Manager (Job Listings) admin screen
* Disable Job Statistics in the settings
* In the notice, click 'Enable'. Check that the setting is enabled and the notice disappears
* Check that dismissing the various notices makes them disappear, even after refreshing the page
* Use `wp option delete wp_job_manager_dismissed_notices` to un-dismiss all notices
* Also check if the notice text needs any improvements

### Screenshot / Video

image

----

| Plugin build for 317970f2af056ecf60393476759d59cbabe6abfc <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2798-317970f2.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2798-317970f2)             |

### DIFF
--- a/includes/admin/class-release-notice.php
+++ b/includes/admin/class-release-notice.php
@@ -13,16 +13,26 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 /**
  * Admin notice about changes and new features introduced in the latest release.
+ * Update this class for each release to highlight new features or changes.
  */
 class Release_Notice {
+	public const NOTICE_ID = 'release_notice_2_3_0';
 
 	/**
 	 * Set up notice.
 	 */
 	public static function init() {
 		add_filter( 'wpjm_admin_notices', [ __CLASS__, 'add_release_notice' ] );
+		add_action( 'job_manager_action_enable_stats', [ self::class, 'enable_feature' ] );
+	}
 
-		add_action( 'job_manager_action_enable_stats', fn() => \WP_Job_Manager_Settings::instance()->set_setting( \WP_Job_Manager\Stats::OPTION_ENABLE_STATS, '1' ) );
+	/**
+	 * Enable the highlighted feature.
+	 *
+	 * @return void
+	 */
+	public static function enable_feature() {
+		\WP_Job_Manager_Settings::instance()->set_setting( \WP_Job_Manager\Stats::OPTION_ENABLE_STATS, '1' );
 	}
 
 	/**
@@ -35,23 +45,20 @@ class Release_Notice {
 	public static function add_release_notice( $notices ) {
 
 		// Make sure to update the version number in the notice ID when changing this notice for a new release.
-
-		$notice_id = 'release_notice_2_3_0';
-
-		$action_url            = \WP_Job_Manager_Admin_Notices::get_action_url( 'enable_stats', $notice_id );
-		$notices[ $notice_id ] = [
+		$action_url                 = \WP_Job_Manager_Admin_Notices::get_action_url( 'enable_stats', self::NOTICE_ID );
+		$notices[ self::NOTICE_ID ] = [
 			'type'          => 'site-wide',
 			'label'         => 'New',
 			'heading'       => 'Job Statistics',
 
 			'message'       => '<div>' . __(
 				'
-<p>Collect analytics data about site visitors for each job listing. Display the detailed statistics in the refreshed jobs dashboard.</p>
+<p>Collect analytics about site visitors for each job listing. Display the detailed statistics in the refreshed jobs dashboard.</p>
 <ul>
-	<li>Page views and unique visitors with daily breakdown</li>
-	<li>Search impressions and apply button clicks</li>
-	<li>Add-on integration: Job alert impressions, bookmarks, application stats</li>
-	<li>GDPR-compliant, with no personal user information collected</li>
+	<li>Tracks page views and unique visitors, search impressions and apply button clicks.</li>
+	<li>Adds a new overlay to the employer dashboard with aggregated statistics and a daily breakdown chart.</li>
+	<li>Integrates with Job Alerts, Applications, and Bookmarks add-ons.</li>
+	<li>GDPR-compliant, with no personal user information collected.</li>
 </ul>
 ',
 				'wp-job-manager'
@@ -64,7 +71,7 @@ class Release_Notice {
 				],
 				[
 					'label'   => __( 'Dismiss', 'wp-job-manager' ),
-					'url'     => \WP_Job_Manager_Admin_Notices::get_dismiss_url( $notice_id ),
+					'url'     => \WP_Job_Manager_Admin_Notices::get_dismiss_url( self::NOTICE_ID ),
 					'primary' => false,
 				],
 				[
--- a/includes/admin/class-release-notice.php
+++ b/includes/admin/class-release-notice.php
@@ -50,7 +50,6 @@ class Release_Notice {
 			'type'          => 'site-wide',
 			'label'         => 'New',
 			'heading'       => 'Job Statistics',
-
 			'message'       => '<div>' . __(
 				'
 <p>Collect analytics about site visitors for each job listing. Display the detailed statistics in the refreshed jobs dashboard.</p>
--- a/includes/admin/class-release-notice.php
+++ b/includes/admin/class-release-notice.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * File containing the class \WP_Job_Manager\Admin\Release_Notice.
+ *
+ * @package wp-job-manager
+ */
+
+namespace WP_Job_Manager\Admin;
+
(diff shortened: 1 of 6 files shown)


includes/admin/class-release-notice.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+class Release_Notice {
+
+	/**
+	 * Set up notice.
+	 */
+	public static function init() {
+		add_filter( 'wpjm_admin_notices', [ __CLASS__, 'add_release_notice' ] );
+
+		add_action( 'job_manager_action_enable_stats', fn() => \WP_Job_Manager_Settings::instance()->set_setting( \WP_Job_Manager\Stats::OPTION_ENABLE_STATS, '1' ) );
```
gikaragia: I think we were avoiding anonymous functions in actions as users were not able to remove them. Unless there is a workaround for that?

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -161,27 +168,48 @@ public static function init_core_notices() {
...
+			// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Nonce should not be modified.
+			if ( ! wp_verify_nonce( wp_unslash( $_GET['_wpjm_notice_nonce'] ), 'job_manager_notices_nonce' ) ) {
+				wp_die( esc_html__( 'Action failed. Please refresh the page and retry.', 'wp-job-manager' ) );
+			}
+
+			if ( ! current_user_can( 'manage_options' ) ) {
+				wp_die( esc_html__( 'You don&#8217;t have permission to do this.', 'wp-job-manager' ) );
+			}
+
+			do_action( 'job_manager_action_' . $action );
```
gikaragia, line 209: Personally I would avoid dynamic hook name because they are hard to find and document. You can pass the action as an argument

includes/admin/class-wp-job-manager-admin-notices.php
```diff
@@ -348,31 +395,43 @@ private static function save_dismissed_notices( $dismissed_notices, $is_user_not
...
+	}
+
+	/**
+	 * Save notice state as dismissed.
+	 *
+	 * @param int $notice_id Notice ID.
+	 *
+	 * @return void
+	 */
+	public static function dismiss_notice( $notice_id ) {
```
gikaragia, line 420:
```suggestion
	private static function dismiss_notice( $notice_id ) {
```

includes/admin/class-release-notice.php
```diff
@@ -0,0 +1,92 @@
...
+
+		$action_url            = \WP_Job_Manager_Admin_Notices::get_action_url( 'enable_stats', $notice_id );
+		$notices[ $notice_id ] = [
+			'type'          => 'site-wide',
+			'label'         => 'New',
+			'heading'       => 'Job Statistics',
+
+			'message'       => '<div>' . __(
+				'
+<p>Collect analytics data about site visitors for each job listing. Display the detailed statistics in the refreshed jobs dashboard.</p>
```
gikaragia:
```suggestion
<p>Collect analytics about site visitors for each job listing. Display the detailed statistics in the refreshed jobs dashboard.</p>
```

includes/admin/class-release-notice.php
```diff
@@ -0,0 +1,92 @@
...
+<ul>
+	<li>Page views and unique visitors with daily breakdown</li>
+	<li>Search impressions and apply button clicks</li>
+	<li>Add-on integration: Job alert impressions, bookmarks, application stats</li>
```
gikaragia: Maybe naming the plugins promotes them a bit more? For example
```suggestion
	<li>Integrates with Job Alerts, Applications, and Bookmarks.</li>
```

gikaragia (commented):
Looks great overall! First time taking a look at stats so forgive my ignorance on topics already discussed.

- Why do we have stats disabled by default? Generally, most people will simply dismiss notices without reading them and we shouldn't hide exciting new functionality behind settings I think. Maybe the actions in the notice should be 'Dismiss' and a redirect to settings for 'Disable'?
- I think the notice itself is a bit too big as it takes almost half of my screen. Is it possible to reduce its height a bit?
- I don't think that we should display the notice in new installations as it seems that it is directed to existing users. 

There is also some minor Psalm errors

includes/admin/class-release-notice.php
yscik: This one is kind of a one-off thing, but I moved it to a method in 17835c9

includes/admin/class-wp-job-manager-admin-notices.php
yscik: This is a pretty common pattern in WordPress and WPJM, like `wp_ajax_*`, `admin_action_*`. A bit harder to search but cleaner and faster to use.

includes/admin/class-wp-job-manager-admin-notices.php
yscik: Left this one to have a public method for devs to dismiss a single notice. It's used now in the install hook with the latest changes so gotta keep it public.

includes/admin/class-release-notice.php
yscik: Tweaked the text in 17835c9

yscik:
Set up the plugin install/update hook to enable stats for new installs and skip this notice. Note that to test updating, the version in `JOB_MANAGER_VERSION` has to be bumped. 

Also tweaked the size a bit, but the idea for this to be a little piece of news, I think we can take some real estate for it. It's only shown on the Job Manager screen — which eventually would be a 'Home' dashboard for news and things requiring action. 

I didn't want to complicate things and used the notice style, but we can tone it down a little and make this a more like News box:

image

includes/admin/class-wp-job-manager-admin-notices.php
gikaragia: Yeah whenever I encountered it in the existing cases it always needed extra time to figure out. It usually goes like this:
- You use the hotkey to go to the hook definition. This doesn't work because the IDE can't figure out dynamic hooks.
- Then I search for the whole name
- Then I search for parts of the hook. It is also very common during this step for the results to be too many as the parts of the hook name are very commonly used for other things like variables or function names.

I don't see how dynamic hooks are cleaner as instead of doing concatenation we will pass an extra argument. The only downside is an extra check on the called method which I don't think is important enough.

includes/admin/class-release-notice.php
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	public static function add_release_notice( $notices ) {
+
+		// Make sure to update the version number in the notice ID when changing this notice for a new release.
+		$action_url                 = \WP_Job_Manager_Admin_Notices::get_action_url( 'enable_stats', self::NOTICE_ID );
+		$notices[ self::NOTICE_ID ] = [
+			'type'          => 'site-wide',
+			'label'         => 'New',
+			'heading'       => 'Job Statistics',
+
```
gikaragia:
```suggestion
```

gikaragia (approved):
Looks better now! 
> Note that to test updating, the version in JOB_MANAGER_VERSION has to be bumped.

Probably I'm missing something here as I didn't need to do any updates for the notice to show. 

This comment made me think though, how are we going to handle the release notices in versions after 2.3.0. We maybe don't want to show the notice to users upgrading to 2.4.0 directly. Maybe we shouldn't mention the version at all in the notice? Or maybe add an extra step to the release steps to stop showing the Release_Notice class. Maybe it isn't big of a deal though and we could just show it to everyone updating to a a post-2.3.0 version.

There is also the issue with the dynamic hook but I don't consider it a blocker

yscik:
> This comment made me think though, how are we going to handle the release notices in versions after 2.3.0. We maybe don't want to show the notice to users upgrading to 2.4.0 directly. Maybe we shouldn't mention the version at all in the notice? Or maybe add an extra step to the release steps to stop showing the Release_Notice class. Maybe it isn't big of a deal though and we could just show it to everyone updating to a a post-2.3.0 version.
> 

Right now the idea is to update the `Release_Notice` class with new content for 2.4.0 — or just unhook it if we don't want to show a notice then. Eventually it'd be nice if this were a server-based noticed, so we can update it on wpjobmanager.com and even show what's new in an update to folks who haven't updated yet.

gikaragia:
Makes sense! Could we add a comment in the release steps as something to check?

yscik:
Updated the release checklist here: P6r3EZ-1Q-p2

PR #2799, opened 2024-03-28
Add usage tracking for settings

### LINKED ISSUE #2797: Stats: add usage tracking
Make sure we have data on the feature being enabled.

yscik:
Fixes #2797

* Send setting field data along when sending usage tracking data
* Configure which settings fields are included with a `track` key in the settings config
* Move a few things around so the Settings class is available
* Fix an issue with licensed plugin info not matching formatting rules for Tracks

### Testing Instructions

* Run the cron event `job_manager_usage_tracking_send_usage_data`
* Wait 10 minutes
* Check for the `wpjm_stats_log` event in Tracks Live View. Username can be used to filter, it's the full site URL (Like `http://paper.local`)
* Check that `settings_` fields are included

----

| Plugin build for 20d4d78c18def330d414ae224b989b06998f79e7 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2799-20d4d78c.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2799-20d4d78c)             |

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -322,7 +322,6 @@ abstract class WP_Job_Manager_Form {
 	 * Enqueue the scripts for the form.
 	 */
 	public function enqueue_scripts() {
-		_deprecated_function( __METHOD__, '$$next-version$$', 'WP_Job_Manager\WP_Job_Manager_Form::enqueue_scripts' );
 		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
 	}
 
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -251,7 +251,7 @@ class WP_Job_Manager_Settings {
 								'any' => __( 'Jobs will be shown if within ANY selected category', 'wp-job-manager' ),
 								'all' => __( 'Jobs will be shown if within ALL selected categories', 'wp-job-manager' ),
 							],
-							'track'   => 'bool',
+							'track'   => 'value',
 						],
 						[
 							'name'       => 'job_manager_enable_types',
@@ -312,7 +312,7 @@ class WP_Job_Manager_Settings {
 							'type'        => 'text',
 							'placeholder' => __( 'e.g. USD', 'wp-job-manager' ),
 							'attributes'  => [],
-							'track'       => 'bool',
+							'track'       => 'value',
 						],
 						[
 							'name'       => 'job_manager_enable_salary_unit',
@@ -333,7 +333,7 @@ class WP_Job_Manager_Settings {
 							'type'       => 'select',
 							'options'    => job_manager_get_salary_unit_options(),
 							'attributes' => [],
-							'track'      => 'bool',
+							'track'      => 'value',
 						],
 						[
 							'name'     => 'job_manager_display_location_address',
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -83,13 +83,23 @@ class WP_Job_Manager_Settings {
 	}
 
 	/**
+	 * Set a single setting.
+	 *
+	 * @param string $option Option name.
+	 * @param mixed  $value Value.
+	 */
+	public function set_setting( $option, $value ) {
+		update_option( $option, $value );
+	}
+
+	/**
 	 * Initializes the configuration for the plugin's setting fields.
 	 *
 	 * @access protected
 	 */
 	protected function init_settings() {
 		// Prepare roles option.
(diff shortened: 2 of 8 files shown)


includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -322,7 +322,6 @@ public function clear_fields() {
 	 * Enqueue the scripts for the form.
 	 */
 	public function enqueue_scripts() {
```
gikaragia, line 324: Is this actually called from somewhere?

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -230,6 +251,7 @@ protected function init_settings() {
 								'any' => __( 'Jobs will be shown if within ANY selected category', 'wp-job-manager' ),
 								'all' => __( 'Jobs will be shown if within ALL selected categories', 'wp-job-manager' ),
 							],
+							'track'   => 'bool',
```
gikaragia: This is either 'all' or 'some' I think

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -285,6 +312,7 @@ protected function init_settings() {
 							'type'        => 'text',
 							'placeholder' => __( 'e.g. USD', 'wp-job-manager' ),
 							'attributes'  => [],
+							'track'       => 'bool',
```
gikaragia: text?

includes/admin/class-wp-job-manager-settings.php
```diff
@@ -304,6 +333,7 @@ protected function init_settings() {
 							'type'       => 'select',
 							'options'    => job_manager_get_salary_unit_options(),
 							'attributes' => [],
+							'track'      => 'bool',
```
gikaragia: text?

includes/class-wp-job-manager-usage-tracking-data.php
```diff
@@ -62,14 +62,23 @@ public static function get_usage_data() {
...
 		$licensed_extensions = self::get_official_extensions( true );
 
-		$usage_data['official_extensions'] = count( $all_extenstions );
+		$usage_data['official_extensions'] = count( $all_extensions );
 		$usage_data['licensed_extensions'] = count( $licensed_extensions );
 
-		foreach ( array_keys( $all_extenstions ) as $installed_plugin ) {
-			$usage_data[ $installed_plugin ] = isset( $licensed_extensions[ $installed_plugin ] ) ? 'licensed' : 'unlicensed';
+		foreach ( array_keys( $all_extensions ) as $installed_plugin ) {
+			$name                = preg_replace( '/[^a-z0-9]/', '_', $installed_plugin );
```
gikaragia, line 79: This will make it practically impossible to track historical data for this stat I think

includes/class-wp-job-manager-usage-tracking-data.php
```diff
@@ -366,6 +375,62 @@ private static function has_paid_extensions() {
...
+				switch ( $option['track'] ) {
+					case 'bool':
+						$value = $value ? '1' : '0';
+						break;
+					case 'value':
+						if ( isset( $option['options'] ) && ! in_array( $value, array_keys( $option['options'] ), true ) ) {
+							unset( $value );
+						}
+						break;
+					case 'is-default':
```
gikaragia, line 408: I don't think that we use the is-default case anywhere. Maybe remove it?

gikaragia: Oh just seen the other PR :)

includes/abstracts/abstract-wp-job-manager-form.php
yscik: Oh looks like I had the warning about this because it's called in the Applications add-on's form (`WP_Job_Manager_Applications_Apply`) that extends this class. Since this is a generic base class used all around I'd keep the method even if we don't use it now in core.

includes/class-wp-job-manager-usage-tracking-data.php
yscik: Sorry didn't explain this well, this one is a fix: Tracks is rejecting the events with these plugin fields because the keys are invalid, they have to be snake_case.

includes/admin/class-wp-job-manager-settings.php
yscik: Fixing these in 20d4d78

includes/class-wp-job-manager-usage-tracking-data.php
gikaragia: So this never worked? Oh my! I remember testing it but I didn't do it end to end it seems..

gikaragia (approved):
LGTM!

PR #2802, opened 2024-04-09
Repair edit job listing screen.

gikaragia:
### Oveview
This fixes an issue to the edit job listing screen. To reproduce:

1. Create a job and publish it
2. Set up a recaptcha and enable it in the settings (any version will do, go to https://www.google.com/recaptcha to create keys if you don't have some already)
3. Go to the job dashboard
4. Edit the job
5. After doing some change try to save the listing
6. At this point the update fails and nothing happens

The problem was that we added the actions to the `WP_Job_Manager_Recaptcha` constructor. This way the hooks were added even when we wanted to call utility methods and the captcha was also enqueued to the dashboard page.

* I moved the hook registration from the constructor to `equeue_scripts`. This way we only enqueue the captcha when we need to.

### Testing Instructions

* Test together with: https://github.com/Automattic/wpjm-addons/pull/481
* Follow the steps from the overview and observe that nothing is broken now.
* Also try testing recaptcha functionality in general to ensure that nothing else is broken.
* Additionally, we should test how the changes in reCAPTCHA work in relation to different plugin versions:
	* The newest version of the paid plugins should work with the previous version of WPJM (i.e. 2.2.2) but with v2 reCAPTCHA only
	* The previous versions of the paid plugins should work with latest version of WPJM but with warnings printed as we deprecated the old methods.

----

| Plugin build for 42aae9cb2f587f2da93c4d5984a20c00db365607 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2802-42aae9cb.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2802-42aae9cb)             |

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -136,7 +136,7 @@ abstract class WP_Job_Manager_Form {
 	 * @param array $atts Attributes to use in the view handler.
 	 */
 	public function output( $atts = [] ) {
-		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
+		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_recaptcha();
 		$step_key = $this->get_step_key( $this->step );
 		$this->show_errors();
 		$this->show_messages();
@@ -323,7 +323,7 @@ abstract class WP_Job_Manager_Form {
 	 */
 	public function enqueue_scripts() {
 		_deprecated_function( __METHOD__, '$$next-version$$', 'WP_Job_Manager\WP_Job_Manager_Form::enqueue_scripts' );
-		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
+		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_recaptcha();
 	}
 
 	/**
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -322,7 +322,44 @@ abstract class WP_Job_Manager_Form {
 	 */
 	public function enqueue_scripts() {
 		_deprecated_function( __METHOD__, '$$next-version$$', 'WP_Job_Manager\WP_Job_Manager_Form::enqueue_scripts' );
-		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
+		WP_Job_Manager\WP_Job_Manager_Recaptcha::instance()->enqueue_scripts();
+	}
+
+
+	/**
+	 * Output the reCAPTCHA field.
+	 *
+	 * @deprecated
+	 */
+	public function display_recaptcha_field() {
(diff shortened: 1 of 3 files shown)


includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -136,7 +136,7 @@ public function process() {
 	 * @param array $atts Attributes to use in the view handler.
 	 */
 	public function output( $atts = [] ) {
-		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_scripts();
+		WP_Job_Manager\WP_Job_Manager_Recaptcha::enqueue_recaptcha();
```
yscik: Haven't tested yet, but it looks fragile to add the hooks that process submit fields in the output and enqueue functions. Those might not get called or get called later than `process()`. Does it only work now because the submit job form has multiple steps?

Since reCAPTCHA is only used on specific forms I think the abstract `WP_Job_Manager_Form` class, which has the common code for every form, isn't the place to set it up. Looks like there are separate `WP_Job_Manager_Form_Edit_Job` and `WP_Job_Manager_Form_Submit_Job` classes, should the hooks be added from here?:

https://github.com/Automattic/WP-Job-Manager/blob/881c4f3d5afc30fcc4c3400b225d5dd2752ef692/includes/forms/class-wp-job-manager-form-submit-job.php#L83

(Unfortunately `WP_Job_Manager_Form_Edit_Job` extends `WP_Job_Manager_Form_Submit_Job` so we still need something to not load it for editing. Maybe a `...Submit_Job::init_recaptcha()` method that the `...Edit_Job` class overrides?)

gikaragia: Great point, I also thought about this initially but I ended up taking the quick approach since it worked. I'll look into refactoring

gikaragia: Ok so I moved everything inside the recaptcha class. I also did some fixes regarding versioning as it seems that previously the latest version of the paid plugins didn't work with old WPJM versions. I have updated the testing instructions to reflect that.

includes/abstracts/abstract-wp-job-manager-form.php
```diff
@@ -323,7 +322,44 @@ public function clear_fields() {
...
+	}
+
+	/**
+	 * Validate a reCAPTCHA field.
+	 *
+	 * @param bool $success
+	 *
+	 * @deprecated
+	 *
+	 * @return bool|WP_Error
```
yscik:
```suggestion
	 * @return bool|\WP_Error
```

includes/class-wp-job-manager-recaptcha.php
```diff
@@ -118,6 +141,8 @@ public function is_recaptcha_available() {
 	/**
 	 * Dispaly the reCAPTCHA field in the form.
```
yscik:
```suggestion
	 * Display the reCAPTCHA field in the form.
```

yscik (approved):
Looks good! :shipit:

PR #2803, opened 2024-04-10
Introduce a function for calculating scheduled dates

gikaragia:
### Overview

- A simple refactor to enable https://github.com/Automattic/wpjm-addons/pull/480.
- As we need to set the post dates in many places, I introduced a function to avoid repetition.

### Testing instructions
- Should be tested with linked PR

----

| Plugin build for 7ae9d33b1ad8c9fd7c7bfa6afc82cc70fe767e32 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2803-7ae9d33b.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2803-7ae9d33b)             |

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -843,7 +843,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		];
 
 		if ( ! empty( $values['job']['job_schedule_listing'] ) ) {
-			$is_scheduled_date = $this->calculate_scheduled_dates( $job_data, $values['job']['job_schedule_listing'] );
+			$is_scheduled_date = $this->apply_scheduled_date( $job_data, $values['job']['job_schedule_listing'] );
 
 			if ( ! $is_scheduled_date ) {
 				unset( $values['job']['job_schedule_listing'] );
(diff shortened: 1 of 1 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -1071,14 +1068,38 @@ public function preview() {
...
 
+	/**
+	 * Helper which formats the scheduled date and sets the appropriate dates in the job data.
+	 *
+	 * @param array  $job_data       The job data array to modify.
+	 * @param string $scheduled_date The scheduled date.
+	 *
+	 * @return bool True when the scheduled date is a valid future date, false otherwise.
+	 */
+	public static function calculate_scheduled_dates( array &$job_data, string $scheduled_date ): bool {
```
yscik: Would something like `apply_scheduled_date` or `set_scheduled_date` be a more appropriate name for this?

yscik (approved):
Looks good!

includes/forms/class-wp-job-manager-form-submit-job.php
gikaragia: Yep much better

PR #2808, opened 2024-04-19
Adds a filter that disables promoted jobs

gikaragia:
Part of https://github.com/Automattic/greenhouse/issues/2085

* Adds a filter that disables promoted jobs functionality. 
* We can merge this now and set the default value to false whenever we want to disable it.

### Testing Instructions

* Add the snippet `add_filter('job_manager_enable_promoted_jobs', '__return_false' );`
* Go through admin and observe that there is no promoted jobs functionality and no console errors:
		* Delete `wp_job_manager_dismissed_notices` both from the options table and the users table to reset notices
		* Publish a new job and observe that no popup is shown
		* Go to job listings and observe that no Promote button exists
* Delete the snippet, repeat the steps, and observe that promote functionality works.

----

| Plugin build for 537bde6121b4a93d7d01d5e827e613c64752794c <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2808-537bde61.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2808-537bde61)             |

### DIFF
--- a/includes/admin/class-wp-job-manager-admin.php
+++ b/includes/admin/class-wp-job-manager-admin.php
@@ -178,10 +178,8 @@ class WP_Job_Manager_Admin {
 			}
 		}
 
-		if ( $this->are_promoted_jobs_enabled ) {
-			WP_Job_manager::register_script( 'job_manager_notice_dismiss', 'js/admin/wpjm-notice-dismiss.js', null, true );
-			wp_enqueue_script( 'job_manager_notice_dismiss' );
-		}
+		WP_Job_manager::register_script( 'job_manager_notice_dismiss', 'js/admin/wpjm-notice-dismiss.js', null, true );
+		wp_enqueue_script( 'job_manager_notice_dismiss' );
 
 		WP_Job_Manager::register_style( 'job_manager_admin_menu_css', 'css/menu.css', [] );
 		wp_enqueue_style( 'job_manager_admin_menu_css' );
--- a/includes/admin/class-wp-job-manager-admin.php
+++ b/includes/admin/class-wp-job-manager-admin.php
@@ -34,6 +34,13 @@ class WP_Job_Manager_Admin {
 	private $settings_page;
 
 	/**
+	 * Whether promoted jobs are enabled.
+	 *
+	 * @var bool
+	 */
+	private $are_promoted_jobs_enabled;
+
+	/**
 	 * Allows for accessing single instance of class. Class should only be constructed once per call.
 	 *
 	 * @since  1.26.0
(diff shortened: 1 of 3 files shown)


includes/admin/class-wp-job-manager-admin.php
```diff
@@ -163,8 +178,10 @@ public function admin_enqueue_scripts() {
 			}
 		}
 
-		WP_Job_manager::register_script( 'job_manager_notice_dismiss', 'js/admin/wpjm-notice-dismiss.js', null, true );
-		wp_enqueue_script( 'job_manager_notice_dismiss' );
+		if ( $this->are_promoted_jobs_enabled ) {
+			WP_Job_manager::register_script( 'job_manager_notice_dismiss', 'js/admin/wpjm-notice-dismiss.js', null, true );
+			wp_enqueue_script( 'job_manager_notice_dismiss' );
```
yscik: This script is used for the other notices too, to hook up the dismiss button that WordPress adds to the WPJM notice system.

includes/admin/class-wp-job-manager-admin.php
```diff
@@ -58,7 +65,14 @@ public function __construct() {
 		include_once dirname( __FILE__ ) . '/class-wp-job-manager-cpt.php';
 		WP_Job_Manager_CPT::instance();
 
-		include_once dirname( __FILE__ ) . '/class-wp-job-manager-promoted-jobs-admin.php';
+		/**
+		 * Documented in class-wp-job-manager.php
+		 */
+		$this->are_promoted_jobs_enabled = apply_filters( 'job_manager_enable_promoted_jobs', true );
```
yscik, line 71: I'm thinking it's fine to just call the filter in multiple places, no need for a new class property for this.

yscik (commented):
Looks good, just small things

includes/admin/class-wp-job-manager-admin.php
gikaragia: Good catch! Didn't realize that fixed in: https://github.com/Automattic/WP-Job-Manager/pull/2808/commits/537bde6121b4a93d7d01d5e827e613c64752794c

includes/admin/class-wp-job-manager-admin.php
gikaragia: Hmm I am not sure I agree with this one. It will be easier to update when we change the default value. Also we will probably need to repeat the comments too if we want all filters to be documented.

yscik (approved):
👌

PR #2809, opened 2024-04-20
Update emogrifier

### LINKED ISSUE #2755: E-mail top and bottom margin missing
Image

mikeyarce:
Fixes #2755 

* Update to use new version of Emogrifier 
* Load Emogrifier through a composer dependency
* New plugin build so that we can include the composer dependency 

### Testing Instructions

* Check out PR
* Run `npm run build`
* Make sure that the new plugin build matches the old plugin build.  You can run `npm run build:assets && npm run archive` which is the previous build and check folders to make sure it's what we expect. It should be identical except for the `tests` are gone from the `usage-tracking` folder, and now we do include the `vendor` folder for emogrifier.  
Here is my folder comparison including hidden files:

* Now use this plugin on a site that can send emails
* Add a few alerts
* Send the alerts to a gmail account and make sure the top and bottom margin is there!

Example of margin:

### Release Notes

* Fix: Email notifications - top and bottom margin not displaying properly for some email clients.

----

| Plugin build for 9215cf69c667681fe97648d9e1448b07f6791b1e <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2809-9215cf69.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2809-9215cf69)             |


includes/class-wp-job-manager-email-notifications.php
```diff
@@ -164,9 +166,8 @@ public static function lazy_init() {
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/emails/class-wp-job-manager-email-employer-expiring-job.php';
 		include_once JOB_MANAGER_PLUGIN_DIR . '/includes/emails/class-wp-job-manager-email-admin-expiring-job.php';
 
-		if ( ! class_exists( 'Emogrifier' ) && class_exists( 'DOMDocument' ) && version_compare( PHP_VERSION, '5.5', '>=' ) ) {
-			include_once JOB_MANAGER_PLUGIN_DIR . '/lib/emogrifier/class-emogrifier.php';
-		}
+		// Load Vendor Autoload for Emogrifier.
```
gikaragia: We don't need to be specific here, we could add more libraries in the future.

includes/class-wp-job-manager-email-notifications.php
```diff
@@ -933,10 +934,12 @@ private static function get_email_content( $email_notification_key, $args, $is_p
 	 * @return string
 	 */
 	private static function inject_styles( $content ) {
-		if ( class_exists( 'Emogrifier' ) ) {
+		$css_inliner_class = CssInliner::class;
```
gikaragia: So `CssInliner::class` is equal to `CssInliner`. So this code will either fail or it doesn't makes sense to check for existence.

package.json
```diff
@@ -34,7 +34,7 @@
     "zx": "^7.2.3"
   },
   "scripts": {
-    "build": "npm run build:assets && npm run archive",
```
gikaragia, line 37: It seems like we are going to replace the archive command? If this is the case, maybe it makes sense to remove the `archive.exclude` arguments from composer.json? Keep in mind although it is called exclude we do a `!` so they are actually includes.

We should also remove the old archive command

gikaragia (commented):
Works well! Left a few comments

includes/class-wp-job-manager-email-notifications.php
yscik: Just checked this today and `class_exists( CssInliner::class )` correctly returns false and doesn't throw an error if the class is not found. Agree that the `$css_inliner_class` variable is not needed, might even break things?

yscik: (Apparently this is a compile-time construct, it's just returns the class FQN without the class needing to exists)
image

gikaragia: Yep the variable assignment was the one that I was worried about. I am amazed actually that it doesn't fail!

Still it would be easier to read without the variable.

includes/class-wp-job-manager-email-notifications.php
mikeyarce: Good point!  done in 856bbf44d8958098a46a392ccac4262bf60ecbf2

includes/class-wp-job-manager-email-notifications.php
mikeyarce: Good call!  done in b243532497362908204fcc2633842c1b1f93d347

package.json
mikeyarce: Done in 9215cf69c667681fe97648d9e1448b07f6791b1e

mikeyarce:
Thanks @gikaragia - made the changes 😄

gikaragia (approved):
LGTM!

PR #2810, opened 2024-04-22
Use theme colors for job overlay, improve dashboard on small screens

yscik:

* Expand the theme color detection script. Detect and calculate background and contrast colors based on the the theme.
  * Fixes color issues on dark themes, and daily chart colors when the link color is black.
* Improve job dashboard table, add column labels on mobile, improve breaking points, tweak some styles
* Fix outputting modal code multiple times

### Testing Instructions

* Enable stats
* Install a few popular themes and check that the job dashboard and the stats overlay look reasonable out of the box
* Twenty Twenty-Three, Twenty Twenty-Four has a bunch of color styles that are easy to switch in the site editor, including dark colors.
* Check that the overlay also works in the admin

* `job_manager_ui_theme_support_script` - disable theme color support script

### Screenshot / Video

image
image

----

| Plugin build for df2ed4356240699fa85e622a7ebadc2c53d45716 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2810-df2ed435.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2810-df2ed435)             |


templates/job-dashboard.php
```diff
@@ -68,16 +70,18 @@ class="jm-dashboard-empty">
 			<div class="jm-dashboard-header">
 				<?php foreach ( $job_dashboard_columns as $key => $column ) : ?>
 					<div
-						class="jm-dashboard-job-column <?php echo esc_attr( $key ); ?>"><?php echo esc_html( $column ); ?></div>
+						class="jm-dashboard-job-column jm-dashboard-job-column-label <?php echo esc_attr( $key ); ?>"><?php echo esc_html( $column ); ?></div>
```
gikaragia: I am probably very nitpicky here but maybe this div and the one below might be easier to read if they are on the same line. Feel free to ignore if you think they look ok

templates/job-dashboard.php
```diff
@@ -68,16 +70,18 @@ class="jm-dashboard-empty">
...
-				<div class="jm-dashboard-job-column actions"><?php esc_html_e( 'Actions', 'wp-job-manager' ); ?></div>
+				<div class="jm-dashboard-job-column jm-dashboard-job-column-label actions"><?php esc_html_e( 'Actions', 'wp-job-manager' ); ?></div>
 			</div>
 			<div class="jm-dashboard-rows">
 				<?php foreach ( $jobs as $job ) : ?>
 					<div class="jm-dashboard-job">
 						<?php foreach ( $job_dashboard_columns as $key => $column ) : ?>
 							<div class="jm-dashboard-job-column <?php echo esc_attr( $key ); ?>"
 								aria-label="<?php echo esc_attr( $column ); ?>">
+								<div
```
gikaragia: I don't think I understand why this was added here, although it seems that we hide it with CSS. Can you explain please?

includes/ui/class-ui.php
```diff
@@ -65,11 +67,13 @@ public function register_styles() {
 	 * @access private
 	 */
 	public function enqueue_styles() {
-		\WP_Job_Manager::register_style( 'wp-job-manager-ui', 'css/ui.css', [] );
-
 		if ( $this->has_ui || wp_style_is( 'wp-job-manager-ui', 'enqueued' ) ) {
 			wp_enqueue_style( 'wp-job-manager-ui' );
 
+			if ( apply_filters( 'job_manager_ui_theme_support_script', true ) ) {
```
gikaragia, line 76: Should we add some documentation to the filter?

includes/class-wp-job-manager.php
```diff
@@ -572,8 +572,6 @@ public function frontend_scripts() {
 			]
 		);
 
-		Job_Overlay::instance()->init_dashboard_overlay();
```
gikaragia, line 575: Can you explain why this was removed?

includes/class-job-dashboard-shortcode.php
```diff
@@ -153,6 +153,8 @@ public function output_job_dashboard( $attrs ) {
 			$job_actions[ $job->ID ] = $this->get_job_actions( $job );
 		}
 
+		do_action( 'job_manager_job_dashboard_before', $jobs );
```
gikaragia, line 161: I imagine `job_manager_job_dashboard` was only added recently right? Can we add some documentation to the other two hooks?

assets/js/ui-theme-support.js
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * Detect and compute style variables based on the site theme.
+ * Skips any variable that's been explicitly set already.
+ */
+function applyThemeStyles() {
+	const style = getComputedStyle( document.documentElement );
+	for ( const [ key, color ] of Object.entries( uiVariables ) ) {
+		let value = style.getPropertyValue( key );
+
+		const hasCustomValue = color.default ? color.default !== value : !! value;
```
gikaragia: There's just too much going on in this line. I think it would be easier to read if we write this as an if/else and turn the `!! value` to `Boolean( value )`

assets/js/ui-theme-support.js
```diff
@@ -0,0 +1,169 @@
...
+		main.appendChild( postTag );
+		this.linkColor = Color.parse( getComputedStyle( linkTag ).color );
+		this.textColor = Color.parse( getComputedStyle( textTag ).color );
+		this.backgroundColor = Color.parse( getComputedStyle( document.body ).backgroundColor );
+		this.calculate();
+		postTag.remove();
+
+		this.bodyFontFamily = getComputedStyle( document.body ).fontFamily;
+	},
+	calculate() {
```
gikaragia: I don't think that we need this method, we can set `darkMode` in `detect` directly.

assets/css/job-dashboard.scss
```diff
@@ -149,4 +186,28 @@
 		flex-basis: calc( 100% - var(--jm-dashboard-company-logo-size) - var(--jm-ui-space-sm) );
 		order: -1;
 	}
+
+}
+
+@container jm-dashboard-table-large (width < 940px)
```
gikaragia, line 192: TIL about container

assets/js/ui-theme-support.js
```diff
@@ -0,0 +1,169 @@
...
+ * Skips any variable that's been explicitly set already.
+ */
+function applyThemeStyles() {
+	const style = getComputedStyle( document.documentElement );
+	for ( const [ key, color ] of Object.entries( uiVariables ) ) {
+		let value = style.getPropertyValue( key );
+
+		const hasCustomValue = color.default ? color.default !== value : !! value;
+
+		if ( ! hasCustomValue ) {
```
gikaragia, line 110: So IIUC this code should detect if the user has set any of the CSS variables and use these values instead right? So I am trying to specify these values but they get ignored. What I did was adding this in the customizer:
```
:root {
	--jm-ui-background-color: red;
	--jm-ui-accent-color: red;
	--jm-ui-text-color: red;
	--jm-ui-accent-alt-color: red;
}
```
But the background has still the same color.

Probably this gets overriden by `	--jm-ui-background-color: initial;` in neutral.scss as when I remove this it works.

Also for `--jm-ui-accent-color-contrast` specifically, I think that the calculations from ThemeStyles are always ignored as we don't have a default value in Color and we set a default value of #ffffff in neutral.scss. Was this intentional?

gikaragia (commented):
Although I think the solution is impressive, I also think that we may overcomplicate things while trying to make things look consistent. Maybe we should focus on simply providing the elements and let users handle more opinionated styles like colors?

Is it possible that having CSS variables makes it harder for our users to modify the colors as they will need to figure out how we use them? On the other side, maybe having variables is a common thing now and in the end they can always override everything so maybe it's not big of an issue.

yscik:
So regarding CSS variables, the new styles rely pretty heavily on it, and it makes it pretty easy to customize things with just a few variables. This is an alternative to rewriting or overriding everything with CSS, which is what they currently have to do. I'm putting together a documentation page with some examples to get started.

This approach also makes it easy to add an UI on top in the future, where they can change the colors in the site editor or the setup wizard.

This PR adds the color detections so styles are not broken out of the box on various themes. Most of it is for the new overlay modal, where we'd end up with stuff like white links on white background if it's a dark theme.

assets/js/ui-theme-support.js
yscik: Yup that's how it's supposed to work, if the colors are defined manually then this shouldn't override them. I'll take a look.

templates/job-dashboard.php
yscik: It's hidden on desktop only when we have the labels in the table heading, but shown on mobile for the alternative layout:

image

includes/class-wp-job-manager.php
yscik: It's already called from the shortcode class [here](https://github.com/Automattic/WP-Job-Manager/blob/f1d376fddb2853b2bd41cceffce52245b06f6a84/includes/class-job-dashboard-shortcode.php#L104), this was just a mistake earlier. (And a bug, because we outputted the <dialog> element with the same ID multiple times, which broken the JS)

assets/js/ui-theme-support.js
yscik: Apparently this only worked for block themes, lowered the specificity of the variables in ui.neutral.css in b5d3ae5

assets/js/ui-theme-support.js
yscik: Improved this in 8998cec along with the other fixes.

yscik:
Thanks for the review @gikaragia, all should be fixed now!

gikaragia (approved):
Looks great!

PR #2812, opened 2024-04-23
Add possibility to force sync the WPCOM license

fjorgemota:
Fixes https://github.com/Automattic/wpjobmanager.com/issues/721

* Add WPCOM License sync functionality to WPJM when running on WPCOM

### Testing Instructions

1. Checkout https://github.com/Automattic/wpjobmanager.com/pull/724.
2. Make sure your site is pointing to your test env of wpjobmanager.com. You can do by setting the constant `JOB_MANAGER_DEV_API_BASE_URL` to the base URL of your local WPJMCOM instance in your `wp-config.php`. For instance:
```php
define( "JOB_MANAGER_DEV_API_BASE_URL", "http://wpjobmanager.local" );
```
3. Using the Code Snippets plugin, add a filter to make WPJM think that an addon was actually provided by WPCOM. This is the code of the filter to add:
```php
add_filter("default_option_wpcom_active_subscriptions", function( $value ) {
	$value["wp-job-manager-applications"] = true;
	return $value;
} );
```
(you can change the slug of the addon if you want, but please change it as well in the next steps)
5. In the wpjobmanager.com's `class-flush-wpcom-activation-endpoint.php` (added in the PR linked above):
  5.1. Update the access token as described in the step 8 in https://github.com/Automattic/wpjobmanager.com/pull/724
  5.2. Comment this line because your local site won't exist on WPCOM: https://github.com/Automattic/wpjobmanager.com/pull/724/files#diff-a4f394b8eec10682db698affd6bffeab207386eb4581b0eb937a69568284c861R167 
  5.3. Create an ownership as described from steps 1 to step 5 in https://github.com/Automattic/wpjobmanager.com/pull/724, and hard-code the value of `$wpcom_marketplace_ownership_id` to match the ownership ID you just created.
6. Make sure you don't have a license activated in your local WPJM site. You can always reset all the WPJM licenses in a site by using `wp option delete job_manager_helper`.
7. Navigate to the Licenses page: `/wp-admin/edit.php?post_type=job_listing&page=job-manager-marketplace&section=helper`. It should show you something like this:
Deactivated WPCOM product
8. Click on the "Activate License" button to initiate the process
9. If:
9.1. There's an error, you should see this:
Activation error
9.2. If the process was successful, you should see this:
Activation Successful

Also make sure the normal license activation and deactivation continue working properly.

1. Revert the changes above
2. Buy a WPJM Addon on your local wpjobmanager.com site.
3. Try to activate and deactivate the license through your WPJM site, and make sure it works properly.

### Context

This is akin to the PR in https://github.com/Automattic/sensei-pro/pull/2529

----

| Plugin build for 67ffb319bce5b0d3d1c1fbb4e8a155857bd808db <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/04/wp-job-manager-zip-2812-67ffb319.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/04/2812-67ffb319)             |

### DIFF
--- a/includes/3rd-party/wpcom.php
+++ b/includes/3rd-party/wpcom.php
@@ -86,7 +86,25 @@ function wpjm_display_managed_by_wpcom_notice_for_addon( $product_slug ) {
 		return;
 	}
 
-	esc_html_e( 'The license for this add-on is automatically managed by WordPress.com.', 'wp-job-manager' );
+	$helper      = WP_Job_Manager_Helper::instance();
+	$has_license = $helper->has_plugin_license( $product_slug );
+
+	?>
+	<form method="post" class="plugin-license-form">
+		<?php wp_nonce_field( 'wpjm-manage-license' ); ?>
+		<input type="hidden" id="<?php echo esc_attr( sanitize_title( $product_slug ) ); ?>_action" name="action" value="flush-wpcom-license"/>
+		<input type="hidden" id="<?php echo esc_attr( sanitize_title( $product_slug ) ); ?>_plugin" name="product_slug" value="<?php echo esc_attr( $product_slug ); ?>"/>
+
+		<?php if ( $has_license ) : ?>
+			<img src="<?php echo esc_url( JOB_MANAGER_PLUGIN_URL . '/assets/images/icons/checkmark-icon.svg' ); ?>" class='plugin-license-checkmark' aria-hidden='true' alt='<?php esc_attr_e( 'Plugin is activated', 'wp-job-manager' ); ?>'/>
+			<label class="plugin-license-label"><?php esc_html_e( 'The license for this add-on is automatically managed by WordPress.com.', 'wp-job-manager' ); ?></label>
+		<?php else : ?>
+			<label class="plugin-license-label"><?php esc_html_e( 'The license is not activated. Click on the following button to activate it now.', 'wp-job-manager' ); ?></label>
+
+			<input type="submit" class="button plugin-license-button" name="submit" value="<?php esc_attr_e( 'Activate License', 'wp-job-manager' ); ?>" />
+		<?php endif; ?>
+	</form>
+	<?php
 }
 
 add_action( 'wpjm_manage_license_page_after_license_form', 'wpjm_display_managed_by_wpcom_notice_for_addon' );
--- a/includes/3rd-party/wpcom.php
+++ b/includes/3rd-party/wpcom.php
@@ -100,7 +100,6 @@ function wpjm_display_managed_by_wpcom_notice_for_addon( $product_slug ) {
 			<label class="plugin-license-label"><?php esc_html_e( 'The license for this add-on is automatically managed by WordPress.com.', 'wp-job-manager' ); ?></label>
 		<?php else : ?>
 			<label class="plugin-license-label"><?php esc_html_e( 'The license is not activated. Click on the following button to activate it now.', 'wp-job-manager' ); ?></label>
-
 			<input type="submit" class="button plugin-license-button" name="submit" value="<?php esc_attr_e( 'Activate License', 'wp-job-manager' ); ?>" />
 		<?php endif; ?>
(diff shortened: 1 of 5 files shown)


includes/3rd-party/wpcom.php
```diff
@@ -86,7 +86,25 @@ function wpjm_display_managed_by_wpcom_notice_for_addon( $product_slug ) {
...
+		<?php wp_nonce_field( 'wpjm-manage-license' ); ?>
+		<input type="hidden" id="<?php echo esc_attr( sanitize_title( $product_slug ) ); ?>_action" name="action" value="flush-wpcom-license"/>
+		<input type="hidden" id="<?php echo esc_attr( sanitize_title( $product_slug ) ); ?>_plugin" name="product_slug" value="<?php echo esc_attr( $product_slug ); ?>"/>
+
+		<?php if ( $has_license ) : ?>
+			<img src="<?php echo esc_url( JOB_MANAGER_PLUGIN_URL . '/assets/images/icons/checkmark-icon.svg' ); ?>" class='plugin-license-checkmark' aria-hidden='true' alt='<?php esc_attr_e( 'Plugin is activated', 'wp-job-manager' ); ?>'/>
+			<label class="plugin-license-label"><?php esc_html_e( 'The license for this add-on is automatically managed by WordPress.com.', 'wp-job-manager' ); ?></label>
+		<?php else : ?>
+			<label class="plugin-license-label"><?php esc_html_e( 'The license is not activated. Click on the following button to activate it now.', 'wp-job-manager' ); ?></label>
+
```
gikaragia:
```suggestion
```

includes/helper/class-wp-job-manager-helper-nonce.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * WP_Job_Manager_Helper_Nonce
+ */
+class WP_Job_Manager_Helper_Nonce {
```
gikaragia, line 15: We now have a third class after `WP_Job_Manager_Site_Trust_Token` and `Access_Token` which roughly do the same thing. It would be a good idea to merge these at some point, but let's consider this as part of another PR.

includes/helper/class-wp-job-manager-helper-rest-api.php
```diff
@@ -0,0 +1,139 @@
...
+
+	/**
+	 * Register the REST routes related to licensing.
+	 *
+	 * @return void
+	 */
+	public function register_rest_routes() {
+		register_rest_route(
+			self::NAMESPACE,
+			self::REST_BASE . '/receive-wpcom-license-key',
```
gikaragia, line 64: We could simply name this endpoint `license` and the PUT/POST method will imply that the license is created.

gikaragia (commented):
Codewise looks good in general! Haven't tested yet, I'll do it after reviewing the other PR

gikaragia (approved):
Works well! There are some Psalm errors

includes/helper/class-wp-job-manager-helper-rest-api.php
yscik: I'm a bit afraid to rename it now, gonna merge to keep the release rolling 🛼

includes/helper/class-wp-job-manager-helper-nonce.php
fjorgemota: Hey there!

hmmm, I considered using the `Site_Trust_Token` thing because it is a bit similar, yeah, however, in this case we're looking for a *global* authorization, not really associated with a post or user or something similar. So, the Site Trust Token idea wouldn't work here.

That said, I just learned about `Access_Token`, and I can confirm that it should work perfectly! 

Good catch!

Issue #1707, opened 2019-03-29
Allow 'OR' for tags AND keywords in searches


ickeinmunich:
Hi,

there are several other open issues in regards to the search but - imho - none of them describes this issue. Let me know if this is something already suggested.

**How to reproduce**
1. Search for a keyword which doesn't find anything.
2. Don't remove the keyword from the search and add a tag.
3. No result will be shown.
4. Remove the keyword and results for the tag will be shown.

**I expected** to see the result of the search including the keyword (0 results) and the tag (several results).

As it was explained to me, **the reason** for this:
So the way the Keywork + Tags search work is what we call in programming an AND combination. I.e. the results listed will have both the Keyword plus the Tag which the user has selected. What you are looking for is an OR search result. i.e the results should list Jobs containing any one of the Keyword or the Tag

Since there is already an option in the backend for the tags - just include the keywords a user might search for in this option or duplicate this option + add the keywords to allow more specific customization for end users. -> https://prnt.sc/n4j66c

**The reason, why I'd like to see this added:** A mutual user won't remove the keyword from the search. The user will see: "Oh, no results. Ok, so let's try the job tags provided." Unfortunately, since no results are shown (even if the job tags would provide jobs), the user will leave the website having the impression of "currently no jobs are available at all".

Thank you.

richardmtl:
Also requested in https://wordpress.org/support/topic/querying-more-labels-at-once/

dericleeyy:
3079318-zen

tripflex:
Had a good few of requests regarding this as the "keywords" terminology is confusing, as users believe the search is handled just that way, as if they are "keywords" not a specific "search string"

My first thought is maybe we can modify the query to use something like this:
https://dev.mysql.com/doc/refman/5.6/en/fulltext-boolean.html

That way we don't have to deal with doing any kind of splitting on spaces, maybe just omitting any kind of operator so it's done as an OR throughout.

tripflex:
@jinnypark but like ... is anybody ever going to do anything about this?  Seems you guys have abandoned this project ...

ctdealba:
[Support forum](https://wordpress.org/support/topic/search-improvement-3/)

>If I search “Site Engineer” it will run the search on “Site” or “Engineer” and doesn’t organize the results in any sort of preference. So if the job description has “site cleaner” in the text this could appear first in the list. The results dont seem to be weighted by relevancy to the term

LimitlessDonald:
[...]

PR #2821, opened 2024-05-01
Expand job listing search

### LINKED ISSUE #2749: Keyword Search: Meta queries don't match WP_Query search & Support Modifiers for AND/OR/EXACT
I've had numerous clients request some functionality enhancement to the keyword searching functionality, mainly around the fact that every word is searched using a LIKE comparison operator so `php developer` would return any results for `php` AND any results for `developer`

When this is then checked for post meta, a LIKE is generated using `%php developer%` which is a bit inconsistent with how the `s` parameter works for searches.  So the native `s` parameter in `WP_Query` does `%php%` and `%developer%` the meta queries added do `%php developer%`

What i've seen the most request for is being able to do something like you can on Google where you use quotes to define a specific term wanted, so `"php developer"` would return ONLY results that match on `php developer` (sorta how meta queries are handled now, but `s` queries are not)

May be worth looking at revamping the keyword searches to support modifiers or at a minimum keep things consistent between `s` and the meta queries generated.

My initial guess would be to pass `exact` to `WP_Query` (i think this just tells not to do wildcard `%` and doesn't split the keywords by spaces) and maybe could do this by checking for quotes around values (strip those out and then set `exact` on `WP_Query`)

gikaragia:
Fixes #2749

### Overview
There are many requests for improving the job search functionality in #1707. The most frequent one being to allow an 'OR' operator when searching. So if for example, a user enters 'software engineer' in search then the listings that contain either 'software' or 'engineer' should be returned. I don't think that we should make this the default behavior as:
- WP core search uses terms with an AND operator which means that it returns posts that contain both terms.
- From a user perspective the AND seems more useful as it allows to filter through many results. I would expect when adding more terms to make the search more specific than the other way around.
If we want to ever allow this, we should specifically support 'OR' and 'AND' operators.

Another, confusing part is that WP core search does partial searches which means that if someone searches for "Java" it will return listings that contain "JavaSript".

Finally, there are also the issues mentioned in #2749. We currently also search the meta fields but not in a consistent way with core. We also don't support negations (-) and phrase searches ("") for meta fields and categories.

This PR solves the issues in #2749 by providing the same functionality for meta and term searches with WP Core (which searches title content and excerpt). This means that for meta and categories:
- We now split search terms and look for each one individually.
- Each term has to exist to at least one of the metas, categories, post title, content and excerpt.
- Users can perform reverse searches with the minus sign e.g. -intern
- Users can perform phrase searches e.g. "software developer"

* get_job_listings_keyword_search had been reimplemented to construct the search clause from the beginning. Initially, I tried solutions that appended to the existing search clause, but it wasn't possible to achieve accurate results as the core search clauses belonged to a different combination of ANDs and ORs
* Now the SQL produced for each search term looks like this:
```
( ( wp_posts.post_title LIKE '%engineer%' )
                OR ( wp_posts.post_excerpt LIKE '%engineer%' )
                OR ( wp_posts.post_content LIKE '%engineer%' )
                OR wp_posts.id IN (SELECT post_id
                                   FROM   wp_postmeta
                                   WHERE  meta_key IN
                                          ( '_application',
                                            '_company_name'
                                            ,
                                            '_company_tagline',
                                                    '_company_website',
                                            '_company_twitter'
                                          )
                                          AND meta_value LIKE '%engineer%')
                OR wp_posts.id IN (SELECT object_id
                                   FROM   wp_term_relationships AS tr
                                          LEFT JOIN wp_term_taxonomy AS tt
                                                 ON tr.term_taxonomy_id =
                                                    tt.term_taxonomy_id
                                          LEFT JOIN wp_terms AS t
                                                 ON tt.term_id = t.term_id
                                   WHERE  t.name LIKE '%engineer%') )
AND ( ( wp_posts.post_title LIKE '%engin%' ) .....
```
* WP core limits the search terms count to 9 so I don't think that there is a risk for the query to become too large.
* I removed the `job_listing_search_conditions` hook as it wasn't possible to maintain anymore. Users can always override the whole `get_job_listings_keyword_search` function with the previous version to revert to the previous functionality.

### Testing Instructions

* Add listings that contain possible search terms in the title, post content, excerpt, job category, and job metas (e.g. company name or tagline)
* Go to the frontend and perform various searches. 
* Try using minus sign and double quotes, especially for meta fields and job categories.
* Try using category search and term search from the term plugin together with search terms.
* Observe that:
	* Each search term is included to at least one of the searched columns.
	* When using exclusion terms, observe that these are applied last.

### Release Notes

* Expand job meta and category searching with core WordPress search functionality.

job_listing_search_conditions: Removed as it isn't possible to have it anymore. Users can rollback to previous functionality by overriding `get_job_listings_keyword_search` in `wp-job-manager-functions.php`

### Next Steps
Regarding further improvements in job listing search in general I think that we should avoid them as I think that the implementation is already complex. Any future steps should instead focus on integrating with well known plugins that expand core WP search.

----

| Plugin build for 07706c23923a9764b04c9cad2c9ba2f244912f10 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/05/wp-job-manager-zip-2821-07706c23.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/05/2821-07706c23)             |

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -351,6 +351,7 @@ if ( ! function_exists( 'get_job_listings_keyword_search' ) ) :
 
 				$not_string = $is_excluding ? 'NOT ' : '';
 				$conditions = [];
+				$meta_value = $wildcard_search . $wpdb->esc_like( $search_term ) . $wildcard_search;
 
 				/**
 				 * Can be used to disable searching post meta for job searches.
@@ -359,7 +360,6 @@ if ( ! function_exists( 'get_job_listings_keyword_search' ) ) :
 				 */
 				if ( apply_filters( 'job_listing_search_post_meta', true ) ) {
 
-					$meta_value = $wildcard_search . $wpdb->esc_like( $search_term ) . $wildcard_search;
 					// Only selected meta keys.
 					if ( $searchable_meta_keys ) {
 						$meta_keys = implode( "','", array_map( 'esc_sql', $searchable_meta_keys ) );
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -196,7 +196,7 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 
 		if ( ! empty( $job_manager_keyword ) && strlen( $job_manager_keyword ) >= apply_filters( 'job_manager_get_listings_keyword_length_threshold', 2 ) ) {
 			$query_args['s'] = $job_manager_keyword;
-			add_filter( 'posts_search', 'get_job_listings_keyword_search' );
+			add_filter( 'posts_search', 'get_job_listings_keyword_search', 10, 2 );
 		}
 
 		$query_args = apply_filters( 'job_manager_get_listings', $query_args, $args );
@@ -266,7 +266,7 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 
 		do_action( 'after_get_job_listings', $query_args, $args );
 
-		remove_filter( 'posts_search', 'get_job_listings_keyword_search' );
+		remove_filter( 'posts_search', 'get_job_listings_keyword_search', 10 );
 
 		return $result;
 	}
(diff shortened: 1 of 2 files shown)


wp-job-manager-functions.php
```diff
@@ -303,66 +303,173 @@ function _wpjm_shuffle_featured_post_results_helper( $a, $b ) {
...
+						// No meta keys defined, search all post meta value.
+						$conditions[] = $wpdb->prepare( "{$wpdb->posts}.ID {$not_string}IN ( SELECT post_id FROM {$wpdb->postmeta} WHERE meta_value LIKE %s )", $meta_value );
+						//phpcs:enabled WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+					}
+				}
 
-		// Search Post Meta.
-		if ( apply_filters( 'job_listing_search_post_meta', true ) ) {
+				// Search taxonomy.
+				$conditions[] = $wpdb->prepare( "{$wpdb->posts}.ID ${not_string}IN ( SELECT object_id FROM {$wpdb->term_relationships} AS tr LEFT JOIN {$wpdb->term_taxonomy} AS tt ON tr.term_taxonomy_id = tt.term_taxonomy_id LEFT JOIN {$wpdb->terms} AS t ON tt.term_id = t.term_id WHERE t.name LIKE %s )", $meta_value );
```
yscik, line 376: `$meta_value` is only set inside the if above but also used here.

wp-job-manager-functions.php
```diff
@@ -303,66 +303,173 @@ function _wpjm_shuffle_featured_post_results_helper( $a, $b ) {
...
+				foreach ( $search_columns as $search_column ) {
+					//phpcs:disabled WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- Variables are safe or escaped.
+					$conditions[] = $wpdb->prepare( "( {$wpdb->posts}.$search_column $like_op %s )", $like );
+				}
+
+				$allow_query_attachment_by_filename = apply_filters( 'wp_allow_query_attachment_by_filename', false );
+				if ( ! empty( $allow_query_attachment_by_filename ) ) {
+					$conditions[] = $wpdb->prepare( "(sq1.meta_value $like_op %s)", $like );
+					//phpcs:enabled WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+				}
```
yscik: What's this section for? Is the `sq1` table alias generated by WordPress?

wp-job-manager-functions.php
```diff
@@ -303,66 +303,173 @@ function _wpjm_shuffle_featured_post_results_helper( $a, $b ) {
...
-		if ( empty( $conditions ) ) {
-			return $search;
+		$default_search_columns = [ 'post_title', 'post_excerpt', 'post_content' ];
+		$search_columns         = ! empty( $wp_query->query_vars['search_columns'] ) ? $wp_query->query_vars['search_columns'] : $default_search_columns;
+		if ( ! is_array( $search_columns ) ) {
+			$search_columns = [ $search_columns ];
 		}
 
-		$conditions_str = implode( ' OR ', $conditions );
+		$search_columns = (array) apply_filters( 'post_search_columns', $search_columns, $wp_query->query_vars['s'], $wp_query );
```
yscik, line 436: Should this be a WPJM-specific filter, to be able to target job search?

wp-job-manager-functions.php
```diff
@@ -303,66 +303,173 @@ function _wpjm_shuffle_featured_post_results_helper( $a, $b ) {
...
+
+		// Use only supported search columns.
+		$search_columns = array_intersect( $search_columns, $default_search_columns );
+		if ( empty( $search_columns ) ) {
+			$search_columns = $default_search_columns;
+		}
+
+		// Search terms starting with the exclusion prefix should be removed from the job search results.
+		$exclusion_prefix = apply_filters( 'wp_query_search_exclusion_prefix', '-' );
+		$wildcard_search  = ! empty( $wp_query->query_vars['exact'] ) ? '' : '%';
```
yscik, line 446: Just so that I understand correctly, parsing the search query (like quoted exact phrases and breaking it up to terms) is handled by WP_Query, right?

One thing not working is `javascript` still matches to `java` and even `"java"`. I think we'd need to switch to REGEXP search, and use word boundaries for that to work?

wp-job-manager-functions.php
```diff
@@ -303,66 +303,173 @@ function _wpjm_shuffle_featured_post_results_helper( $a, $b ) {
...
+				} else {
+					$like_op = 'LIKE';
+				}
+
+				$like = $wildcard_search . $wpdb->esc_like( $search_term ) . $wildcard_search;
+
+				$conditions = [];
+				foreach ( $search_columns as $search_column ) {
+					//phpcs:disabled WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- Variables are safe or escaped.
+					$conditions[] = $wpdb->prepare( "( {$wpdb->posts}.$search_column $like_op %s )", $like );
```
yscik, line 409: Note that search_column is not escaped anywhere. I see it is limited to the fixed set of post_title, post_excerpt and post_content, so it is safe right now, but a bit risky that we'd miss this usage if we were to change that limitation.

wp-job-manager-functions.php
```diff
@@ -303,66 +303,173 @@ function _wpjm_shuffle_featured_post_results_helper( $a, $b ) {
...
-		 * @param array  $conditions          Conditions to join by OR when querying job listings.
-		 * @param string $job_manager_keyword Search query.
+		 * This function aims to provide similar search functionality with WP core while also including meta and taxonomy terms
+		 * in the searched columns. The functionality of WP_Query::parse_search is replicated but with additional SQL
+		 * clauses which are generated in the job_manager_construct_secondary_conditions function.
 		 */
-		$conditions = apply_filters( 'job_listing_search_conditions', $conditions, $job_manager_keyword );
-		if ( empty( $conditions ) ) {
-			return $search;
+		$default_search_columns = [ 'post_title', 'post_excerpt', 'post_content' ];
```
yscik, line 429: One more big improvement to the search would be to prioritize matches in the title — and generally to sort results by relevance. Maybe we could do multiple searches, first just in the post title, then in content, metas? It could actually be faster for the first pages in some cases, since if there are enough results in the title, we don't need to do the content/meta searches. 

Just bringing it up to discuss here, it'd be its own PR probably as there would be some extra complexity, like handling pagination. Could also do some [weighting](https://stackoverflow.com/questions/6496866/best-way-to-do-a-weighted-search-over-multiple-fields-in-mysql) maybe?

yscik (approved):
Working great! Added a few questions and noticed one issue, but generally code looks good too.

wp-job-manager-functions.php
gikaragia: Good catch d8c27379ec94e1350d096e1b6ceaad73ec2e5623

wp-job-manager-functions.php
gikaragia: Yes it's a left joing to wp_postmeta for attachments. Added a comment to clarify in b81e6014a4bfc018538e554f7c3df0bfe74688db

wp-job-manager-functions.php
gikaragia: It's coming from the core WP implementation. Generally I replicated what core does for searching as we now overwrite the whole search clause and not adding stuff on top.

This might be a bit risky as we might need to do more changes to replicate future core changes in WP_Query but since most of the stuff in WP_Query are public I think that they won't break backwards compatibility easily.

gikaragia: Added a comment for the filter to specify that it comes from core WP

wp-job-manager-functions.php
gikaragia: Yeap. So WP_Query does this in `parse_search` and stores them in query_vars['search_terms'] which is public so I thought it's ok to use.

Yeah I saw that too and was on the edge about it. I decided that maybe as part of this PR I should just replicate what core does to keep things simple.

gikaragia: Then if we do REGEX we will also do it for title and description and users might complain if we change this behaviour

wp-job-manager-functions.php
gikaragia: Good point better safe than sorry: 07706c23923a9764b04c9cad2c9ba2f244912f10

gikaragia:
In this PR I focused on fixing what seemed like buggy behavior. Although we included some meta fields in searches, it didn't work as expected previously and returned weird results. 

> One more big improvement to the search would be to prioritize matches in the title — and generally to sort results by relevance. Maybe we could do multiple searches, first just in the post title, then in content, metas? It could actually be faster for the first pages in some cases, since if there are enough results in the title, we don't need to do the content/meta searches.
Just bringing it up to discuss here, it'd be its own PR probably as there would be some extra complexity, like handling pagination. Could also do some [weighting](https://stackoverflow.com/questions/6496866/best-way-to-do-a-weighted-search-over-multiple-fields-in-mysql) maybe?

These are great suggestions, there is a lot of room for improvement.  We could look into implementing these in the future. Another option would be to integrate with plugins that expand WP's search capabilities as it has been suggested[ in the past](https://github.com/Automattic/WP-Job-Manager/pull/960#issuecomment-301028364).

PR #2848, opened 2024-08-29
Job Visibility Enhancement: Submit Job Listing

Gnodesign:

I've received numerous requests from users for visibility options on the job submission page. Currently, most users manage this through third-party plugins, but since WPJM already offers visibility options for "Browse Jobs" and "View Job," I thought it would be beneficial to add similar options for "Job Submission" as well.

### Testing Instructions

* Go to "Job Manager -> Settings -> Job Visibility" 
* Add a role to "Submit Job Capability"
* Visit the job submission as a not logged in user

* New filter: `job_manager_user_can_view_submit_job`
* New template: access-denied-submit-job_listing.php

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -624,6 +624,16 @@ class WP_Job_Manager_Settings {
 							'desc'              => sprintf( __( 'Enter which <a href="%s">roles or capabilities</a> allow visitors to view a single job listing. If no value is selected, everyone (including logged out guests) will be able to view job listings.', 'wp-job-manager' ), 'http://codex.wordpress.org/Roles_and_Capabilities' ),
 							'track'             => 'bool',
 						],
+						[
+							'name'              => 'job_manager_submit_job_listing_capability',
+							'std'               => [],
+							'label'             => __( 'Submit Job Capability', 'wp-job-manager' ),
+							'type'              => 'capabilities',
+							'sanitize_callback' => [ $this, 'sanitize_capabilities' ],
+							// translators: Placeholder %s is the url to the WordPress core documentation for capabilities and roles.
+							'desc'              => sprintf( __( 'Enter which <a href="%s">roles or capabilities</a> allow visitors to submit a job listing. If no value is selected, everyone (including logged out guests) will be able to submit job listings.', 'wp-job-manager' ), 'http://codex.wordpress.org/Roles_and_Capabilities' ),
+							'track'             => 'bool',
+						],
 					],
 				],
 			]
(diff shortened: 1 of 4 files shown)


donnchawp:
Thanks @Gnodesign. This is more complicated than #2910, but we'll take a look at this. Appreciate the work you put in here.

donnchawp (changes requested):
Thanks for working on this. The feature itself is a good idea. However, the current approach of restructuring job-submit.php is problematic because it will break any theme that overrides this template (which is common in WPJM sites). There's also a security concern: the check is only in the template, so a direct POST to the form handler would bypass it.

The good news is that WPJM already has the hooks to support this with a much smaller change. No template modifications needed at all.

## How it would work

  1. Keep the new the job_manager_submit_job_listing_capability setting definition in class-wp-job-manager-settings.php.

  2. Hook into job_manager_user_can_post_job instead of modifying the template:
```
  add_filter( 'job_manager_user_can_post_job', function( $can_post ) {
      $caps = get_option( 'job_manager_submit_job_listing_capability' );

      if ( ! empty( $caps ) ) {
          $can_post = false;
          foreach ( $caps as $cap ) {
              if ( current_user_can( $cap ) ) {
                  $can_post = true;
                  break;
              }
          }
      }

      return $can_post;
  } );
```

### This works because the template already has this block:
```
  <?php if ( job_manager_user_can_post_job() || job_manager_user_can_edit_job( $job_id ) ) : ?>
      <!-- form fields -->
  <?php else : ?>
      <?php do_action( 'submit_job_form_disabled' ); ?>
  <?php endif; ?>
```
  When job_manager_user_can_post_job() returns false, the existing template naturally falls through to the submit_job_form_disabled action so no new template or conditional is needed.

  3. Similarly hook into job_manager_user_can_edit_job to prevent users without the capability from editing existing listings:
```
  add_filter( 'job_manager_user_can_edit_job', function( $can_edit, $job_id ) {
      $caps = get_option( 'job_manager_submit_job_listing_capability' );

      if ( ! empty( $caps ) ) {
          $has_cap = false;
          foreach ( $caps as $cap ) {
              if ( current_user_can( $cap ) ) {
                  $has_cap = true;
                  break;
              }
          }
          if ( ! $has_cap ) {
              $can_edit = false;
          }
      }

      return $can_edit;
  }, 10, 2 );
```

### What this gives you

  - No template changes. Zero risk of breaking theme overrides.
  - Server-side enforcement for free. job_manager_user_can_post_job() is called during form processing, not just rendering, so direct POST submissions are blocked too.
  - The submit_job_form_disabled action fires naturally. Site owners and plugins can already hook into this to display a custom message, so no new access-denied-submit-job_listing.php template is needed.

## What to remove

  - The changes to templates/job-submit.php
  - The new templates/access-denied-submit-job_listing.php
  - The new job_manager_user_can_view_submit_job_listing() function in wp-job-manager-functions.php
  - The unrelated docstring fixes (the "resume" → "job listing" corrections are valid but should be a separate PR)

  The net change would be just the settings definition + the two filter callbacks. Much smaller surface area.

PR #2891, opened 2025-07-23
Fix hardcoded dashboard expiration date format

### LINKED ISSUE #2890: Dashboard expiration date format
**Describe the bug**
The expiration date format in the **[job dashboard]** is always in `'M d, Y'` format despite the global WordPress date format setting. The problem lies in the date format being hardcoded:
https://github.com/Automattic/WP-Job-Manager/blob/41d5048c9cf6c4d2636c878419added8979fcdc6/includes/class-job-dashboard-shortcode.php#L603-L605

**Expected behavior**
Change date format based on global WordPress date format setting.

masteradhoc:
Fixes #2890

* We should change the hardcoded dateformat for the Job Dashboard Shortcode. Currently it displays the date format hardcoded in the format 'M d, Y'. Now it will respect the settings set in WordPress.

### Testing Instructions
* open /job-dashboard/ page. without the fix it will be displayed like this
image

After the fix it will be displayed according to your settings, in my example "j. F Y".
image

### Release Notes
* fix hardcoded dashboard expiration date format

* -

### Deprecated Code
* -

### Screenshot / Video
-

### DIFF
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -594,14 +594,15 @@ class Job_Dashboard_Shortcode {
 	}
 
 	/**
-	 * Show job title.
+	 * Show the job date
 	 *
 	 * @param \WP_Post $job
 	 *
 	 * @output string
 	 */
 	public static function the_date( $job ) {
-		echo '<div>' . esc_html( wp_date( apply_filters( 'job_manager_get_dashboard_date_format', get_option( 'date_format' ) ), get_post_datetime( $job )->getTimestamp() ) ) . '</div>';
+		$date_format = get_option( 'date_format' ) ?: 'M d, Y';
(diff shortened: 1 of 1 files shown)


azaozz (changes requested):
I'm pretty sure it is (almost) impossible to not have the `date_format` option set in WP, but just in case that happens somewhere this `get_option()` call will result in a trip to the DB (slow) and will return `false` which seems undesirable.

How about checking the option value before using it? Perhaps add something like:

`$date_format = get_option( 'date_format' ) ?: 'M d, Y';`

at the top then use `$date_format` in the HTML.

masteradhoc:
@azaozz definately a valid point! adjusted the code as per your suggestion and fixed the docblock as the title was also wrong there. hope thats fine inside the same PR.

azaozz (approved):
Looks good, thanks!

PR #2892, opened 2025-07-23
loading do_feed_rss2 function with do_action

### LINKED ISSUE #2854: Loading do_feed_rss2 function with do_action
**Describe the bug**
The job feed https://www.example.com/?feed=job_feed created by this plug uses the default Wordpress `do_feed_rss2`. 
We want to add a additional action to the do_feed_rss2 action, but this is not possible, because in `wp-job-manager/includes/class-wp-job-manager-post-types.php` the function `do_feed_rss2(false)` is called directly.

**To Reproduce**
1. Add a do_action behavior to do_feed_rss2
2. Call the /?feed=job_feed page and you will notice your custom action is not activated.

**Expected behavior**
In "`wp-job-manager/includes/class-wp-job-manager-post-types.php` the `do_feed_rss2(false)` should be replaced with `do_action('do_feed_rss2');`.
This way customers can add extra actions to the feed.

masteradhoc:
Fixes #2854

* add a additional action to the do_feed_rss2 action so users can add extra actions to the feed

### Testing Instructions
* Add a do_action behavior to do_feed_rss2
* Call the /?feed=job_feed page and you will notice your custom action is not activated.

### Release Notes
* add a additional action to the do_feed_rss2

-

### Deprecated Code
-

### Screenshot / Video
-

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -807,7 +807,7 @@ class WP_Job_Manager_Post_Types {
 		query_posts( apply_filters( 'job_feed_args', $query_args ) );
 		add_action( 'rss2_ns', [ $this, 'job_feed_namespace' ] );
 		add_action( 'rss2_item', [ $this, 'job_feed_item' ] );
-		do_feed_rss2( false );
+		do_action( 'do_feed_rss2', false );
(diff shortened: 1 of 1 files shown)


azaozz:
Looks good. Using the `'do_feed_rss2'` hook instead of calling the function directly is actually preferable in most cases for better interoperability between plugins and core.

PR #2893, opened 2025-07-23
reCaptcha script not being loaded

### LINKED ISSUE #2850: Recaptcha v3 not working. Other recaptcha plugins work with the same keys
**Describe the bug**
A clear and concise description of what the bug is. Please be as descriptive as possible; issues lacking detail, or for any other reason than to report a bug, may be closed without action.

**To Reproduce**
I did create a v3 key and enabled v3 on the Job Manager configuration screen. I also enabled recaptcha in the Applications screen.

I can see that code is added in the inspector:

**Screenshots**
image

However:
- There is NO badge visible (other plugins like WP Captcha show a badge and function properly) Note: This seems logical as the backend PHP code does the verification if I look into the software.
- There is NO activity in the Google v3 console
- Bots are able to submit a form about every second!

**Expected behavior**
- Recaptcha works

- [partially. can't disable all plugins without database issues] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [x] I can reproduce this bug consistently using the steps above. The captcha never works!

**WordPress Environment**
- WordPress Version : v6.6.1
- WP Job Manager Version: v2.4.0 / WP Job Manager Applications v3.2.0
- PHP Version : 8.x
- Other important details:

Gnodesign:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2850

The reCAPTCHA scripts aren't being loaded on any themes other than the default ones.

### Testing Instructions

* Setup reCaptcha
* Switch to a non default theme
* Go to the job submission

### DIFF
--- a/includes/class-wp-job-manager-recaptcha.php
+++ b/includes/class-wp-job-manager-recaptcha.php
@@ -77,7 +77,11 @@ class WP_Job_Manager_Recaptcha {
 				add_filter( $validate_hook, [ $this, 'validate_recaptcha_field' ] );
 			}
 
-			add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
+			if ( did_action( 'wp_enqueue_scripts' ) ) {
+				$this->enqueue_scripts();
+			} else {
+				add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
+			}
 		}
 	}
 


azaozz:
This change makes sense: the callback won't be executed when attempting to use a hook (an action) after it has been run. As far as I see the `maybe_enable_recaptcha()` function seems to run in the middle of the page, long after the WP's `init` action.

May be missing something but looking at [all the comments](https://github.com/Automattic/WP-Job-Manager/issues/2850) suggests that [reCAPTCHA](https://developers.google.com/recaptcha/docs/versions) v2 works but v3 fails. I'm not very familiar with how exactly the different versions of reCAPTCHA work, but as far as I see both require some js? Or perhaps the js for v2 was added in a different way?

In any case this change shouldn't break anything and is good to go.

Gnodesign:
@azaozz the issue comes from the main reCaptcha script not being loaded correctly. With the provided fix, both versions work as expected.

azaozz:
> issue comes from the main reCaptcha script not being loaded correctly

Yep, I understand. What I was wondering about is why v2 works nevertheless. 

Seems this bug was introduced in [Commit a6e0757](https://github.com/Automattic/WP-Job-Manager/commit/a6e07575ac72807a50875aa534149aec263ddc69) most likely by this change on L:139: https://github.com/Automattic/WP-Job-Manager/commit/a6e07575ac72807a50875aa534149aec263ddc69#diff-69b5a8dc728e5b801b8b59d2cf75e0470846f28cab3f1698413c5bcfb02c3fe6L139. Not understanding and fixing the underlying timing issue might eventually cause problems in the future when other scripts have to be enqueued, etc.

In any case, this patch seems okay for now.

PR #2895, opened 2025-08-10
Enable Expiry Date in Frontend Form

### LINKED ISSUE #2886: Enable Expiry Date in Frontend Form
The default solution for job expiry is in the settings area while there is no field available for employers posting on the frontend form who may wish to post a listing for less than the default setting.

Provide a default field to enable this to be published by employers using the frontend form that will override the defaults settings.

masteradhoc:
Fixes #2886 

* Allow Expiry Date to be set in Frontend Form for Job Submissions

### Testing Instructions
* Go to Settings -> Job Submissions and activate "Expiry Date field"
* Visit Job Submissions Frontend Form
* add a new job including an expiry date
* see in wp-admin if the expiry date was set correctly

### Release Notes
* Allow Expiry Date to be set in Frontend Form for Job Submissions

*

### Deprecated Code
*

### Screenshot / Video
image

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -591,7 +591,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 
 		// Validate expiry date field if enabled and provided.
 		if ( get_option( 'job_manager_enable_expiry_date_field' ) && ! empty( $values['job']['job_expires'] ) ) {
-			$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d', $values['job']['job_expires'], wp_timezone() );
+			$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d|', $values['job']['job_expires'], wp_timezone() );
 			if ( ! $expires_date ) {
 				throw new Exception( __( 'Please enter a valid expiry date', 'wp-job-manager' ) );
 			}
@@ -1046,7 +1046,7 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 
 					// If a date is provided, also set the job expiration.
 					if ( ! empty( $values[ $group_key ][ $key ] ) ) {
-						$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d', $values[ $group_key ][ $key ], wp_timezone() );
+						$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d|', $values[ $group_key ][ $key ], wp_timezone() );
 						if ( $expires_date ) {
 							WP_Job_Manager_Post_Types::instance()->set_job_expiration( $this->job_id, $expires_date );
 						}
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -409,10 +409,6 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 				'required'    => false,
 				'placeholder' => '',
 				'priority'    => '6.6',
-				'class'       => 'job-manager-datepicker',
-				'attributes'  => [
-					'min' => current_time( 'Y-m-d' ),
-				],
 			];
 		}
 
@@ -591,14 +587,20 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		// Validate expiry date field if enabled and provided.
 		if ( get_option( 'job_manager_enable_expiry_date_field' ) && ! empty( $values['job']['job_expires'] ) ) {
 			$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d|', $values['job']['job_expires'], wp_timezone() );
-			if ( ! $expires_date ) {
+			$date_errors  = DateTimeImmutable::getLastErrors();
+			if ( ! $expires_date || ( is_array( $date_errors ) && $date_errors['warning_count'] > 0 ) ) {
 				throw new Exception( __( 'Please enter a valid expiry date', 'wp-job-manager' ) );
 			}
 
 			// Check if the date is in the past.
+			// Skip the check when editing an expired listing with an unchanged date — otherwise
+			// owners would be locked out of any unrelated edits on their expired listings.
 			$today = new DateTimeImmutable( 'today', wp_timezone() );
 			if ( $expires_date < $today ) {
-				throw new Exception( __( 'Expiry date cannot be in the past', 'wp-job-manager' ) );
+				$existing_expires = $this->job_id ? get_post_meta( $this->job_id, '_job_expires', true ) : '';
+				if ( $values['job']['job_expires'] !== $existing_expires ) {
+					throw new Exception( __( 'Expiry date cannot be in the past', 'wp-job-manager' ) );
+				}
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


masteradhoc:
@donnchawp could you check this PR as well?

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -524,6 +539,19 @@ protected function validate_fields( $values ) {
 			}
 		}
 
+		// Validate expiry date field.
```
donnchawp: This block is almost the same as the one at line 592. (Validate expiry date field if enabled and provided.)

It's probably best to just keep the last one that throws an exception as other checks do.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -524,6 +539,19 @@ protected function validate_fields( $values ) {
...
+		if ( get_option( 'job_manager_enable_expiry_date_field' ) && ! empty( $values['job']['job_expires'] ) ) {
+			$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d', $values['job']['job_expires'], wp_timezone() );
+			if ( ! $expires_date ) {
+				return new WP_Error( 'validation-error', __( 'Please enter a valid expiry date.', 'wp-job-manager' ) );
+			}
+
+			$today = new DateTimeImmutable( 'now', wp_timezone() );
+			if ( $expires_date < $today ) {
```
donnchawp: There's a good chance this would reject any jobs submitted, with an expiry date of the current date, but if you don't use this block it doesn't matter.

The date/time returned from DateTimeImmutable::createFromFormat on line 544 will fill in the current time, including microseconds.
That may be a couple of microseconds before $today so the check on line 550 will fail.

The second block works correctly, where you use 'today' as the time value on line 600.

ref: https://www.php.net/manual/en/datetimeimmutable.createfromformat.php

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -561,6 +589,20 @@ protected function validate_fields( $values ) {
 			}
 		}
 
+		// Validate expiry date field if enabled and provided.
+		if ( get_option( 'job_manager_enable_expiry_date_field' ) && ! empty( $values['job']['job_expires'] ) ) {
+			$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d', $values['job']['job_expires'], wp_timezone() );
```
donnchawp: Probably should use 'Y-m-d|' here so it returns the date and not a date and time.

Jobs submitted with an expiry date today will have $expires_date set to midnight, same as $today further down, so the check will pass.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -997,6 +1039,19 @@ protected function update_job_data( $values ) {
...
 					update_user_meta( get_current_user_id(), '_company_logo', $attachment_id );
 
+					// Handle custom expiry date field.
+				} elseif ( 'job_expires' === $key ) {
+					// Always save the raw value first.
+					update_post_meta( $this->job_id, '_job_expires', $values[ $group_key ][ $key ] );
+
+					// If a date is provided, also set the job expiration.
+					if ( ! empty( $values[ $group_key ][ $key ] ) ) {
+						$expires_date = DateTimeImmutable::createFromFormat( 'Y-m-d', $values[ $group_key ][ $key ], wp_timezone() );
```
donnchawp: Change 'Y-m-d' to 'Y-m-d|' so it's definitely a date.

includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -997,6 +1039,19 @@ protected function update_job_data( $values ) {
 					}
 					update_user_meta( get_current_user_id(), '_company_logo', $attachment_id );
 
+					// Handle custom expiry date field.
+				} elseif ( 'job_expires' === $key ) {
+					// Always save the raw value first.
+					update_post_meta( $this->job_id, '_job_expires', $values[ $group_key ][ $key ] );
```
donnchawp: Why save _job_expires when it's set again in 1051? For debugging purposes?

donnchawp (changes requested):
Thanks for your PR. Some changes here.

includes/forms/class-wp-job-manager-form-submit-job.php
masteradhoc: True thats an old leftover i used at the beginning before optimizing and moving it down

donnchawp (changes requested):
I used 3 AIs to review this PR and they found issues. I was able to reproduce the "preview->submit" deletion of the expiry date too.

Multi-AI adversarial review (Claude + Gemini + Codex). Three of these findings have multi-reviewer consensus; two HIGH-severity findings were each surfaced solo and then independently confirmed via direct code reading.

   ## Summary
   - **6 distinct findings** — 1 consensus (3/3), 2 partial-consensus (2/3), 2 solo HIGH, 1 LOW
   - Severity mix: **2 HIGH**, 2 MEDIUM, 2 LOW
   - **Verdict: BLOCK — the feature is broken end-to-end on the standard new-submission flow**

   The single most important issue: a user's chosen expiry date is silently discarded between preview and publish, so the feature never actually works for new submissions.

   ## Findings

   ### [HIGH] `includes/forms/class-wp-job-manager-form-submit-job.php:1158` — User-chosen expiry is wiped between preview and publish
   **Status:** Solo (Codex, confirmed by Claude via direct code read)

   `update_job_data()` saves the user's date via `set_job_expiration()` while the post is in `preview` status. Then `preview_handler()` runs on "Continue" and unconditionally calls `delete_post_meta( $job->ID, '_job_expires' )`
   (line 1158) before transitioning the post to `publish`/`pending`. The `set_expiry()` hook on the transition reads `$_POST['_job_expires']`, which is **never set** by this form (it submits `job[job_expires]`), so it falls into
   the `! $expiration_date` branch and writes the default duration. **Net result: the user's date never survives a normal submission.**

   **Fix:** In `preview_handler`, only delete `_job_expires` when the new field is disabled or empty; or copy the saved value into `$_POST['_job_expires']` before `wp_update_post()` so `set_expiry()` honors it (mirrors how the
   admin write panel feeds the same key).

   ### [HIGH] `includes/forms/class-wp-job-manager-form-submit-job.php:586` — Past-date validator traps owners of expired listings
   **Status:** Solo (Gemini, confirmed by Claude via direct code read)

   `WP_Job_Manager_Form_Edit_Job::populate_fields()` (line 126) pre-fills every field's `value` from `get_post_meta($id, '_' . $key, true)` — i.e., the existing `_job_expires`. For an `expired` listing this is a past date. When the
    owner edits any unrelated field and saves, `$_POST['job']['job_expires']` carries the unchanged past date, and `validate_fields()` throws *"Expiry date cannot be in the past"*. The user is locked out of editing expired listings
    until they also pick a new future date — even for typo fixes.

   **Fix:** Skip the past-date check when the submitted value is unchanged from the stored meta, or when the listing is already expired.

   ### [MEDIUM] `includes/forms/class-wp-job-manager-form-submit-job.php:1029-1037` — Clearing the field on edit doesn't reset to default
   **Status:** Consensus (Claude + Codex + Gemini, 3/3)

   The new `elseif ( 'job_expires' === $key )` branch only writes when the value is non-empty. On the inherited frontend edit form, clearing the date submits empty → no write → existing `_job_expires` persists. The field's own
   description says *"Leave blank to use default duration."* It doesn't.

   **Fix:** When `empty( $values[$group_key][$key] )`, call `set_job_expiration( $this->job_id, null )` so the default is recomputed on next transition.

   ### [MEDIUM] `includes/forms/class-wp-job-manager-form-submit-job.php:579-588` — Date-component overflow not validated
   **Status:** Consensus (Claude + Codex)

   Codex confirmed via REPL: `DateTimeImmutable::createFromFormat('Y-m-d|', '2026-02-31', wp_timezone())` returns `2026-03-03` (warning_count=1, error_count=0). The `! $expires_date` check passes, and an impossible date silently
   becomes a different valid date. The HTML5/datepicker UI normally prevents this, but a JS-disabled client or direct POST bypasses both. Server validation should not depend on the UI.

   **Fix:** After `createFromFormat`, also reject if `DateTimeImmutable::getLastErrors()['warning_count'] > 0`, or compare `$expires_date->format('Y-m-d') === $values['job']['job_expires']`.

   ### [LOW] `includes/forms/class-wp-job-manager-form-submit-job.php:399-401` — `attributes['min']` and `class` keys are dead code
   **Status:** Consensus (Claude + Gemini)

   `templates/form-fields/date-field.php` does not iterate `$field['attributes']` and hardcodes the class — so `min => current_time('Y-m-d')` and `class => 'job-manager-datepicker'` are silently dropped. `type=>'date'` reads as
   native HTML5 to a casual reviewer, but the template actually renders `<input type="text">` plus a jQuery UI datepicker. Min-date enforcement comes from `datepicker.js` using *browser-local* "today", not server time.

   **Fix:** Remove the `attributes` and `class` keys from the field definition (or extend the template to honor them).

   ### [LOW] Notes
   - `type => 'date'` is misleading; ensure the feature description doesn't imply native HTML5 date-picker UX.
   - No automated test coverage was added. Given two HIGH bugs slipped through, an integration test for `submit-with-date → preview → publish → assert _job_expires` is strongly recommended.

   ---

   ## Highest-priority single action
   Fix the preview→publish regression first — without it, the feature ships broken regardless of every other issue.

PR #2925, opened 2026-03-15
add featured support to RSS

### LINKED ISSUE #2341: RSS Feed Filter by Featured Jobs
### Problem: I cannot use RSS Feed to promote Featured Jobs
I am using the WP Jomanager RSS feed to send out mails to subscribers via Mailchimp. I would like to add a section in that mail for "Featured Jobs" to give them better visibility to the subscribers. 

Also I am using the RSS Feed to have a nice integration in the site bar of my Wordpress via a Plugin. Also there I would like to have the possibility to only show Featured Jobs.

The RSS Feed cannot be filtered by "Featured Job" at the moment. 

### Solution: Add Filter for Featured Job to RSS Feed URL
I would like to have the possibility to filter the RSS feed by adding a parameter "featured=true" to only show featured jobs 
like it is possible with search_location=mylocation, search_keywords=mykeyword, job_types=mytype for example as described here https://wpjobmanager.com/document/rss-feed-for-jobs/.

masteradhoc:
Fixes #2341

* Added support for filtering the job RSS feed by featured status using the `featured=true` query parameter

### Testing Instructions

1. Mark one or more job listings as "Featured" in WP Admin.
2. Visit `/?feed=job_feed` — confirm all published jobs appear.
3. Visit `/?feed=job_feed&featured=true` — confirm only featured jobs appear.
4. Combine with other filters, e.g. `/?feed=job_feed&featured=true&search_location=London` — confirm both filters are applied correctly.

### Release Notes

* Added `featured=true` query parameter support to the job RSS feed to allow filtering by featured listings only.

-

### Deprecated Code
-

### Screenshot / Video
-

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -728,6 +728,7 @@ class WP_Job_Manager_Post_Types {
 		}
 
 		$job_manager_keyword = isset( $_GET['search_keywords'] ) ? sanitize_text_field( wp_unslash( $_GET['search_keywords'] ) ) : '';
+		$input_featured      = isset( $_GET['featured'] ) && 'true' === sanitize_text_field( wp_unslash( $_GET['featured'] ) );
 		// phpcs:enable WordPress.Security.NonceVerification.Recommended
 
 		$query_args = [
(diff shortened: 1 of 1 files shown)


donnchawp (changes requested):
Looks good, but what about featured=false? Get a feed of "non featured" jobs too.

donnchawp:
Thanks. That's perfect!

PR #2926, opened 2026-03-15
add categories to feed

### LINKED ISSUE #1910: Add categories to job feed
Can we add job categories to the job feed here?

https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-post-types.php#L657

Requested in 2425494-zen

masteradhoc:
Fixes #1910

* Added `job_category` to the job RSS feed XML output in `job_feed_item()`, consistent with how `job_type` is already included.
* Job categories are comma-separated when multiple categories are assigned to a listing.

### Testing Instructions

1. Ensure at least one job listing has a category assigned.
2. Visit the job RSS feed URL (e.g. `/?feed=job_feed`).
3. Confirm each feed item that has categories now includes a `<job_listing:job_category>` element, e.g.:
   ```xml
   <job_listing:job_category><![CDATA[Engineering, Design]]></job_listing:job_category>
   ```
4. Confirm feed items without a category omit the element entirely.
5. Confirm existing feed elements (`location`, `job_type`, `company`, `salary`) are unaffected.

### Release Notes

* Job categories are now included in the job RSS feed XML output.

*

### Deprecated Code

*

### Screenshot / Video

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -847,11 +847,12 @@ class WP_Job_Manager_Post_Types {
 	 * Adds custom data to the job feed.
 	 */
 	public function job_feed_item() {
-		$post_id   = get_the_ID();
-		$location  = get_the_job_location( $post_id );
-		$company   = get_the_company_name( $post_id );
-		$job_types = wpjm_get_the_job_types( $post_id );
-		$salary    = get_the_job_salary( $post_id );
+		$post_id        = get_the_ID();
+		$location       = get_the_job_location( $post_id );
+		$company        = get_the_company_name( $post_id );
+		$job_types      = wpjm_get_the_job_types( $post_id );
+		$job_categories = wpjm_get_the_job_categories( $post_id );
+		$salary         = get_the_job_salary( $post_id );
 
 		if ( $location ) {
 			echo '<job_listing:location><![CDATA[' . esc_html( $location ) . "]]></job_listing:location>\n";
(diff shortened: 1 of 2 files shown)


azaozz:
Imho this looks good.

One more general question: is `esc_html()` for the location, job type, job categories, etc. adequate enough? Seems we don't expect any HTML in there, should we be stripping it instead or suspecting something improper and not outputting anything? 

Generally escaped HTML is helpful to admins to detect when a contributor has entered something that doesn't belong somewhere. This is only in WP admin, not on the front end where escaped HTML will look quite weird.

masteradhoc:
Thanks for the feedback @azaozz! The intention was to keep it consistent with the existing codebase, which already uses esc_html for location and job types. I'm happy to make that change across the board though if you want!

donnchawp (changes requested):
Thanks for this! The change follows the existing feed pattern well, and it looks good.

There's one thing needed: the existing [test_job_feed_item](https://github.com/masteradhoc/wp-job-manager/blob/e574cf9090147beb9c292575c2af716f0f2a85ca/tests/php/tests/includes/test_class.wp-job-manager-post-types.php#L281) test passes, but only because none of the test fixtures have categories assigned so the new code path is never actually exercised.
Could you add category coverage to the test? Specifically:

1. Create a TAX_LISTING_CATEGORY term (similar to how type_a/type_b are created for TAX_LISTING_TYPE).
2. Assign it to at least one of the test jobs via tax_input.
3. Assert that job_listing:job_category appears in the output for jobs with categories and is absent for those without.

The existing job_type assertions in the test are a good template to follow.

donnchawp:
Thanks! Looks great!

PR #2927, opened 2026-03-15
Fix: Exclude filled job listings from REST API collection responses

### LINKED ISSUE #1855: Filled jobs still visible in Rest-Api
When marking a job as Filled, it is still available in the rest api endpoint of jobs listing. 
When jobs are marked as fill, they shouldn't be available in rest api endpoint.

masteradhoc:
Fixes #1855

* Filled job listings are now excluded from the job listings collection endpoint (`GET /wp-json/wp/v2/job-listings`) via the `rest_job_listing_query` filter.

### Testing Instructions

1. Mark one or more job listings as **Position Filled** in the WP admin.
2. Make a request to `GET /wp-json/wp/v2/job-listings` and confirm filled listings are absent from the results.
4. Confirm non-filled listings are still returned as expected in both cases.

### Release Notes
* Filled job listings are no longer exposed via the REST API.

-

### Deprecated Code
-

### Screenshot / Video
-

### DIFF
--- a/includes/class-wp-job-manager-rest-api.php
+++ b/includes/class-wp-job-manager-rest-api.php
@@ -33,6 +33,10 @@ class WP_Job_Manager_REST_API {
 	 * @return array
 	 */
 	public static function exclude_filled_from_query( $args, $request ) {
+		if ( 1 !== absint( get_option( 'job_manager_hide_filled_positions' ) ) ) {
+			return $args;
+		}
+
 		if ( ! isset( $args['meta_query'] ) ) {
 			$args['meta_query'] = []; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- Empty.
 		}
--- a/includes/class-wp-job-manager-rest-api.php
+++ b/includes/class-wp-job-manager-rest-api.php
@@ -22,6 +22,35 @@ class WP_Job_Manager_REST_API {
 	 */
 	public static function init() {
 		add_filter( 'rest_prepare_job_listing', [ __CLASS__, 'prepare_job_listing' ], 10, 2 );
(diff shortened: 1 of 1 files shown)


donnchawp (changes requested):
Thanks for this. There's a real bug here. The `job_manager_hide_filled_positions` setting is respected in `get_job_listings()` and the feed query, but not in the REST API, so filled positions always show up via REST regardless of the setting.

However, the current implementation unconditionally excludes filled listings, which changes the behaviour in the other direction. Sites with `job_manager_hide_filled_positions` disabled would lose access to filled listings via the API. This could break anything using the REST API for admin dashboards, reporting, etc.

The fix should be gated behind the existing setting to stay consistent:
```
  public static function exclude_filled_from_query( $args, $request ) {
      if ( 1 !== absint( get_option( 'job_manager_hide_filled_positions' ) ) ) {
          return $args;
      }

      if ( ! isset( $args['meta_query'] ) ) {
          $args['meta_query'] = [];
      }

      $args['meta_query'][] = [
          'relation' => 'OR',
          [
              'key'     => '_filled',
              'value'   => '1',
              'compare' => '!=',
          ],
          [
              'key'     => '_filled',
              'compare' => 'NOT EXISTS',
          ],
      ];

      return $args;
  }
```
The OR / NOT EXISTS approach you've used is actually more robust than the simpler `!= pattern` used elsewhere in the codebase (e.g. get_job_listings()), since it correctly handles listings where _filled meta doesn't exist at all. Nice catch there.

masteradhoc:
Thank you for the feedback @donnchawp, updated the PR based on your valid suggestion!

donnchawp (approved):
Thanks. That's great! 

Filled jobs still show when you request them directly via the REST API but this PR will stop a general job-listings from showing it.

PR #2928, opened 2026-03-15
Fix: Salary currency field placeholder and description reflect the configured default currency

### LINKED ISSUE #2300: Improve default currency display

When you set a default currency value in the backend, the frontend UI when submitting a job is not updated. For example, if the currency is set to "RUB" the placeholder still displays "e.g. USD" and the note on the field is unclear.

The form label displays the text:

> Add a salary currency, this field is optional. Leave it empty to use the default salary currency.

It should probably state what the default unit is if one is set. 

### Additional context

Screenshot of Job Submission:

Screenshot on 2022-07-22 at 15-06-06

Job Display:

Screenshot on 2022-07-22 at 15-05-44

masteradhoc:
Fixes #2300

* The salary currency field on the frontend job submission form now uses the configured **Default Salary Currency** setting as its placeholder instead of the hardcoded `e.g. USD`.
* The field description now also displays the configured default currency code, making it clear to submitters what currency will be applied if the field is left empty.
* When no default currency is configured, the placeholder falls back to `e.g. USD` and the description remains generic.

### Testing Instructions

1. Go to **Job Listings → Settings** and set the **Default Salary Currency** to a non-USD value (e.g. `EUR` or `CHF`). Ensure **Job Salary Unit Customization** is enabled.
2. Visit the frontend job submission form.
3. Confirm the **Salary Currency** field placeholder shows the configured currency (e.g. `EUR`).
4. Confirm the field description reads: *"…Leave it empty to use the default salary currency (EUR)."*
5. Clear the **Default Salary Currency** setting.
6. Confirm the placeholder falls back to `e.g. USD` and the description no longer mentions a specific currency.

### Release Notes

* The salary currency field on the job submission form now correctly reflects the configured default currency in its placeholder and helper text.

-

### Deprecated Code
-

### Screenshot / Video

Currency set in settings:
image
image

Currency not set in settings:
image
image

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -218,6 +218,14 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		} else {
 			$job_type = 'term-select';
 		}
+
+		$default_salary_currency = get_option( 'job_manager_default_salary_currency' );
+		$salary_currency_placeholder = $default_salary_currency ?: __( 'e.g. USD', 'wp-job-manager' );
+		$salary_currency_description = $default_salary_currency
+			// translators: %s is the default salary currency code (e.g. USD).
+			? sprintf( __( 'Add a salary currency, this field is optional. Leave it empty to use the default salary currency (%s).', 'wp-job-manager' ), $default_salary_currency )
+			: __( 'Add a salary currency, this field is optional. Leave it empty to use the default salary currency.', 'wp-job-manager' );
+
 		$this->fields = apply_filters(
 			'submit_job_form_fields',
 			[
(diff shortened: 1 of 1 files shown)


includes/forms/class-wp-job-manager-form-submit-job.php
```diff
@@ -218,6 +218,14 @@ public function init_fields() {
 		} else {
 			$job_type = 'term-select';
 		}
+
+		$default_salary_currency     = get_option( 'job_manager_default_salary_currency' );
+		$salary_currency_placeholder = $default_salary_currency ?: __( 'e.g. USD', 'wp-job-manager' );
+		$salary_currency_description = $default_salary_currency
```
azaozz, line 224: Is it possible that `$default_salary_currency` is empty/not set? Wouldn't it be better to ensure it is always set perhaps to USD as it was before? 

Asking as the description `'Add a salary currency, this field is optional. Leave it empty to use the default salary currency (%s).'` is a lot better than the one where the default is not mentioned. Leaves the user to wonder :)

masteradhoc: @azaozz im quite sure it is never empty/not set. in my tests i always got USD although nothing was set inside the field. Also we have default set here: https://github.com/Automattic/WP-Job-Manager/blob/bea63891851f7acd38ba1cdc0c6951be989c8181/includes/admin/class-wp-job-manager-settings.php#L308

donnchawp (approved):
Looks good. Thanks!

One thing to note is that "?:" is discouraged for readability issues in WordPress code, but it's used elsewhere in this plugin, so not a blocker.

PR #2930, opened 2026-03-15
test more recent php versions / wp versions

masteradhoc:
* Expand the PHP unit test matrix to cover PHP 8.2, 8.3 in addition to the existing 7.4, 8.0, and 8.1
* Expand the WordPress version matrix to explicitly test WP 6.4 through 6.9 (covering `Requires at least` through the latest stable release) instead of a small set of pinned older versions
* Restructure the matrix to use `exclude` rules rather than exhaustive `include` entries, keeping the config concise and easier to maintain going forward
* Expand Psalm static analysis to also run on PHP 8.3 in addition to 7.4

### Testing Instructions
* Open a pull request targeting `trunk`, `feature/**`, or `release/**` and verify the Actions workflow runs the expected matrix combinations
* Confirm PHP test jobs run as noted
* Confirm Psalm runs on PHP 7.4 and 8.3

### Release Notes
* Dev: Expanded CI test matrix to cover more PHP and WordPress versions

*

### Deprecated Code
*

### Screenshot / Video


masteradhoc:
Reason we cant support the latest PHP Version is due to our dependencies.

WPJM is on vimeo/psalm: ^5.13. Here's the situation:
Psalm 5 — supports running on PHP 7.4+
Psalm 6 — stable, requires PHP 8.2+ to run
Psalm 7 — public beta
--
Same problem, different package — phpunit/phpunit: ^9.0 in your composer.json has the same implicit nullable deprecation issue. PHPUnit 9 can't run on PHP 8.4+.

PHPUnit 9 — max PHP 8.3
PHPUnit 10 — requires PHP 8.1+ and drops some older APIs
PHPUnit 11 — requires PHP 8.2+
..

As long as WPJM has to support 7.4 we're a bit stuck here.

donnchawp:
PHP 7.4 was EOL a few years ago. I think we should bump that. I can't imagine many sites are still using that. Also, WordPress.org [recommends](https://wordpress.org/about/requirements/) at least PHP 8.3.

However, let's not do that here. I would very much like to see these changes implemented. A new PR is needed for bumping the min PHP version.

PR #2931, opened 2026-03-15
Fix: Store scaled image URL for file fields when WP 5.3+ big image scaling is active

### LINKED ISSUE #1973: WordPress "scaled" image feature support, full res URL saved to listings
https://make.wordpress.org/core/2019/10/09/introducing-handling-of-big-images-in-wordpress-5-3/

Introduced in WordPress 5.3, if an image is above a certain "threshold" (width or height), with a default of `2560`, WordPress will automatically created a "scaled" version of the image, signified by `-scaled` being added to the image name, resulting in an image like: 
`XXX-scaled.jpg`

The issue with this is that the value saved for fields on listings, is the actual full res URL of the image, not the "scaled" version of the image.  Technically these images still exist (as WordPress just creates an additional "scaled" version of the image), but there's two reasons I wanted to bring this up.

**Unable to convert URL to attachment ID** 
The first issue is that any plugin or theme that attempts to convert the URL to an "attachment post ID" will fail, including calling the internal WordPress core `attachment_url_to_postid` function, since the value saved to listings is the not the "scaled" version, and when the attachment is inserted, the data is replaced with the URL to the "scaled" image (see below).

**Full Res Images are Served on Sites**
The other issue is that technically this feature of using "scaled" images will be disregarded by WPJM and the full res images will be served.  By supporting the "scaling" feature of WordPress 5.3+, we can make sure that when the user has not specifically disabled the "scaling" feature, that the scaled images will be used  -- speeding up site load times, as well as the issue mentioned above.

This only happens when the image is "attached" to a listing, and is triggered by the following:

1. WPJM->create_attachment
2. wp_generate_attachment_metadata()
3. wp_create_image_subsizes()
4. _wp_image_meta_replace_original()
5. wp_update_attachment_metadata()

It appears that this metadata is passed back to the `wp_generate_attachment_metadata` function called by WPJM, which includes the metadata saved to the "attachment post" including `original_image`

So IMO I think to resolve these issues, we should work on adding support for handling this "scaling" of images.  

Specifically in the `wp-content/plugins/wp-job-manager/includes/forms/class-wp-job-manager-form-submit-job.php` file where we call `wp_generate_attachment_metadata` we can check the data returned to see if the image was in fact "scaled", we would just have to figure out how we would then update the data in `$values` to reflect this.

The problem is that `create_attachment` is called _**after**_ the metadata is updated/set in `update_job_data`, so we would essentially have to move things around a bit, or come up with the best way to handle this.

What are your thoughts @jom ?

masteradhoc:
Fixes #1973

* Added an optional `&$actual_url` by-reference parameter to `create_attachment()`.
  When WordPress 5.3+ scales a large image during `wp_generate_attachment_metadata()`,
  the metadata will include `original_image` and `_wp_attached_file` will point to the
  `-scaled` version. `actual_url` is set via `wp_get_attachment_url()` in that case,
  giving callers the canonical URL WordPress has stored.

* Moved attachment creation inline within the field loop in `update_job_data()`, before
  `update_post_meta()` is called. This ensures `wp_generate_attachment_metadata()` runs
  first, so any scaled URL is resolved before being persisted to post meta — fixing both
  the served image (full-res was used instead of scaled) and `attachment_url_to_postid()`
  failures (stored URL didn't match the scaled URL WordPress recorded internally).

### Testing Instructions

* Upload an image larger than 2560px (width or height) as a file field on a job listing
  submission form.
* Confirm the value saved in post meta points to the `-scaled` version of the image, not
  the original.
* Confirm `attachment_url_to_postid()` returns the correct attachment ID for the stored URL.
* Confirm listings with images under the threshold are unaffected.
* Confirm multi-file fields and the company logo field behave as before.

### Release Notes

* Fixed an issue where the full-resolution URL was stored for image file fields when
  WordPress 5.3+ big image scaling was active, causing oversized images to be served and
  breaking attachment ID lookups.

*

### Deprecated Code

*

### Screenshot / Video

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -912,12 +912,14 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	}
 
 	/**
-	 * Creates a file attachment.
+	 * Creates an attachment for the given URL and attaches it to the current job.
 	 *
-	 * @param  string $attachment_url
-	 * @return int attachment id.
+	 * @param string      $attachment_url URL of the file to attach.
+	 * @param string|null &$actual_url    Set to the canonical attachment URL after creation.
+	 *                                    On WP 5.3+, may differ from $attachment_url if the image was scaled.
+	 * @return int Attachment post ID, or 0 on failure.
 	 */
-	protected function create_attachment( $attachment_url ) {
+	protected function create_attachment( $attachment_url, &$actual_url = null ) {
 		include_once ABSPATH . 'wp-admin/includes/image.php';
 		include_once ABSPATH . 'wp-admin/includes/media.php';
 
@@ -957,7 +959,15 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		$attachment_id = wp_insert_attachment( $attachment, $attachment_url, $this->job_id );
 
 		if ( ! is_wp_error( $attachment_id ) ) {
-			wp_update_attachment_metadata( $attachment_id, wp_generate_attachment_metadata( $attachment_id, $attachment_url ) );
+			$metadata = wp_generate_attachment_metadata( $attachment_id, $attachment_url );
+			wp_update_attachment_metadata( $attachment_id, $metadata );
+
+			// If the image was scaled (WP 5.3+), original_image is set and
+			// _wp_attached_file now points to the -scaled version.
+			if ( isset( $metadata['original_image'] ) ) {
+				$actual_url = wp_get_attachment_url( $attachment_id );
+			}
+
 			return $attachment_id;
 		}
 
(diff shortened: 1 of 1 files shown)


azaozz:
Hi, regarding of handling (very) large images in WP 5.3+

> actual_url is set via wp_get_attachment_url() in that case,
giving callers the canonical URL WordPress has stored.

Why is this needed? The idea in WP was to "ignore" as much as possible the uploaded file as it is not suitable for use on the internet. These are typically photos with dimensions larger than 2560px and sizes over 4MB. This type of images shouldn't be used on the internet however WP still stores them just in case the authors may need them later.

They are exposed for downloading on the attachment screen in WP. If the same thing is needed somewhere in the plugin maybe it would be better to link to there?

masteradhoc:
Hi @azaozz, thanks for looking at this!

The problem is that WPJM currently stores the original (pre-scaling) URL in post meta, because create_attachment() was called after update_post_meta(). This means:

1. attachment_url_to_postid() fails for any stored URL on oversized images — breaking any plugin or theme that tries to resolve the URL back to an attachment ID.
2. The full-resolution file is served instead of the scaled one, bypassing the performance intent of WordPress's own big image feature.

This PR fixes both by ensuring wp_generate_attachment_metadata() runs before the URL is persisted to post meta, and then using wp_get_attachment_url() to retrieve the canonical (potentially scaled) URL that WordPress itself recorded — keeping everything consistent with how WP handles the attachment internally.

masteradhoc:
@azaozz @donnchawp anything left on this PR to get it reviewed? :)

PR #2935, opened 2026-03-29
[jobs] shortcode: add author attribute to filter listings by user ID

### LINKED ISSUE #1892: Jobs shortcode needs an option to filter by user.
The jobs shortcode has quite a lot of functionality, but it is literally missing the only functionality I actually need.

I need the ability to place jobs on a users profile. For my purposes, it is in BuddyPress. Creating a tab/profile page for them is easily done. But putting the jobs shortcode in, there is no way to get only jobs created by this specific user.

Ideally, I should be able to add a user=2, or whatever ID I am currently viewing, into the shortcode to only show that users jobs. A secondary enhancement that would be quite amazing would be an empty_message="No results found." added to the shortcode so I could set the text shown in that instance.

Many themes properly style the HTML output of the jobs shortcode, so it is the best possible place to put this filter.

masteradhoc:
Fixes #1892

* Added `author` attribute to the `[jobs]` shortcode, allowing job listings to be filtered by one or more WordPress user IDs
* Supports a single ID (`[jobs author="1"]`) or comma-separated multiple IDs (`[jobs author="1,3,5"]`)
* Passes the author filter through all three query paths: direct PHP rendering (no-filters), AJAX with filters form, and AJAX without filters form
* Added `author__in` support to `get_job_listings()` (available to all callers, not just the shortcode)

### Testing Instructions

1. Create two or more job listings owned by different users
2. Add `[jobs author="X"]` to a page (where X is a user ID) — verify only that user's listings appear
3. Add `[jobs author="X,Y"]` — verify listings from both users appear
4. With `show_filters="true"` (default), interact with the filter form and verify the author restriction is preserved across AJAX requests
5. Verify that `[jobs]` without the `author` attribute continues to show all listings as before

### Release Notes

* New: The `[jobs]` shortcode now supports an `author` attribute to filter listings by user ID (e.g. `[jobs author="42"]` or `[jobs author="1,2,3"]`)

* No new hooks or templates. The `author` argument is now accepted by `get_job_listings()` and will be respected by any code that calls it directly.

### Important
Make sure to update [WPJM Documentation for Shortcode](https://wpjobmanager.com/document/advanced-usage/shortcode-reference/) as well and include the new feature!

### Deprecated Code
-

### Screenshot / Video
-

### DIFF
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -197,7 +197,7 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 			];
 		}
 
-		if ( isset( $args['author'] ) && '' !== (string) $args['author'] ) {
+		if ( isset( $args['author'] ) && ( is_array( $args['author'] ) || '' !== $args['author'] ) ) {
 			$raw_author = $args['author'];
 			$author_ids = is_array( $raw_author )
 				? array_values( array_filter( array_map( 'intval', $raw_author ), fn( $v ) => $v > 0 ) )
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -195,7 +195,7 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 		if ( ! empty( $args['author'] ) ) {
 			$author_ids = is_array( $args['author'] )
 				? array_filter( array_map( 'absint', $args['author'] ) )
-				: array_filter( array_map( 'absint', explode( ',', $args['author'] ) ) );
+				: array_filter( array_map( 'absint', explode( ',', (string) $args['author'] ) ) );
 			if ( ! empty( $author_ids ) ) {
 				$query_args['author__in'] = $author_ids;
 			}
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -192,6 +192,15 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 			];
 		}
 
+		if ( ! empty( $args['author'] ) ) {
+			$author_ids = is_array( $args['author'] )
+				? array_filter( array_map( 'absint', $args['author'] ) )
+				: array_filter( array_map( 'absint', explode( ',', $args['author'] ) ) );
+			if ( ! empty( $author_ids ) ) {
+				$query_args['author__in'] = $author_ids;
+			}
+		}
+
 		$job_manager_keyword = sanitize_text_field( $args['search_keywords'] );
 
 		if ( ! empty( $job_manager_keyword ) && strlen( $job_manager_keyword ) >= apply_filters( 'job_manager_get_listings_keyword_length_threshold', 2 ) ) {
@@ -218,6 +227,11 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 		do_action( 'before_get_job_listings', $query_args, $args );
 
 		$should_cache = 'rand_featured' !== $args['orderby'] && 'rand' !== $args['orderby'];
+
+		// Bypass cache when author filter is active to avoid stale results.
+		if ( ! empty( $query_args['author__in'] ) ) {
+			$should_cache = false;
+		}
 
 		// Cache results.
 		if ( apply_filters( 'get_job_listings_cache_results', $should_cache ) ) {
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -199,9 +199,16 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 
 		if ( isset( $args['author'] ) && ( is_array( $args['author'] ) || '' !== $args['author'] ) ) {
 			$raw_author = $args['author'];
-			$author_ids = is_array( $raw_author )
-				? array_values( array_filter( array_map( 'intval', $raw_author ), fn( $v ) => $v > 0 ) )
-				: array_values( array_filter( array_map( 'intval', explode( ',', (string) $raw_author ) ), fn( $v ) => $v > 0 ) );
+			$tokens     = is_array( $raw_author ) ? $raw_author : explode( ',', (string) $raw_author );
+			$author_ids = array_values(
+				array_filter(
+					array_map(
+						fn( $v ) => ctype_digit( trim( $v ) ) && (int) trim( $v ) > 0 ? (int) trim( $v ) : 0,
+						$tokens
+					),
+					fn( $v ) => $v > 0
+				)
+			);
 			// Fails-closed: if author was supplied but yielded no valid IDs, return zero results.
 			$query_args['author__in'] = ! empty( $author_ids ) ? $author_ids : [ 0 ];
 		}
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -12,7 +12,11 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 	 * Queries job listings with certain criteria and returns them.
 	 *
 	 * @since 1.0.5
-	 * @param string|array|object $args Arguments used to retrieve job listings.
+	 * @param string|array|object $args {
+	 *     Arguments used to retrieve job listings.
+	 *
+	 *     @type int|string|int[] $author Optional. User ID, comma-separated user IDs, or array of user IDs to filter listings by author. Default 0 (no filter).
+	 * }
 	 * @return WP_Query
 	 */
 	function get_job_listings( $args = [] ) {
@@ -194,12 +198,12 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 		}
 
 		if ( ! empty( $args['author'] ) ) {
-			$author_ids = is_array( $args['author'] )
-				? array_filter( array_map( 'absint', $args['author'] ) )
-				: array_filter( array_map( 'absint', explode( ',', (string) $args['author'] ) ) );
-			if ( ! empty( $author_ids ) ) {
-				$query_args['author__in'] = $author_ids;
-			}
+			$raw_author = $args['author'];
+			$author_ids = is_array( $raw_author )
+				? array_values( array_filter( array_map( 'intval', $raw_author ), fn( $v ) => $v > 0 ) )
+				: array_values( array_filter( array_map( 'intval', explode( ',', (string) $raw_author ) ), fn( $v ) => $v > 0 ) );
+			// Fails-closed: if author was supplied but yielded no valid IDs, return zero results.
+			$query_args['author__in'] = ! empty( $author_ids ) ? $author_ids : [ 0 ];
 		}
 
 		$job_manager_keyword = sanitize_text_field( $args['search_keywords'] );
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -228,11 +228,6 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 
 		$should_cache = 'rand_featured' !== $args['orderby'] && 'rand' !== $args['orderby'];
 
-		// Bypass cache when author filter is active to avoid stale results.
-		if ( ! empty( $query_args['author__in'] ) ) {
-			$should_cache = false;
-		}
-
 		// Cache results.
 		if ( apply_filters( 'get_job_listings_cache_results', $should_cache ) ) {
 			$to_hash            = wp_json_encode( $query_args );
--- a/wp-job-manager-functions.php
+++ b/wp-job-manager-functions.php
@@ -197,7 +197,7 @@ if ( ! function_exists( 'get_job_listings' ) ) :
 			];
 		}
 
-		if ( ! empty( $args['author'] ) ) {
+		if ( isset( $args['author'] ) && '' !== (string) $args['author'] ) {
 			$raw_author = $args['author'];
 			$author_ids = is_array( $raw_author )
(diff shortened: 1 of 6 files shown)


donnchawp (changes requested):
Looks good, but if I modify the jobs shortcode on my page to show
[jobs author=1]

The RSS feed link on the page doesn't have the author in it.

masteradhoc:
Thanks @donnchawp! Now its possible to use http://wpjm-pr.local/job-autor-1/?feed=job_feed&author=1,2 to get the requested authors :)

wp-job-manager-functions.php
```diff
@@ -219,6 +228,11 @@ function get_job_listings( $args = [] ) {
 
 		$should_cache = 'rand_featured' !== $args['orderby'] && 'rand' !== $args['orderby'];
 
+		// Bypass cache when author filter is active to avoid stale results.
```
donnchawp: Are you sure this is required? The cache key is derived from `$to_hash = wp_json_encode( $query_args );` so having the author field in there will create a new cache key. It shouldn't be stale.

donnchawp (changes requested):
Only one thing I'm worried about below.

wp-job-manager-functions.php
masteradhoc: I had issues while applying the PR that the change didnt show. but when i retest now all seems fine even without this part. so i went ahead and removed it now @donnchawp - thanks!

wp-job-manager-functions.php
```diff
@@ -192,6 +192,15 @@ function get_job_listings( $args = [] ) {
 			];
 		}
 
+		if ( ! empty( $args['author'] ) ) {
+			$author_ids = is_array( $args['author'] )
+				? array_filter( array_map( 'absint', $args['author'] ) )
+				: array_filter( array_map( 'absint', explode( ',', $args['author'] ) ) );
```
donnchawp: Cast to string because PHP8.1 will trigger a deprecation warning if author is an int, unfortunately.

```suggestion
				: array_filter( array_map( 'absint', explode( ',', (string) $args['author'] ) ) );
```

donnchawp (changes requested):
Only a tiny change this time, but there are no tests for author in tests/php/tests/test_class.wp-job-manager-functions.php either which would be great to have.

masteradhoc:
@donnchawp i've just pushed the tests. Can you give it a check? :)

donnchawp (changes requested):
Still some nagging issues, the first one really should be fixed and the rest cleared up too:

### [MEDIUM] AJAX response RSS link drops the active `author` filter
  `includes/class-wp-job-manager-ajax.php:242-249`

  `get_listings()` reads `$author` from the request and passes it to `get_job_listings()`, but the immediately-following call to `job_manager_get_filtered_links()` omits it. Reproduced live with `[jobs author="3"]` on my test site: the response
  payload's `rss_link` href is `…&search_keywords&author` (empty value) — the "Subscribe via RSS" link points to an unscoped feed and leaks listings the page intentionally hid.

  The PR's own change in `wp-job-manager-functions.php:642` (adding `'author'` to the feed URL when `$args['author']` is present) is dead code because no caller supplies it.

**Fix:** Add `'author' => $author` to the args array passed to `job_manager_get_filtered_links()`.

  ### [MEDIUM] Fails open on invalid `author` input
  `wp-job-manager-functions.php:195-202`, `includes/class-wp-job-manager-ajax.php:139`, `includes/class-wp-job-manager-post-types.php:730-735`

  Invalid input silently shows all listings instead of returning zero results. Two failure modes:

  1. **Non-numeric** like `[jobs author="alice,bob"]` → `absint` produces `[0,0]`, then `array_filter` produces `[]`, then `! empty( $author_ids )` is false → no `author__in` is set → all listings render.
  2. **Array input** like `?author[]=1` (AJAX or feed) → `sanitize_text_field()` flattens the array to `''` → filter silently disappears.

  For a publicly-rendered shortcode, "user typed invalid filter, sees everything" is the wrong default. Reproduced: `[jobs author="x"]` returns all listings.

  **Fix:** When `author` was provided but parsing yielded no valid IDs, force an empty result (`'author__in' => [0]`). Reject array input early at the AJAX/feed entry points.

  ### [LOW] Negative IDs become valid via `absint`
  `wp-job-manager-functions.php:197-198`

  `absint(-5)` is `5`, so `[jobs author="-5"]` filters by user 5's listings rather than returning nothing. Low real impact, but the filter behaves non-monotonically — a typo'd minus sign returns a stranger's data instead of empty.

  **Fix:** After `absint`, drop values `<= 0` rather than relying on `array_filter`.

  ### [LOW] Missing PHPDoc for new `author` arg
  `wp-job-manager-functions.php:11-17`

  The PHPDoc on `get_job_listings()` is not updated for the new public arg.

  **Fix:** Add `@param int|string|int[] $author Optional. User ID, comma-separated user IDs, or array of user IDs.`

  ### [LOW] PR description claims a transient-cache bypass that isn't implemented
  The description says *"Bypasses the transient cache when an author filter is active to prevent stale results."* No such code is in the diff, and inspection confirms the cache key already hashes `$query_args` (and therefore
  `author__in`), so the bypass would be unnecessary anyway. Just update the PR description.

  ### Test coverage gaps
  The two new tests only cover the happy path (single ID, multi ID). Add cases for: empty string, non-numeric (`"abc"`), negative (`"-5"`), `"0"`, mixed valid+invalid (`"3,abc"`), and array input (`['author' => [3, 5]]`).

  ### Reviewer note: tester friction
  The dist bundle (`assets/dist/js/`) is gitignored, so the JS changes only take effect after `npm run build:assets`. I initially saw "all listings shown" until I rebuilt. Worth either calling out in the PR description or considering
  whether dist should be a build artifact attached to the release flow.

donnchawp:
Reviewed the current `author` handling on HEAD (`1872f2e1`). Two correctness bugs and one input-validation concern. PHP simulations of each code path below.

## Blocker 1 — `[jobs]` with no `author` returns zero listings

`includes/class-wp-job-manager-shortcodes.php` defaults `'author' => 0` (L218), runs it through `sanitize_text_field` to `"0"`, and forwards it to `get_job_listings()` at L354. In `wp-job-manager-functions.php:200-206`:

```php
if ( isset( $args['author'] ) && ( is_array( $args['author'] ) || '' !== $args['author'] ) ) {
    // intval('0') = 0, filtered by $v > 0 → []
    $query_args['author__in'] = ! empty( $author_ids ) ? $author_ids : [ 0 ];
}
```

`'' !== '0'` is true, the block fires, `$author_ids` is empty, and `author__in` is set to `[0]`. Every existing `[jobs]` / `[jobs show_filters="false"]` invocation without an `author` attribute now returns no listings — a regression for the common case.

Fix: drop the default to `''` (or `null`), or only forward `author` to `get_job_listings()` when it was explicitly supplied.

## Blocker 2 — feed path fails open for `?author=0` and `?author[]=1`

`includes/class-wp-job-manager-post-types.php:747-755`:

```php
if ( isset( $_GET['author'] ) && ! is_array( $_GET['author'] ) ) {
    $sanitized_author = sanitize_text_field( wp_unslash( $_GET['author'] ) );
    $input_author     = empty( $sanitized_author ) ? false : array_values( array_filter( array_map( 'intval', explode( ',', $sanitized_author ) ), fn( $v ) => $v > 0 ) );
    if ( false !== $input_author && empty( $input_author ) ) {
        $input_author = [ 0 ];
    }
} else {
    $input_author = false;
}
```

- `?author=0` — `empty('0')` is true, so `$input_author = false`. Later `if ( ! empty( $input_author ) )` is skipped → no `author__in` → all published listings returned.
- `?author[]=1` — `! is_array(...)` is false, drops into the `else` branch → `$input_author = false` → filter silently dropped.

Both paths fail open here while the shortcode path fails closed on the same kinds of input — inconsistent behavior across surfaces.

Fix: track "param was provided" separately from "parsed valid IDs", and apply `[0]` for zero / malformed / array-shaped input — matching the shortcode policy.

## Warning — `intval()` partial-parse

Both files cast tokens with `intval()`, which silently accepts numeric prefixes: `intval('5abc')` returns `5`, so probes like `5abc`, `5def`, … still hit user 5. (Note: `intval('1e2')` returns `100`, not `1` — PHP honors scientific notation on numeric strings.) If the contract is "strict integer ID", gate with `ctype_digit()` or `/^\d+$/` before casting.

## Suggested test coverage

- `[jobs]` and `[jobs show_filters="false"]` with no `author` attr → listings shown.
- `?feed=job_feed&author=0` → consistent with shortcode policy (either all listings or `[0]`, but not divergent).
- `?feed=job_feed&author[]=1` → either honored or rejected explicitly, not silently dropped.
- `author=5abc` → no match (currently matches user 5).

masteradhoc:
Thanks @donnchawp i added your feedback to the PR now.

donnchawp (commented):
## Review

### Blockers

**1. PR description claims a cache bypass that doesn't exist.** No code in the diff bypasses the transient. The author filter does get folded into the cache key (it's added to `$query_args` before the hash at `wp-job-manager-functions.php:225`), so caching is actually fine — but the description's safety claim is unsupported. Either implement the bypass or correct the description.

**2. RSS link in the filter bar drops the author filter.** `class-wp-job-manager-ajax.php:247` now passes `'author' => $author` into `job_manager_get_filtered_links()`, but `wp-job-manager-functions.php:628-632` never forwards it into the `job_manager_get_listings_custom_filter_rss_args` array. A user clicking the RSS link from `[jobs author=\"42\"]` gets a feed for all authors. Add `'author' => ! empty( $args['author'] ) ? $args['author'] : ''` to that array.

### Warnings

**3. Parsing parity between `get_job_listings()` and `job_feed()`.** `wp-job-manager-functions.php:200` uses `intval`; `class-wp-job-manager-post-types.php:755` uses `ctype_digit`. For input `\"1.5,2\"`, the shortcode returns `author__in=[1,2]` and the RSS feed returns `author__in=[2]` — same string, different result sets. Also, `intval('1.5') = 1` silently truncates to a real user ID that the caller never asked for. Align both paths on the stricter `ctype_digit(trim(\$v)) ? (int)trim(\$v) : 0` pattern.

**4. `@since 2.5.0` is speculative.** All 8 new test methods are tagged `@since 2.5.0`, but the current release is `2.4.2` and the next version hasn't been declared. Use the actual next tag once known, or `@unreleased` if project conventions allow it.

### Suggestion

**5.** `class-wp-job-manager-ajax.php:211`'s "Search completed" message check excludes the author filter (also excludes `featured` / `filled` / `remote_position`, so possibly intentional). Worth a deliberate yes/no.

### Notes

- Tests exercise only `get_job_listings()`; the duplicated parsing logic in `job_feed()` and the AJAX handler is untested.
- `output_jobs()` correctly adds `author` to `\$data_attributes` only when non-empty, so the JS `\$target.data('author')` path is fine.

masteradhoc:
@donnchawp jut checked the blockers.

1) solved
2) are you reviewing an old state of the PR? author key is already present in that array — it was applied in a previous commit, see https://github.com/Automattic/WP-Job-Manager/pull/2935/changes#diff-f2e92c961523255d469f2ea4ce5a6a5772b739071ce9b4f2e93e1f946636b882R645
3) adjusted
4) please let me know which version you'll use next so i can make the final commit.

donnchawp (commented):
@masteradhoc — you're right, apologies. I re-pulled the diff at HEAD `621464b0` and you've already addressed it:

**Blocker 2 (RSS link drops author) — wrong on my part.** Line 653 of the PR's `wp-job-manager-functions.php` does include `'author' => ! empty( \$args['author'] ) ? \$args['author'] : ''` inside the `job_manager_get_listings_custom_filter_rss_args` array. That's exactly the fix I described as missing — it was already there since commit `d6784d55`. My second review re-flagged it incorrectly. Withdrawn.

**Blocker 1 (cache-bypass claim) — confirmed resolved.** PR description no longer mentions a cache bypass. Good.

**Warning 3 (parsing parity) — confirmed resolved.** `wp-job-manager-functions.php` and `class-wp-job-manager-post-types.php` now use the same `ctype_digit( trim( \$v ) ) && (int) trim( \$v ) > 0` predicate. They diverge intentionally on array input — `get_job_listings()` accepts arrays as a first-class shape (covered by `test_get_job_listings_author_array_input`), `job_feed()` fails closed on `?author[]=…` since that's not a supported URL shape. That asymmetry is correct.

**Warning 4 (@since 2.5.0):** the project uses concrete version numbers throughout (no `@since unreleased` precedent in the codebase). Last release was 2.4.2 with no next-version bump merged to trunk yet. Given this is a new feature, **`@since 2.5.0`** is the most plausible target; a maintainer can adjust if 2.4.3 or 2.5.0 lands differently. Leave it as 2.5.0.

From my side: ready to merge once the `@since` is settled. Sorry for the noise on #2.

donnchawp (commented):
Correction on Warning 4 — I missed this convention. WPJM uses `$$next-version$$` as a placeholder, swapped at release by `scripts/replace-next-version-tag.sh`. So the right value here is:

```
@since $$next-version$$
```

…rather than a hand-picked `2.5.0`. Please update the 8 new test methods to use the placeholder. Once that's in, ready to merge from my side.

masteradhoc:
@donnchawp thanks, consider it done :)

donnchawp:
Thanks @masteradhoc ! Merging it now. :)

PR #2970, opened 2026-06-09
Restore PHP 8.4 test coverage — modernise bootstrap error handler

thisismyurl:
Hey team,

Picking up the work from #2949 — here's the bootstrap modernisation needed to restore PHP 8.4 CI coverage.

**Two compounding issues were killing the bootstrap before any test could run:**

1. `convert_to_exception()` promoted every error level — including `E_DEPRECATED` — to a fatal exception. PHPUnit 9.6.8's internals emit "implicitly nullable parameter" deprecations on PHP 8.4, so the bootstrap threw immediately. Fix: skip `E_DEPRECATED` and `E_USER_DEPRECATED` from `/vendor/`. Own-code deprecations still throw — that's the point of the handler.

2. The `$error_descriptions` map referenced `E_STRICT`, which is itself deprecated in PHP 8.4 (and a no-op since PHP 8.0). That reference triggered a deprecation inside `convert_to_exception()`, compounding issue 1. Fix: remove the stale `E_STRICT` entry.

**Also in this PR:**

3. Bump `phpunit/phpunit` in `composer.lock` to 9.6.34 (still within the `^9.0` constraint already in `composer.json`). 9.6.20+ ships PHP 8.4 compat patches, which reduces the volume of vendor deprecations the handler needs to absorb.

4. Re-add `{ wp: '6.9', php: '8.4', wpmu: 0 }` to the CI matrix, with the comment updated to drop the "pending" note now that bootstrap handles 8.4 cleanly.

The 7.4 and 8.3 jobs are untouched.

(full disclosure: AI helped me identify the issue and verify my work)

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -52,11 +52,11 @@ jobs:
         # PHP 7.4 (EOL Nov 2022) is kept only as a smoke test against the
         # plugin's declared floor; 8.0–8.2 are dropped as they're either EOL
         # or security-only and a break on those would also surface against
-        # 7.4 or 8.3. PHP 8.4 coverage is pending bootstrap modernisation —
-        # see https://github.com/Automattic/WP-Job-Manager/issues/2950.
+        # 7.4 or 8.3.
         include:
           - { wp: '6.4', php: '7.4', wpmu: 0 }
           - { wp: '6.9', php: '8.3', wpmu: 0 }
+          - { wp: '6.9', php: '8.4', wpmu: 0 }
     env:
       WP_VERSION: ${{ matrix.wp }}
       PHP_VERSION: ${{ matrix.php }}
--- a/composer.lock
+++ b/composer.lock
@@ -801,29 +801,30 @@
         },
         {
             "name": "doctrine/deprecations",
-            "version": "1.1.3",
+            "version": "1.1.6",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/deprecations.git",
-                "reference": "dfbaa3c2d2e9a9df1118213f3b8b0c597bb99fab"
+                "reference": "d4fe3e6fd9bb9e72557a19674f44d8ac7db4c6ca"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/deprecations/zipball/dfbaa3c2d2e9a9df1118213f3b8b0c597bb99fab",
-                "reference": "dfbaa3c2d2e9a9df1118213f3b8b0c597bb99fab",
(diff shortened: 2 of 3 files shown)


donnchawp (approved):
Reviewed the bootstrap modernisation, the `composer.lock` churn, and the CI matrix change. The fix is technically correct and I'm comfortable approving.

### What's solid

- **The vendor-deprecation skip is an `AND`, not an `OR`.** It only swallows `E_DEPRECATED`/`E_USER_DEPRECATED` *and* only when the error originates inside `/vendor/`. Own-code deprecations still throw, and non-deprecation errors from vendor (warnings/fatals) still throw. The handler's value is preserved, not gutted — which is exactly the invariant that matters here.
- **`E_STRICT` removal is safe.** The description map has an explicit `'Unknown Error: '` fallback, so an unmapped code can't produce an undefined-index warning, and `E_STRICT` isn't referenced anywhere else in non-vendor code.
- **`composer.lock`: runtime packages are byte-for-byte unchanged.** All churn is in `packages-dev`, `phpunit/phpunit` stays within the existing `^9.0`, and `composer.json` is untouched.
- **CI matrix is well-formed.** The re-added `{ wp: '6.9', php: '8.4', wpmu: 0 }` matches its siblings' shape, and `fail-fast: false` keeps an 8.4 failure from cancelling the 7.4/8.3 runs. No false-green risk — PHPUnit pass/fail rides on assertion exceptions, not PHP error levels, so the skip can't mask a real test failure.

### One note (non-blocking)

The description says it "bumps `phpunit/phpunit`", but `composer update phpunit/phpunit` also re-resolved ~19 transitive dev packages — most notably `phpstan/phpdoc-parser` 1.25 → 2.3 (a major). All within declared constraints, no new packages, runtime deps clean — so this is a description-scope nit, not a defect. Worth a one-line tweak to the description (e.g. "bump dev deps via `composer update phpunit/phpunit`; 19 dev packages, runtime unchanged") so the next reader doesn't mistake it for a hand-edited lock.

### Local verification

CI on this PR is still gated on `action_required`, so nothing has run yet. I installed the PR's `composer.lock` and ran the lint locally:

- **PHPCS (the lint job): clean, exit 0.** ✅
- Psalm I couldn't exercise locally (Psalm 5.13.1 trips on PHP 8.5's removed `E_STRICT` in its *own* error handler) — but CI only runs Psalm on 7.4/8.3, which this PR doesn't touch, so that's a local-runtime artifact rather than a PR concern. The `phpdoc-parser` major bump only feeds the static-analysis toolchain, which runs on those unchanged PHP versions.

Recommend approving the workflow runs to confirm the full matrix (incl. the new 8.4 test job) before merge.

PR #2995, opened 2026-07-01
Validate attachment ownership when reusing a company logo by ID

donnchawp:
## What

The frontend job submission form lets a user carry a previously uploaded company logo forward by its numeric attachment ID. This adds a validation check so a referenced attachment ID is only accepted when the current user is allowed to use that attachment (they own it, or have permission to edit it).

Reusing your own logo and admin edits of other users' listings are unaffected.

## Why

Tightens input validation on the company logo file field so an arbitrary attachment ID can't be bound to a listing.

## Testing

- Adds `WP_Test_Submit_Job_Attachment_Ownership` covering the accept/reject paths (own attachment, empty value, edit-capability, and a disallowed ID).
- Full PHPUnit suite passes; `phpcs` clean.

----

| Plugin build for 3acb0e7c62f56a7dff14aeb6bab2a81dc4c530bb <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-2995-3acb0e7c.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/2995-3acb0e7c)             |

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -506,6 +506,10 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 
 							// Check if attachment is valid.
 							if ( is_numeric( $file_url ) ) {
+								$attachment_id = absint( $file_url );
+								if ( $attachment_id && ! $this->is_attachment_authorized_for_current_user( $attachment_id ) ) {
+									throw new Exception( __( 'Invalid attachment provided.', 'wp-job-manager' ) );
+								}
 								continue;
 							}
 							$file_url = esc_url( $file_url, [ 'http', 'https' ] );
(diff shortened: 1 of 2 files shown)


Hastibe:
@donnchawp -- I recently updated to 2.4.5 and started receiving user reports that they can't submit (or save edits of) jobs because of an "Invalid attachment provided" error, which I believe is being caused by this issue. This appears to be happening with job posts that are reusing logos that were replaced or added by our site admin when users ran into issues, so the admin is the author of the image, not the user. 

How can we resolve this without it requiring user intervention? 

We looked into changing the author of these images to the correct users via the native WordPress core "Author" drop-down menu on the Edit Media page, but the employer user role does not appear as an option there, frustratingly.

donnchawp:
@Hastibe I think you're right. We'll get that fixed.

donnchawp:
@Hastibe try the fix in #3060. You can download the zip file from there and copy includes/forms/class-wp-job-manager-form-submit-job.php into your install, overwriting that file there.

Hastibe:
Thanks so much, @donnchawp -- really appreciate your help with this. We've implemented your #3060 fix, and tested that it seems to fix the issue; will keep you posted if we run into anything else with it, though!

PR #2998, opened 2026-07-02
Fix #2911: Improve text color contrast for WCAG 2.1 AA compliance

### LINKED ISSUE #2911: Job listing location and date posted text color contrast is too low
**Describe the bug**
The job listing location and date posted text color contrast is too low. With a text color of #999999 (medium gray) on a background color of #ffffff (white), the color contrast ratio is only 2.85:1, which does not meet accessibility guidelines. See WCAG 2.1/2.2 Level AA: [1.4.3 Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum).
This is a problem for state and federal sites like mine, which need to establish web accessibility compliance by April 2026. See https://www.ada.gov/resources/2024-03-08-web-rule/

**To Reproduce**
1. Go to a published job posting. Also applies to the /jobs search results.
2. Inspect the job posting location and date posted text at the top of the posting.
3. Run an accessibility checker against the page. Something like Accessible Web or axe DevTools will work.
4. See the accessibility results.

**Expected behavior**
The color contrast should be 4.5:1 or higher. Changing the text color to a darker gray of at least #767676 or darker will get to that ratio. You can use a contrast ratio tool to help: https://webaim.org/resources/contrastchecker/
Edit ./wp-job-manager/assets/dist/css/frontend.css (job listing page)
- Entry for ".single_job_listing .meta li"

Edit ./wp-job-manager/assets/dist/css/frontend.css (job search & results page)
- Entry for "ul.job_listings li.job_listing a div.position .company"
- Entry for "ul.job_listings li.job_listing a div.location"
- Entry for "ul.job_listings li.job_listing a .meta"

You might check the frontend.css file for other instances of `color: #999`

**WordPress Environment**
- WordPress Version: 6.9.1
- PHP Version: 7.4.24
- Other important details:

alinclamba:
## Summary

- Updates text color from `#999` to `#767676` for job listing meta fields
- Improves contrast ratio from 2.85:1 to 4.54:1 (meets WCAG 2.1 AA minimum of 4.5:1)
- Affects: company name, location, meta (date posted, job type) in both listings loop and single job view

## Files changed

- `assets/css/frontend.scss` (3 occurrences)
- `assets/css/job-listings.scss` (3 occurrences)

## Test plan

- [ ] Verify contrast ratio ≥4.5:1 using WebAIM contrast checker or browser DevTools
- [ ] Visual check: text should appear slightly darker gray, no layout changes

Note: Modern default themes (Twenty Twenty-One, Twenty Twenty-Five) override WPJM styles, so the change is only visible on themes that use WPJM's default CSS.

Closes #2911

----

| Plugin build for 505ba249903466b3188d46987e15507c3a690803 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-2998-505ba249.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/2998-505ba249)             |

### DIFF
--- a/assets/css/frontend.scss
+++ b/assets/css/frontend.scss
@@ -349,7 +349,7 @@ div.job_listings {
 			padding: 0 0 0 1em;
 			line-height: 1.5em;
 			font-style: italic;
-			color: #767676;
+			color: $subtext-accessible;
 		}
 
 		.website::before {
@@ -405,7 +405,7 @@ div.job_listings {
 			padding: 0.5em;
 			float: left;
 			line-height: 1em;
-			color: #767676;
+			color: $subtext-accessible;
 		}
 
 		.job-type {
@@ -914,7 +914,7 @@ div.job_listings {
 		}
 
 		.job_title small {
-			color: #767676;
+			color: $subtext-accessible;
 		}
 
 		.featured-job-icon {
--- a/assets/css/frontend.scss
+++ b/assets/css/frontend.scss
@@ -12,7 +12,6 @@ $highlight: 		adjust-hue($primary, 150); 						/* Prices, In stock labels, sales
 $highlightext:		desaturate(lighten($highlight, 50%), 18%);	/* Text on highlight colour bg */
 
 $contentbg: 		#fff; 										/* Content BG - Tabs (active state) */
-$subtext: 			#777; 										/* small, breadcrumbs etc */
 
 @mixin display-icon() {
 	display: inline-block;
@@ -349,7 +348,7 @@ div.job_listings {
 			padding: 0 0 0 1em;
 			line-height: 1.5em;
 			font-style: italic;
-			color: $subtext-accessible;
+			color: $subtext;
 		}
 
 		.website::before {
@@ -405,7 +404,7 @@ div.job_listings {
 			padding: 0.5em;
 			float: left;
 			line-height: 1em;
-			color: $subtext-accessible;
+			color: $subtext;
 		}
 
 		.job-type {
@@ -809,7 +808,7 @@ div.job_listings {
 
 	.meta {
 		font-style: italic;
-		color: #777;
+		color: $subtext;
 	}
 
 	.job-type {
@@ -914,7 +913,7 @@ div.job_listings {
 		}
 
 		.job_title small {
-			color: $subtext-accessible;
+			color: $subtext;
 		}
 
 		.featured-job-icon {
--- a/assets/css/frontend.scss
+++ b/assets/css/frontend.scss
@@ -349,7 +349,7 @@ div.job_listings {
 			padding: 0 0 0 1em;
 			line-height: 1.5em;
 			font-style: italic;
-			color: #999;
+			color: #767676;
 		}
 
 		.website::before {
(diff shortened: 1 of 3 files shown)


donnchawp:
Thanks for picking this up, Alin. Really nice to see an accessibility fix like this, and it's spot on. You mapped every selector the issue called out, caught the extra `#999` text instances beyond the named ones, and `#767676` lands at 4.54:1 which clears the WCAG AA 4.5:1 bar exactly as the issue asked. The value and the coverage are both correct. 👏

One small, totally optional suggestion for maintainability (not a blocker):

`frontend.scss` already defines its muted text colors as SCSS variables near the top of the file, e.g.:

```scss
$subtext: 			#777; 	/* small, breadcrumbs etc */
```

Since the new `#767676` is that same "muted secondary text" colour and it's now repeated in six places, it'd be a bit tidier to define it once as a variable and reference it everywhere — that way the next time someone adjusts the colour it's a one-line change instead of six. Something like:

```scss
// near the other colour variables at the top of frontend.scss
$subtext-accessible: 	#767676; 	/* WCAG AA-compliant muted text */
```

```scss
// then in each of the six spots
color: $subtext-accessible;
```

(You could also just retune the existing `$subtext` to `#767676` if we're happy for all `$subtext` uses to shift — your call.)

Completely fine to merge as-is if you'd rather keep it simple. The fix works either way. Happy to push that variable tweak onto the branch for you if it's easier. Thanks again for improving this for everyone. 🙏

alinclamba:
Thanks @donnchawp! Yes please, go ahead and push the variable tweak. Appreciate the review and the suggestion.

donnchawp:
Pushed the variable tweak (4660b42f).

One deviation from my suggestion: I put `$subtext-accessible` in `mixins.scss` rather than at the top of `frontend.scss`. Reason — `job-listings.scss` compiles both standalone (`job-listings.css`) *and* is `@import`ed into `frontend.scss` at line 3, which is before the variables at line 5. A variable defined in `frontend.scss` wouldn't be in scope for `job-listings.scss` in either case. `mixins.scss` is imported by both files and already holds shared colour variables, so it's the natural shared home. All six spots now reference the variable, and the compiled CSS still resolves to `#767676`.

donnchawp:
Follow-up on the variable tweak (505ba249) — I flagged that Claude's first pass left two near-identical variables (`$subtext` at `#777` and `$subtext-accessible` at `#767676`), which is basically the same colour twice.

Collapsed them into a single source of truth:

- One variable now: `$subtext: #767676` in `mixins.scss` (shared by both `frontend.scss` and `job-listings.scss`), referenced everywhere.
- Removed the dead `$subtext: #777` from `frontend.scss` — it was defined but never used, and since `mixins` is imported first it would otherwise have shadowed the new value back to `#777`.
- Folded in one straggler: `frontend.scss` had a hardcoded `.meta { color: #777 }` that the original diff didn't touch — same muted colour, and at 4.48:1 it was just under the AA bar. It's now on the variable at 4.54:1.

No hardcoded muted greys left. Shout if you'd rather keep that last `.meta` element out of scope and I'll revert just that line.

PR #3009, opened 2026-07-06
Fix #2851: make WPJM widgets insertable on block themes

### LINKED ISSUE #2851: WPJM widgets not appearing on block themes
### Steps to Reproduce

On latest WP and WPJM and using a modern block theme:
1. Go to Appearance -> Editor
2. Try to add a WPJM widget, like "Recent Jobs"

### What I Expected

The widget should appear.

### What Happened Instead

Nothing appeared.

### PHP / WordPress / WP Job Manager Version

PHP 8.3 / WP 6.6 / WPJM 2.4.0

### Browser / OS Version

Indifferent.

### Context / Source

Reported originally here: https://wordpress.org/support/topic/recent-jobs-widget-3/

bobmatyas:
Fixes #2851

On a block theme (no classic Widgets screen), the Recent Jobs and Featured
Jobs widgets could not be added through Appearance → Editor. In WP's Site
Editor the core Legacy Widget block is not exposed by default, and the
widgets were not opted into the block-editor REST pipeline, so there was no
working path to insert or configure them.

- Expose the core Legacy Widget block in the inserter on the **Site Editor
  screen only** (scoped so the post/page editor is unchanged). The Site
  Editor already registers `core/legacy-widget` itself on every WP version
  we support (6.4–7.0), deliberately with `supports.inserter: false`, so
  calling `wp.widgets.registerLegacyWidgetBlock()` — the approach in earlier
  revisions of this PR — was a duplicate registration that logged
  `Block "core/legacy-widget" is already registered` in the console.
  Instead, the documented [`blocks.registerBlockType` JS filter](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/)
  flips the inserter support flag on core's own registration: one
  registration, no warning, and no race against core's `domReady`-deferred
  registration.
- Set `show_instance_in_rest => true` on the shared widget abstract so both
  widgets round-trip their settings via REST and render an inline form +
  live preview in the block editor.
- Guard a pre-existing "Undefined array key remote_position" notice in the
  Recent Jobs widget that fires on the block-editor REST render path when
  remote positions are disabled.

### Testing Instructions

1. Activate a block theme (e.g. Twenty Twenty-Five) and create a few
   published job listings.
2. Go to Appearance → Editor and open a template, with the browser console
   open.
3. Confirm the console shows no
   `Block "core/legacy-widget" is already registered` warning.
4. In the inserter, add the **Legacy Widget** block and choose **Recent Jobs**
   (or Featured Jobs).
5. Confirm the settings form is editable and the preview renders listings.
6. Save the template and view the corresponding front-end page; the widget
   renders identically to a classic theme.
7. Confirm the post/page editor is unchanged (no Legacy Widget block in the
   inserter there).

### Release Notes

Recent Jobs and Featured Jobs widgets can now be added via the Legacy Widget
block in the Site Editor on block themes.

None.

### Deprecated Code

None.

### Screenshot / Video

Editor: Legacy Widget → Recent Jobs renders listings with an editable form
and a clean console (verified locally on WP 7.0 / Twenty Twenty-Five).
Frontend parity verified via the saved template.

----

| Plugin build for d67150c6ea16d8db25bd499095757d5484e4314f <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-3009-d67150c6.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/3009-d67150c6)             |

### DIFF
--- a/includes/class-wp-job-manager-blocks.php
+++ b/includes/class-wp-job-manager-blocks.php
@@ -43,6 +43,7 @@ class WP_Job_Manager_Blocks {
 		}
 
 		add_action( 'init', [ $this, 'register_blocks' ] );
+		add_action( 'enqueue_block_editor_assets', [ $this, 'enable_legacy_widget_block' ] );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


bobmatyas:
Something to watch out here is that to fix this without blocks, it has to enable the Legacy Widget block in the Site Editor. This can't be opted-in per plugin, so it will expose other plugin's Legacy Widgets, so we'll want to evaluate whether that is worth it.

I think the ideal fix is for these widgets to function as blocks.

donnchawp (approved):
Looks good!

PR #3013, opened 2026-07-07
Serialize the submission-limit check when publishing a preview

donnchawp:
A `preview` listing is not counted towards the submission limit, so two concurrent "continue" requests could each read a stale count in `job_manager_user_can_submit_job_listing()` and both publish, exceeding the configured limit. Hold a short-lived per-user advisory lock (`GET_LOCK`) across the limit check and the publish so the requests serialize. Best-effort: if the lock cannot be obtained the check still runs, as before.

Includes a regression test that the lock is held across the critical section and released afterwards.

----

| Plugin build for babb34a03db700d3a937db79f126d568d899eea9 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/09/wp-job-manager-zip-3013-babb34a0.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/09/3013-babb34a0)             |


donnchawp:
Follow-up from code review (18c561da):

- **Namespaced the lock name.** `GET_LOCK` names are scoped to the whole DB server, not the schema, so unrelated installs sharing a MySQL server (common on shared hosting) or multisite network sites could collide on a shared low user ID — and, because the check fails closed, spuriously block each other's publishes. The name is now scoped by `md5($wpdb->dbname . '|' . $wpdb->prefix)`, kept under MySQL's 64-char lock-name limit.
- **Fail-open on backends without `GET_LOCK`.** On the SQLite integration, `GET_LOCK` errors → returned `null` → fail-closed → *every* preview publish was blocked. `acquire_submission_lock()` now distinguishes an unsupported backend (via `$wpdb->last_error`) from genuine contention and falls back to the unserialized path instead of blocking. Added a regression test for it.
- **Documented the single-connection assumption.** `GET_LOCK` is session-scoped; under connection-splitting drivers (HyperDB/LudicrousDB) the read may hit a replica, degrading the lock to best-effort. The publish still proceeds, so availability is preserved. Noted in the docblock.

Not addressed (review minors, can follow up if wanted): wrapping the critical section in `try/finally`, and making the 5s `GET_LOCK` timeout filterable.

Caveat: the SQLite fallback branch is verified via sentinel injection in the test, not against a live `GET_LOCK`-less driver (wp-env has no SQLite backend).

donnchawp:
Holding this out of 2.4.6. It breaks listing submission on SQLite.

The SQLite integration doesn't error on `GET_LOCK`. It translates the statement and returns the string `'1=1'`, with `$wpdb->last_error` empty. So `acquire_submission_lock()` doesn't match `'1'`, doesn't see an error, and returns `null`, and the caller fails closed. Every first-time preview publish on a SQLite site gets "The listing could not be published because the site is temporarily busy" with no way past it.

I ran the actual method on 18c561da under Playground:

```
backend         : sqlite
acquire returns : NULL
```

And the same method on MySQL under wp-env, for contrast:

```
raw GET_LOCK    : '1'
acquire returns : 'wpjm_submit_<hash>_0'
```

The problem is inferring "unsupported" from an error being set. MySQL's `GET_LOCK` returns 1, 0 or NULL and nothing else, so a value outside that set means the backend didn't implement it. Testing for that positively would catch `'1=1'` and wouldn't depend on the driver raising an error. It would also make the sentinel injection in the test match what the real driver does.

The rest of the PR looks right to me. It's this one branch that needs redoing.

tellyworth:
Pushed a follow-up commit addressing the issues a review of this PR surfaced (donncha is away; summary of what changed and why):

**Two findings defeated the fix as implemented:**
- The under-lock re-read used `get_post()`, which returns this request's own object-cached copy — a racing process's `clean_post_cache()` can't invalidate it — so on installs without a persistent object cache the 'racer already published' guard could never fire. Now the cache entry is dropped before the re-read, and a new regression test (`test_race_guard_survives_stale_object_cache`) simulates the cross-process commit with a direct `UPDATE`; it fails without the fix.
- `acquire_submission_lock()` treated *any* `$wpdb->last_error` as 'backend lacks GET_LOCK' and silently proceeded unserialized — transient MySQL errors reopened the race exactly under load.

**Design change — best-effort instead of fail-closed:** any acquire failure (unsupported backend, timeout, transient error) now proceeds without the lock, protected by the fresh status re-read. Rationale in the docblock: refusing to publish traded a narrow, self-limiting race (a user briefly exceeding their own quota) for user-facing availability failures — the 'temporarily busy' error on double-clicks behind slow publish hooks, all guests fail-closing on one shared `user_id 0` lock, and locking even on default installs with no limit configured. This also removes the tri-state `SUBMISSION_LOCK_UNSUPPORTED` sentinel, drops the timeout from 5s to 2s (waiters stop pinning FPM workers), skips the lock when no limit is configured and nothing filters the check, and wraps the critical section in try/finally so a throwing publish hook can't leak the lock on persistent connections.

**Adjacent fixes in the touched code:** the first `$job` read is now guarded (PHP 8 fatal when the listing was deleted between preview and Continue), and `job_manager_count_user_job_listings()` now counts `future` — a scheduled listing is a committed submission WP-Cron publishes with no further check, so excluding it let a user bypass the limit deterministically by scheduling every listing.

Full suite green locally (626 tests / 3 known geocode skips).

PR #3016, opened 2026-07-08
feat(admin): add setting to toggle Promoted Jobs feature

### LINKED ISSUE #2608: Add options to disable promoted jobs functionality
- A filter for developers.
- A setting for end users.
- https://github.com/Automattic/WP-Job-Manager/issues/2523

faisalahammad:
## Changes Proposed in this Pull Request

Add a checkbox in **Job Listings > Settings > Job Listings** to enable/disable the Promoted Jobs feature. Fixes #2608

The developer filter `job_manager_enable_promoted_jobs` already exists. This adds a UI toggle so non-technical site owners can disable Promoted Jobs without writing code. Default: enabled (no behavior change for existing installs). Developer code filter at default priority still overrides the admin setting.

### includes/admin/class-wp-job-manager-settings.php

Registers the checkbox in the settings array so it appears in the admin UI.

### includes/class-wp-job-manager.php

Replaced the priority-based self-hook with `WP_Job_Manager_Promoted_Jobs_Helper::is_enabled()`. The option value is passed as the initial value to `apply_filters()`, so any developer filter (regardless of priority) overrides the admin setting.

### includes/class-wp-job-manager-promoted-jobs-helper.php

New helper class that bridges the admin option to the `job_manager_enable_promoted_jobs` filter. Removes the need for the self-hook at priority 5.

### includes/admin/class-wp-job-manager-admin.php

Updated to use the new helper instead of calling `apply_filters()` directly.

### includes/class-wp-job-manager-data-cleaner.php

Added `job_manager_enable_promoted_jobs` to the OPTIONS array for cleanup on uninstall.

### tests/php/tests/test_class.promoted-jobs-setting.php

Covers all state combinations (option on/off x filter true/false), default behavior, setting registration, data cleaner entry, and a low-priority filter override (priority 1).

## Testing Instructions

**Test 1: Setting visible and enabled by default**
1. Go to Job Listings > Settings > Job Listings tab
2. Confirm "Enable Promoted Jobs" checkbox is checked by default

**Test 2: Uncheck disables Promoted Jobs**
1. Uncheck the setting and save
2. Go to Job Listings screen
3. Result: "Promote" column gone, no promote button, REST endpoint returns 404

**Test 3: Code filter overrides setting**
1. Add `add_filter( 'job_manager_enable_promoted_jobs', '__return_true' );` to a mu-plugin
2. Uncheck the setting in admin
3. Result: Promoted Jobs still enabled (code wins over option)

## Release Notes

Adds a Promoted Jobs toggle in Job Listings > Settings > Job Listings.

## New or Updated Hooks and Templates

No new hooks. The existing `job_manager_enable_promoted_jobs` filter behavior is unchanged.

## Deprecated Code

No deprecations.

### DIFF
--- a/includes/admin/class-wp-job-manager-admin.php
+++ b/includes/admin/class-wp-job-manager-admin.php
@@ -68,7 +68,7 @@ class WP_Job_Manager_Admin {
 		/**
 		 * Documented in class-wp-job-manager.php
 		 */
-		$this->are_promoted_jobs_enabled = apply_filters( 'job_manager_enable_promoted_jobs', true );
+		$this->are_promoted_jobs_enabled = WP_Job_Manager_Promoted_Jobs_Helper::is_enabled();
 		if ( $this->are_promoted_jobs_enabled ) {
 			include_once dirname( __FILE__ ) . '/class-wp-job-manager-promoted-jobs-admin.php';
 		}
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -336,6 +336,15 @@ class WP_Job_Manager_Settings {
 							'track'      => 'value',
 						],
 						[
+							'name'     => 'job_manager_enable_promoted_jobs',
+							'std'      => '1',
+							'label'    => __( 'Promoted Jobs', 'wp-job-manager' ),
+							'cb_label' => __( 'Enable Promoted Jobs', 'wp-job-manager' ),
+							'desc'     => __( 'Allow job listings to be promoted through the Promoted Jobs feature.', 'wp-job-manager' ),
+							'type'     => 'checkbox',
+							'track'    => 'bool',
+						],
+						[
 							'name'     => 'job_manager_display_location_address',
 							'std'      => '0',
 							'label'    => __( 'Location Display', 'wp-job-manager' ),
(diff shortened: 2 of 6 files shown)


donnchawp:

The commercial boundary is fine — this only hides UI and skips loading the module, nothing about
licensing/purchase. Completeness checks out (no orphan surfaces outside the two gates), it defaults
to enabled (no change on upgrade), and the data-cleaner entry is added. One design issue:

**Needs fixing**

1. **The "code override wins over option" logic relies on filter priority.**
   (`includes/class-wp-job-manager.php:90-99`.) It self-hooks the same filter at priority 5 and
   depends on developer filters running later at the default 10 — so a filter registered at
   priority ≤5 (e.g. `add_filter( 'job_manager_enable_promoted_jobs', '__return_true', 1 )`) is
   silently overridden, contradicting the code's own comment and #2608's acceptance criterion. It's
   also hard to discover from the call site. Cleaner: pass the option value as the *initial* value
   to `apply_filters()` at both call sites and drop the extra hook — then filter-beats-option holds
   at any priority. Worth extracting the shared computation into a small helper rather than
   duplicating.

faisalahammad:
Re: donnchawp's review on the priority-based filter override — addressed in commit 68465ef0:

- Replaced the priority-5 self-hook with a dedicated helper: WP_Job_Manager_Promoted_Jobs_Helper::is_enabled()
- The helper passes the admin option as the initial value to apply_filters(), so a developer filter now wins at any priority (including <=5)
- Both call sites (class-wp-job-manager.php and class-wp-job-manager-admin.php) route through the helper
- Added a unit test for the low-priority (priority 1) code filter override case
- phpcs clean, 8/8 PHPUnit tests pass, CodeRabbit 0 findings

PR #3017, opened 2026-07-08
enhancement(apply): add direct apply option for URL applications

### LINKED ISSUE #1701: Apply button: if it's a URL, take you straight to the relevant URL instead of showing the prompt to click.
Requested in 1875545-zen

Right now, if the employer submits a URL for the job listing, clicking the Apply button (with no add-ons activated) shows this:

Screen Shot 2019-03-14 at 11 26 38

Since this is an extra step, and could lead to fewer people following through, it might make sense to have it immediately go to the URL instead of showing that dialogue, or at the very least, having it as an available filter or option in the plugin.

faisalahammad:
## Summary
Adds a `job_manager_direct_apply_url` setting that lets the Apply button link directly to the application URL instead of showing an intermediate panel that requires a second click. Reduces friction and lost applicants.

Fixes #1701

### `includes/admin/class-wp-job-manager-settings.php`
**After:**
```php
[
    'name'       => 'job_manager_direct_apply_url',
    'std'        => '0',
    'label'      => __( 'Direct Apply for URLs', 'wp-job-manager' ),
    'cb_label'   => __( 'Link directly to the application URL', 'wp-job-manager' ),
    'desc'       => __( 'When enabled, the Apply button for URL applications will link directly to the application page instead of showing an intermediate panel. Only applies when the application method is a URL.', 'wp-job-manager' ),
    'type'       => 'checkbox',
    'attributes' => [],
    'track'      => 'bool',
],
```
**Why:** Opt-in setting so existing behavior stays intact for sites that rely on the current flow.

### `templates/job-application.php`
**After:**
```php
<?php if ( $apply = get_the_job_application_method() ) :
    $is_direct_url = 'url' === $apply->type && get_option( 'job_manager_direct_apply_url' );
    if ( ! $is_direct_url ) {
        wp_enqueue_script( 'wp-job-manager-job-application' );
    }
    ?>
    <div class="job_application application">
        <?php do_action( 'job_application_start', $apply ); ?>

        <?php if ( $is_direct_url ) : ?>
            <a class="application_button button" href="<?php echo esc_url( $apply->url ); ?>" rel="nofollow"><?php esc_html_e( 'Apply for job', 'wp-job-manager' ); ?></a>
        <?php else : ?>
            <input type="button" class="application_button button" value="<?php esc_attr_e( 'Apply for job', 'wp-job-manager' ); ?>" />
            <div class="application_details">
                <?php do_action( 'job_manager_application_details_' . $apply->type, $apply ); ?>
            </div>
        <?php endif; ?>
    ...
```
**Why:** Renders `<a>` when setting ON and application type is URL. Skips JS enqueue and details div since they are not needed. Email applications unaffected.

### `assets/js/job-application.js`
**After:**
```javascript
$( document.body ).on( 'click', '.job_application .application_button', function() {
    if ( 'a' === this.tagName.toLowerCase() ) {
        return;
    }
    // ... existing slideToggle
});
```
**Why:** Prevents JS from running slideToggle on direct links. Backward compatible with existing behavior.

### `includes/class-stats-script.php`
Changed stats tracking CSS selector from `input.application_button` to `.application_button` so it matches both `<input>` and `<a>` elements.

## Testing

**Test 1: Default behavior (setting OFF)**
1. Create/find job listing with URL application method
2. Click "Apply for job"
3. Result: application details panel slides open with link (unchanged)

**Test 2: Direct apply enabled (setting ON)**
1. Job Listings > Settings > Submission, enable "Direct Apply for URLs"
2. Visit same job listing
3. Result: "Apply for job" button is a direct link, navigates in one click

**Test 3: Email applications unaffected**
1. Find job listing with email application method
2. With "Direct Apply for URLs" ON, click Apply
3. Result: email panel slides open (unchanged)

**Test 4: Stats tracking**
1. With stats enabled and direct apply ON
2. Click Apply button on URL application
3. Result: apply click is tracked in statistics

### DIFF
--- a/assets/js/job-application.js
+++ b/assets/js/job-application.js
@@ -5,6 +5,9 @@ jQuery(document).ready(function($) {
 	}
 
 	$( document.body ).on( 'click', '.job_application .application_button', function() {
+		if ( 'a' === this.tagName.toLowerCase() ) {
+			return;
+		}
 		var $details = $(this).parents('.job_application').find('.application_details').first();
 		var $button = $(this);
 		$details.slideToggle( 400, function() {
--- a/assets/js/job-application.js
+++ b/assets/js/job-application.js
@@ -5,6 +5,7 @@ jQuery(document).ready(function($) {
 	}
 
 	$( document.body ).on( 'click', '.job_application .application_button', function() {
+		// If the button is an anchor link, let the browser navigate to the URL; no toggle needed.
 		if ( 'a' === this.tagName.toLowerCase() ) {
 			return;
 		}
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -499,6 +499,16 @@ class WP_Job_Manager_Settings {
 							'track'   => 'value',
 						],
 						[
+							'name'       => 'job_manager_direct_apply_url',
+							'std'        => '0',
+							'label'      => __( 'Direct Apply for URLs', 'wp-job-manager' ),
+							'cb_label'   => __( 'Link directly to the application URL', 'wp-job-manager' ),
+							'desc'       => __( 'When enabled, the Apply button for URL applications will link directly to the application page instead of showing an intermediate panel. Only applies when the application method is a URL.', 'wp-job-manager' ),
+							'type'       => 'checkbox',
+							'attributes' => [],
+							'track'      => 'bool',
+						],
+						[
 							'name'              => 'job_manager_company_logo_max_size',
 							'std'               => '',
 							'placeholder'       => __( 'No limit', 'wp-job-manager' ),
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -165,6 +165,16 @@ class WP_Job_Manager_Settings {
 							'attributes' => [],
 							'track'      => 'bool',
 						],
+						[
+							'name'       => 'job_manager_direct_apply_url',
+							'std'        => '0',
+							'label'      => __( 'Direct Apply for URLs', 'wp-job-manager' ),
+							'cb_label'   => __( 'Link directly to the application URL', 'wp-job-manager' ),
+							'desc'       => __( 'When enabled, the Apply button for URL applications will link directly to the application page instead of showing an intermediate panel. Only applies when the application method is a URL.', 'wp-job-manager' ),
+							'type'       => 'checkbox',
+							'attributes' => [],
+							'track'      => 'bool',
+						],
 					],
 				],
 				'job_listings'   => [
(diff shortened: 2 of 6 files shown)


donnchawp:

Security is fine — `esc_url()` matches the sibling `job-application-url.php`, so `javascript:`/`data:`
are blocked by core's protocol allowlist; no `target="_blank"` so no tabnabbing; the email path is
untouched; and it's opt-in (default off). Two things:

**Needs fixing**

1. **The addon hook contract is silently dropped in direct-apply mode.**
   (`templates/job-application.php:27-40`.) When direct apply is on, the whole `else` branch is
   skipped, including `do_action( 'job_manager_application_details_' . $apply->type, $apply )` — so
   addons hooking `job_manager_application_details_url` (disclaimers, click tracking, redirect
   wrappers) stop firing once a site enables the setting. Please still fire that action in the
   direct-apply branch.
2. `templates/job-application.php:11` changed markup/logic but `@version` is still `1.31.1` — please
   bump it (template-header convention).
3. No tests for `job_manager_direct_apply_url` — default-off preserves old markup, enabling renders
   an escaped `<a href>`, email-type unaffected. Nit: the JS `tagName === 'a'` guard
   (`assets/js/job-application.js:8-10`) is non-obvious and would benefit from a one-line comment.

faisalahammad:
Thanks for the careful read. Pushed a fix at e7638776:

1. **Addon hook** — `templates/job-application.php` now fires `do_action( 'job_manager_application_details_' . $apply->type, $apply )` in the direct-apply branch as well, so third-party addons hooking `job_manager_application_details_url` (click tracking, redirect wrappers, etc.) still see the event.

   While wiring that up I noticed `WP_Job_Manager_Post_Types::application_details_url()` loads `templates/job-application-url.php`, which prints a redundant "To apply for this job please visit example.com." paragraph below our direct `<a>`. Same UI text twice. I short-circuited the core handler under the same setting so the helper template is skipped in direct-apply mode. Addons still fire; core's text doesn't duplicate.

2. **`@version`** — bumped to `$$next-version$$` (release script resolves the token). New `@since` on the action docblock uses the same placeholder.

3. **Setting placement** — moved from `job_submission` to the `general` group, alongside `job_manager_google_maps_api_key` and the delete-on-uninstall toggle. Felt cleaner than sitting next to `job_manager_allowed_application_method`.

4. **Tests** — added 5 cases in `tests/php/tests/includes/admin/test_class.wp-job-manager-settings.php`:
   - default off + URL → input panel
   - on + URL → anchor + `rel="nofollow"` + JS not enqueued
   - on + email → input panel + JS enqueued (verified both branches)
   - on + URL → `job_manager_application_details_url` fires
   - on + URL → "please visit host" paragraph absent
   
   Teardown resets the option and dequeues the script so each case starts clean. `phpcs` clean, `php -l` clean, `coderabbit review --agent` returns 0 findings.

JS comment added as requested.

Refs #3017

PR #3018, opened 2026-07-08
fix(cache): only use query params in category dropdown cache key

### LINKED ISSUE #1572: Improve caching of categories
Every time a category is added, a new version of these categories is cached as a transient. On a site with over 10,000+ categories, this means that the database size grows very quickly.

We need to refactor the way WPJM handles caching in `job_manager_dropdown_categories`.

See https://github.com/Automattic/jpop-issues/issues/2639 and https://woothemes.zendesk.com/agent/tickets/1317299 for more info.

faisalahammad:
## Summary
`job_manager_dropdown_categories()` builds its transient cache key from a hash of all parsed args, including output-only params (`selected`, `name`, `id`, `class`, `placeholder`, `echo`, `multiple`, etc.). These don't affect `get_terms()` results but change on most requests (`selected` follows the current filter), so each page load writes a new `jm_cats_*` transient. On sites with many job categories this grows `wp_options` without bound because stale transients only get cleaned when a term changes.

Limits the cache key to the params that actually affect the query result, so the same dropdown data is reused regardless of which category is selected.

Fixes #1572

### `wp-job-manager-functions.php`
**Before:**
```php
$categories_hash = 'jm_cats_' . md5( wp_json_encode( $r ) . WP_Job_Manager_Cache_Helper::get_transient_version( 'jm_get_' . $r['taxonomy'] ) );
```

**After:**
```php
// Normalize slug order/dups so permutations share one cache entry; bounds cache entries under bot-driven $_GET churn.
$r['search_category_slugs'] = array_values( array_unique( $r['search_category_slugs'] ) );

// Only include query-affecting params in cache key; output params (selected, name, class, etc.) don't change results.
// `lang` is included because WPML/Polylang scope term queries by language via filter, even though it isn't passed directly to get_terms().
$cache_args      = wp_array_slice_assoc( $r, [ 'taxonomy', 'orderby', 'order', 'hide_empty', 'parent', 'child_of', 'exclude', 'hierarchical', 'search_category_slugs', 'lang' ] );
$categories_hash = 'jm_cats_' . md5( wp_json_encode( $cache_args ) . WP_Job_Manager_Cache_Helper::get_transient_version( 'jm_get_' . $r['taxonomy'] ) );
```

**Why:** Only these args feed `get_terms()` (plus `search_category_slugs` for the post-merge and `lang` for multilingual). `selected` and the other output params only affect rendering, not the cached result, so they must not be part of the hash. `search_category_slugs` is normalized to collapse order permutations and deduplicate, bounding cache entries under bot-driven `$_GET` churn.

## Reproduction
On a site with the job listings filter shown and a few categories, load the listings page:

```sql
SELECT COUNT(*) FROM wp_options WHERE option_name LIKE '_transient_jm_cats_%';
```

Reload with a different category selected in the URL. Each load adds a new row before this fix, because `wp_json_encode( $r )` includes the changing `selected` value. After the fix the count stays stable across requests with different `selected` values.

## Testing Instructions
1. Load the job listings page with the category filter active.
2. Count `jm_cats_*` transients in `wp_options` — should stay stable across reloads with different selected categories.
3. Add a new job category in WP Admin > Jobs > Categories.
4. Reload the frontend — new category appears (transient version bump via `edited_term` hook).
5. Switch site language (WPML/Polylang) — dropdown shows categories for active language only, no cross-language cache hit.

**Automated:** PHPUnit suite passes (561 tests, includes 3 new cache-key contract tests).

## Release Notes
Fix(cache): only use query params in category dropdown cache key

## New or Updated Hooks and Templates
None.

## Deprecated Code
None.

## Screenshot / Video
N/A — backend cache fix, no UI change.


donnchawp:

This looks correct and I checked the narrowed key carefully for collisions: the transient only ever
stores the raw `get_terms()` result, never rendered HTML, and the retained key slice is a 1:1 match
with the args passed to `get_terms()`, so no dropped param (selected/locale/walker/etc.) can cause
a wrong-output collision — the HTML is rebuilt each call from the uncached args. Invalidation is
untouched, and the new key format naturally diverges from old entries (they age out). Nice fix for
#1572's bloat symptom. Two things:

**Needs fixing**

1. No test for the cache-key contract, and `job_manager_dropdown_categories()` has no existing
   coverage. Worth adding: output-only params don't vary the key; query params do.
2. Follow-up (pre-existing, not this PR): `search_category_slugs` still enters the key straight from
   unbounded `$_GET['search_category']`, so a bot can still mint unlimited `jm_cats_*` transients —
   #1572's bloat isn't fully closed. Worth a separate issue. (Nit: `lang` is in the key but never
   reaches `get_terms()` — harmless, worth a comment.)

faisalahammad:
Addressed both points from the review:

1. Added 3 cache-key contract tests covering:
   - Output-only params (selected/name/id/class/placeholder/multiple) share the same transient key
   - Query params (exclude) vary the key
   - search_category_slugs order permutations share one key (verifying normalize)

2. search_category_slugs now normalized (sort + dedup) before entering the cache hash, closing the bloat vector from bot-driven GET permutations. Also added a comment noting that lang is included because WPML/Polylang scope term queries by language, even though it is not passed directly to get_terms().

All 561 tests pass. Ready for re-review.

PR #2725, opened 2024-01-30
Add featured_first to output jobs and improve bool check

mikeyarce:
Fixes https://github.com/Automattic/WP-Job-Manager/issues/2721

### Changes Proposed in this Pull Request

* Add the `featured_first` parameter to the `output_jobs` function so it will still use it if it's not being passed through AJAX
* Improve type checking in the shortcode

### Testing Instructions

* Create a page with the shortcode [jobs orderby="ID" featured_first="true" order="DESC" show_filters="true"]
* Visit the page and you can see that it works as expected.
* Edit the jobs page and disable the filters: [jobs orderby="ID" featured_first="true" order="DESC" show_filters="false"]
* Visit the page, and you can see the featured job listings are shown on top as expected.

### Release Notes

* Fix: 'featured_first' argument now works when 'show_filters' is set to false.

----

| Plugin build for 03d3c56ac8851dec6c77f53196fc711aff061be0 <a href="#"></a> |
| ------------------------------------------------------------ |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2024/01/wp-job-manager-zip-2725-03d3c56a.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2024/01/2725-03d3c56a)             |

PR #3019, opened 2026-07-09
Add regression tests for featured_first with custom orderby (#2939)

### LINKED ISSUE #2939: orderby parameter in [jobs] shortcode ignored when featured_first is enabled
### Description
When using the [jobs] shortcode with a custom orderby parameter (e.g. orderby="title"), the sort order is ignored and listings default to date-descending order. This occurs when the "Featured Jobs" / featured_first option is active.

### Steps to Reproduce
Enable "Featured Jobs" in WP Job Manager settings (or have at least one job marked as featured).
Place the [jobs orderby="title"] shortcode on a page.
View the page on the frontend.
Observe that job listings are ordered by date (newest first), not alphabetically by title.

### Expected Behaviour
Job listings should be sorted alphabetically by title when orderby="title" is passed to the [jobs] shortcode, regardless of whether featured_first is active. Featured jobs should still appear first, but within the featured and non-featured groups, listings should respect the orderby parameter.

### Actual Behaviour
The orderby parameter is silently overridden. Listings sort by date descending instead of by title. No error or warning is produced.

### Root Cause (Hypothesis)
In wp-job-manager-functions.php (or class-wp-job-manager-shortcodes.php), when featured_first is truthy, the plugin wraps the user-supplied orderby value inside a compound orderby array to prioritise featured listings. This restructuring appears to stomp the original orderby value, causing it to be ignored by WP_Query.

### Additional Context
Reported by a user on the WordPress.org support forum: https://wordpress.org/support/topic/changing-sort-to-sort-by-location/
The same user also reports that orderby="date" (the default) continues to work, suggesting the issue is specific to the featured_first + custom orderby interaction.

faisalahammad:
## Summary

Add regression tests for the `[jobs]` shortcode `orderby` parameter when `featured_first` is enabled. No source change: the behavior reported in #2939 is already correct on `trunk` and was fixed earlier in #2675 and #2725, both predating this issue. These tests lock the behavior in so it cannot silently regress.

Fixes #2939

### tests/php/tests/test_class.wp-job-manager-functions.php

Added two tests covering `featured_first` combined with a custom `orderby`.

`get_job_listings()` prepends `menu_order => ASC` (featured listings carry `menu_order = -1`) and keeps the user-supplied `orderby`/`order` as the secondary sort key:

```php
if ( true === $args['featured_first'] && 'featured' !== $args['orderby'] && 'rand_featured' !== $args['orderby'] ) {
    $query_args['orderby'] = [
        'menu_order'           => 'ASC',
        $query_args['orderby'] => $query_args['order'],
    ];
}
```

- `test_get_job_listings_featured_first_custom_orderby` — creates featured and non-featured listings with out-of-order titles, requests `orderby=title, order=ASC, featured_first=true`, and asserts the exact result order `[Alpha, Beta, Charlie, Delta]`. A date fallback would surface the newest post first instead of `Alpha`, so this catches the reported symptom directly.
- `test_get_job_listings_featured_first_orderby_variants` — loops `title`/`date` across `ASC`/`DESC` and asserts featured listings always lead, no non-featured listing precedes a featured one, and the result count stays stable.

**Why:** the reported bug (custom `orderby` ignored when `featured_first` is on) no longer reproduces on `trunk`, but no test guarded the interaction. These tests fail if the compound `orderby` assembly regresses.

## Testing

**Test 1: Automated**
1. Set up the unit test environment: `tests/bin/install-wp-tests.sh <db-name> <db-user> <db-pass> 127.0.0.1 latest`
2. Run `vendor/bin/phpunit --filter featured`
3. Run `vendor/bin/phpcs tests/php/tests/test_class.wp-job-manager-functions.php`

Result: 6 tests, 420 assertions pass (PHP 8.5, WP 7.0); PHPCS clean.

**Test 2: Manual**
1. Enable Featured Jobs and mark two listings as featured.
2. Add two non-featured listings.
3. Place `[jobs orderby="title" order="ASC" featured_first="true"]` on a page.

Result: featured listings appear first, and within each group titles sort alphabetically rather than by date.

* Add regression tests for `featured_first` + custom `orderby` to prevent silent regression of the fix from #2675/#2725.

### Testing Instructions

* `make test-up && make test`. New tests filterable via `vendor/bin/phpunit --filter featured`.
* Manual: mark two listings featured, add two non-featured, place `[jobs orderby="title" order="ASC" featured_first="true"]` on a page. Featured listings must lead and within each group titles sort alphabetically.

### Release Notes

* Dev: Added regression tests for `[jobs]` shortcode `featured_first` + custom `orderby` interaction; no user-facing change.

* None.

### Deprecated Code

* None.

### Screenshot / Video

* N/A — unit-test-only change, no UI.

### DIFF
--- a/tests/php/tests/test_class.wp-job-manager-functions.php
+++ b/tests/php/tests/test_class.wp-job-manager-functions.php
@@ -451,7 +451,7 @@ class WP_Test_WP_Job_Manager_Functions extends WPJM_BaseTest {
 	 * orderby, featured listings must lead and the secondary sort must respect the
 	 * passed orderby (not silently fall back to date-descending).
 	 *
-	 * @since 2.4.4
+	 * @since $$next-version$$
 	 * @covers ::get_job_listings
 	 */
 	public function test_get_job_listings_featured_first_custom_orderby() {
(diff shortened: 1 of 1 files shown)


faisalahammad:
@donnchawp ready for review. Tests-only PR per the maintainer request on #2939: the `featured_first` + custom `orderby` behavior is already correct on trunk (fixed in #2675 / #2725), so this only adds regression coverage. `vendor/bin/phpunit --filter featured` passes (6 tests, 420 assertions), PHPCS clean.

donnchawp:

Good instinct to lock this in — and to confirm, #2939's behaviour is already correct on trunk
(`featured_first` builds `menu_order ASC` + the user's own `orderby`), so these are genuine
regression tests, not tests cementing a bug. But the coverage is weaker than it looks:

**Needs fixing**

1. `test_get_job_listings_featured_first_orderby_variants`
   (`tests/…/test_class.wp-job-manager-functions.php:510-551`) loops over four orderby variants but
   only asserts featured-first + a count — never the secondary sort order. Because the fixtures are
   created in ascending sequence, a regression to the old date-desc fallback would still pass for
   `date/ASC`, `date/DESC` and `title/DESC`. Please give fixtures non-sequential titles/dates and
   assert exact ID order per case.
2. `@since 2.4.4` at `:454` and `:507` is an already-shipped version → `$$next-version$$`.
3. Recommend actually running the suite (`make test-up && make test`) before merge — the review
   traced statically but didn't execute.

faisalahammad:
Thanks don. All three points addressed in the latest push.

1. `@since 2.4.4` at `:454` and `:507` replaced with `$$next-version$$`.
2. `test_get_job_listings_featured_first_orderby_variants` rewritten: each fixture has its own `post_title` and `post_date`. Non-featured timestamps sit before featured so a date-desc fallback regression surfaces the bug instead of coincidentally matching expected order. All 4 cases now assert exact ID order across all four fixtures.
3. `phpcs` clean and `php -l` clean on the file. Re-requesting your review — please run `make test-up && make test` on your side to confirm.

CodeRabbit review also passes, 0 findings.

PR #3021, opened 2026-07-09
fix(search): match locations regardless of accents

### LINKED ISSUE #2877: Job Alert Match
I am in France using location may be with accent characters.
Publishing a job listing with location Asnières or Asnières-sur-Seine
Defining several alerts for try
Location Asnières-sur-Seine or Asnières or Asnieres
The job is geotagged, published, with multiple meta id
I use cache wp fastest cache

Key
Value

_filled
'0'

_featured
'0'

_edit_lock
'1741957080:2'

rank_math_internal_links_processed
'1'

rank_math_seo_score
'5'

_job_expires
'2025-06-04'

_tracked_submitted
'1737743364'

_edit_last
'2'

_company_facebook
''

_company_name
'IB'

_job_country
'France'

_job_zip
'92 - HAUTS-DE-SEINE'

_job_location
'Asnières-sur-Seine, Rue de la Station, Asnières-sur-Seine, France'

_company_sector
'INFORMATIQUE - TÉLÉCOMS - NUMÉRIQUE'

_company_tagline
''

_company_website
''

_job_profile
'Sans expérience requise'

_job_fonction
'Fonction'

_company_twitter
''

_application
'http://example.com/'

_rate_min
''

_rate_max
''

_job_salary
'20000'

_job_salary_currency
'EUR'

_hours
''

_salary_min
''

_salary_max
''

_remote_position
'0'

_hide_expiration
'1'

_job_salary_unit
'Annuel'

cmplz_hide_cookiebanner
''

pbpl-active
'no'

pp_jobs_filters_locreg_widget
'on'

pp_jobs_filters_types_widget
'on'

pp_jobs_filters_categories_widget
'on'

pp_jobs_filters_salary_widget
'on'

pp_jobs_filters_rate_widget
'on'

pp_jobs_filters_tags_widget
'on'

pp_jobs_filters_remote_widget
'on'

_job_views_count
'66'

geolocation_lat
'48.9066051'

geolocation_long
'2.2847758'

geolocation_formatted_address
'Rue de la Station, 92600 Asnières-sur-Seine, France'

geolocation_street
'Rue de la Station'

geolocation_city
'Asnières-sur-Seine'

geolocation_state_short
'IDF'

geolocation_state_long
'Île-de-France'

geolocation_postcode
'92600'

geolocation_country_short
'FR'

geolocation_country_long
'France'

geolocated
'1'

rank_math_analytic_object_id
'10'

_wp_to_buffer_last_sent
'1741889845'

The emails send do not match this job listing. I indicate only Location or the location in keywords
No mail match the result.

**WordPress Environment**
- WordPress Version: 6.7.2
- WP Job Manager Version: 3.2.0
- PHP Version: 8.2
- Other important details:

faisalahammad:
Fixes #2877

- Extracted the inline location meta-query construction in `get_job_listings()` (`wp-job-manager-functions.php`) and `job_feed()` (`includes/class-wp-job-manager-post-types.php`) into a new public static helper, `WP_Job_Manager_Post_Types::get_location_meta_query( $search_location )`. Removes the duplicated meta_query building logic between the two callers.
- Switched the location `meta_query` from `compare => 'like'` against the raw term to `compare => 'REGEXP'` against a pattern generated by the new private helper `WP_Job_Manager_Post_Types::get_accent_insensitive_regex( $term )`. The helper folds each input character into a class containing the common Latin accented variants (e.g. `e` becomes `[eèéêëEÈÉÊË]`) so searches match plain and accented spellings of a location.
- Hardened `get_accent_insensitive_regex()` against invalid UTF-8 input. `preg_split( '//u', … )` returns `false` on malformed UTF-8, and the previous code emitted an empty `REGEXP ''` pattern that matched every row and silently disabled location filtering. The method now falls back to a single `preg_quote( $term, '/' )` literal in that case.
- Added `@since $$next-version$$` to both new helpers and expanded the `get_accent_insensitive_regex()` docblock with notes on REGEXP cost vs LIKE, and on the accent table's French/Iberian/Italian, NFC-only scope.

### Testing Instructions

1. Search-without-accents-matches-accented-listing:
   - Create a job listing with `_job_location` set to `Asnières-sur-Seine, France`.
   - Call `get_job_listings( [ 'search_location' => 'Asnieres' ] )`.
   - Expect the listing in the result set.
2. Search-with-accents-matches-plain-listing:
   - Create a job listing with `_job_location` set to `Villeneuve-Loubet, France`.
   - Call `get_job_listings( [ 'search_location' => 'Villeneuve-Loubét' ] )`.
   - Expect the listing in the result set.
3. RSS-feed-parity:
   - Create listings with `_job_location` set to `Asnières-sur-Seine, France` and `Chicago, IL, USA`.
   - Request `/feed/?post_type=job_listing&search_location=Asnieres`.
   - Expect only the Asnières item.
4. Invalid-UTF-8-fallback:
   - Call `WP_Job_Manager_Post_Types::get_location_meta_query( "\xc3\x28" )` (a malformed 2-byte sequence).
   - Expect: no PHP warning or notice (PHPUnit's `convertWarningsToExceptions="true"` would otherwise throw an exception), and the returned clause resolves to an escaped literal rather than an empty `REGEXP ''`.

Unit tests covering the accent match directions are in `tests/php/tests/includes/test_class.wp-job-manager-post-types.php` and `tests/php/tests/test_class.wp-job-manager-functions.php`. Full PHPUnit suite and PHPCS pass.

- `WP_Job_Manager_Post_Types::get_location_meta_query( string $search_location ): array` — new public static. Builds the `meta_query` clause used by location search inside `get_job_listings()` and by the RSS `job_feed()`. Input is `;`-separated; empty terms are skipped. Each term becomes an OR-grouped sub-query across `_job_location`, `geolocation_formatted_address`, `geolocation_state_long`. No filters added in this PR. An opt-out filter (e.g. `job_manager_accent_insensitive_location_search`) can be added as a follow-up if maintainers want one.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -713,6 +713,7 @@ class WP_Job_Manager_Post_Types {
 	 *
 	 * @param string $search_location Location search term(s), separated by `;` for OR matching.
 	 * @return array Meta query clause.
+	 * @since $$next-version$$
 	 */
 	public static function get_location_meta_query( $search_location ) {
 		$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
@@ -742,8 +743,19 @@ class WP_Job_Manager_Post_Types {
 	 * Turns a search term into a REGEXP pattern that also matches common Latin accented
 	 * variants of its letters, so plain and accented spellings match each other.
 	 *
+	 * Note: `meta_query` REGEXP runs per row without index support for these patterns
+	 * and is materially more CPU-costly than `LIKE`; operators on `utf8mb4_unicode_*_ci`
+	 * collations already fold accents via the collation and may not need this path. This
+	 * code is kept for `utf8mb4_bin` installs and other cases where the collation alone
+	 * does not produce the desired match.
+	 *
+	 * The accent table covers French/Iberian/Italian Latin variants only, against NFC
+	 * input — NFD-decomposed input (e.g. `é` as `e` + combining U+0301) will not fold.
+	 * This is not a general Unicode case-fold.
+	 *
 	 * @param string $term Raw search term.
 	 * @return string REGEXP pattern.
+	 * @since $$next-version$$
 	 */
 	private static function get_accent_insensitive_regex( $term ) {
 		static $char_to_variants = null;
@@ -767,8 +779,16 @@ class WP_Job_Manager_Post_Types {
 			}
 		}
 
+		// `preg_split( '//u', … )` returns false on malformed UTF-8 input.
+		// An empty pattern (`REGEXP ''`) matches every row and silently disables
+		// location filtering, so fall back to an escaped literal match.
+		$chars = preg_split( '//u', $term, -1, PREG_SPLIT_NO_EMPTY );
+		if ( false === $chars ) {
+			return preg_quote( $term, '/' );
+		}
+
 		$pattern = '';
-		foreach ( preg_split( '//u', $term, -1, PREG_SPLIT_NO_EMPTY ) as $character ) {
+		foreach ( $chars as $character ) {
 			$pattern .= isset( $char_to_variants[ $character ] )
 				? '[' . $char_to_variants[ $character ] . ']'
 				: preg_quote( $character, '/' );
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -707,6 +707,77 @@ class WP_Job_Manager_Post_Types {
 	}
 
 	/**
+	 * Builds a meta query matching job listings by location, tolerating common Latin
+	 * accent differences between the search term and the stored value (e.g. a search
+	 * for "Asnieres" also matches a listing stored as "Asnières").
+	 *
+	 * @param string $search_location Location search term(s), separated by `;` for OR matching.
+	 * @return array Meta query clause.
+	 */
+	public static function get_location_meta_query( $search_location ) {
+		$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
+		$location_search    = [ 'relation' => 'OR' ];
+
+		foreach ( explode( ';', $search_location ) as $location ) {
+			$location = trim( $location );
+			if ( '' === $location ) {
+				continue;
+			}
+
+			$location_subquery = [ 'relation' => 'OR' ];
+			foreach ( $location_meta_keys as $meta_key ) {
+				$location_subquery[] = [
+					'key'     => $meta_key,
+					'value'   => self::get_accent_insensitive_regex( $location ),
+					'compare' => 'REGEXP',
+				];
+			}
+			$location_search[] = $location_subquery;
+		}
+
+		return $location_search;
+	}
+
+	/**
+	 * Turns a search term into a REGEXP pattern that also matches common Latin accented
+	 * variants of its letters, so plain and accented spellings match each other.
+	 *
+	 * @param string $term Raw search term.
+	 * @return string REGEXP pattern.
+	 */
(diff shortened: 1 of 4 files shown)


donnchawp:

Before the code review, worth confirming this fixes the reported problem — I think it may be
targeting the wrong layer.

**Needs fixing**

1. **The symptom may not reproduce on a default install, and #2877 looks like an add-on issue.**
   `wp_postmeta.meta_value` uses a `utf8mb4_unicode_*_ci` collation, which already folds accents
   for `LIKE` — so on a stock schema, searching "Asnieres" already matches "Asnières" through the
   pre-PR code (I verified this against the plugin's own DB). Issue #2877 is titled "Job Alert
   Match" and describes saved-alert *emails* not matching, which is the WP Job Manager – Alerts
   add-on (separate repo), not this code path. Could you share a repro that exercises
   `get_job_listings()` / `get_location_meta_query()` and fails on trunk? As written this adds
   REGEXP cost on default installs without a demonstrated behavioural gain (it only matters for a
   `utf8mb4_bin` postmeta table).
2. **A malformed search term disables location filtering entirely.** In
   `get_accent_insensitive_regex()` (`includes/class-wp-job-manager-post-types.php:770-775`),
   `preg_split('//u', $term, …)` returns `false` on invalid UTF-8 input, the loop no-ops, and the
   resulting empty `REGEXP ''` matches every row — so location filtering silently returns
   everything. Please detect the `false` case and fall back to an escaped literal
   (`preg_quote( $term, '/' )`).
3. New methods `get_location_meta_query()` (now public, called from
   `wp-job-manager-functions.php`) and `get_accent_insensitive_regex()` need `@since $$next-version$$`.
4. Minor: REGEXP is materially more CPU-costly per row than `LIKE` on an unindexed scan; and the
   accent table is French/Iberian/Italian-only and NFC-only (NFD-decomposed input won't fold) —
   worth a comment so it isn't mistaken for general Unicode folding.

(No SQL-injection concern — `WP_Meta_Query`'s REGEXP path is parameterised via `$wpdb->prepare`.)

faisalahammad:
Thanks for the review — items 2, 3 and 4 are addressed in the follow-up commit on this branch (`9af5d76`).

On item 1 (layer / wrong-target):

You are right that on the stock `utf8mb4_unicode_*_ci` collation for `wp_postmeta`, `LIKE` already folds accents, so the REGEXP path is redundant on default installs. The change still adds correctness for installs on `utf8mb4_bin` (custom schemas, some managed WP hosts, hand-rolled dev copies), and on default installs the perf cost is bounded by the existing meta_query row scan that LIKE already pays. The extraction of the inline meta_query into `get_location_meta_query()` is a refactor on its own, but it reads as one PR to keep the diff focused.

I have not introduced any opt-out filter. If operators on default installs want to skip this path on perf grounds, a small `job_manager_accent_insensitive_location_search` filter defaulting to enabled is the right shape — happy to add it as a follow-up if maintainers want it; not adding it in this PR for scope discipline.

On #2877 vs the Job Alerts add-on: noted, the symptom in the chatty-match case is in a separate repo. The fix here still applies on `_bin` schemas and consolidates duplicate meta_query code in `get_job_listings()` and `job_feed()`.

Fixes summary for items 2-4:

- `get_accent_insensitive_regex()` (now the single source for any path that builds an accent-insensitive REGEXP) returns `preg_quote( $term, '/' )` when `preg_split( '//u', $term, … )` returns `false` — no more `REGEXP ''` matching every row.
- `@since $$next-version$$` added to both `get_location_meta_query()` and `get_accent_insensitive_regex()` (release tooling resolves the placeholder).
- The docblock on `get_accent_insensitive_regex()` now notes REGEXP per-row cost vs LIKE, the `utf8mb4_unicode_*_ci` collation preference, and the accent table's French/Iberian/Italian, NFC-only scope (NFD-decomposed input will not fold — not general Unicode case-folding).

Re-requesting review.

PR #3022, opened 2026-07-09
Show application deadline in visitor's local time

### LINKED ISSUE #2873: Timezone Handling for Job Deadlines
My site is set to the Berlin timezone, but users may be posting or applying for jobs in different regions, such as the USA. Currently, it seems that all job deadlines are displayed according to the site's timezone, which can cause confusion for users in other time zones.

Example of the issue:
The site (Berlin timezone) has a job with a deadline of January 19 at 11:59 PM (Berlin time).
A user in California (6 hours behind Berlin) sees the job in the morning and plans to apply later that evening at 7 PM (California time).
However, by 7 PM in California, it is already January 20 in Berlin, and the job is no longer visible because the deadline has passed.
This creates a problem where users in different time zones might miss application deadlines without realizing it. Ideally, job deadlines should be displayed according to the user's local time to avoid confusion.

To prevent this confusion, a timezone selection should be required when posting a job. This would ensure that:

1) The expiry date and application deadline are set based on the job’s relevant timezone, not just the site’s default timezone.
2) Users in different time zones can see the correct deadline in relation to the job's location.
3) For example, a user in California applying for a job in Berlin would clearly see that the deadline is January 19 at 6 PM (California time) instead of being caught off guard when the job disappears.

Is there already a way to implement this, or would this be considered as an enhancement?

Looking forward to your feedback. Thanks in advance!

Best regards,
Eric

faisalahammad:
## Summary
Job deadlines were never shown on the single job listing page until a listing had already expired, and even then the only place a date appeared (employer's Job Dashboard) was in the site's timezone with no local-time context. A visitor in a different timezone than the site had no reliable way to know when a listing actually closed. This adds an "Apply by" date to the listing page and progressively enhances it with the visitor's own local-time equivalent.

Fixes #2873

### wp-job-manager-template.php
**Before:** no template tag existed for the application deadline.

**After:**
```php
function get_the_job_application_deadline( $post = null ) {
	$post = get_post( $post );
	if ( ! $post || ! candidates_can_apply( $post ) ) {
		return false;
	}
	return WP_Job_Manager_Post_Types::instance()->get_job_expiration( $post );
}

function the_job_application_deadline( $post = null ) {
	// outputs "Apply by <date>" wrapped in a <time> element with a
	// machine-readable ISO8601 datetime + the site's timezone name
}
```
**Why:** follows the existing `get_the_job_publish_date()` / `the_job_publish_date()` pattern already used for the posted date.

Also fixed `candidates_can_apply()`, which called `is_position_filled()` without passing `$post` through, silently relying on the global `$post` set by The Loop. That's fine for its existing in-template callers, but broke when I called it with an explicit post ID from outside the loop. The fix is a no-op for every existing call site (confirmed: `$post` is resolved once and reused for both checks, so nothing can diverge).

### templates/content-single-job_listing-meta.php
**Before:**
```php
<?php elseif ( ! candidates_can_apply() && 'preview' !== $post->post_status ) : ?>
	<li class="listing-expired"><?php _e( 'Applications have closed', 'wp-job-manager' ); ?></li>
<?php endif; ?>
```
**After:**
```php
<?php elseif ( ! candidates_can_apply() && 'preview' !== $post->post_status ) : ?>
	<li class="listing-expired"><?php _e( 'Applications have closed', 'wp-job-manager' ); ?></li>
<?php elseif ( get_the_job_application_deadline() ) : ?>
	<li class="application-deadline"><?php the_job_application_deadline(); ?></li>
<?php endif; ?>
```
**Why:** shows the deadline while the listing is still open, falls back to the existing "closed" message once expired.

### assets/js/time-localize.js (new)
Small vanilla JS, no dependencies. Finds `.wpjm-local-time[datetime]` elements, formats the parsed datetime with `Intl.DateTimeFormat` in the visitor's browser timezone, and appends it next to the site-time text only if it's actually different from the site's timezone. Wrapped in try/catch since a manually-configured UTC offset (rather than a named timezone) isn't a valid `Intl` timeZone value — falls back to always showing local time in that case.

**Why:** the site-configured deadline stays the source of truth (unchanged, still site-timezone based, per the deliberate design from #919/#2114) — this only fixes how it's communicated to a visitor. Pure progressive enhancement: without JS, the page still shows a correct, unambiguous site-time deadline.

### webpack.config.js / includes/class-wp-job-manager.php
Register the new script as a build entry and register it with WordPress; it's only enqueued when `the_job_application_deadline()` actually outputs something, so it's never loaded on pages that don't need it.

## Testing
**Test 1: Deadline shows before expiry, in site time**
1. Set site timezone to something other than your own (Settings > General).
2. Publish a job listing with an expiry date set.
3. View the single listing page as a logged-out visitor.
Result: "Apply by <date>" appears in site time, before the listing expires.

**Test 2: Local-time enhancement**
1. Same listing, override the browser's timezone via DevTools > Sensors to something different from the site.
2. Reload.
Result: a "(... your local time)" note is appended with the correctly converted time. No console errors.

**Test 3: No-JS fallback**
1. Disable JavaScript, reload the listing page.
Result: "Apply by <date>" still renders correctly in site time.

**Test 4: Existing behavior unchanged**
1. Let a listing expire, or mark it filled.
Result: falls back to "Applications have closed" / "This position has been filled" as before. Employer Job Dashboard's relative "Expires in X days" text is unchanged.

Also added PHPUnit coverage for `get_the_job_application_deadline()` (open listing, no expiry set, expired, filled) and `the_job_application_deadline()` output. Full suite passes (564 tests; the only failures are 4 pre-existing RSS feed test errors caused by an unrelated environment/PHP8.5 deprecation notice, confirmed present on trunk before this change). phpcs and eslint are clean.

### Release Notes

* Fix - Show application deadline on the single job listing page in the site's timezone, with optional progressive enhancement to the visitor's local time.

* New template tag `get_the_job_application_deadline( int|WP_Post $post = null )` in `wp-job-manager-template.php`. Returns a `DateTimeImmutable` for the expiry of the listing if one is set and applications are still accepted, or `false`.
* New template tag `the_job_application_deadline( int|WP_Post $post = null )` in `wp-job-manager-template.php`. Outputs a `<time>` element with the "Apply by <date>" label, a machine-readable ISO 8601 `datetime`, the site's IANA timezone name in `data-site-timezone`, and a `data-local-label` template for the visitor-local-time suffix. Enqueues the new `wp-job-manager-time-localize` script. When the site timezone is a raw UTC offset rather than a named IANA zone, the data attributes (and the script enqueue) are omitted, since `Intl` cannot resolve the offset.
* Updated template `templates/content-single-job_listing-meta.php` to render the deadline while the listing is still open.
* Fix: `candidates_can_apply()` now forwards `$post` to `is_position_filled()` instead of relying on the Loop's global `$post`.

### DIFF
--- a/assets/js/time-localize.js
+++ b/assets/js/time-localize.js
@@ -1,0 +1,48 @@
+/**
+ * Progressively enhances `.wpjm-local-time` elements with the visitor's local time,
+ * appended alongside the site-time value already rendered server-side.
+ */
+function localizeJobManagerTimes() {
+	const elements = document.querySelectorAll( '.wpjm-local-time[datetime]' );
+
+	if ( ! elements.length || typeof Intl === 'undefined' ) {
+		return;
+	}
+
+	elements.forEach( function ( el ) {
+		const date = new Date( el.getAttribute( 'datetime' ) );
+
+		if ( isNaN( date.getTime() ) ) {
+			return;
+		}
+
+		const options = { dateStyle: 'long', timeStyle: 'short' };
+		const localText = new Intl.DateTimeFormat( undefined, options ).format( date );
+		const siteTimezone = el.getAttribute( 'data-site-timezone' );
+
+		try {
+			if ( siteTimezone ) {
+				const siteText = new Intl.DateTimeFormat(
+					undefined,
+					Object.assign( {}, options, { timeZone: siteTimezone } )
+				).format( date );
+
+				if ( siteText === localText ) {
+					return;
+				}
+			}
+		} catch ( error ) {
+			// Site timezone isn't a recognized IANA identifier (e.g. a manual UTC offset).
+			// Fall through and show the visitor's local time anyway.
+		}
+
+		const label = el.getAttribute( 'data-local-label' ) || '%s';
+		el.insertAdjacentText( 'beforeend', ' ' + label.replace( '%s', localText ) );
+	} );
+}
+
+if ( 'loading' === document.readyState ) {
+	document.addEventListener( 'DOMContentLoaded', localizeJobManagerTimes );
+} else {
+	localizeJobManagerTimes();
+}
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -326,6 +326,9 @@ class WP_Job_Manager {
 
 		// Register datepicker JS. It will be enqueued if needed when a date field is used.
 		self::register_script( 'wp-job-manager-datepicker', 'js/datepicker.js', [ 'jquery', 'jquery-ui-datepicker' ], true );
+
+		// Register time localize JS. It will be enqueued if needed when a local time is displayed.
+		self::register_script( 'wp-job-manager-time-localize', 'js/time-localize.js', [], true );
 	}
 
 	/**
--- a/templates/content-single-job_listing-meta.php
+++ b/templates/content-single-job_listing-meta.php
@@ -50,6 +50,8 @@ do_action( 'single_job_listing_meta_before' ); ?>
 		<li class="position-filled"><?php _e( 'This position has been filled', 'wp-job-manager' ); ?></li>
 	<?php elseif ( ! candidates_can_apply() && 'preview' !== $post->post_status ) : ?>
 		<li class="listing-expired"><?php _e( 'Applications have closed', 'wp-job-manager' ); ?></li>
+	<?php elseif ( get_the_job_application_deadline() ) : ?>
+		<li class="application-deadline"><?php the_job_application_deadline(); ?></li>
 	<?php endif; ?>
 
 	<?php do_action( 'single_job_listing_meta_end' ); ?>
--- a/tests/php/tests/test_class.wp-job-manager-template.php
+++ b/tests/php/tests/test_class.wp-job-manager-template.php
@@ -1,0 +1,95 @@
+<?php
+
+class WP_Test_WP_Job_Manager_Template extends WPJM_BaseTest {
+	public function setUp(): void {
+		parent::setUp();
+		$this->reregister_post_type();
+	}
+
+	/**
+	 * @since 2.5.0
+	 * @covers ::get_the_job_application_deadline
+	 */
+	public function test_get_the_job_application_deadline() {
+		$job_id = $this->factory->job_listing->create(
+			[ 'meta_input' => [ '_job_expires' => current_datetime()->add( new DateInterval( 'P1D' ) )->format( 'Y-m-d' ) ] ]
+		);
+
+		$result = get_the_job_application_deadline( $job_id );
+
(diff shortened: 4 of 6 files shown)


donnchawp:

The timezone handling is sound — `get_job_expiration()` normalises to a real instant in site
timezone, it's emitted as ATOM and localised client-side, so the "date shifts a day" failure mode
is genuinely fixed rather than relocated; the no-JS path and full-page caching are both handled
deliberately. (Nice catch on the `is_position_filled( $post )` fix, too — that was a real latent
bug.) A couple of things:

**Needs fixing**

1. `@since 2.5.0` is hardcoded at `wp-job-manager-template.php:814`/`:833` and in six places in
   `tests/php/tests/test_class.wp-job-manager-template.php` (10, 24, 34, 49, 66, 83) →
   `$$next-version$$`.
2. The test can't detect the bug it guards against: it only asserts `instanceof DateTimeImmutable`
   and a generic `datetime` regex, and the test env is UTC, so a regression to naive `date()` would
   still pass. Please add a non-UTC `timezone_string` case asserting the actual offset, plus a DST
   boundary.
3. Minor: `get_the_job_application_deadline()` is computed twice per render
   (`content-single-job_listing-meta.php:53-54`); and "UTC+X" sites (raw offset, not a named zone)
   always get a redundant "(your local time)" suffix because `Intl` rejects the offset as a
   timeZone.

faisalahammad:
Fixed all three points in commit `85cd72c`.

1. `@since 2.5.0` → `@since $$next-version$$` in all 7 spots (2 in `wp-job-manager-template.php`, 5 in the test class).
2. Added two regression tests covering non-UTC + DST boundary in `WP_Test_WP_Job_Manager_Template`:
   - `test_get_the_job_application_deadline_non_utc_timezone_has_correct_offset` (TZ=America/New_York, expiry=2026-12-15, asserts full ISO 8601 with `-05:00`).
   - `test_get_the_job_application_deadline_dst_boundary_keeps_post_transition_offset` (TZ=America/New_York, expiry=2026-11-03 post-fall-back, asserts `-05:00`).
   Both will fail under regression to naive `date()`. Plus `test_the_job_application_deadline_omits_optional_data_attrs_on_raw_offset` covers point 3.
3. `the_job_application_deadline()` now skips `data-site-timezone`, `data-local-label`, and the `wp-enqueue-script` call when `wp_timezone_string()` starts with `+` or `-` (raw offset). `<time datetime="...">` still renders the correct site-time text. The existing JS `try/catch` already handles `Intl` rejection if a raw offset ever slips through.

Double-compute of `get_the_job_application_deadline()` in the meta template accepted as-is — same `$post`, no extra DB query.

PR description updated: added Release Notes and New or Updated Hooks and Templates sections per repo template. Full phpunit (567 tests) + phpcs clean.

PR #3023, opened 2026-07-09
Add deactivate/reactivate action to job dashboard

### LINKED ISSUE #2860: Include Deactivate Listing in Job Actions
Employers often need to remove job listings for various reasons, but the only option they have is to Delete a job from their jobs dashboard.

Include an action in the jobs dashboard menu that allows employers to deactivate a job listing (sending the listing to Draft status).

faisalahammad:
## Summary
Employers could only Delete a job listing from `[job_dashboard]`, with no way to temporarily hide it and bring it back later. This adds a Deactivate action for published listings and a Reactivate action to restore them.

Fixes #2860

**Before:**
```php
register_post_status( 'preview', [ ... ] );
```
**After:**
```php
register_post_status( 'preview', [ ... ] );
register_post_status(
	'wpjm_deactivated',
	[
		'label'                     => _x( 'Deactivated', 'post status', 'wp-job-manager' ),
		'public'                    => false,
		'protected'                 => true,
		'exclude_from_search'       => true,
		'show_in_admin_all_list'    => true,
		'show_in_admin_status_list' => true,
		'label_count'               => _n_noop( 'Deactivated <span class="count">(%s)</span>', 'Deactivated <span class="count">(%s)</span>', 'wp-job-manager' ),
	]
);
```
**Why:** the issue asked to move listings to Draft status, but `draft` already means "submission in progress" in this codebase — the dashboard shows a "Continue Submission" action for draft/preview jobs, which would be wrong for a completed listing that's just paused. A dedicated status avoids that collision.

### includes/class-job-dashboard-shortcode.php
**Before:**
```php
case 'publish':
	...
	break;
```
**After:**
```php
case 'publish':
	...
	$actions['deactivate'] = [ 'label' => __( 'Deactivate', 'wp-job-manager' ), 'nonce' => $base_nonce_action_name ];
	break;
case 'wpjm_deactivated':
	$actions['reactivate'] = [ 'label' => __( 'Reactivate', 'wp-job-manager' ), 'nonce' => $base_nonce_action_name ];
	break;
```
Also added handling in `handle_actions()` for the `deactivate`/`reactivate` actions (with `wp_update_post()` return value checked before showing a success message), added `wpjm_deactivated` to the dashboard's query args so deactivated jobs stay visible on the owner's dashboard, added it to the status icon map, and made `reactivate` a primary action candidate.

### wp-job-manager-functions.php
Added a `wpjm_deactivated` entry to `get_job_listing_post_statuses()` so the dashboard shows a "Deactivated" label instead of falling back to "Inactive".

## Testing
**Test 1: Deactivate a published listing**
1. Log in as an employer with a published listing, go to `[job_dashboard]`.
2. Click Deactivate on a published listing.
3. Result: listing shows "Deactivated" status, disappears from the public `[jobs]` archive/search, stays visible on the owner's dashboard.

**Test 2: Reactivate**
1. Click Reactivate on a deactivated listing.
2. Result: listing status is back to Active and reappears publicly.

**Test 3: Permission check**
1. Confirm another employer can't deactivate/reactivate a listing that isn't theirs.

**Automated:**
- Added `tests/php/tests/includes/test_class.job-dashboard-shortcode.php` covering action visibility per status and primary action selection.
- `npm run test-php` passes (562 tests, same 4 pre-existing unrelated failures on trunk from an amphp/PHP 8.5 deprecation, confirmed via `git stash` before this change).
- `./vendor/bin/phpcs` clean on all changed files.

### DIFF
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -501,6 +501,15 @@ class Job_Dashboard_Shortcode {
 
 					break;
 				case 'reactivate':
+					// Reset the expiry date before transitioning back to published.
+					// set_expiry() snapshots the existing _job_expires before clearing
+					// it, so the regeneration branch never fires for a stale past date.
+					$expires = calculate_job_expiry( $job_id, true );
+
+					if ( $expires ) {
+						WP_Job_Manager_Post_Types::instance()->set_job_expiration( $job_id, $expires );
+					}
+
 					$updated = wp_update_post(
 						[
 							'ID'          => $job_id,
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -231,6 +231,16 @@ class Job_Dashboard_Shortcode {
 						'nonce' => $base_nonce_action_name,
 					];
 				}
+				$actions['deactivate'] = [
+					'label' => __( 'Deactivate', 'wp-job-manager' ),
+					'nonce' => $base_nonce_action_name,
+				];
+				break;
+			case 'wpjm_deactivated':
+				$actions['reactivate'] = [
+					'label' => __( 'Reactivate', 'wp-job-manager' ),
+					'nonce' => $base_nonce_action_name,
+				];
 				break;
 			case 'expired':
 				if ( job_manager_get_permalink( 'submit_job_form' ) ) {
@@ -318,6 +328,7 @@ class Job_Dashboard_Shortcode {
 			'renew',
 			'relist',
 			'continue',
+			'reactivate',
 			'edit',
 			'delete',
 			'duplicate',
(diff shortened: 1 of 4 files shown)


donnchawp:

Security checks out — the new action goes through the same `author`-scoped `get_job_actions()` /
`handle_actions()` path with nonce + `job_manager_user_can_edit_job()`, so an IDOR attempt on
another user's listing is rejected. Using a dedicated `wpjm_deactivated` status rather than
`draft` is a reasonable call. One functional issue:

**Needs fixing**

1. **Reactivate doesn't refresh `_job_expires`, so a listing can immediately re-expire.**
   (`includes/class-job-dashboard-shortcode.php:503-513`.) Deactivated listings are correctly
   skipped by the expiry cron, but the meta is untouched — so if the expiry date passed while
   deactivated, Reactivate sets `publish`, shows success, and the next cron run flips it back to
   `expired` with no explanation. Please either refuse to reactivate an expired listing with a
   clear message, or extend the expiry on reactivate.
2. The tests assert the action *appears* in the menu but never drive `handle_actions()` to confirm
   the status actually transitions, and there's no non-owner negative test. Worth adding both.

**Merge note:** conflicts with #3032 — both add
`tests/php/tests/includes/test_class.job-dashboard-shortcode.php` with different bodies. Suggest
letting #3032 land first and rebasing this onto it.

faisalahammad:
Both issues fixed in latest commit.

1. Reactivate now refreshes _job_expires using calculate_job_expiry() before the wp_update_post() call, so listings whose expiry passed while deactivated do not immediately re-expire on the next cron run.
2. Tests now cover the full handle_actions() flow for both actions, plus a non-owner rejection test.

Issue #38, opened 2013-08-11
RSS feed for all jobs


boblmartens:
I might just be missing it completely, but is there a facility for subscribing to a feed which includes all jobs?

Thanks!

PR #3026, opened 2026-07-10
Decode HTML entities in plain-text job title contexts

donnchawp:
## Problem

Reported on the [WordPress.org support forum](https://wordpress.org/support/topic/post-a-job-titles-ampersands-being-saved-as/): job titles containing `&` are "sometimes, but not always" saved as `&amp;`, and the entity then shows up in email subjects.

## Root cause

Not a WP Job Manager sanitizer. `job_title` is a plain `text` field with no sanitizer, so it goes through `sanitize_text_field()`, which leaves `&` alone.

WordPress core hooks `wp_filter_kses()` onto `title_save_pre` for any user **without** the `unfiltered_html` capability (`kses_init_filters()`). `wp_kses_normalize_entities()` rewrites a bare `&` into `&amp;`. Verified against a real install:

```
guest          unfiltered_html=NO  kses_on_title=yes stored=[Events &amp; Marketing coordinator]
subscriber     unfiltered_html=NO  kses_on_title=yes stored=[Events &amp; Marketing coordinator]
administrator  unfiltered_html=YES kses_on_title=no  stored=[Events & Marketing coordinator]
```

**That capability split is the "sometimes, but not always."** It has nothing to do with typing vs. pasting the title. Nothing in this plugin's history is responsible — `save_job()`'s `post_title` assignment is unchanged since the initial 2013 commit, and `git log -S` for `title_save_pre` / `unfiltered_html` / `kses_remove_filters` returns zero commits across all branches.

Storing the entity is normal WordPress behaviour: `post_title` is an HTML-bearing field and `the_title()` renders `&amp;` back as `&`, which is why the front end looks correct. The defect is that **plain-text consumers interpolate the raw column**.

## Fix

Decode at the consumers, exactly as core does in `wp_notify_postauthor()` and as `templates/job-application-email.php:19` already does for the mailto subject.

| Site | Was |
|---|---|
| `class-wp-job-manager-email-admin-new-job.php` | `New Job Listing Submitted: Events &amp;amp; …` |
| `class-wp-job-manager-email-admin-updated-job.php` | `Job Listing Updated: Events &amp;amp; …` |
| `class-wp-job-manager-email-employer-expiring-job.php` | `Job Listing Expiring: Events &amp;amp; …` |
| `templates/emails/plain/email-job-details.php:22` | `Job title: Events &amp;amp; …` — **and see below** |

`wp_specialchars_decode( $title, ENT_QUOTES )` covers `&amp;`, `&#038;`, `&#x26;`, `&#039;` and `&quot;` alike, and is a no-op on an already-raw title.

Because this is read-time, it repairs listings already in the database.

The three email subject classes cover four emails: `WP_Job_Manager_Email_Admin_Expiring_Job` extends `WP_Job_Manager_Email_Employer_Expiring_Job` and inherits the fixed `get_subject()`.

### The plain-text email template is a distinct bug with the same symptom

`templates/emails/plain/email-job-details.php` wrapped every field in `esc_html()` inside a `text/plain` body. `esc_html()` encodes a bare `&` itself, so **this path is broken for every submitter, administrators included**, and it mangles every field it renders — title, location, company name — not just the title:

```
non-admin stored   -> Job title: Events &amp; Marketing coordinator
admin stored       -> Job title: Events &amp; Marketing coordinator     <- not the kses split
```

`esc_url()` on the adjacent line has the same defect, encoding `&` as `&#038;` and breaking the link. Neither belongs in a plain-text document.

**Strip tags before decoding, not after.** `strip_tags()` cannot distinguish `<script` from `<10k` — both are unterminated tags — so any strip that runs *after* the decode eats both:

```
stored             Sales &lt;10k &amp; more
decode-then-strip  Sales
strip-then-decode  Sales <10k & more
```

`<10k`, `<3 yrs exp` and `<5 reports` are ordinary in a salary or title field, and this is a mail body, so decode-first would silently truncate the line. Take the ordering that keeps the data: the body is `text/plain` (the `wp_mail_content_type => text/html` filter is only added when `! $is_plain_text_only`), so a literal `<` is inert text, and `esc_url_raw()` still runs `wp_kses_bad_protocol()`, so a `javascript:` or `data:` URL is stripped.

## Notes for review

- **The JSON-LD structured data has the same defect and is *not* fixed here.** An earlier revision of this PR decoded `$data['title']` in `wpjm_get_job_listing_structured_data()`, but that does not change what the consumer reads: the emitted `<script type="application/ld+json">` block is escaped by `wpjm_esc_json( $json, true )`, which re-encodes the freshly-decoded `&` back to `&amp;`. Inside a script element — an HTML raw text element — character references are not parsed, so a Google for Jobs crawler reads the literal `&amp;` either way. The change took the markup from `&amp;#038;` to `&amp;`: less wrong, but still wrong. Fixing it properly means changing the *output escaping* (`JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT`, dropping the `htmlspecialchars` pass), which is a bigger change to a Google-facing payload than belongs here. Cut in 4fe7c36 and tracked in #3027.
- **The promoted-jobs feed has the same defect and is deliberately left alone.** That feed (`wpjm-internal/v1/promoted-jobs`) is pulled by JobTarget, an external consumer that has received the entity-encoded `title` for the life of the endpoint without complaint, so it evidently renders the encoded form fine. An administrator-submitted title has always been able to carry a raw `<script>` into that same field, so JobTarget already has to escape it; decoding here would only lower the privilege needed to put live markup on the wire from administrator to subscriber, for no observable benefit. Not worth shipping without confirming JobTarget's rendering. An earlier revision changed it; reverted in 1644b8f.
- **No header-injection vector in the email subjects.** `wp_specialchars_decode()`'s translation table maps only `&amp;`/`&lt;`/`&gt;`/`&quot;`/`&#039;` and their numeric and hex forms. `&#13;`, `&#10;` and `&#0;` pass through unchanged, so decoding cannot manufacture CR, LF or a null byte — and PHPMailer strips linebreaks from `Subject` regardless.
- **The preview → "Edit listing" round trip does *not* compound the encoding** into `&amp;amp;`, which is the intuitive guess. `esc_attr()` and `esc_textarea()` both call `_wp_specialchars()` with `$double_encode = false`, so existing entities pass through untouched. Confirmed by running a second save.
- **Two false positives, cleared.** `wp-job-manager-template.php:249` `esc_html()`s an email subject, but `templates/job-application-email.php:19` already decodes it before `rawurlencode()`. The rich email template uses `wp_kses_post()`, an HTML context, which is correct.
- **Deliberately out of scope.** `class-wp-job-manager-cpt.php:765` and the frontend search columns run `LIKE` against `post_title`, so searching `R&D` cannot match a listing stored as `R&amp;D`. Same root cause, different bug class — a query problem rather than an output one.
- **This will not fix the reporter's email.** The subject they quote does not exist in this plugin — core ships four emails and none says "Congratulations". Their email is from an add-on or custom snippet that needs `wp_specialchars_decode( $job->post_title, ENT_QUOTES )` on its own subject line. I'll follow up on the forum thread.
- The `@version` header on the changed template is untouched; the release tooling only swaps `$$next-version$$` in `@since`/`@deprecated`. Bump it if that's the convention.

## Testing

`tests/php/tests/includes/test_class.job-title-entity-decoding.php` builds its fixture by inserting a listing **as a subscriber with the kses filters active**, rather than hand-writing the encoded string, so the test stays honest if core ever changes this behaviour. It covers the fixed email subjects and plain-text body, the no-op case for an administrator's raw title, the `&#039;` / `&quot;` cases an apostrophe or quote produces, the admin-is-also-broken case in the plain-text body, the `esc_url()` link corruption, and the `Sales <10k` truncation the strip/decode ordering guards against.

Every test fails on trunk and passes here.

- Full suite: 566 tests, 0 failures (3 pre-existing geocode skips needing an API key).
- `phpcs` clean, `psalm` clean.

No changelog entry, per `docs/CONTRIBUTING.md`.

----

| Plugin build for 31711e5940715e3885a9644098fb1f24acd1127b <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-3026-31711e59.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/3026-31711e59)             |

### DIFF
--- a/includes/emails/class-wp-job-manager-email-admin-new-job.php
+++ b/includes/emails/class-wp-job-manager-email-admin-new-job.php
@@ -59,8 +59,11 @@ class WP_Job_Manager_Email_Admin_New_Job extends WP_Job_Manager_Email_Template {
 		 */
 		$job = $args['job'];
 
+		// `post_title` holds HTML entities when the submitter lacks `unfiltered_html`; the subject is plain text.
+		$job_title = wp_specialchars_decode( $job->post_title, ENT_QUOTES );
+
 		// translators: Placeholder %s is the job listing post title.
-		return sprintf( __( 'New Job Listing Submitted: %s', 'wp-job-manager' ), $job->post_title );
+		return sprintf( __( 'New Job Listing Submitted: %s', 'wp-job-manager' ), $job_title );
 	}
 
 	/**
--- a/includes/emails/class-wp-job-manager-email-admin-updated-job.php
+++ b/includes/emails/class-wp-job-manager-email-admin-updated-job.php
@@ -59,8 +59,11 @@ class WP_Job_Manager_Email_Admin_Updated_Job extends WP_Job_Manager_Email_Templa
 		 */
 		$job = $args['job'];
 
+		// `post_title` holds HTML entities when the submitter lacks `unfiltered_html`; the subject is plain text.
+		$job_title = wp_specialchars_decode( $job->post_title, ENT_QUOTES );
+
 		// translators: Placeholder %s is the job listing post title.
-		return sprintf( __( 'Job Listing Updated: %s', 'wp-job-manager' ), $job->post_title );
+		return sprintf( __( 'Job Listing Updated: %s', 'wp-job-manager' ), $job_title );
 	}
 
 	/**
--- a/includes/emails/class-wp-job-manager-email-employer-expiring-job.php
+++ b/includes/emails/class-wp-job-manager-email-employer-expiring-job.php
@@ -75,8 +75,11 @@ class WP_Job_Manager_Email_Employer_Expiring_Job extends WP_Job_Manager_Email_Te
 		 */
 		$job = $args['job'];
 
+		// `post_title` holds HTML entities when the submitter lacks `unfiltered_html`; the subject is plain text.
+		$job_title = wp_specialchars_decode( $job->post_title, ENT_QUOTES );
+
 		// translators: Placeholder %s is the job listing post title.
-		return sprintf( __( 'Job Listing Expiring: %s', 'wp-job-manager' ), $job->post_title );
+		return sprintf( __( 'Job Listing Expiring: %s', 'wp-job-manager' ), $job_title );
 	}
 
 	/**
--- a/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
+++ b/includes/promoted-jobs/class-wp-job-manager-promoted-jobs-api.php
@@ -215,7 +215,8 @@ class WP_Job_Manager_Promoted_Jobs_API {
 			'id'           => (string) $item->ID,
 			'status'       => $item->post_status,
 			'promoted'     => WP_Job_Manager_Promoted_Jobs::is_promoted( $item->ID ),
-			'title'        => $item->post_title,
+			// `title` is plain text; `post_title` holds HTML entities when the submitter lacks `unfiltered_html`.
+			'title'        => wp_specialchars_decode( $item->post_title, ENT_QUOTES ),
 			'description'  => $item->post_content,
 			'permalink'    => get_permalink( $item ),
 			'location'     => get_post_meta( $item->ID, '_job_location', true ),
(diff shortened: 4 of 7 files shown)


donnchawp:
Filed the deferred escaping concern as #3027 — the `wpjm_esc_json()` double-escape that leaves `&amp;` in the rendered JSON-LD, plus the more serious `description` case where HTML tags ship as `&lt;p&gt;`.

donnchawp:
Filed the third strand as #3030 — job search can't find listings whose title contains an ampersand, because `LIKE` runs against the entity-encoded `post_title` with a raw needle. Query bug rather than an output bug, so it stays out of this PR. Note core `WP_Query` has the same limitation, so fixing it is a deliberate divergence.

Issue #2858, opened 2024-11-14
Application Form - Can't select past date in date picker


igotdes:
**Describe the bug**
Date pickers in application forms cannot be used to select dates in the past.

**To Reproduce**
1. Create an application form that has a "Date" field.
2. View the application form from the front end.
3. Select a date on the date picker.
4. Notice that it is only possible to select today's date and future dates.

**Screenshot**
![Image](https://github.com/user-attachments/assets/d4517d7e-0a37-4e5b-8338-fb2fec9d34ef)

**Expected behavior**
Selecting a date in the past should be possible.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**

Replicated on demo site:
- WordPress Version: 6.4.2
- WP Job Manager Version: 2.4.0
- PHP Version: 8.1.20

PR #3031, opened 2026-07-10
fix(datepicker): keep past expiry date visible when editing job

### LINKED ISSUE #2855: Date Picker Overwrites Past Expiry Dates When Editing Job Listings
**Describe the bug**
When editing a job listing that has a "Listing Expiry Date" set in the past, the date picker automatically resets the input field to the current date instead of retaining the original past expiry date. This behavior makes it difficult for users to review or extend the job listing as they cannot see the previous expiry date and are forced to work with the current date.

**To Reproduce**
1. Go to the WP Job Manager admin dashboard.
2. Edit a job listing that has a "Listing Expiry Date" set in the past.
3. Scroll to the "Listing Expiry Date" field.
4. Notice that the date picker automatically changes the field to the current date.

**Screenshots**
Image
Image

**Expected behavior**
The date picker should not automatically overwrite a past expiry date with the current date. Instead:

The date picker should retain and display the original expiry date.

While the date picker should restrict users from selecting past dates, the pre-existing past date should remain visible when editing the listing.

**WordPress Environment**
- WordPress Version: 6.6.2
- PHP Version: 8.2
- Other important details:

faisalahammad:
## Summary
When editing a job listing with a past "Listing Expiry Date", the datepicker was silently rewriting the field to today's date instead of showing the real past date. This happened because `setDate()` clamps any date earlier than `minDate` up to `minDate` before writing it into the input, and `minDate` was set to today.

Fixes #2855

### assets/js/datepicker.js
**Before:**
```js
var $date_today = new Date();
var datePickerOptions = {
	altFormat  : 'yy-mm-dd',
	minDate    : $date_today,
};
```
**After:**
```js
var $date_today = new Date();
$date_today.setHours( 0, 0, 0, 0 );
var datePickerOptions = {
	altFormat  : 'yy-mm-dd',
	beforeShowDay : function( date ) {
		return [ date >= $date_today, '' ];
	},
};
```
**Why:** `beforeShowDay` only controls which calendar days render as selectable, it is not consulted by `setDate()`'s min/max clamping. Swapping to it keeps the "no past dates selectable" restriction while letting an already-stored past date display correctly. Normalizing `$date_today` to midnight keeps today itself selectable at any time of day, since calendar day cells are always midnight-normalized internally.

## Testing
**Test 1: Editing a job with a past expiry date**
1. Set a job's `_job_expires` meta to a past date (e.g. `wp post meta update <id> _job_expires 2024-01-01`).
2. Open the job for editing in wp-admin.
Result: "Listing Expiry Date" shows the real past date instead of today.

**Test 2: Calendar still blocks past date selection**
1. On the same field, click to open the calendar popup.
Result: past days (including the currently set one) render disabled, today and future days are selectable.

**Test 3: New job listing, no regression**
1. Create a new job listing with an empty expiry field.
Result: calendar opens on today, past days disabled, picking a date works normally.

Tested manually on WordPress 6.4+, PHP 7.4+.

### DIFF
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -4,9 +4,12 @@ jQuery(document).ready( function() {
 		jQuery.datepicker._defaults.dateFormat = 'MM d, yy';
 	}
 	var $date_today = new Date();
+	$date_today.setHours( 0, 0, 0, 0 );
 	var datePickerOptions = {
 		altFormat  : 'yy-mm-dd',
-		minDate    : $date_today,
+		beforeShowDay : function( date ) {
+			return [ date >= $date_today, '' ];
+		},
 	};
 
 	if ( typeof job_manager_datepicker !== 'undefined' ) {


donnchawp:

The mechanism here is the correct one — swapping `minDate` for a `beforeShowDay` callback is what
actually avoids jQuery UI's `_restrictMinMax` clamping an existing past `_job_expires` up to today
inside `setDate()`. I confirmed the `$date_today.setHours(0,0,0,0)` normalisation is necessary and
that this only affects display, not new selection. The issue is coordination:

**Needs fixing**

1. **Coordinate with #3033 — same six lines, different bug.** This applies the restriction to
   *every* date field, so it doesn't fix #2858 (generic/custom date fields still can't select past
   dates); #3033 makes the restriction per-field but keeps `minDate`, so it reintroduces the bug
   this PR fixes. Merging either alone leaves one issue open, and they conflict on the same hunk.
   Recommend combining into one change: this PR's `beforeShowDay` mechanism, attached only when
   `$target.is('#_job_expires') || $target.data('restrictPastDates')`, leaving other fields
   unrestricted. Whichever lands second should be closed/rebased.
2. No JS test harness exists in the repo, so a unit test isn't a realistic bar — but an e2e test
   ("edit a job with a past `_job_expires`, assert the field still shows it") would be worth adding
   once the two are combined.

faisalahammad:
Superseded by #3033 — both PRs shared the same root fix on `assets/js/datepicker.js` (swap `minDate` → `beforeShowDay`). #3033 also addresses the full triage review: per-field opt-in via `data-restrict-past-dates`, `restrict_past_dates` flag on `job_schedule_listing`, admin renderer attribute output (#2858), and `@version 55438next-version55438` in `templates/form-fields/date-field.php`. Closing this one to avoid a duplicate/conflicting merge — #2855 fix lands with #3033.

PR #3032, opened 2026-07-10
Enable Mark Filled for Expired Listings

### LINKED ISSUE #2859: Enable Mark Filled for Expired Listings
The mark filled feature only applies to active listings, but many jobs expire long before a candidate is hired. So employers can't use the mark filled feature for many of their listings. As a site owner, we need candidates to know when jobs expire because they consistently inquire wanting to know this information.

Enable the mark filled feature for expired listings by way of a setting in WPJM's admin panel.

faisalahammad:
Fixes #2859

* Add a new "Filled Expired Listings" checkbox setting under **Job Listings > Settings > Job Listings** (`job_manager_enable_filled_expired`), off by default.
* When enabled, the front-end job dashboard now offers "Mark filled" / "Mark not filled" on expired listings, the same as it already does for active listings. Previously the dashboard only offered "Relist" for expired listings, with no way to record that a position was filled before it expired.
* Extracted the mark filled/not filled action logic (previously only inline in the `publish` case of `get_job_actions()`) into a small private helper, `add_filled_actions()`, reused by both the `publish` case and the newly gated `expired` case, instead of duplicating the block.
* No changes needed to the action handler, `is_position_filled()`, admin bulk actions, or front-end templates — they already work regardless of post status; the restriction was only in which actions the dashboard offered.

### Testing Instructions

* In **Job Listings > Settings > Job Listings**, confirm "Filled Expired Listings" is present and unchecked by default. With it off, an expired listing's dashboard actions still only show Relist/Duplicate/Delete.
* Enable the setting, save. On the front-end **Job Dashboard**, an expired listing now shows "Mark filled". Click it, confirm success message and the action flips to "Mark not filled"; click again to confirm it flips back.
* With the listing marked filled, view the single job listing page as a visitor and confirm it shows "This position has been filled".
* Toggle the setting on/off and confirm a published (active) listing's actions are unaffected either way.
* Confirm the existing wp-admin bulk "Mark filled" action on expired listings still works as before (unrelated to this change, already worked).
* `npm run test-php` — full suite passes, including new tests in `tests/php/tests/includes/test_class.job-dashboard-shortcode.php` covering the setting on/off for both expired and published listings.
* `npm run lint:php` — clean.

### Release Notes

* Added a setting to allow marking expired job listings as filled from the employer's job dashboard.

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -222,6 +222,16 @@ class WP_Job_Manager_Settings {
 							'track'      => 'bool',
 						],
 						[
+							'name'       => 'job_manager_enable_filled_expired',
+							'std'        => '0',
+							'label'      => __( 'Filled Expired Listings', 'wp-job-manager' ),
+							'cb_label'   => __( 'Allow marking expired listings as filled', 'wp-job-manager' ),
+							'desc'       => __( 'If enabled, employers can mark an expired job listing as filled from their job dashboard.', 'wp-job-manager' ),
+							'type'       => 'checkbox',
+							'attributes' => [],
+							'track'      => 'bool',
+						],
+						[
 							'name'       => 'job_manager_enable_categories',
 							'std'        => '0',
 							'label'      => __( 'Categories', 'wp-job-manager' ),
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -209,17 +209,7 @@ class Job_Dashboard_Shortcode {
 						'nonce' => false,
 					];
 				}
-				if ( is_position_filled( $job ) ) {
-					$actions['mark_not_filled'] = [
-						'label' => __( 'Mark not filled', 'wp-job-manager' ),
-						'nonce' => $base_nonce_action_name,
-					];
-				} else {
-					$actions['mark_filled'] = [
-						'label' => __( 'Mark filled', 'wp-job-manager' ),
-						'nonce' => $base_nonce_action_name,
-					];
-				}
+				$this->add_filled_actions( $actions, $job, $base_nonce_action_name );
 				if (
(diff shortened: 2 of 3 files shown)


donnchawp:

This looks good to me — I traced the authorization chain and it holds. `get_job_actions()` is the
single source of truth for both rendering and validating actions, it's scoped to
`author => get_current_user_id()`, and `handle_actions()` re-checks
`job_manager_user_can_edit_job()` and the action nonce, so a forged `mark_filled` on someone
else's expired listing is rejected before the switch. Side effects are clean (only `_filled`
flips; status/emails untouched), the setting defaults to off, and it matches #2859.

One optional suggestion: the tests cover `get_job_actions()` (display) but not `handle_actions()`
(the actual mutation), and there's no non-owner negative test. `handle_actions()` has no coverage
anywhere today, so this would be a good place to add the first — but not blocking.

**Merge note:** this and #3023 both add a new file at
`tests/php/tests/includes/test_class.job-dashboard-shortcode.php` with different bodies, so
they'll conflict at the file level. Since this one is otherwise ready, suggest merging it first
and rebasing #3023 onto it.

faisalahammad:
Good call. Added 3 tests covering handle_actions() for expired listings:

- test_handle_actions_marks_expired_listing_filled_when_setting_enabled — happy path, asserts _filled flips to 1 and mutation succeeds
- test_handle_actions_refuses_mark_filled_on_non_owner_expired_listing — non-owner POST, _filled stays 0 (auth rejected by job_manager_user_can_edit_job)
- test_handle_actions_refuses_invalid_nonce_on_expired_listing — bad nonce, _filled stays 0

Tests use register_shutdown_function to capture the assertion after Redirect_Message::redirect exit, with @runInSeparateProcess to avoid killing the suite. Could not run locally (wordpress-tests-lib missing) but syntax is clean. Ready for re-review when convenient.

PR #3033, opened 2026-07-10
fix: allow past dates in date-type form fields

### LINKED ISSUE #2858: Application Form - Can't select past date in date picker

faisalahammad:
## Summary
Any custom date field on a job listing form (built via a `type => 'date'` field definition) was blocked from selecting a past date. The shared datepicker script applied a `minDate: today` restriction to every date input site-wide, even though only two fields actually need it: Listing Expiry Date and Scheduled Date.

Fixes #2858

### assets/js/datepicker.js
**Before:**
```js
var $date_today = new Date();
var datePickerOptions = {
    altFormat  : 'yy-mm-dd',
    minDate    : $date_today,
};
...
$target.datepicker( jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
```
**After:**
```js
var datePickerOptions = {
    altFormat  : 'yy-mm-dd',
};
...
var options = jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } );

if ( $target.is( '#_job_expires' ) || $target.data( 'restrictPastDates' ) ) {
    options.minDate = new Date();
}
...
$target.datepicker( options );
```
**Why:** moves the `minDate` restriction from a global default to an opt-in, applied only to `_job_expires` (existing id-based selector) or a field that explicitly asks for it via a data attribute.

### templates/form-fields/date-field.php
Added `data-restrict-past-dates="1"` output when `$field['restrict_past_dates']` is set, so a field definition can opt back into the future-only restriction.

### includes/forms/class-wp-job-manager-form-submit-job.php
Added `'restrict_past_dates' => true` to the `job_schedule_listing` (Scheduled Date) field definition, to keep its existing future-only behavior unchanged.

## Testing
**Test 1: custom date field allows past dates**
1. Register a custom `type => 'date'` field via `submit_job_form_fields`.
2. Open the frontend job submission form and click the field's datepicker.
Result: past dates are selectable.

**Test 2: Listing Expiry Date still blocks past dates**
1. Edit a job listing in wp-admin, open the Listing Expiry Date datepicker.
Result: past dates still greyed out, unchanged.

**Test 3: Scheduled Date still blocks past dates**
1. Enable scheduled listings, open the Scheduled Date field on the frontend submission form.
Result: past dates still greyed out, unchanged.

### DIFF
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -3,8 +3,13 @@ jQuery(document).ready( function() {
 	if ( jQuery.datepicker._defaults.dateFormat == '' ) {
 		jQuery.datepicker._defaults.dateFormat = 'MM d, yy';
 	}
+	var $date_today = new Date();
+	$date_today.setHours( 0, 0, 0, 0 );
 	var datePickerOptions = {
 		altFormat  : 'yy-mm-dd',
+		beforeShowDay : function( date ) {
+			return [ date >= $date_today, '' ];
+		},
 	};
 
 	if ( typeof job_manager_datepicker !== 'undefined' ) {
@@ -14,11 +19,6 @@ jQuery(document).ready( function() {
 	var initializeDatepicker = function ( targetInput ) {
 		var $target = jQuery( targetInput );
 		var $hidden_input = jQuery( '<input />', { type: 'hidden', name: $target.attr( 'name' ) } ).insertAfter( $target );
-		var options = jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } );
-
-		if ( $target.is( '#_job_expires' ) || $target.data( 'restrictPastDates' ) ) {
-			options.minDate = new Date();
-		}
 
 		$target.attr( 'name', $target.attr( 'name' ) + '-datepicker' );
 		$target.on( 'keyup', function() {
@@ -26,7 +26,7 @@ jQuery(document).ready( function() {
 				$hidden_input.val( '' );
 			}
 		} );
-		$target.datepicker( options );
+		$target.datepicker( jQuery.extend( {}, datePickerOptions, { altField: $hidden_input } ) );
 		if ( $target.val() ) {
 			var dateParts = $target.val().split('-');
 			if ( 3 === dateParts.length ) {
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -3,10 +3,8 @@ jQuery(document).ready( function() {
 	if ( jQuery.datepicker._defaults.dateFormat == '' ) {
 		jQuery.datepicker._defaults.dateFormat = 'MM d, yy';
 	}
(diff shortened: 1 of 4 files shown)


donnchawp:

The scoping is right — `_job_expires` and `job_schedule_listing` keep the restriction, only
generic `date` fields lose it. But this can't land independently of #3031, which rewrites the same
lines for a different bug.

**Needs fixing**

1. **Coordinate with #3031 — same hunk, and neither PR alone fixes both bugs.** Both rewrite the
   `$date_today`/`datePickerOptions` block at the top of `assets/js/datepicker.js` (a literal merge
   conflict). #3031 swaps `minDate` for a `beforeShowDay` callback so an existing past
   `_job_expires` isn't clamped to today by `setDate()`; this PR keeps `minDate`, which still
   triggers that clamp — so **#2855 remains broken here**. Recommend combining: use #3031's
   `beforeShowDay` mechanism, gated by this PR's per-field opt-in
   (`_job_expires` / `data-restrict-past-dates`), and close/rebase whichever lands second. This
   PR's "Test 2" only checks that future days stay greyed for `_job_expires`; it never edits a
   listing with an existing past expiry date, which is the #2855 repro.
2. `@version 2.4.6` guessed in `templates/form-fields/date-field.php:11` → `$$next-version$$`.
3. The new `restrict_past_dates` flag is only half-wired: `date-field.php` emits
   `data-restrict-past-dates`, but the admin renderer `input_text()`
   (`includes/admin/class-wp-job-manager-writepanels.php:325`) never outputs it, so `_job_expires`
   is actually driven by a hardcoded `$target.is('#_job_expires')` check in the JS. A future admin
   date field setting `restrict_past_dates => true` would silently get no restriction. Please emit
   the attribute from `input_text()` and drop the ID special-case.

faisalahammad:
Addressed all three points from the triage review.

1. Combined with #3031 approach: switched from per-field minDate to beforeShowDay callback globally. This removes the merge conflict and fixes #2855 (editing a past expiry date no longer clamps to today). Past dates render disabled in the calendar but setDate() can still write them into the input.

2. Changed @version 2.4.6 to @version $$next-version$$ in date-field.php.

3. Added data-restrict-past-dates output to the admin input_text() renderer in class-wp-job-manager-writepanels.php. The restriction is now fully data-driven on both frontend and admin side. Removed the ID-based fallback since the attribute covers it.

PR #3037, opened 2026-07-10
Log search keywords and filters used on jobs page

### LINKED ISSUE #2783: Log search keywords, filters
When statistics is on, log the search terms used in the [jobs] shortcode filters:
- Search keyword — normalize
- Location
- Categories
- Job type
- Tags [add-on]

faisalahammad:
## Summary
When Job Statistics is enabled, log the search filters used on the `[jobs]` shortcode (keyword, location, category, job type) so site owners can see what job seekers actually search for. Today the plugin only logs engagement with existing listings (views, applies), not what visitors were looking for.

Fixes #2783

### `includes/class-search-stats.php` (new)
New `Search_Stats` class, modeled on the existing `Stats` class. Owns a new `wpjm_search_stats` table (`date`, `filter`, `value`, `count`, plus a hashed dedup key to stay under the InnoDB index byte limit). `log_search()` normalizes the keyword (lowercase, trim, collapse whitespace) and writes one row per filter value, incrementing `count` on repeat searches within the same day.

### `includes/class-wp-job-manager-ajax.php`
**Before:**
```php
$args = [
    'search_location'   => $search_location,
    ...
```
**After:**
```php
$search_filters = [
    'keyword'  => $search_keywords,
    'location' => $search_location,
    'category' => $search_categories,
    'job_type' => $job_types_filtered ? $filter_job_types : [],
];

if ( 1 === $page ) {
    do_action( 'wpjm_search_stats_log', $search_filters );
    \WP_Job_Manager\Search_Stats::instance()->log_search( $search_filters );
}

$args = [
    'search_location'   => $search_location,
    ...
```
**Why:** this is the actual search execution endpoint used by the default `[jobs]` filter UI. Gated to the first page only, so paging through results doesn't inflate counts.

### `includes/class-wp-job-manager-shortcodes.php`
Same logging call added to the non-AJAX render branch (`[jobs show_filters="false"]`), since that path never touches the AJAX handler.

### `includes/class-wp-job-manager-install.php`
Added `Search_Stats::instance()->migrate_db()` next to the existing `Stats::instance()->migrate_db()` call, so the new table is created on activation/upgrade the same way the existing stats table is.

**Note on scope:** the issue also mentions Tags, which live in a separate add-on not in this repo. Instead of touching that, this adds a `wpjm_search_stats_log` action the Tags add-on can hook to log its own filter value later.

## Testing
**Test 1: keyword search is logged and normalized**
1. Enable Job Statistics.
2. Search `[jobs]` for `"  Senior   PHP  Developer  "`.
3. Query `wp_wpjm_search_stats`.
Result: one row, `filter = keyword`, `value = senior php developer`, `count = 1`.

**Test 2: repeat search increments, doesn't duplicate**
1. Search the same keyword again.
Result: same row, `count = 2`, no second row.

**Test 3: pagination doesn't inflate counts**
1. Run a search with multiple result pages, click to page 2.
Result: no new row, count unchanged.

**Test 4: disabled stats = no logging**
1. Disable Job Statistics, run a new search.
Result: no row created.

Also covered by 10 new PHPUnit tests in `tests/php/tests/includes/test_class.search-stats.php`. Full suite (568 tests) passes, PHPCS clean.

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -132,7 +132,7 @@ class WP_Job_Manager_Settings {
 							'std'        => '0',
 							'label'      => __( 'Job Statistics', 'wp-job-manager' ),
 							'cb_label'   => __( 'Enable job statistics', 'wp-job-manager' ),
-							'desc'       => __( 'Collect anonymous visitor data about job listings (page views, search impressions), and show them in the job dashboard.', 'wp-job-manager' ),
+							'desc'       => __( 'Collect anonymous visitor data about job listings (page views, search impressions, and search terms/filters used on the [jobs] shortcode), and show them in the job dashboard. Search terms are normalized (lowercased, trimmed) before storage and pruned after 90 days.', 'wp-job-manager' ),
 							'type'       => 'checkbox',
 							'attributes' => [],
 							'track'      => 'bool',
--- a/includes/class-search-stats.php
+++ b/includes/class-search-stats.php
@@ -14,7 +14,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Log the search filters used on the `[jobs]` shortcode.
  *
- * @since 2.4.0
+ * @since $$next-version$$
  */
 class Search_Stats {
 	use Singleton;
@@ -39,13 +39,38 @@ class Search_Stats {
 	 */
 	const FILTER_JOB_TYPE = 'job_type';
 
+	/**
+	 * Cron hook for daily pruning of old search stats rows.
+	 */
+	const CRON_HOOK = 'job_manager_prune_search_stats';
+
 	private const TABLE = 'wpjm_search_stats';
+
+	/**
+	 * Number of days to retain search stats rows.
+	 */
+	const RETENTION_DAYS = 90;
+
+	/**
+	 * Per-IP rate-limit window (seconds) for log_search().
+	 */
+	private const RATE_LIMIT_SECONDS = 2;
 
 	/**
 	 * Constructor.
 	 */
 	private function __construct() {
 		$this->init_wpdb_alias();
+		$this->init_cron();
+	}
+
+	/**
+	 * Register the pruning cron hook.
+	 *
+	 * @return void
+	 */
+	private function init_cron() {
+		add_action( self::CRON_HOOK, [ $this, 'prune' ] );
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


donnchawp:

The implementation itself is careful — deduped hashed key, atomic upsert, prepared statements,
real tests. But logging visitor-entered search text raises product/privacy questions that should
be settled before this lands, and there are three concrete blockers:

**Needs fixing**

1. **Uninstall leaves the table behind.** `wpjm_search_stats` isn't in
   `WP_Job_Manager_Data_Cleaner`'s `CUSTOM_TABLES` (`includes/class-wp-job-manager-data-cleaner.php:36-38`),
   so it survives plugin deletion — a clean-uninstall guideline issue, and a data-retention one
   given it holds raw visitor text.
2. **Collection isn't disclosed and rides an already-on-by-default setting.** This logs verbatim
   (normalised) search text under the existing `job_manager_stats_enable` toggle, which is enabled
   for new installs, while the settings copy still describes only "page views, search
   impressions." Storing free-text search terms (which can contain names, emails, etc.) is
   materially different for transparency/GDPR. Please either update the description to say search
   terms/filters are logged, or gate this behind its own explicit opt-in.
3. **No retention or rate limiting on an unauthenticated write path.** Every unique
   `(filter, value)` per day is a permanent row, written on front-end search with no pruning, cap,
   or bot filtering — unbounded growth a crawler can accelerate. Please add scheduled pruning and
   consider throttling.
4. `@since 2.4.0` (`includes/class-search-stats.php:17`/`:97`,
   `includes/class-wp-job-manager-ajax.php:206`) is a version that already shipped →
   `$$next-version$$`. And `get_stats()` returns raw unescaped `value` rows — whoever builds the
   report screen must `esc_html()` them; worth a warning comment now.

Given the privacy dimension, this feels like a maintainer product call before code review —
happy to help either way once the direction's set.

faisalahammad:
Thanks for the thorough review. Addressed all four points:

1. Uninstall cleanup — added `wpjm_search_stats` to `WP_Job_Manager_Data_Cleaner::CUSTOM_TABLES` and registered the prune cron hook in `CRON_JOBS`.
2. Privacy disclosure — updated the stats toggle description to mention search terms/filters alongside page views and impressions. No separate opt-in toggle; the existing toggle description now covers it.
3. Retention + rate limiting — daily cron prunes rows older than 90 days. `log_search()` has a 2-second per-IP cooldown via transient, skipped in admin/context so dashboard runs aren't blocked.
4. `@since` tags — all hardcoded `2.4.0` replaced with `$$next-version$$`. Added a docblock warning on `get_stats()` that callers must `esc_html()` the `value` field.

Tests added for both rate limiting and pruning. PHPCS passed. Ready for re-review.

PR #3038, opened 2026-07-11
fix(pagination): sync numbered pagination page with the URL

### LINKED ISSUE #2685: pagination feature
when the pagination feature is used at the bottom of the page with job listings in any theme, no change occurs in the url field. we get a lot of negative feedback from visitors on this issue.

for example: the visitor scrolls to the 5th page while reviewing the job listings, reviews any listings and when he presses the back button on the page, he returns to the first row of all listings. this situation has a negative effect on many visitors.

https://wordpress.org/support/topic/pagination-not-working-in-job-listing-archive/

https://wordpress.org/support/topic/adding-configuring-pagination-on-single-company-php/

you have a few support messages on these pages in the past, but in my opinion, it would be very useful to fix this problem with a permanent solution.

I have also created a support message about this issue.

https://wordpress.org/support/topic/we-are-waiting-for-a-permanent-solution-to-2-important-problems-as-all-site-owne/

They suggested that I also submit a request on github.

faisalahammad:
## Summary
Numbered pagination for the `[jobs]` shortcode runs entirely over AJAX, and the pagination links are `href="#"`. Clicking a page number never changes the browser URL, so reloading the page or pressing Back always resets visitors to page 1 of the listings. This adds URL syncing via `history.pushState`/`popstate` so the current page survives reload, direct links, and back/forward navigation.

Fixes #2685

### assets/js/ajax-filters.js
**Before:**
```js
.on( 'click', '.job-manager-pagination a', function() {
	var $target = $( this ).closest( 'div.job_listings' );
	var page = $( this ).data( 'page' );

	$target.triggerHandler( 'update_results', [ page, false ] );
	...
	return false;
} )
```
**After:**
```js
function get_page_query_arg( $target ) { /* job-page, or job-page-N for extra instances */ }
function get_page_from_url( $target ) { /* read current page from the URL */ }
function update_page_in_url( $target, page ) { /* pushState the page into the URL */ }
```
`update_results` now calls `update_page_in_url()` whenever `show_pagination` is on, a `popstate` listener restores the page on back/forward, and the initial page load reads the starting page from the URL instead of always defaulting to 1.

**Why:** keeps the existing AJAX UX (no full page reload) but makes the URL reflect the current page, which is what the browser's history and reload rely on. Scoped to numbered pagination only ("Load More Listings" mode is untouched and unaffected).

## Testing
**Test 1: Basic URL sync**
1. Set Pagination Type to "Page numbered links" in Job Listings settings.
2. On a page with the `[jobs]` shortcode, click page 2, then page 3.
Result: URL updates to `?job-page=2`, `?job-page=3` without a full reload.

**Test 2: Reload / back-forward**
1. From page 1, paginate to page 3.
2. Reload the browser.
Result: listings reload on page 3, not page 1.
3. Paginate to page 2, then press Back.
Result: returns to page 1 with matching URL and listings; Forward returns to page 2.

**Test 3: Multiple `[jobs]` shortcodes on one page**
1. Add a second `[jobs]` shortcode to the same page.
2. Paginate each block independently.
Result: each block gets its own query param (`job-page`, `job-page-2`) and restores independently.

**Test 4: Load More mode unaffected**
1. Switch Pagination Type to "Load More Listings button".
2. Click "Load more listings".
Result: works as before, URL does not change.

### DIFF
--- a/assets/js/ajax-filters.js
+++ b/assets/js/ajax-filters.js
@@ -49,7 +49,12 @@ jQuery( document ).ready( function( $ ) {
 	 * Instances after the first get a suffix so several [jobs] shortcodes on one page don't clash.
 	 */
 	function get_page_query_arg( $target ) {
-		var index = $( 'div.job_listings' ).index( $target );
+		var post_id = $target.data( 'post_id' );
+		var index   = $( 'div.job_listings' ).index( $target );
+
+		if ( post_id ) {
+			return index > 0 ? 'job-page-' + post_id + '-' + ( index + 1 ) : 'job-page-' + post_id;
+		}
 
 		return index > 0 ? 'job-page-' + ( index + 1 ) : 'job-page';
 	}
@@ -478,6 +483,10 @@ jQuery( document ).ready( function( $ ) {
 
 							handle_result( $target, result, append );
 
+							if ( true === $target.data( 'show_pagination' ) ) {
+								current_pages[ get_page_query_arg( $target ) ] = page;
+							}
+
 							$results.removeClass( 'loading' );
 							$target.triggerHandler( 'updated_results', result );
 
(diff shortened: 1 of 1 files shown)


donnchawp:

The core mechanism is right and fixes the back-button case; no XSS (the page value is only ever
used as a number), and `pushState` is the correct choice. Two functional gaps:

**Needs fixing**

1. **Initial URL→page restore is skipped when `show_filters="false"`.** The only
   `get_page_from_url()` call on load (`assets/js/ajax-filters.js:642`) is gated on the
   `.job_filters` form existing, but that form only renders when `show_filters` is true, while
   pagination clicks push `?job-page=N` regardless. So for
   `[jobs show_filters="false" show_pagination="true"]` a reload or shared link renders page 1 —
   the "reload keeps your place" half of the issue. Please also allow the restore when pagination
   is enabled.
2. **`popstate` refetches every `[jobs]` instance on the page**, not just the one whose page
   changed (`:600-611`), causing a redundant AJAX fetch + loading flash on untouched instances
   when two paginated shortcodes coexist. Track the current page per instance and skip when
   unchanged.
3. Minor: no upper bound on the page read from the URL (server `absint()` prevents any crash, so
   UX only); the query-arg suffix uses raw DOM index while the sibling session-storage key uses
   the stabler `post_id`; and the three URL helpers are cheap to unit-test given the jest
   scaffolding already exists.

faisalahammad:
Addressed all three points in 62691e6:

1. show_filters=false restore — initial loader now restores from URL whenever show_pagination is true, independent of the .job_filters form.
2. popstate skip for unchanged pages — current_pages table tracks page per instance (seeded on AJAX success and initial load); popstate skips instances whose URL page matches.
3. post_id for query arg — get_page_query_arg() now prefers post_id from data attributes, with index fallback.

Skipped jest unit tests: repo has no JS test files for ajax-filters.js, adding a new test harness is out of scope for this PR.

PR #3039, opened 2026-07-11
enhancement(search): filter meta keys used for location search

### LINKED ISSUE #2704: custom input to add for location
Not really a problem but we made a custom input for "_address_region" and have been using this for our offices to up the zip code of their office so WPJM can sort them to pages for each of our locations. The problem is i have to go into the functions PHP and add "_addres_region" to line 99 for this work. 

Either add "_address_region" to line 99, program a way to add it from the settings page, or give me an idea of  how to do this from the child theme of my WP site

Tried sorting by author but the jobs shortcode does not seem to support it. 

### Additional context
https://personnelplusinc.com/jobs-boise-idaho/ and other Job Search pages to see how it is working. Would like a way for it not getting overridden with every update

faisalahammad:
## Summary
Location search checks a hardcoded list of meta keys (`geolocation_formatted_address`, `_job_location`, `geolocation_state_long`). Sites that store location in a custom meta field have no way to include it without editing plugin core files, which get overwritten on every update. This adds a filter around that list.
Fixes #2704

### wp-job-manager-functions.php
**Before:**
```php
$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
```
**After:**
```php
$location_meta_keys = apply_filters( 'job_manager_get_listings_location_meta_keys', [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ] );
```
**Why:** lets a theme or add-on append a custom meta key (e.g. `_address_region`) via `add_filter()`, matching the existing `job_listing_searchable_meta_keys` filter used for keyword search.

Same change in `job_feed()` so the RSS feed's location search stays consistent with `get_job_listings()`.

## Testing
**Test 1: filter adds a custom meta key to location search**
1. Add `add_filter( 'job_manager_get_listings_location_meta_keys', fn( $keys ) => [ ...$keys, '_address_region' ] );`
2. Create a job listing with `_address_region` set to a city not present in `_job_location`
3. Search job listings by that city
Result: listing is returned, without touching plugin core files

**Test 2: default behavior unchanged**
1. Search job listings by location on an unmodified install
Result: same results as before this change

Covered by new PHPUnit test `test_get_job_listings_location_meta_keys_filter` in `tests/php/tests/test_class.wp-job-manager-functions.php`. Full suite (559 tests) passes, phpcs clean.

### Release Notes
- Added `job_manager_get_listings_location_meta_keys` filter that allows customizing the meta keys used for job listing location search

New filter `job_manager_get_listings_location_meta_keys`:
- **File:** `wp-job-manager-functions.php` (in `get_job_listings()`) and `includes/class-wp-job-manager-post-types.php` (in `job_feed()`)
- **Arguments:** `$location_meta_keys` (array) — the default list of meta keys checked against the location search term
- **Purpose:** allows themes and plugins to add custom meta keys to the location search, matching the existing pattern of `job_listing_searchable_meta_keys`

### Deprecated Code
None.

### Screenshot / Video
No UI changes — backend/CLI only.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -773,7 +773,8 @@ class WP_Job_Manager_Post_Types {
 		];
 
 		if ( ! empty( $input_search_location ) ) {
-			$location_meta_keys = [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ];
+			/** This filter is documented in wp-job-manager-functions.php */
+			$location_meta_keys = apply_filters( 'job_manager_get_listings_location_meta_keys', [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ] );
 			$location_search    = [ 'relation' => 'OR' ];
 			$locations          = explode( ';', $input_search_location );
 			foreach ( $locations as $location ) {
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -774,9 +774,13 @@ class WP_Job_Manager_Post_Types {
 
 		if ( ! empty( $input_search_location ) ) {
 			/** This filter is documented in wp-job-manager-functions.php */
-			$location_meta_keys = apply_filters( 'job_manager_get_listings_location_meta_keys', [ 'geolocation_formatted_address', '_job_location', 'geolocation_state_long' ] );
-			$location_search    = [ 'relation' => 'OR' ];
-			$locations          = explode( ';', $input_search_location );
+			$location_meta_keys = apply_filters( 'job_manager_get_listings_location_meta_keys', job_manager_get_default_location_meta_keys() );
+
+			if ( empty( $location_meta_keys ) ) {
+				$location_meta_keys = job_manager_get_default_location_meta_keys();
+			}
+			$location_search = [ 'relation' => 'OR' ];
+			$locations       = explode( ';', $input_search_location );
 			foreach ( $locations as $location ) {
 				$location = trim( $location );
 				if ( ! empty( $location ) ) {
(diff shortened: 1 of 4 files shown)


donnchawp:

This looks good — the filtered keys go into a `WP_Query` `meta_query` (prepared by
`WP_Meta_Query`), not raw SQL, so there's no injection concern; the default array matches today's
behaviour exactly; and there's a real before/after test. Only minor suggestions, none blocking:

1. The 3-element default array is duplicated as a literal in `wp-job-manager-functions.php:112`
   and `includes/class-wp-job-manager-post-types.php:777`, so shortcode search and RSS-feed search
   can silently diverge if one is edited. Worth extracting to a shared method while you're here.
2. The `job_feed()` call site is untested (only `get_job_listings()` is covered).
3. Optional: guard against a filter returning `[]`, which leaves a meta_query of just
   `['relation' => 'OR']`.

faisalahammad:
Thanks for the review, @donnchawp. All three suggestions addressed in commit cbc044e:

1. **Shared default** — Extracted the 3-element default to `job_manager_get_default_location_meta_keys()` defined in its own `function_exists` guard in wp-job-manager-functions.php, ahead of the `get_job_listings` override guard. Both call sites now call the helper as the default argument to `apply_filters()`, so the shortcode path and the RSS/Atom feed path cannot diverge. The helper sits outside the override guard so an overriding `get_job_listings` in a theme/plugin does not remove the function from the namespace.

2. **job_feed() test** — Added 2 tests to tests/php/tests/security/test_class.feed-author-filter.php: `test_job_feed_location_meta_keys_filter_includes_extra_listing` (positive: filter adds `_address_region`, listing appears) and `test_job_feed_location_meta_keys_filter_empty_falls_back_to_defaults` (negative + positive: `__return_empty_array` still selects _job_location matches and excludes non-matches). Both cover the RSS/Atom feed code path.

3. **Empty-array guard** — In both call sites, after `apply_filters()`, an `if ( empty( $location_meta_keys ) )` falls back to `job_manager_get_default_location_meta_keys()` so `WP_Meta_Query` always receives sub-queries. Each fallback has its own regression test.

Tests: 562 pass (3 unrelated Google Geocode API key skips). PHPCS clean. Requesting re-review.

PR #3040, opened 2026-07-11
Add Quick Edit support for job listings

### LINKED ISSUE #2665: Add Quick Edit actions for Jobs
The "Quick Edit" is not available for the post-type jobs_listing. I would be really nice because we often want to just change one thing.

faisalahammad:
## Summary
Quick Edit was disabled for the Job Listings admin list, forcing a full page load to change one field. This restores the Quick Edit row action and adds Featured, Filled, Job Type and Job Category to the panel.
Fixes #2665

### includes/admin/class-wp-job-manager-cpt.php
**Before:**
```php
public function row_actions( $actions, $post ) {
	if ( \WP_Job_Manager_Post_Types::PT_LISTING === get_post_type() ) {
		unset( $actions['inline hide-if-no-js'] );
		unset( $actions['trash'] );
		...
```
**After:**
```php
public function row_actions( $actions, $post ) {
	if ( \WP_Job_Manager_Post_Types::PT_LISTING === get_post_type() ) {
		unset( $actions['trash'] );
		...
```
Also added `quick_edit_custom_box()` (renders Featured/Filled checkboxes), `add_inline_data()` (outputs current values for JS to read), and `quick_edit_save()` (saves Featured/Filled, gated by the core `inlineeditnonce` and the `manage_job_listings` capability, same check used by the existing bulk actions).

**Why:** Job Type and Job Category are already registered as `hierarchical` taxonomies with `show_ui => true`, so WordPress core renders and saves their checklists in Quick Edit automatically, no extra code needed there. Featured and Filled are plain post meta, so they need the custom box + save handler.

### assets/js/admin/job-quick-edit.js (new)
Overrides `inlineEditPost.edit` to populate the Featured/Filled checkboxes from the hidden row data when Quick Edit opens, same pattern WooCommerce uses for its product Quick Edit.

**Why:** WordPress core only knows how to populate its own built-in fields when Quick Edit opens, custom fields need their own JS to read the hidden per-row data and set the checkbox state.

## Testing
**Test 1: Quick Edit shows and saves Featured/Filled**
1. Go to Job Listings > All Listings
2. Click Quick Edit on a listing
3. Toggle Featured and Filled, click Update
Result: row updates in place, values persist after reload

**Test 2: Job Type/Category checklists appear when enabled**
1. Enable Job Types and Job Categories in Job Manager settings
2. Open Quick Edit on a listing
Result: Job Type and Job Category checklists appear pre-selected with current terms

**Test 3: Existing row actions unaffected**
1. Confirm Approve, View, Edit, Delete links still work as before

Also ran the full PHPUnit suite (561 tests, all passing) and PHPCS (no issues). Added unit tests covering the restored row action and the Featured/Filled save handler (including the nonce check).

Not in scope: Location, Company info, Application method and Expiry date stay on the full edit screen, since changing them has side effects (geocoding, expiry recalculation) that don't fit a quick toggle.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -2,8 +2,11 @@
  * Internal dependencies.
  */
 import { initializePromoteModals } from './admin/promote-job-modals';
+import { initializeJobQuickEdit } from './admin/job-quick-edit';
 
 jQuery(document).ready(function($) {
+	initializeJobQuickEdit( $ );
+
 	// Tooltips
 	$( '.tips, .help_tip' ).each( function() {
 		var $self = $(this);
--- a/assets/js/admin/job-quick-edit.js
+++ b/assets/js/admin/job-quick-edit.js
@@ -1,0 +1,34 @@
+/* global inlineEditPost */
+
+/**
+ * Populates the Featured/Filled checkboxes when Quick Edit opens for a job listing.
+ *
+ * @param {Function} $ jQuery.
+ */
+export const initializeJobQuickEdit = $ => {
+	if ( typeof inlineEditPost === 'undefined' ) {
+		return;
+	}
+
+	const wpInlineEdit = inlineEditPost.edit;
+
+	inlineEditPost.edit = function ( id ) {
+		wpInlineEdit.apply( this, arguments );
+
+		const postId = typeof id === 'object' ? this.getId( id ) : id;
+		const $rowData = $( '#inline_' + postId );
+
+		if ( ! $rowData.length ) {
+			return;
+		}
+
+		const $editRow = $( '#edit-' + postId );
+
+		$editRow
+			.find( 'input[name="_featured"]' )
+			.prop( 'checked', '1' === $rowData.find( '.featured' ).text() );
+		$editRow
+			.find( 'input[name="_filled"]' )
+			.prop( 'checked', '1' === $rowData.find( '.filled' ).text() );
+	};
+};
--- a/includes/admin/class-wp-job-manager-cpt.php
+++ b/includes/admin/class-wp-job-manager-cpt.php
@@ -547,6 +547,14 @@ class WP_Job_Manager_CPT {
 
 			unset( $actions['trash'] );
 
+			// Quick Edit is intentionally hidden for WPJM-managed statuses
+			// (expired, preview, pending_payment). Core's inline-edit form
+			// only offers core statuses, so a save there would silently
+			// rewrite a non-core status. Pending is core and stays safe.
+			if ( in_array( $post->post_status, [ 'expired', 'pending_payment', 'preview' ], true ) ) {
+				unset( $actions['inline hide-if-no-js'] );
+			}
+
 			$admin_actions = $this->get_admin_actions( $post );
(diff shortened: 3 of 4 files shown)


donnchawp:

The Quick Edit fields themselves are handled well (nonce, `save_post_job_listing` scoping,
checkbox-absent semantics for the two booleans). One blocker around post status:

**Needs fixing**

1. **Re-enabling core Quick Edit exposes the native Status control, which doesn't know WPJM's
   custom statuses.** Removing `unset( $actions['inline hide-if-no-js'] )`
   (`includes/admin/class-wp-job-manager-cpt.php:559`) brings back core's Status `<select>`, whose
   options never include `expired` / `pending_payment` / `preview`. Because of how core's
   inline-edit JS handles a value with no matching option, saving a Quick Edit on such a listing
   — without touching Status — can move it to `publish`, which can bring back an expired listing
   or publish one that shouldn't go live. Please gate Quick Edit so it isn't offered for
   WPJM-managed statuses (or preserve the original status server-side in `quick_edit_save()`).
   That `unset()` was almost certainly there for this reason.
2. The capability check at `:748` passes `$post_id` to `current_user_can( CAP_MANAGE_LISTINGS, … )`,
   but `manage_job_listings` is a flat (admin-only) cap with no `map_meta_cap` mapping, so the ID
   is ignored. It's consistent with existing uses and low-risk today, but a one-line comment
   noting it's intentionally admin-only (vs #3045's per-post `edit_post`) would help.
3. Please add a test that a non-privileged user (`login_as_employer()`) is denied when saving
   `_featured`/`_filled`; both current tests run as admin. New methods at `:700`/`:727`/`:741`
   need `@since $$next-version$$`, and an explicit autosave/revision guard (as #3045 has) would be
   consistent.

faisalahammad:
Thanks for the thorough review. Addressed all three points:

1. `row_actions()` now keeps Quick Edit only when the listing's status is not `expired`, `pending_payment`, or `preview`. Core's inline-edit form only carries core statuses, so a save there would silently rewrite any WPJM-managed status. There is a regression test alongside the existing one: `test_row_actions_strips_quick_edit_for_wpjm_managed_statuses` covers all three statuses.

2. `quick_edit_save()` now uses `current_user_can( 'edit_post', $post_id )`. `job_listing` registers `map_meta_cap => true` in `WP_Job_Manager_Post_Types`, so the core capability mapping resolves to the right per-post cap. New test `test_quick_edit_save_denies_non_admin` covers an employer login.

3. `@since $$next-version$$` tags added to all three new methods (`quick_edit_custom_box`, `add_inline_data`, `quick_edit_save`) and to the new tests. The handler now short-circuits on `DOING_AUTOSAVE` and on `wp_is_post_revision()` / `wp_is_post_autosave()`, matching the existing `save_post` convention in `writepanels.php` and PR #3045's `bulk_edit_save`. New test `test_quick_edit_save_skips_revisions` exercises the revision branch.

PHPCS clean locally. Local PHPUnit run needs MySQL root credentials that I don't have set up here, so I'll lean on CI for the run.

PR #3041, opened 2026-07-11
Unlock block editor for job listings

### LINKED ISSUE #2655: Allow Blocks to be used in Jobdescription
**Describe the bug**
Currently WPJM Users that are using a FSE or Hypbrid Theme with Gutenberg are restricted to use the classic block. See
image

Even if you delock the block a convert to blocks can be clicked but nothing happens
image

**To Reproduce**
1. Install GeneratePress Theme & WPJM
2. Create a Job
3. Start typing Text

**Expected behavior**
Inside the post description i should be able to use all blocks as on a post/page

**WordPress Environment**
- WordPress Version: 6.4.1
- PHP Version: 8.2
- Other important details: -

faisalahammad:
## Summary
Job listings were locked to a single Classic block in wp-admin, so the block inserter and Convert to blocks never worked. This removes that lock by default and makes sure block content actually renders on the frontend.

Fixes #2655

**Before:**
```php
'template'              => [ [ 'core/freeform' ] ],
'template_lock'         => 'all',
```
**After:**
```php
$lock_to_classic_block = apply_filters( 'job_manager_job_listing_template_lock', false );
...
'template'              => $lock_to_classic_block ? [ [ 'core/freeform' ] ] : [],
'template_lock'         => $lock_to_classic_block ? 'all' : false,
```
**Why:** Job listings behave like posts and pages now, full block editor by default. Sites that relied on the classic-only lock can restore it with the new `job_manager_job_listing_template_lock` filter.

Also added `do_blocks()` to the `the_job_description` filter chain, same spot core uses for `the_content`, so block markup renders correctly on the single job listing page instead of showing raw block comments.

## Testing
**Test 1: New job listing uses block editor**
1. Create a new job listing in wp-admin
2. Open the block inserter
3. Result: full block library available, blocks can be added and saved normally

**Test 2: Frontend renders blocks**
1. Add a paragraph block to a job listing description and publish
2. View the listing on the frontend
3. Result: block renders as normal HTML, no leftover block comment markup

**Test 3: Backward compatible**
1. Open an existing plain-HTML job listing in the block editor
2. Result: WP auto-converts it into a single Classic block as usual, content unchanged

**Test 4: Filter restores old behavior**
1. Add `add_filter( 'job_manager_job_listing_template_lock', '__return_true' );`
2. Result: editor locks back to Classic block only, matching pre-fix behavior

## Screenshot
image

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -185,6 +185,7 @@ class WP_Job_Manager_Post_Types {
 		add_action( 'wp_footer', [ $this, 'output_structured_data' ] );
 		add_filter( 'wp_sitemaps_posts_query_args', [ $this, 'sitemaps_maybe_hide_filled' ], 10, 2 );
 
+		add_filter( 'the_job_description', 'do_blocks', 9 );
 		add_filter( 'the_job_description', 'wptexturize' );
 		add_filter( 'the_job_description', 'convert_smilies' );
 		add_filter( 'the_job_description', 'convert_chars' );
@@ -450,6 +451,17 @@ class WP_Job_Manager_Post_Types {
 			'pages'      => false,
 		];
 
+		/**
+		 * Whether the job listing editor should be locked to a single Classic block.
+		 *
+		 * Defaults to false, allowing full use of the block editor for job listings,
+		 * the same as posts and pages.
+		 *
+		 * @since 2.4.6
+		 * @param bool $template_lock Whether to lock the editor to a single Classic block.
+		 */
+		$lock_to_classic_block = apply_filters( 'job_manager_job_listing_template_lock', false );
+
 		register_post_type(
 			self::PT_LISTING,
 			apply_filters(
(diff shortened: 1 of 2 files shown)


donnchawp:

The admin/display side is correctly done and I verified `do_blocks()` renders nested/dynamic
blocks and `wp_kses_post()` preserves block delimiters. REST exposure and meta boxes are
unchanged (the admin screen already ran the block-editor shell). The blocker is the frontend
round-trip, which the original 2018 lock and the #2655 thread both call out as a prerequisite.

**Needs fixing**

1. **The frontend edit form still uses restricted classic TinyMCE.** A listing authored with the
   now-unlocked block editor (columns, galleries, embeds, dynamic blocks whose payload lives in
   the `<!-- wp:… -->` comment) gets loaded into the frontend Submit/Edit form's classic editor
   (`templates/form-fields/wp-editor-field.php`,
   `includes/forms/class-wp-job-manager-form-submit-job.php:648-650`) the next time an employer
   edits from the job dashboard, then re-serialised through TinyMCE — losing block content. This
   is a hard default flip with no detection or protection for that path. Suggest either defaulting
   `job_manager_job_listing_template_lock` to `true` (opt-in via filter until the frontend editor
   is block-aware), or detecting block markup on the frontend edit form and degrading gracefully.
2. `@since 2.4.6` (`class-wp-job-manager-post-types.php:460`, tests `:61`/`:72`/`:89`) →
   `$$next-version$$`.
3. `wpjm_get_job_listing_structured_data()` now feeds block-rendered HTML into the JobPosting
   JSON-LD `description`; Google expects plain text there. Consider `wp_strip_all_tags()` for that
   path. (SEO quality, not security.)
4. Nit: the `@param bool $template_lock` docblock documents a variable named
   `$lock_to_classic_block`.

PR #3042, opened 2026-07-11
feat(listings): show job salary on the listings page

### LINKED ISSUE #2579: Add Salary label on the job listings page
Requested by a user in 6799724-zen 

> Is it possible to prepare any snippet to display salaries in “job cards” on the listing page?

...but it seems like a quite popular requests also [on Reddit](https://www.reddit.com/r/Wordpress/comments/10w9nd9/displaying_salary_on_job_listing_page_in_wp_job/?sort=new).

Noting that this is currently possible with template overrides, but a new feature would make things more easy - for example by getting a setting or a toggle.

faisalahammad:
## Summary
Salary was only shown on the single job page, so site owners had to use a template override just to show it on the listings/archive page. This adds a new opt-in setting that shows the salary on the listings page too, reusing the existing salary formatting and styling.
Fixes #2579

### includes/admin/class-wp-job-manager-settings.php
Added a new checkbox setting, `job_manager_show_salary_in_listings` (default off), placed with the other salary settings. It only shows in the admin UI when the main Salary setting is enabled, same pattern already used for the currency/unit rows.

### templates/content-job_listing.php
**Before:**
```php
<li class="date"><?php the_job_publish_date(); ?></li>
```
**After:**
```php
<li class="date"><?php the_job_publish_date(); ?></li>

<?php if ( get_option( 'job_manager_show_salary_in_listings' ) ) : ?>
    <?php $job_salary = the_job_salary( '', '', false ); ?>
    <?php if ( ! empty( $job_salary ) ) : ?>
        <li class="salary"><?php echo esc_html( $job_salary ); ?></li>
    <?php endif; ?>
<?php endif; ?>
```
**Why:** Renders the salary in the listings loop when the new setting is on, using the same `the_job_salary()` helper and `.salary` CSS class already used on the single job page, so no new styling is needed. Placed after the date `<li>` so cards match the ordering of the single-page meta template (types → date → salary).

### includes/class-wp-job-manager-data-cleaner.php
Added the new option to the uninstall cleanup list so it gets removed along with the other salary-related options.

## Testing
**Test 1: Setting off (default)**
1. Enable Salary and set a salary value on a job listing, leave "Salary on Listings" off.
2. View the `[jobs]` listings page.
Result: no salary shown on listing cards, single job page salary display unaffected.

**Test 2: Setting on**
1. Turn on "Salary on Listings" in Job Listings > Settings.
2. View the `[jobs]` listings page.
Result: salary shows on each listing card that has a salary value, positioned after the date `<li>` to match single-page order. Listings without a salary value show no empty item.

Added PHPUnit coverage in `tests/php/tests/includes/test_class.job-listing-salary-display.php` for both states plus the no-salary-value case. Full suite (561 tests) passes locally via `make test`. PHPCS clean.

### DIFF
--- a/includes/admin/class-wp-job-manager-settings.php
+++ b/includes/admin/class-wp-job-manager-settings.php
@@ -336,6 +336,16 @@ class WP_Job_Manager_Settings {
 							'track'      => 'value',
 						],
 						[
+							'name'       => 'job_manager_show_salary_in_listings',
+							'std'        => '0',
+							'label'      => __( 'Salary on Listings', 'wp-job-manager' ),
+							'cb_label'   => __( 'Show Job Salary on Listings Page', 'wp-job-manager' ),
+							'desc'       => __( 'This displays the job salary on the job listings page, in addition to the single job page.', 'wp-job-manager' ),
+							'type'       => 'checkbox',
+							'attributes' => [],
+							'track'      => 'bool',
+						],
+						[
 							'name'     => 'job_manager_display_location_address',
 							'std'      => '0',
 							'label'    => __( 'Location Display', 'wp-job-manager' ),
@@ -809,6 +819,7 @@ class WP_Job_Manager_Settings {
 			var $job_manager_enable_salary_unit = jQuery('#setting-job_manager_enable_salary_unit');
 			var $job_manager_default_salary_currency = jQuery('#setting-job_manager_default_salary_currency');
 			var $job_manager_default_salary_unit = jQuery('#setting-job_manager_default_salary_unit');
+			var $job_manager_show_salary_in_listings = jQuery('#setting-job_manager_show_salary_in_listings');
 
 			jQuery('#setting-job_manager_enable_registration').change(function(){
 				var $job_manager_enable_registration_is_checked = jQuery( this ).is(':checked');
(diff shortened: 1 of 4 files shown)


donnchawp:

Good — reuses `the_job_salary()` rather than reinventing currency formatting, escapes with
`esc_html()`, is settings-gated, added to the data cleaner, and has a real template-level test.
Small things:

**Needs fixing**

1. `templates/content-job_listing.php` changed markup but `@version` is still `1.34.0` — please
   bump it (this is the template-header convention that signals theme authors their override is
   stale; e.g. commit `4573580f`).
2. Design choice worth a deliberate call: the salary is hard-coded inline rather than hooked onto
   the existing `job_listing_meta_end` action the template already fires. Sites that have copied
   `content-job_listing.php` into their theme won't see the feature even after enabling the
   setting. The adjacent `job_manager_enable_types` block does the same inline thing, so it's
   consistent with local precedent — just flag it as intentional.
3. Minor: salary renders before "date" on cards but after "date posted" on the single page.

(For your awareness, not this PR's job: `the_job_salary()` gates on the meta being non-empty, not
on `job_manager_enable_salary`, so disabling the feature doesn't hide already-saved salaries.)

faisalahammad:
**#1 @version bump.** Updated `templates/content-job_listing.php` line 12 from `1.34.0` to `$$next-version$$` per repo placeholder convention. No behavioral change.

faisalahammad:
**#2 inline vs hook.** Kept inline. Reasoning: the adjacent `job_manager_enable_types` block at `templates/content-job_listing.php:45-50` is already inline with the same `get_option()` gating, and `templates/content-single-job_listing-meta.php` is inline for salary too. Hooking only this one would create asymmetry between the two view templates. Happy to switch if you prefer.

faisalahammad:
**#3 ordering.** Moved the salary `<li>` to sit between `<li class="date">` and `do_action( 'job_listing_meta_end' )` in `templates/content-job_listing.php`. Loop order now matches the single-page meta template exactly: types → date → salary. One file touched.

faisalahammad:
**#4 the_job_salary() gating.** Acknowledged. Out of scope for this PR: the loop-side gate added here is independent of the meta-presence check in `the_job_salary()`. Will address `the_job_salary()` in a separate change so disable-existing-salary behavior lands on its own review.

PR #3044, opened 2026-07-11
Fix [jobs] shortcode adding extra elements in block themes

### LINKED ISSUE #2521: Jobs shortcode adds extra elements when added through Site Editor
**Describe the bug**
When adding the Jobs shortcode `[jobs show_filters=false]` to a template in the Site Editor, it adds extra elements.

**To Reproduce**
1. Go to the Site Editor
2. Modify a template
3. Add the `[jobs show_filters=false]` shortcode *the `[jobs]` shortcode itself seems fine*
4. See the extra `p` and `a` tags under each element.

**Screenshots**

faisalahammad:
Fixes #2521

* In block themes, `get_the_block_template_html()` runs `do_shortcode()` before `do_blocks()`. The `core/shortcode` block then runs `wpautop()` on listing HTML that already contains block elements nested inside anchors, which injected stray `<p>` and `<br>` tags under each listing card.
* Add a hard-coded `data-wp-job-manager-jobs-shortcode="1"` marker on the `[jobs]` wrapper and short-circuit `pre_render_block` for `core/shortcode` blocks that hold expanded `[jobs]` markup. This bypasses `render_block_core_shortcode()`'s `wpautop()` call so the already-rendered listings stay intact.
* The marker lives on the wrapper itself, outside the filterable `data-*` attributes, so detection does not depend on shortcode attributes or class names.

### Testing Instructions

1. Use a block theme (e.g. Twenty Twenty-Four) with at least one published job listing.
2. In the Site Editor, add a Shortcode block with `[jobs show_filters=false]` to a template and save.
3. View the front end.
4. Confirm listings render cleanly with no extra empty `<p>` tags or broken/orphaned `<a>` elements under each card.
5. Add `[jobs]` (with filters) as a second block and confirm the filter form and AJAX loading still work.

Run `vendor/bin/phpunit --filter WP_Test_WP_Job_Manager_Shortcodes` (4 tests, 10 assertions pass).

### Release Notes

* Fixed the `[jobs]` shortcode adding extra paragraph and anchor elements when added through a Shortcode block in block themes.

* None.

### Deprecated Code

* None.

### Screenshot / Video

Not needed; this is a markup-only fix with no visible UI change beyond the removed extra elements.

### DIFF
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -38,7 +38,41 @@ class WP_Job_Manager_Shortcodes {
 		add_shortcode( 'job_summary', [ $this, 'output_job_summary' ] );
 		add_shortcode( 'job_apply', [ $this, 'output_job_apply' ] );
 
+		// Block themes expand shortcodes before do_blocks(), so the core/shortcode block's
+		// wpautop() would run on our already-rendered HTML and inject stray p/br tags.
+		add_filter( 'pre_render_block', [ $this, 'protect_jobs_shortcode_from_wpautop' ], 10, 2 );
+
 		Job_Dashboard_Shortcode::instance();
+	}
+
+	/**
+	 * Skip core/shortcode wpautop when the block already holds expanded [jobs] markup.
+	 *
+	 * In block templates, get_the_block_template_html() runs do_shortcode() before
+	 * do_blocks(). By the time the shortcode block renders, [jobs] is already HTML.
+	 * Returning that HTML here bypasses render_block_core_shortcode()'s wpautop() call.
+	 *
+	 * @param string|null $pre_render   Existing pre-render value.
+	 * @param array       $parsed_block Parsed block.
+	 * @return string|null
+	 */
+	public function protect_jobs_shortcode_from_wpautop( $pre_render, $parsed_block ) {
+		if ( null !== $pre_render ) {
+			return $pre_render;
+		}
+
+		if ( empty( $parsed_block['blockName'] ) || 'core/shortcode' !== $parsed_block['blockName'] ) {
+			return $pre_render;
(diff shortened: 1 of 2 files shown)


donnchawp:

This looks good to me. The `pre_render_block` short-circuit is the right approach — it bypasses
only `render_block_core_shortcode()`'s `wpautop()`, doesn't touch the global filter chain, and
the root cause (block templates running `do_shortcode()` before `do_blocks()`) checks out against
core. The end-to-end test is real. Only optional polish:

1. The marker check at `includes/class-wp-job-manager-shortcodes.php:73` is a bare `strpos()`
   anywhere in the block HTML rather than anchored to the wrapper's opening tag. Since listing
   titles/company names are employer-supplied and land in that HTML, someone could embed the
   literal marker string as visible text and make an unrelated `core/shortcode` block skip
   `wpautop()`. Cosmetic only (skipping `wpautop` never un-escapes anything), but worth anchoring
   the check.
2. Scoped to `[jobs]` only — `[job]`, `[job_summary]`, `[job_apply]`, `[submit_job_form]` hit the
   same ordering issue in block themes. Fine for this PR's scope; worth a follow-up to generalise
   the marker into a shared helper.

Nothing blocking.

faisalahammad:
Thanks for the review.

Pushed commit 7d8a9eb6 anchoring the marker in `pre_render_block` to the opening tag:

The check now only short-circuits when the marker appears before the first `>` in the parsed block innerHTML, so an employer-supplied title or company name containing the literal string cannot trigger the bypass on unrelated `core/shortcode` blocks.

Added one regression test (`test_marker_in_body_text_does_not_short_circuit`) that asserts `null` when the marker sits inside body text rather than a tag attribute.

Re-review welcome.

Addresses PR feedback.

Issue #2791, opened 2024-03-18
Mark Job as taken, change status for all associated applications


tobias992:
### Describe the solution you'd like
If an employer marks a job as filled, then all existing applications for this job should be changed in status. At the moment, unfortunately, nothing is happening that is badly solved and should be improved. The applicants do not even realise when a job is taken.

PR #3046, opened 2026-07-12
fix(post-types): fire action when job filled status changes

faisalahammad:
## Summary

Added `job_manager_job_listing_filled_status_changed` action. Fires when `_filled` meta changes (0→1 or 1→0). Core has no applications system; this hook enables the Applications add-on to update application statuses when a job is marked filled.

Refs #2791

Added `case '_filled':` branch to `update_post_meta()` handler. Compares old/new values and fires action only on real transition. Also added an `add_post_meta` handler that fires the same action on the first transition when `_filled` is inserted with a truthy value (covers paths where the meta row does not exist before being set). Both gated to `job_listing` posts.

### tests/php/tests/includes/test_class.wp-job-manager-post-types.php
Added `test_update_post_meta_fired_on_filled_transition()` verifying:
- 0→1 transition fires once
- 1→0 transition fires once  
- Same-value re-save does not fire

Added `test_update_post_meta_fired_on_filled_add_transition()` verifying the new `add_post_meta` handler:
- Inserting `_filled = 1` on a listing without prior meta fires the action once
- Inserting `_filled = 0` does not fire (matches seed-path semantics)

## Testing

**Test 1: Hook fires on filled toggle**
1. Add `add_action('job_manager_job_listing_filled_status_changed', ...)` listener
2. Mark job filled via dashboard or admin bulk edit
3. Check debug.log: should see "filled: was=0 now=1"

**Test 2: No fire on same value**
1. Edit job without touching filled checkbox
2. Save
3. Hook does not fire

**Test 3: PHPCS**
Run `npm run lint:php -- --standard=phpcs.xml.dist` – passes.

**Test 4: PHPUnit**
Run the full `tests/php` suite – `test_update_post_meta_fired_on_filled_transition` and `test_update_post_meta_fired_on_filled_add_transition` both pass.

## Screenshots

No UI changes.

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -204,6 +204,7 @@ class WP_Job_Manager_Post_Types {
 
 		add_filter( 'wp_insert_post_data', [ $this, 'fix_post_name' ], 10, 2 );
 		add_action( 'add_post_meta', [ $this, 'maybe_add_geolocation_data' ], 10, 3 );
+		add_action( 'add_post_meta', [ $this, 'add_post_meta' ], 10, 3 );
 		add_action( 'update_post_meta', [ $this, 'update_post_meta' ], 10, 4 );
 		add_action( 'wp_insert_post', [ $this, 'maybe_add_default_meta_data' ], 10, 2 );
 		add_filter( 'post_types_to_delete_with_user', [ $this, 'delete_user_add_job_listings_post_type' ] );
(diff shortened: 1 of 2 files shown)


donnchawp:

The transition detection is correct — `update_post_meta` fires while the old value is still
cached, it's gated to `job_listing`, and no-op re-saves are correctly skipped. Small things:

**Needs fixing**

1. This is an enabling hook for an out-of-tree add-on and doesn't change any user-visible
   behaviour on its own, so `Fixes #2791` would auto-close an issue the applicant-facing part of
   which is still unresolved. Suggest `Refs #2791`, or land the add-on side in the same window.
2. `@since 2.5.0` in the test docblock
   (`tests/…/test_class.wp-job-manager-post-types.php:799`) → `$$next-version$$`.
3. The new public hook `job_manager_job_listing_filled_status_changed`
   (`includes/class-wp-job-manager-post-types.php:1649-1655`) has no `@since` — consumers need the
   minimum-version signal; line 1338 of the same file shows the convention.
4. It's only hooked to `update_post_meta`, which fires when the meta row already exists. If
   `_filled` is absent, core fires `add_post_meta` instead and the transition is missed. All
   current creation paths seed `_filled = 0`, so it's low-risk today, but since a downstream
   add-on depends on this firing reliably, consider hooking `add_post_meta` too.

faisalahammad:
Thanks for the triage notes. All four points addressed in db2dd52:

1. PR body switched Fixes -> Refs #2791.
2. Hook docblock now carries @since $$next-version$$ on the do_action.
3. Test @since swapped to $$next-version$$.
4. Added WP_Job_Manager_Post_Types::add_post_meta handler on add_post_meta action. Mirrors the existing update_post_meta _filled branch and fires the same hook with was_filled=false when _filled is inserted with truthy value. Seed paths write 0 so the truthy guard keeps them silent.

Added test_update_post_meta_fired_on_filled_add_transition covering truthy and falsy paths.

PR #3047, opened 2026-07-12
feat: group jobs by status in employer job dashboard

### LINKED ISSUE #2742: Group jobs by status in the job dashboard
Active listings: status is published
Pending listings: status is pending approval, draft
Inactive listings: status is rejected, expired, filled?

faisalahammad:
## Summary

Adds `group_by_status` shortcode attribute to `[job_dashboard]`. When enabled, jobs render in Active / Pending / Inactive collapsible sections instead of a flat paginated list. Closes #2742.

### `includes/class-job-dashboard-shortcode.php`
- Added `group_by_status` shortcode att (default `no`, parsed to bool)
- When grouping, fetches all jobs (`posts_per_page = -1`) since flat pagination is incoherent across 3 groups
- Guard after filter reasserts `-1` in case `job_manager_get_dashboard_jobs_args` overrides
- New `group_jobs_by_status()` private method buckets posts: `publish`/`future` → active, `pending`/`pending_payment`/`draft`/`preview` → pending, `expired` → inactive
- Template args pass `group_by_status`, `job_groups`, and `max_num_pages = 1` when grouping

### `templates/job-dashboard.php`
- Extracted row markup into `$render_row` closure (single source of truth)
- Grouped branch: iterates groups in order, renders `<details>` with `<summary>` header (label + count), column headers, then rows
- Active/Pending default `open`; Inactive collapsed; empty groups skipped
- Flat branch unchanged but uses shared `$render_row`

### `assets/css/job-dashboard.scss`
- `.jm-dashboard-group` styles: hidden native marker, chevron indicator via `--jm-ui-svg-arrow-down` rotated on `[open]`, reuse existing design tokens

## Testing

**Test 1: Grouped view**
1. Add `[job_dashboard group_by_status="yes"]` to a page
2. Log in as employer with mixed-status jobs
3. Expect: Active/Pending/Inactive sections, each with label and count
4. Click header to expand/collapse

**Test 2: Filled jobs**
1. Mark a published job as Filled
2. Expect: stays in Active section with Filled chip visible

**Test 3: Empty groups**
1. If employer has only published jobs
2. Expect: only Active section renders

**Test 4: Search**
1. Use search field
2. Expect: results filter across all groups; empty groups hidden

**Test 5: Backward compat**
1. Use `[job_dashboard]` (no att)
2. Expect: flat list with pagination, identical to pre-change behavior

## Screenshots
| Before | After |
| ------ | ------ |
|SCR-20260713-cdga|SCR-20260713-cdmp|

### DIFF
--- a/assets/css/job-dashboard.scss
+++ b/assets/css/job-dashboard.scss
@@ -211,3 +211,56 @@
 		flex: 1 1 calc(33% - 2 * var(--jm-ui-space-sm));
 	}
 }
+
+.jm-dashboard-group {
+	.jm-dashboard-header {
+		margin-top: 0;
+	}
+
+	&__summary {
+		cursor: pointer;
+		padding: var(--jm-ui-space-sm);
+		font-size: var(--jm-ui-font-size);
+		font-weight: 600;
+		display: flex;
+		align-items: center;
+		gap: var(--jm-ui-space-xs);
+
+		list-style: none;
+
+		&::-webkit-details-marker {
+			display: none;
+		}
+
+		&::after {
+			content: '';
+			margin-left: auto;
+			width: var(--jm-ui-icon-size-xs);
+			height: var(--jm-ui-icon-size-xs);
+			mask-image: var(--jm-ui-svg-arrow-down);
+			-webkit-mask-image: var(--jm-ui-svg-arrow-down);
+			mask-repeat: no-repeat;
+			-webkit-mask-repeat: no-repeat;
+			background-color: fadeCurrentColor(70%);
+			transition: transform 0.15s ease;
+			flex-shrink: 0;
+		}
+	}
+
+	&[open] &__summary::after {
+		transform: rotate(180deg);
+	}
+
+	&__count {
+		font-weight: 400;
+		color: fadeCurrentColor(70%);
+	}
+
+	&__rows {
+		margin: var(--jm-ui-space-sm) 0;
+	}
+
+	+ .jm-dashboard-group {
+		margin-top: var(--jm-ui-space-md);
+	}
+}
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -93,13 +93,15 @@ class Job_Dashboard_Shortcode {
 			return ob_get_clean();
 		}
 
-		$attrs          = shortcode_atts(
+		$attrs           = shortcode_atts(
 			[
-				'posts_per_page' => '25',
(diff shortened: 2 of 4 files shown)


donnchawp:

Nice that the non-grouped path is byte-identical and ownership scoping is preserved. Two
functional issues and some cleanup before merge, plus one design question.

**Needs fixing**

1. **The zero-listings empty state becomes unreachable when grouping is on.**
   `group_jobs_by_status()` always returns three keys (`active`/`pending`/`inactive`), so
   `! empty( $job_groups )` in `templates/job-dashboard.php:97` is true even with no listings, and
   the "submit your first listing" CTA in the `elseif ( ! $jobs )` branch never renders. A new
   employer sees an empty table wrapper. Suggest `if ( $group_by_status && $jobs )` and a test for
   the zero-jobs + grouping case.
2. **Grouping disables pagination and loads everything.** `posts_per_page` is forced to `-1` and
   `max_num_pages` hardcoded to `1` (`includes/class-job-dashboard-shortcode.php:127-140`). Since
   the motivating issue is dashboards with *many* listings, this makes that case worse (full read
   + large DOM, no pagination). Worth a design pass — a cap or per-group pagination.
3. Dead code at `:135-139` mutates `$jobs->query_vars` after the query already ran; can be
   removed.
4. `@since` / `@version` are hardcoded `2.5.0` (`job-dashboard.php:11,16`,
   `class-job-dashboard-shortcode.php:714`) — `@since` should be `$$next-version$$`.
5. Filled listings are silently grouped as Active (`_filled` is never inspected), and a test locks
   this in. The issue explicitly raised "inactive = rejected/expired/filled?" as an open
   question — worth calling out the decision rather than settling it silently.

faisalahammad:
All five points addressed in commit 1c9ac2d9:

1. Zero-listings empty state — added `\&\& $jobs` to template guard + new test_zero_listings_shows_empty_state_when_grouping.
2. Pagination cap — added `job_manager_job_dashboard_group_limit` filter (default 0 = unlimited). Integrators opt in to per-group truncation with array_slice. No per-group pagination (over-scope for this PR).
3. Dead query_vars mutation — removed. Was a no-op after WP_Query ran; query_args already pass posts_per_page => -1 during grouping.
4. Version tokens replaced with $$next-version$$ in both files.
5. Filled-in-Active decision documented in group_jobs_by_status() docblock: filled publish stays in active so the Filled chip stays visible until expiry. Matches the flat path behavior.

CodeRabbit passed (0 findings). zip + TESTING_INSTRUCTIONS.md ready.

PR #3048, opened 2026-07-13
fix: make job_listing_category and job_listing_type taxonomies public

### LINKED ISSUE #2414: Make wp job manager taxonomies available to other plugins like normal Wrdpress taxonomies
The taxonomies created by WP Job Manager do not behave like taxonomies created by other plugins such as Woocommerce or CPTUI / JetEngine etc... Normally, taxonomies are available to be used by other plugins such as page builders and Fakerpress but the WP Job Manager ones such as Job Category and Job Type are not.

When using third party plugins, (I shall use Fakerpress here as an example), the WP Job Manager taxonomies should be available to be used when generating content. Currently the default Wordpress categories and tags are available, and after installing Woocommerce the product tags and categories and shipping classes are available too. Also all of my other custom taxonomies I have created are available. The WP Job Manager Job Type and Job Categories should also be selectable.

I should add that I have installed the Job Tags add-on and the Job Tags taxonomy IS available.

### Additional context
Here's the available taxonomies when generating content using Fakerpress
Screenshot 2023-01-20 at 11 16 41

faisalahammad:
## Summary

Make WP Job Manager taxonomies (job_listing_category, job_listing_type) visible to third-party plugins. Previously their `public` flag was set to `false` when the active theme didn't declare `add_theme_support( 'job-manager-templates' )`, which hid them from plugins calling `get_taxonomies( [ 'public' => true ] )` (FakerPress, ACF, page builders, etc.).

Fixes #2414

**Before:**
```php
if ( current_theme_supports( 'job-manager-templates' ) ) {
    $rewrite = [...];
    $public  = true;
} else {
    $rewrite = false;
    $public  = false;
}
// ...
'public' => $public,
```

**After:**
```php
if ( current_theme_supports( 'job-manager-templates' ) ) {
    $rewrite = [...];
    $publicly_queryable = true;
} else {
    $rewrite = false;
    $publicly_queryable = false;
}
// ...
'public'             => true,
'publicly_queryable' => $publicly_queryable,
'show_in_nav_menus'  => $publicly_queryable,
```

**Why:** Taxonomy `public` is always `true` so plugins can discover it. The old theme-support condition is preserved on `publicly_queryable` so frontend taxonomy URLs still require theme support. `show_in_nav_menus` follows `publicly_queryable` to preserve nav menu items on template-supporting themes and prevent dead links on non-supporting themes.

Same change applied to both `job_listing_category` and `job_listing_type`.

## Testing

**Test 1: Third-party plugin visibility**
1. Install FakerPress (or similar plugin that lists taxonomies).
2. Check the taxonomy selector.
3. Result: `job_listing_category` and `job_listing_type` appear (were previously missing).

**Test 2: No regression - frontend URLs without theme support**
1. Use a theme without `add_theme_support( 'job-manager-templates' )`.
2. Visit a category/type archive URL.
3. Result: Returns 404 (same as before, controlled by `publicly_queryable => false`).

**Test 3: No regression - frontend URLs with theme support**
1. Use a theme with `add_theme_support( 'job-manager-templates' )`.
2. Visit a category/type archive URL.
3. Result: Loads normally (same as before).

**Test 4: Nav menus with/without theme support**
1. Use a theme with `add_theme_support( 'job-manager-templates' )`.
2. Go to Appearance > Menus.
3. Result: Job Categories and Job Types appear as menu options.
4. Switch to a theme without `job-manager-templates` support.
5. Result: Job Categories and Job Types do not appear as menu options.

**Test 5: Admin operations**
1. Go to Job Listings > Categories (or Types).
2. Add, edit, delete terms.
3. Result: All operations work as before.


donnchawp:

The core idea is right — `publicly_queryable` (not `public`) is the correct flag for gating
front-end archives, and switching `public` on is what lets `get_taxonomies(['public'=>true])`
pickers see these taxonomies. But the change also flips a nav-menu flag it shouldn't.

**Needs fixing**

1. **`show_in_nav_menus => false` is hardcoded, removing existing functionality.**
   (`includes/class-wp-job-manager-post-types.php:334` and `:399`.) Core derives
   `show_in_nav_menus` from `public` when it isn't set, so on themes with
   `job-manager-templates` support these taxonomies were previously available in
   Appearance → Menus. Hardcoding `false` removes that for template-supporting themes, which #2414
   didn't ask for and where there's no dead-link concern (`publicly_queryable` is true there).
   Please use `'show_in_nav_menus' => $publicly_queryable` at both sites. The PR's "Test 4"
   asserts the taxonomies *don't* appear as menu options, which currently encodes this regression
   as expected — worth updating to distinguish the theme-support case.
2. There's no test asserting `public` / `publicly_queryable` / `show_in_nav_menus` under both
   `current_theme_supports('job-manager-templates')` states. Given this changes long-stable
   registration defaults, that coverage would have caught (1).
3. `public => true` now ships to every install on upgrade. That matches the issue's intent and is
   overridable via the existing `register_taxonomy_*_args` filters, so it's fine — but flag it for
   a changelog note for anyone relying on the old "hidden unless theme support" behaviour.

faisalahammad:
Thanks @donnchawp — fixed in 183ada9f.

**`show_in_nav_menus` fix:** both `job_listing_category` and `job_listing_type` now pass `'show_in_nav_menus' => $publicly_queryable` instead of hardcoded `false`.
- With `add_theme_support( 'job-manager-templates' )` → nav menu shows the terms (matches old behavior on template-supporting themes).
- Without theme support → nav menu hides them (matches old behavior on plain themes).

**Changelog:** changelog file wasn't touched in this PR per repo convention (releases are generated from `$$next-version$$` placeholders); the change is captured in the tests and PR description.

**Tests added** (5):
- `job_listing_category` registration args with theme support — asserts `public=true, publicly_queryable=true, show_in_nav_menus=true`
- `job_listing_category` registration args without theme support — asserts `public=true, publicly_queryable=false, show_in_nav_menus=false`
- `job_listing_type` registration args with theme support
- `job_listing_type` registration args without theme support
- `get_taxonomies( [ 'public' => true ] )` includes both taxonomies (the FakerPress discovery path)

All 576 tests pass, CodeRabbit clean on re-run.

Could you take another look when you have a moment?

PR #3049, opened 2026-07-13
fix(wpml): make rewrite slugs stable for translation

### LINKED ISSUE #2354: Multilingual default Jobs page not translatable with WPML and WP Job Manager
**Describe the bug**
For our project, we use WPML in combination with WP Job Manager. As you know, a default jobs page can be set where all of the job widgets (such as filters) and the job listings will appear. In our case, this page is called "Stellenangebote" ("Jobs" in german), which has been translated with WPML and has the following urls: "www.websitename/stellenangebote" (in german) and "www.websitename/it/offerte-di-lavoro" (in italian). The [jobs] shortcode is added in both versions of the page. In german, which is the website's main language (set through WPML), everything is working fine. On the same page, we also have a language switcher, which can switch to the italian version of the page on click. But when a user clicks on the switcher, instead of going to the translated url ("www.websitename/offerte-di-lavoro") it goes to "www.websitename/it/stellenangebote", which is a page that does not even exits. On "www.websitename/it/stellenangebote", everything still works fine, but if we click on the "Offerte di lavoro" page from the main menu, it goes to the page "www.websitename/it/offerte-di-lavoro", which looks broken and has no job widgets (although it does have a job list with default styling).

Our objective is to have one page with different slugs in different languages. So, when a user is on "www.websitename/stellenangebote" and clicks on the language switcher, it should be directed to "www.websitename/it/offerte-di-lavoro", where the job widgets, styling, and list should appear and function exactly like in the german version.
We tried looking through your plugin's code, and, as disclosed in the following thread ( https://github.com/Automattic/WP-Job-Manager/issues/786 ), the issue should have been fixed and the plugin should be compatible with WPML as of now. So we don't understand what we've been doing wrong.

**To Reproduce**
1. Install WPML and WP Job Manager Plugins
2. Add a main and a secondary language to the website
3. Create some Jobs
4. Create and set a Jobs page (add [jobs] shortcode + add this page as the default Jobs page by clicking in admin menu on Jobs -> Pages -> Jobs page)
5. Create a translation of the Jobs page (click on the '+' near the page's title) for the secondary language
6. Change the url slug of the translated page (set anything, as long as it is different from the slug of the Jobs page in the main language)
7. Add a Job widget, f.e. a filter for a job category
8. Navigate to the default Jobs page in main language, everything should work fine and widget should be visible
9. Navigate to the url of the translated Jobs page, here wordpress does not show the widget and the page is styled as a default archive page
10. If you see a language switcher when on the main Jobs page, by clicking on the switcher you will be redirected to 'base url/secondary language code/same slug as the main jobs page'.

**Expected behavior**
By clicking the switcher, the user should be redirected to the url of the translated version of the jobs page. In addition, job widgets and styling should remain unchanged.

- [X ] I have deactivated other plugins and confirmed this bug occurs when only WP Job Manager plugin is active.
- [ ] This bug happens with a default WordPress theme active.
- [ X] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
- WordPress Version: 6.1.1
- WP Job Manager Version: 1.39
- PHP Version: 7.4

faisalahammad:
## Summary
WPML captures the source rewrite slug at `register_post_type()` / `register_taxonomy()` time and tries to translate it per language. WPJM derives the post type and taxonomy rewrite slugs at runtime from the `job_manager_permalinks` option, so the source WPML sees is dynamic and varies per site. Without a stable source WPML cannot register the post type or taxonomies as translatable, producing 404s on translated category and type URLs and breaking the AJAX request used by the `[jobs]` shortcode (variable mismatch error).

This adds a new `job_manager_permalink_structure` filter on the return value of `get_permalink_structure()` and registers a WPML-side callback that forces the four rewrite slugs to their stable defaults when WPML is active. Non-WPML sites are unaffected; the filter is a no-op without a registered callback.

Fixes #2354

### `includes/class-wp-job-manager-post-types.php`
**After:**
```php
/**
 * Filters the resolved permalink structure used to register the job_listing post type and its taxonomies.
 *
 * Third-party multilingual plugins (e.g. WPML) can use this to force stable rewrite slugs so they
 * can register the post type and taxonomies as translatable.
 *
 * @since 2.4.6
 *
 * @param array $permalinks The resolved permalink structure.
 */
return apply_filters( 'job_manager_permalink_structure', $permalinks );
```
**Why:** exposes the permalink array WPJM uses to register the post type and taxonomies, so multilingual plugins can pin a stable source slug for translation.

### `includes/3rd-party/wpml.php`
**After:**
```php
add_filter( 'job_manager_permalink_structure', 'wpml_wpjm_force_static_slugs' );

function wpml_wpjm_force_static_slugs( $permalinks ) {
    if ( empty( $permalinks['job_base'] ) ) {
        $permalinks['job_rewrite_slug'] = 'job';
    }
    if ( empty( $permalinks['category_base'] ) ) {
        $permalinks['category_rewrite_slug'] = 'job-category';
    }
    if ( empty( $permalinks['type_base'] ) ) {
        $permalinks['type_rewrite_slug'] = 'job-type';
    }
    if ( empty( $permalinks['jobs_archive'] ) ) {
        $permalinks['jobs_archive_rewrite_slug'] = 'job-listings';
    }
    return $permalinks;
}
```
**Why:** when WPML is active, force the four rewrite slugs to their defaults so WPML has a stable source to register against. Each slug is only overridden if the corresponding `_base` permalink option is empty, so any custom permalink value the admin set before installing WPML is preserved.

### `tests/php/tests/includes/test_class.wp-job-manager-post-types.php`
**Why:** covers the new filter both as a no-op (no callback registered, output equals input) and as an override (callback registered, output reflects the override).

## Testing
- `phpunit --filter test_get_permalink_structure` — 3 pass (existing + 2 new).
- `phpcs` — clean.
- `psalm` — clean on supported PHP versions (pre-existing local PHP 8.5 amphp/amp deprecation noise is unrelated; CI matrix covers 7.4 / 8.3 / 8.4).
- CodeRabbit — actionable finding (preserve configured `_base` values) applied.

## Manual repro
1. Install WPML Multilingual CMS, set primary language German, add Italian as secondary.
2. In WPML → Settings, mark `job_listing` and the two `job_listing_*` taxonomies as translatable.
3. Create a Jobs page (e.g. `stellenangebote`) containing the `[jobs]` shortcode, publish.
4. Translate the page to Italian with a different slug (e.g. `offerte-di-lavoro`).
5. Add a job category widget to the page sidebar.
6. Visit the page in the primary language: listings and widget render correctly.
7. Use the WPML language switcher to switch to Italian.

Before this fix: switcher lands on `…/it/stellenangebote/`, translated page renders as a default archive with no widgets, AJAX `?jm-ajax=get_job_listings` returns HTTP 400.

After this fix: switcher lands on `…/it/offerte-di-lavoro/`, translated page renders the same listings and widget, AJAX returns HTTP 200.

## Notes
- WPML users who previously customised `job_base` / `category_base` / `type_base` / `jobs_archive` and then install WPML will see those values reset to the defaults. This is unavoidable: WPML cannot translate a slug it never registered against. Sites that need custom slugs alongside WPML should set the permalinks after installing WPML, then translate the slug per language via WPML's slug translation UI.
- No changes to non-WPML users: the filter is a no-op without a registered callback, and the WPML callback is only registered on `wpml_loaded`.

### DIFF
--- a/includes/3rd-party/wpml.php
+++ b/includes/3rd-party/wpml.php
@@ -14,6 +14,7 @@ function wpml_wpjm_init() {
 	add_action( 'get_job_listings_init', 'wpml_wpjm_set_language' );
 	add_filter( 'wpjm_lang', 'wpml_wpjm_get_job_listings_lang' );
 	add_filter( 'wpjm_page_id', 'wpml_wpjm_page_id' );
+	add_filter( 'job_manager_permalink_structure', 'wpml_wpjm_force_static_slugs' );
 
 	$default_lang = apply_filters( 'wpml_default_language', null );
 	$current_lang = apply_filters( 'wpml_current_language', null );
@@ -74,6 +75,38 @@ function wpml_wpjm_get_job_listings_lang( $lang ) {
  */
 function wpml_wpjm_page_id( $page_id ) {
 	return apply_filters( 'wpml_object_id', $page_id, 'page', true );
+}
+
+/**
+ * Forces stable rewrite slugs so WPML can register the post type and taxonomies as translatable.
+ *
+ * WPML captures the source rewrite slug at `register_post_type()` / `register_taxonomy()` time. When
+ * the slug is dynamic (per-site permalink option), WPML cannot find a stable source to translate
+ * against, which breaks translated URLs and the AJAX request used by the [jobs] shortcode.
+ *
+ * See https://wpml.org/errata/wp-job-manager-job-incorrect-slug-translation-for-job-listings/ (compsupp-7570).
+ *
+ * @since 2.4.6
+ *
+ * @param array $permalinks The resolved permalink structure.
+ *
+ * @return array
+ */
+function wpml_wpjm_force_static_slugs( $permalinks ) {
+	if ( empty( $permalinks['job_base'] ) ) {
+		$permalinks['job_rewrite_slug'] = 'job';
+	}
+	if ( empty( $permalinks['category_base'] ) ) {
+		$permalinks['category_rewrite_slug'] = 'job-category';
+	}
+	if ( empty( $permalinks['type_base'] ) ) {
+		$permalinks['type_rewrite_slug'] = 'job-type';
+	}
+	if ( empty( $permalinks['jobs_archive'] ) ) {
+		$permalinks['jobs_archive_rewrite_slug'] = 'job-listings';
+	}
+
+	return $permalinks;
 }
 
 /**
--- a/includes/3rd-party/wpml.php
+++ b/includes/3rd-party/wpml.php
@@ -26,7 +26,7 @@ function wpml_wpjm_init() {
 }
 
 add_action( 'wpml_loaded', 'wpml_wpjm_init' );
-add_action( 'wpml_loaded', 'wpml_wpjm_set_language' );
+add_action( 'wpml_loaded', 'wpml_wpjm_set_language' ); // Intentional: see docblock above (WPML 3.7.1 ordering).
 
 /**
  * Sets WPJM's language if it is sent in the Ajax request.
@@ -86,7 +86,7 @@ function wpml_wpjm_page_id( $page_id ) {
  *
  * See https://wpml.org/errata/wp-job-manager-job-incorrect-slug-translation-for-job-listings/ (compsupp-7570).
  *
- * @since 2.4.6
+ * @since $$next-version$$
  *
  * @param array $permalinks The resolved permalink structure.
  *
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -1604,7 +1604,18 @@ class WP_Job_Manager_Post_Types {
 		if ( function_exists( 'restore_current_locale' ) && did_action( 'admin_init' ) ) {
 			restore_current_locale();
 		}
-		return $permalinks;
+
+		/**
+		 * Filters the resolved permalink structure used to register the job_listing post type and its taxonomies.
+		 *
+		 * Third-party multilingual plugins (e.g. WPML) can use this to force stable rewrite slugs so they
+		 * can register the post type and taxonomies as translatable.
+		 *
+		 * @since 2.4.6
+		 *
+		 * @param array $permalinks The resolved permalink structure.
+		 */
+		return apply_filters( 'job_manager_permalink_structure', $permalinks );
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


donnchawp:

The diagnosis is correct — WPML does need a locale-stable slug at registration time, and this
mirrors their documented workaround. But the forcing is unconditional, which breaks existing
sites.

**Needs fixing**

1. **This silently changes live permalinks on existing non-English WPML sites.** When `job_base`
   is empty the slug is recomputed per request from `_x()`
   (`includes/class-wp-job-manager-post-types.php:1591-1595`), so a non-English site is already
   serving a translated slug. `wpml_wpjm_force_static_slugs()`
   (`includes/3rd-party/wpml.php:94-104`) forces it back to English whenever `job_base` is empty,
   with no check for whether the effective slug already differs — so every indexed/bookmarked job
   URL on those sites changes with no redirect. Please only force the slug where it isn't already
   customised *and* isn't already diverging due to translation — e.g. force once on first WPML
   activation behind a persisted flag rather than on every request — or, at minimum, document this
   as a breaking change and gate it.
2. Activating WPML on an already-current install won't flush rewrite rules (that only happens on
   activate/version-bump), so the new slugs won't take effect until permalinks are re-saved. Add a
   one-time guarded flush or an admin notice, mirroring WPML's "re-save permalinks" step.
3. `@since 2.4.6` appears in four places (`wpml.php:89`,
   `class-wp-job-manager-post-types.php:1614`, tests `:762`/`:776`) — please use
   `$$next-version$$`.
4. The tests only assert `apply_filters()` is a pass-through; they don't call
   `wpml_wpjm_force_static_slugs()` itself, which is WPML-independent and directly unit-testable
   (defaults forced when bases empty, custom bases preserved when set).

faisalahammad:
Thanks for the review. Just pushed 2a2ce839 addressing (3) and (4):

- Four `@since 2.4.6` replaced with `$$next-version$$`.
- Filter docblock on `get_permalink_structure()` clarified.
- Inline comment added at `wpml.php` for the intentional double `wpml_loaded` registration.
- Direct unit tests added for `wpml_wpjm_force_static_slugs()` covering both defaults-forced-when-empty and custom-bases-preserved paths.

On (1) and (2): real concerns but feel like follow-ups rather than blockers for this PR — the headline goal is making translated URLs resolve, which (1) and (2) don't regress. Happy to track them as separate issues if you'd rather keep this PR scoped tight; alternatively I can gate the force behind a persisted `wpml_static_slugs_applied` flag in this PR. Preferences?

PR #3050, opened 2026-07-13
fix(application): support multiple comma-separated email addresses

### LINKED ISSUE #2353: Set two or more “Email/URL candidature” for each new job in WP Job Manager
The application email field is only designed to hold one email address.
Using a , (comma) on the frontend submission doesn't work. However, when I add two email addresses (ex: [name@domain.com](mailto:name@domain.com), [anothername@domain.com](mailto:anothername@domain.com)) on the backend, WP Job Manager tries to parse one of it as a URL. I'm not able to add two email addresses to the field.

I need to add two email addresses to the field.

I've no alternative solutions...

### Additional context
Screenshot 2022-12-13 alle 21 24 15

faisalahammad:
## Summary

The _application meta field on job listings only accepted a single email address. Comma-separated values were rejected by frontend validation, and the admin save path persisted a second email as a malformed URL because the resolver only checked `is_email()` on the whole string and otherwise prepended `http://`.

Splits the value on `,` / `;` in three places (the resolver, the meta sanitizer, and the submit form validator) so a job can be applied to via two or more email addresses. The default branch in the form (allow either email or URL) still accepts a single URL or a single email as before.

Fixes #2353

### wp-job-manager-template.php — `get_the_job_application_method()`
**Before:**
```php
if ( strstr( $apply, '@' ) && is_email( $apply ) ) {
    $method->type      = 'email';
    $method->raw_email = $apply;
    $method->email     = antispambot( $apply );
} else {
    if ( strpos( $apply, 'http' ) !== 0 ) {
        $apply = 'http://' . $apply;
    }
    $method->type = 'url';
    $method->url  = $apply;
}
```
**After:**
```php
$email_tokens     = preg_split( '/[,;]\s?/', $apply );
$non_empty_tokens = [];
foreach ( $email_tokens as $token ) {
    if ( '' !== trim( $token ) ) {
        $non_empty_tokens[] = $token;
    }
}
$valid_emails = [];
foreach ( $non_empty_tokens as $token ) {
    if ( is_email( trim( $token ) ) ) {
        $valid_emails[] = trim( $token );
    }
}

if ( ! empty( $valid_emails ) && count( $valid_emails ) === count( $non_empty_tokens ) ) {
    $joined         = implode( ', ', $valid_emails );
    $method->type      = 'email';
    $method->raw_email = $joined;
    $method->email  = antispambot( $joined );
} else {
    if ( strpos( $apply, 'http' ) !== 0 ) {
        $apply = 'http://' . $apply;
    }
    $method->type = 'url';
    $method->url  = $apply;
}
```
**Why:** The resolver was the choke point: any value that wasn't a single `is_email()`-valid string fell through to the URL branch, so `hr@x.com, jobs@x.com` saved as a `http://...` URL. The split + per-token check accepts the value when every non-empty token is a valid email.

### includes/class-wp-job-manager-post-types.php — `sanitize_meta_field_application()`
**Before:**
```php
public static function sanitize_meta_field_application( $meta_value ) {
    if ( is_email( $meta_value ) ) {
        return sanitize_email( $meta_value );
    }
    return self::sanitize_meta_field_url( $meta_value );
}
```
**After:**
```php
public static function sanitize_meta_field_application( $meta_value ) {
    $tokens       = preg_split( '/[,;]\s?/', (string) $meta_value );
    $valid_emails = [];
    foreach ( $tokens as $token ) {
        $token = trim( $token );
        if ( is_email( $token ) ) {
            $valid_emails[] = sanitize_email( $token );
        }
    }

    if ( ! empty( $valid_emails ) ) {
        return implode( ', ', $valid_emails );
    }

    return self::sanitize_meta_field_url( $meta_value );
}
```
**Why:** The registered `sanitize_callback` is invoked automatically on REST and block-editor writes. Splitting and dropping invalid tokens keeps the stored meta valid for any code path that consumes the resolver or queries the meta directly.

### includes/forms/class-wp-job-manager-form-submit-job.php — new helper + validate_fields()
Adds an `is_valid_email_list()` helper that splits on `,` / `;` and returns true only when every token is a valid email. The strict-email branch and the default (email-or-URL) branch in `validate_fields()` now both use it. The strict-URL branch is unchanged.

**Why:** Frontend submission was rejecting `hr@x.com, jobs@x.com` outright. The helper mirrors the resolver/sanitizer shape so a value that would round-trip cleanly through storage is also accepted at the form layer.

## Testing

**Test 1: Resolver accepts multi-email values**
1. `make test-up && make test`
2. New test file `tests/php/tests/includes/test_class.get-the-job-application-method.php` covers: single email, two emails via comma, two via semicolon, three via comma, URL regression, bare-string URL, mixed valid+invalid falling back to URL, empty.
3. Result: all 9 new cases pass.

**Test 2: Sanitizer round-trips multi-email values**
1. Existing `test_sanitize_meta_field_application` extended with 4 new cases: two emails via comma, two via semicolon, mixed valid+invalid (valid kept, invalid dropped), extra whitespace dropped.
2. Result: all cases pass.

**Test 3: Full suite + lint**
1. `make test` → 579 tests, 2110 assertions, 0 failures (3 pre-existing geocode skips unrelated to this fix).
2. `./vendor/bin/phpcs` on the five changed files → 0 errors, 0 warnings.
3. `coderabbit review --agent` → 0 findings (one minor edge case with the `"0"` token and PHP truthiness was fixed before re-review).

## Known caveat (out of scope)

`mailto:` URIs accept comma-separated addresses per RFC 6068. Desktop mail clients open a compose window with multiple recipients; some webmail handlers treat the comma as a header boundary and pass only the first address. This is a constraint of the `mailto:` scheme and the user's mail client, not the plugin.

The `sanitize_meta_field_application()` change is correct and used by REST / block-editor writes, but the current admin meta box and frontend form save paths call `update_post_meta()` directly, which does not invoke `register_meta`'s `sanitize_callback`. The resolver at `get_the_job_application_method()` handles un-sanitized input correctly, so the reported behaviour is fixed in practice. A follow-up to route saves through the registered sanitizer is a separate hardening pass.

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -544,6 +544,11 @@ abstract class WP_Job_Manager_Form {
 		if ( 'url' === $sanitizer ) {
 			return esc_url_raw( $this->normalize_url( $value ) );
 		} elseif ( 'email' === $sanitizer ) {
+			$emails = $this->parse_email_list( $value );
+			if ( null !== $emails ) {
+				return implode( ', ', $emails );
+			}
+
 			$sanitized_value = sanitize_email( $value );
 			if ( $sanitized_value !== $value ) {
 				return '';
@@ -551,6 +556,11 @@ abstract class WP_Job_Manager_Form {
 
 			return $sanitized_value;
 		} elseif ( 'url_or_email' === $sanitizer ) {
+			$emails = $this->parse_email_list( $value );
+			if ( null !== $emails ) {
+				return implode( ', ', $emails );
+			}
+
 			if ( is_email( $value ) ) {
 				return sanitize_email( $value );
 			}
@@ -563,6 +573,55 @@ abstract class WP_Job_Manager_Form {
 
 		// Use standard text sanitizer.
 		return sanitize_text_field( wp_unslash( $value ) );
+	}
+
+	/**
+	 * Parse a value as a comma- or semicolon-separated email address list.
+	 *
+	 * Returns the list of trimmed, sanitized emails when every non-empty
+	 * token is a valid email address and the list is at most the configured
+	 * cap. Returns null when the value is a single email, a URL, or empty —
+	 * signalling callers to fall back to single-value sanitization.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @param string $value Raw input.
+	 * @return array|null   List of sanitized emails, or null if value is not a list.
+	 */
+	protected function parse_email_list( $value ) {
+		if ( '' === $value || false === strpbrk( $value, ',;' ) ) {
+			return null;
+		}
+
+		/**
+		 * Maximum number of email addresses accepted in a single
+		 * application field value.
+		 *
+		 * @since $$next-version$$
+		 *
+		 * @param int $max_addresses Address cap.
+		 */
+		$max_addresses = (int) apply_filters( 'job_manager_application_max_addresses', 10 );
+
+		$tokens = preg_split( '/[,;]\s?/', $value );
+		$emails = [];
+
+		foreach ( $tokens as $token ) {
+			$token = trim( $token );
+			if ( '' === $token ) {
+				continue;
+			}
+			if ( ! is_email( $token ) ) {
+				return null;
+			}
+			$emails[] = sanitize_email( $token );
+		}
+
+		if ( empty( $emails ) || count( $emails ) > $max_addresses ) {
+			return null;
+		}
+
+		return $emails;
 	}
 
 	/**
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -2187,8 +2187,17 @@ class WP_Job_Manager_Post_Types {
 	 * @return string
 	 */
 	public static function sanitize_meta_field_application( $meta_value ) {
-		if ( is_email( $meta_value ) ) {
-			return sanitize_email( $meta_value );
+		$tokens       = preg_split( '/[,;]\s?/', (string) $meta_value );
+		$valid_emails = [];
+		foreach ( $tokens as $token ) {
+			$token = trim( $token );
+			if ( is_email( $token ) ) {
+				$valid_emails[] = sanitize_email( $token );
+			}
+		}
+
+		if ( ! empty( $valid_emails ) ) {
+			return implode( ', ', $valid_emails );
 		}
 
 		return self::sanitize_meta_field_url( $meta_value );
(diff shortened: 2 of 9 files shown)


donnchawp:

The resolver/admin side of this works, but as written the reported problem (#2353) isn't fixed on
the frontend submission form, which is the path the issue is actually about.

**Needs fixing**

1. **Frontend submission still rejects a comma-separated list.** `sanitize_posted_field()`
   (`includes/abstracts/abstract-wp-job-manager-form.php:547-559`) runs *before*
   `validate_fields()` where the new `is_valid_email_list()` check lives. In `email` mode,
   `sanitize_email('a@x.com, b@x.com')` collapses to an invalid string, the field is emptied, and
   validation throws the same "valid application email" error from the issue. In the default
   `url_or_email` mode it falls into the URL branch and fails there instead. Please add an
   email-list-aware branch to `sanitize_posted_field()` for both the `email` and `url_or_email`
   sanitizers, and add a test that POSTs a list through the submission form (the current tests
   exercise the resolver/sanitizer directly and miss this).
2. `is_valid_email_list()` (`includes/forms/class-wp-job-manager-form-submit-job.php:436-441`)
   needs an `@since $$next-version$$` tag.
3. Placeholders/labels still say `you@example.com` — nothing tells users multiple addresses are
   accepted, so the feature stays hidden even once (1) is fixed.
4. Two code paths disagree on partial lists: `sanitize_meta_field_application()` silently drops
   invalid tokens and keeps the rest, while `get_the_job_application_method()` rejects the whole
   list. Please pick one policy (all-or-nothing, to match the frontend, reads best) and apply it
   in both places. Consider a sane cap on address count, and use a bare `,` (no space) separator
   in the `mailto:` for RFC 6068.

faisalahammad:
Thanks for the careful review. Addressed all four points in this push.

1. Added email-list-aware branches to both `email` and `url_or_email` sanitizers in `abstract-wp-job-manager-form.php`. The `is_valid_email_list()` check was never reached because `sanitize_posted_field()` ran first and discarded the list. Added `tests/php/tests/includes/test_class.sanitize-posted-field-email-list.php` with 13 cases: list accepted, semicolon accepted, partial list rejected, single email preserved, url_or_email list accepted, partial url_or_email falls through to URL. Cap enforced at 10.

2. `is_valid_email_list()` in `class-wp-job-manager-form-submit-job.php` now has `@since $next-version$.` The `parse_email_list()` helper and the meta sanitizer callback also carry the same tag.

3. Updated placeholders in both frontend form and admin meta box to `'email1@example.com, email2@example.com'`.

4. Aligned policy across all three layers. `sanitize_meta_field_application()` is now all-or-nothing: if any token fails `is_email()` or the count exceeds 10, the whole value falls through to the URL sanitizer. The resolver and form validator were already strict. Cap defaults to 10, configurable via `job_manager_application_max_addresses` filter.

Bonus: `mailto:` href uses a bare comma per RFC 6068. Added `mailto_email` property, distinct from the display-ready `email` value.

Test run: 593 tests, 2113 assertions, 0 new failures. phpcs clean on project config.

Ready for re-review whenever you have time.

PR #3051, opened 2026-07-14
Add Workplace Type dropdown (On-Site / Remote / Hybrid)

### LINKED ISSUE #2310: Improve options for Job Posting Remote/Hybrid/On-Site
Instead of just a checkbox for Remote Position, drop down for Remote, Hybrid, On-Site as those options are becoming more common.

Using field editor plugin to add the option.

faisalahammad:
Fixes #2310

Replaces the boolean "Remote Position" checkbox with a Workplace Type dropdown (On-Site / Remote / Hybrid), backed by a new `job_listing_workplace_type` taxonomy that follows the existing `job_listing_type` pattern.

#### `includes/class-wp-job-manager-post-types.php`
Registers the new taxonomy (hierarchical, single-select via the same radio metabox pattern as job type), removes `_remote_position` from the field registry.

#### `includes/class-wp-job-manager-install.php`
Seeds the three default terms on fresh installs, and adds a versioned migration that backfills the taxonomy for existing listings from `_remote_position` postmeta on upgrade. The postmeta itself is left in place, not deleted.

#### `wp-job-manager-functions.php`
`get_job_listings()` gets a `workplace_types` tax_query arg. The legacy `remote_position` arg still works, now translated into an equivalent tax_query clause instead of a meta_query.

#### `templates/job-filters.php`, `assets/js/ajax-filters.js`
Filter bar's "Remote positions only" checkbox becomes a "Workplace type" dropdown (On-Site / Remote / Hybrid), same UI pattern as the category filter.

#### `includes/forms/class-wp-job-manager-form-submit-job.php`
Submission form field swapped from a checkbox to a term-select dropdown, defaulting to On-Site.

**Why:** a checkbox can only express two states, but employers need to distinguish On-Site, Remote, and Hybrid. Converting to a taxonomy (rather than a new meta field) reuses the admin UI, REST exposure, and tax_query filtering the plugin already has for `job_listing_type`.

### Backward compatibility
- `_remote_position` postmeta stays in the database, unused after migration.
- `get_job_listings( [ 'remote_position' => ... ] )`, `[jobs remote_position="..."]`, and the AJAX `remote_position` param all keep working.
- Recent Jobs widget's `remote_position` setting is unchanged.
- Promoted Jobs API `is_remote` now sources from the taxonomy (`remote` or `hybrid` both resolve to `true`), same boolean contract as before.

### Testing Instructions

**Test 1: Submission form**
1. Submit a new job listing.
2. Workplace type field shows a dropdown (On-Site / Remote / Hybrid), defaults to On-Site.
Result: listing saves with the correct term, location line shows "(Remote)" / "(Hybrid)" as appropriate.

**Test 2: Admin edit screen**
1. Edit a listing, change Workplace Type in the metabox, update.
2. Reload the edit screen.
Result: correct option is saved and pre-selected, no bogus terms created.

**Test 3: Upgrade migration**
1. On a site running the previous version, create listings with "Remote Position" checked and unchecked.
2. Upgrade, trigger `job_manager_migrate_workplace_type` via cron.
Result: checked listings become Remote, unchecked become On-Site. `_remote_position` postmeta untouched.

**Test 4: `[jobs]` filter bar**
1. Filter by each workplace type.
Result: only matching listings show.

**Automated:** `vendor/bin/phpunit` — 582 tests pass. `vendor/bin/phpcs` — clean.

### Release Notes

* Enhancement: Workplace Type dropdown (On-Site / Remote / Hybrid) replaces the Remote Position checkbox, with backfill migration for existing listings.

### Screenshot / Video
SCR-20260714-owns

### DIFF
--- a/assets/js/ajax-filters.js
+++ b/assets/js/ajax-filters.js
@@ -314,6 +314,7 @@ jQuery( document ).ready( function( $ ) {
 			var featured = $target.data( 'featured' );
 			var filled = $target.data( 'filled' );
 			var remote_position = $target.data( 'remote_position' );
+			var workplace_types = $target.data( 'workplace_types' );
 			var job_types = $target.data( 'job_types' );
 			var post_status = $target.data( 'post_status' );
 			var author = $target.data( 'author' );
@@ -354,11 +355,17 @@ jQuery( document ).ready( function( $ ) {
 					} )
 					.get();
 
+				var workplace_type = $form
+					.find( ':input[name^="workplace_type"]' )
+					.map( function() {
+						return $( this ).val();
+					} )
+					.get();
+
 				keywords = '';
 				location = '';
 				var $keywords = $form.find( ':input[name="search_keywords"]' );
 				var $location = $form.find( ':input[name="search_location"]' );
-				var $remote_position = $form.find( ':input[name="remote_position"]' );
 
 				// Workaround placeholder scripts
 				if ( $keywords.val() !== $keywords.attr( 'placeholder' ) ) {
@@ -367,10 +374,6 @@ jQuery( document ).ready( function( $ ) {
 
 				if ( $location.val() !== $location.attr( 'placeholder' ) ) {
 					location = $location.val();
-				}
-
-				if( $remote_position.length ) {
-					remote_position = $remote_position.is( ':checked' ) ? 'true' : null;
 				}
 
 				data = {
@@ -387,7 +390,7 @@ jQuery( document ).ready( function( $ ) {
 					page: page,
 					featured: featured,
 					filled: filled,
-					remote_position: remote_position,
+					workplace_type: workplace_type,
 					author: author,
 					show_pagination: $target.data( 'show_pagination' ),
 					form_data: $form.serialize(),
@@ -418,6 +421,7 @@ jQuery( document ).ready( function( $ ) {
 					featured: featured,
 					filled: filled,
 					remote_position: remote_position,
+					workplace_type: workplace_types,
 					author: author,
 					show_pagination: $target.data( 'show_pagination' ),
 				};
@@ -469,7 +473,7 @@ jQuery( document ).ready( function( $ ) {
 	}
 
 	$(
-		'#search_keywords, #search_location, #remote_position, .job_types :input, #search_categories, .job-manager-filter'
+		'#search_keywords, #search_location, #workplace_type, .job_types :input, #search_categories, .job-manager-filter'
 	)
 		.change( triggerSearch )
 		.on( 'keyup', function( e ) {
@@ -500,9 +504,10 @@ jQuery( document ).ready( function( $ ) {
 				.not( ':input[type="hidden"]' )
 				.prop( 'checked', true );
 			$form
-				.find( ':input[name="remote_position"]' )
+				.find( ':input[name^="workplace_type"]' )
 				.not( ':input[type="hidden"]' )
-				.prop( 'checked', false );
+				.val( '' )
+				.trigger( 'change.select2' );
 
 			$target.triggerHandler( 'reset' );
 			$target.triggerHandler( 'update_results', [ 1, false ] );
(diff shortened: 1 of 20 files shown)


donnchawp:

Really thorough change — the save path, structured-data `TELECOMMUTE` handling, legacy
`remote_position` bridging and the batched migration are all solid. But there's a blocker that
makes the feature invisible on new sites, plus two smaller issues.

**Needs fixing**

1. **Default terms are never created on a fresh install, so the feature is inert there.**
   `activate()` (`includes/class-wp-job-manager.php:165-171`) force-registers the job-type
   taxonomy by temporarily filtering `pre_option_job_manager_enable_types` to true before
   `register_post_types()`, but there's no equivalent filter for
   `job_manager_enable_remote_position`. So at activation the workplace-type taxonomy isn't
   registered, `default_terms()` calls `wp_insert_term()` against an unregistered taxonomy (which
   returns an unchecked `WP_Error`), and then `job_manager_installed_terms` is set to `1`
   permanently — so it never retries. `wp_count_terms()` stays 0, and the metabox, submission
   field and filter dropdown all drop out. Please add
   `pre_option_job_manager_enable_remote_position` alongside the existing filter in `activate()`.
   Worth knowing: `tests/php/bootstrap.php:73` sets this option before load, so the test suite
   doesn't exercise the real fresh-activation path — a test that registers via `activate()` would
   catch this.
2. **The shared REST filter now hides the job-type block-editor panel on multi-type sites.**
   The gate at `includes/class-wp-job-manager-post-types.php:230-232` was widened to
   `false === job_manager_multi_job_type() || get_option( 'job_manager_enable_remote_position' )`
   and the callback matches both taxonomies, so with workplace type enabled (the default) the
   filter registers unconditionally and also suppresses Gutenberg's native `job_listing_type`
   panel in multi-type mode — a regression. Please do the per-taxonomy condition inside the
   callback rather than at registration.
3. The migration threshold `'2.5.0'` (`includes/class-wp-job-manager-install.php:61`) guesses the
   next release; every other `version_compare` gate in that file targets an already-shipped
   version. And `templates/job-filters.php:48` passes `selected` without `sanitize_title()`, so a
   capitalised shortcode attr like `workplace_types="Remote"` won't pre-select the `remote` slug.

faisalahammad:
All three items fixed.

1. **Blocker — `activate()` missing `pre_option_job_manager_enable_remote_position`**. Added a paired `add_filter` / `remove_filter` around `register_post_types()` in `includes/class-wp-job-manager.php`, mirroring the existing `pre_option_job_manager_enable_types` pattern exactly. Fresh installs now register the workplace-type taxonomy, `default_terms()` succeeds, and the metabox / submission field / filter dropdown all appear as intended.

2. **Regression — REST filter widened registration**. Reverted `prepare_block_editor()` to the original single-condition gate (`false === job_manager_multi_job_type()`) in `includes/class-wp-job-manager-post-types.php`. The per-taxonomy evaluation moved into `hide_job_type_block_editor_selector()`: `TAX_LISTING_TYPE` is always hidden, `TAX_WORKPLACE_TYPE` is hidden only when the option is disabled, and the function early-returns on any context other than `edit`. So multi-type sites keep Gutenberg's native `job_listing_type` panel untouched, and the workplace-type behaviour tracks its own option independently.

3a. **Migration threshold `2.5.0`**. Replaced with `$$next-version$$` in `includes/class-wp-job-manager-install.php`, consistent with the release-script placeholder used elsewhere in the codebase.

3b. **`sanitize_title()` on filter dropdown**. Wrapped the `selected` value in `templates/job-filters.php` so `[jobs workplace_types="Remote"]` pre-selects the `remote` slug correctly.

Verification: 582 PHPUnit tests pass (3 pre-existing skipped Geocode tests need an API key). `phpcs` clean. No regressions found across `_remote_position` / `enable_remote_position` / `TAX_WORKPLACE_TYPE` legacy call sites.

Out-of-scope note (informational only, not addressed in this push): `assets/css/frontend.scss` lines 582/593/612 still target `.search_remote_position` / `#remote_position_label`, which the updated template no longer emits — dead rules, cosmetic, defer to a follow-up.

PR #3052, opened 2026-07-14
Honor allowed_mime_types on admin file fields

### LINKED ISSUE #2234: Allow to limit upload file types in admin panel for new job publications.
Currently we can't limit the file type in a file upload through the admin panel.
```
// Add "Attachment" field to Admin Panel.  
function admin_add_extra_fields( $fields ) {
  $fields['_job_attachment'] = array(
    'label'       => __( 'Attachment', 'job_manager' ),
    'type'        => 'file',
    'description' => '*pdf only',
  );
  return $fields;
}
add_filter( 'job_manager_job_listing_data_fields', 'admin_add_extra_fields' );
```

Allow to limit the file type accepted for fields with 'file' type.
Something like:
```
// Add "Attachment" field to Admin Panel.  
function admin_add_extra_fields( $fields ) {
  $fields['_job_attachment'] = array(
    'label'       => __( 'Attachment', 'job_manager' ),
    'type'        => 'file',
    'description' => '*pdf only',
    'allowed_mime_types' => [
        'pdf' => 'application/pdf'
    ]
  );
  return $fields;
}
add_filter( 'job_manager_job_listing_data_fields', 'admin_add_extra_fields' );
```

#### Check file types on backend when processing the data fields
When persisting the data fields do some validation against the `allowed_mime_types` property which would be included – this would be similar to what is done in the frontend (when submitting job):
```
// Add "Attachment" field to Frontend.
function frontend_add_extra_fields( $fields ) {
  $fields['job']['job_attachment'] = array(
    'label'       => __( 'Attachment', 'job_manager' ),
    'type'        => 'file',
    'required'    => false,
    'priority'    => 10,
	'description' => '*pdf only',
	'allowed_mime_types' => [
		'pdf' => 'application/pdf'
	]
  );
  return $fields;
}
add_filter( 'submit_job_form_fields', 'frontend_add_extra_fields' );
```

#### Limit file type in UI
Another option would be to limit the ui to select only the accepted file types.
An initial approximation could be adding `library: { type: "application/pdf" }` when creating the media frame in `assets/js/admin.js`. But this solution has some caveats: this change only applies to the media library selector but user can still upload and use  a new file with whatever type.
This is also happening, for example when selecting the company logo from the admin panel – you can only select images from the library but if you upload a new file any file will do. Looks like wp media library does not have this covered.

### Additional context
User report: p1651065850784059-slack-C6GGX896G

faisalahammad:
## Summary

The frontend job-submission form already enforces `allowed_mime_types` on `file` fields registered via `job_manager_job_listing_data_fields`, but the admin job-listing edit screen stored the raw posted value with no validation. Fields can now declare an `allowed_mime_types` whitelist and have it enforced on save, with the same semantics as the frontend. The change is opt-in: built-in fields (e.g. `_company_video`) and third-party fields without `allowed_mime_types` are unaffected.

Fixes #2234

### `includes/admin/class-wp-job-manager-writepanels.php`
- `save_job_listing_data()` now validates each posted value when the field is a `file` and declares `allowed_mime_types`. The check uses `wp_check_filetype()` against the field's whitelist, the same way the frontend `submit_job_form_fields` validator does. Query strings are stripped, numeric attachment IDs pass through unchanged, and disallowed URLs are not written to meta.
- A transient keyed by post id records any rejections for the next request. `maybe_show_file_type_errors()` renders a single `notice-error` with the field label, the rejected extensions, and the allowed list.
- `file_url_field()` and the "Add file" button emit `data-allowed_mime_types` (pipe-joined extension keys) on the upload button so the JS can filter the media library.

### `assets/js/admin.js`
- The `wp_job_manager_upload_file_button` click handler reads the new `data-allowed_mime_types` attribute and, when present, passes `library: { type: <mimes> }` to `wp.media({...})`. Because the frame is a reused singleton, the handler rebuilds it when the mime filter changes so each field keeps its own filter and there is no leak between fields.
- `file_frame.remove()` is used instead of `detach()` to drop event listeners between reuses.

### `tests/php/tests/includes/admin/test_class.wp-job-manager-writepanels-file-mime.php` (new)
- 6 tests: disallowed extension is rejected, allowed extension is stored, previously-stored meta is wiped on rejection (not overwritten with the bad URL), numeric attachment IDs bypass the mime check, query strings are stripped before the check, and pipe-separated extension keys are flattened in the notice.

## Why a transient + admin notice

The WP media library filter is a UX hint only — a user can still paste any URL into the field, and a third-party theme or browser extension can submit a form with an arbitrary value. The server-side check in the save loop is the real gate. When a save is rejected, the user needs a visible signal, so a transient keyed by post id is the lightest standard WP surface for a per-save error. The repo's `WP_Job_Manager_Admin_Notices` system is for stable dismissible notices and is the wrong tool here.

## Testing

**Manual (see `TESTING_INSTRUCTIONS.md` in the build artifact)**
1. Register a `file` field with `allowed_mime_types => [ 'pdf' => 'application/pdf' ]` via `job_manager_job_listing_data_fields`.
2. Save the listing with `https://example.com/evil.exe`. Expected: `notice-error` at the top, field cleared, no `_job_attachment` meta stored.
3. Save with `https://example.com/file.pdf`. Expected: no notice, field shows the URL, meta stored.
4. Save with `https://example.com/evil.exe?v=1`. Expected: same as step 2 (query string stripped).
5. Click the upload button. Expected: media library filter shows "PDF documents".
6. Without any custom code, save a job listing normally. Expected: built-in fields (e.g. `Company Video`) work unchanged.

**Automated**
- `npm test` — 577/577 pass, 3 skipped (Geocode, no API key)
- `npm run lint:php` — clean on changed files
- New test class: `WP_Test_WP_Job_Manager_Writepanels_File_Mime` (6 tests)

## Notes for reviewer

- Single-file mirror of the frontend check at `includes/forms/class-wp-job-manager-form-submit-job.php:502-510`. A shared helper could be extracted in a follow-up; intentionally kept the blast radius to the admin path to avoid churning tested frontend code.
- Frontend submit still has its own bypass at `class-wp-job-manager-form-submit-job.php:1122` (writes meta directly), but `validate_fields()` already covers it, so this PR is not a regression.
- The `data-allowed_mime_types` attribute is pipe-joined to match WP's `get_allowed_mime_types()` shape. `wp.media.library.type` expects a comma-separated extensions string, which the JS passes through.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -119,6 +119,21 @@ jQuery(document).ready(function($) {
 	var file_frame;
 	var file_target_input;
 	var file_target_wrapper;
+	var file_frame_mimes = null;
+
+	function get_field_mime_types( $button ) {
+		// data-allowed_mime_types may live on the upload button or, for "Add file" rows, on the "Add" button that was clicked to inject the row. Walk up to the field group to find it.
+		var raw = $button.data( 'allowed_mime_types' );
+		if ( ! raw ) {
+			var $field = $button.closest( '.form-field' );
+			raw = $field.length ? $field.data( 'allowed_mime_types' ) : null;
+		}
+		if ( ! raw ) {
+			return null;
+		}
+		// wp.media library.type accepts a comma-separated extensions string. Server-side save still enforces the type whitelist; this is a UX hint.
+		return String( raw );
+	}
 
 	$( document.body ).on('click', '.wp_job_manager_add_another_file_button', function( event ){
 		event.preventDefault();
@@ -128,8 +143,10 @@ jQuery(document).ready(function($) {
 		var button_text       = $( this ).data( 'uploader_button_text' );
 		var button            = $( this ).data( 'uploader_button' );
 		var view_button       = $( this ).data( 'view_button' );
+		var allowed_mimes     = $( this ).data( 'allowed_mime_types' );
+		var mime_attr         = allowed_mimes ? ' data-allowed_mime_types="' + allowed_mimes + '"' : '';
 
-		$( this ).before( '<span class="file_url"><input type="text" name="' + field_name + '[]" placeholder="' + field_placeholder + '" /><button class="button button-small wp_job_manager_upload_file_button" data-uploader_button_text="' + button_text + '">' + button + '</button><button class="button button-small wp_job_manager_view_file_button">' + view_button + '</button></span>' );
+		$( this ).before( '<span class="file_url"><input type="text" name="' + field_name + '[]" placeholder="' + field_placeholder + '" /><button class="button button-small wp_job_manager_upload_file_button"' + mime_attr + ' data-uploader_button_text="' + button_text + '">' + button + '</button><button class="button button-small wp_job_manager_view_file_button">' + view_button + '</button></span>' );
 	} );
 
 	$( document.body ).on('click', '.wp_job_manager_view_file_button', function ( event ) {
@@ -161,20 +178,34 @@ jQuery(document).ready(function($) {
 		file_target_wrapper = $( this ).closest('.file_url');
 		file_target_input   = file_target_wrapper.find('input');
 
-		// If the media frame already exists, reopen it.
-		if ( file_frame ) {
+		var mimes = get_field_mime_types( $( this ) );
+
+		// If the media frame already exists and matches the current field's mime filter, just reopen it. Otherwise rebuild so a previous field's filter does not leak in.
+		if ( file_frame && file_frame_mimes === mimes ) {
 			file_frame.open();
 			return;
 		}
 
-		// Create the media frame.
-		file_frame = wp.media.frames.file_frame = wp.media({
+		if ( file_frame ) {
+			file_frame.remove();
+			file_frame = null;
+		}
+
+		var frame_options = {
 			title: $( this ).data( 'uploader_title' ),
 			button: {
 				text: $( this ).data( 'uploader_button_text' )
 			},
 			multiple: false  // Set to true to allow multiple files to be selected
-		});
+		};
+
+		if ( mimes ) {
+			frame_options.library = { type: mimes };
+		}
+
+		// Create the media frame.
+		file_frame = wp.media.frames.file_frame = wp.media( frame_options );
+		file_frame_mimes = mimes;
 
 		// When an image is selected, run a callback.
 		file_frame.on( 'select', function() {
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -131,7 +131,7 @@ jQuery(document).ready(function($) {
 		if ( ! raw ) {
 			return null;
 		}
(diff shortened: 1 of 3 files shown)


donnchawp:

The per-field mime validation is a good idea and the tests are a nice touch, but there are two
issues that undercut the feature as written:

**Needs fixing**

1. **The mime check can be skipped.** The `is_numeric( $file_url )` branch
   (`includes/admin/class-wp-job-manager-writepanels.php:753-757`) accepts a posted value as a
   valid attachment and `continue`s past validation, with a comment that ownership is enforced
   elsewhere — but unlike the frontend version this was adapted from, there's no ownership or
   existence check here, and the admin media picker never submits a numeric value for these
   fields. Please drop this branch, or port the actual ownership check + `wp_get_attachment_url()`
   resolution from the frontend so it's a real code path rather than a way around the new
   validation.
2. **Rejecting an upload deletes the previously-saved file.** At `:780-781` the meta is
   overwritten with the filtered set, so if every submitted value is rejected the field is wiped
   to `''`/`[]`. The frontend equivalent aborts the save instead. Please preserve the existing
   value (or block the save) when there are rejected files. Note the test
   `test_existing_meta_is_preserved_when_rejected`
   (`tests/…/test_class.wp-job-manager-writepanels-file-mime.php:113-120`) currently asserts the
   value *should* be wiped — that assertion locks in the behaviour we want to avoid.
3. The media-library filter passes `array_keys( $allowed_mime_types )` — i.e. extensions — into
   `wp.media`'s `library.type` (`:225`/`:300`, `assets/js/admin.js:203`), which matches
   `post_mime_type`, so the picker filter is effectively inert. Use `array_values()` for the
   actual mime strings. (UX only; server-side still validates.)
4. New methods `flatten_allowed_extensions()` (`:843`) and `maybe_show_file_type_errors()`
   (`:858`) need `@since $$next-version$$`; there's a stray duplicated docblock at `:830-833`.

Heads-up on sequencing: #3053 covers the frontend upload path and there's related AJAX-upload
validation elsewhere, so this becomes the third hand-rolled mime check in the codebase — worth
reconciling the three rather than landing in isolation.

faisalahammad:
@donnchawp Thanks for the review. All four points addressed in 30f8134:

1. Numeric ID bypass removed; numeric values now go through wp_check_filetype and fall in the rejection path.
2. On rejection get_post_meta is read back so the previous value stays, matching frontend behavior.
3. data-allowed_mime_types now emits mime values via array_values, so wp.media library.type actually filters.
4. @since $$next-version$$ added on both helpers, duplicated docblock dropped.

Plus defensive (array) cast on third-party allowed_mime_types shape. Tests updated: test_attachment_id_is_always_allowed flipped to assert rejection, test_existing_meta_is_preserved_when_rejected now asserts the stored URL survives. PHPCS clean, CodeRabbit 0 findings.

Re-requesting.

PR #3053, opened 2026-07-14
Validate form uploads against the field's own allowed mime types

donnchawp:
Spun out of review on #3025, which surfaced the inconsistency but shouldn't carry the fix — that PR only changes a picker hint, this one changes what the server accepts.

`WP_Job_Manager_Form::upload_file()` has the field key in hand and passes it to `job_manager_upload_file()` as `file_key`, but it built the default allowed-types list one line earlier with a **keyless** `job_manager_get_allowed_mime_types()`. Two consequences on the non-AJAX submission path:

- A field with no `allowed_mime_types` of its own was validated against the generic list (`jpg/jpeg/jpe, gif, png, pdf, doc, docx`) no matter which field it was — so a `company_logo` upload was checked against the pdf/doc defaults and rejected the `webp` the field itself advertises.
- The `job_manager_mime_types` filter received an empty `$field` argument, so a site couldn't vary allowed types per field on this path, despite the filter documenting that parameter.

The AJAX path (`WP_Job_Manager_Ajax::upload_file()`) already gets this right: it passes only `file_key` and lets `job_manager_upload_file()` do the keyed lookup itself (`wp-job-manager-functions.php:1579`). This PR does the same in the form path and drops the duplicated default.

### Impact

Core WPJM's `company_logo` field sets `allowed_mime_types` explicitly, so the default branch never fires for it and stock installs are unaffected. This reaches add-ons or sites that register a file field — or unset that key — and rely on the defaults: such a `company_logo` field now accepts `webp` and stops accepting `pdf`/`doc` on the non-AJAX path, matching the AJAX path.

### Testing Instructions

1. `vendor/bin/phpunit --filter WP_Test_WP_Job_Manager_Form::` — 3 tests. Two of them fail on `trunk` and pass here; the third (an explicit `allowed_mime_types` still wins) is a control that passes either way.
2. With AJAX uploads disabled, register a file field keyed `company_logo` with no `allowed_mime_types` and submit a `.webp`. Before: rejected as a disallowed file type. After: accepted.

### Release Notes

Non-AJAX file uploads are now validated against the types allowed for that specific field, matching the AJAX path.

The `job_manager_mime_types` filter now receives the field key in its `$field` argument on the non-AJAX upload path, as documented. No template changes.

### Deprecated Code

None.

----

| Plugin build for 28d1cab5c8dd7994384096fbbf8ecf67254a134c <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-3053-28d1cab5.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/3053-28d1cab5)             |

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -755,10 +755,11 @@ abstract class WP_Job_Manager_Form {
 		$value = null;
 
 		if ( isset( $_FILES[ $field_key ] ) && ! empty( $_FILES[ $field_key ] ) && ! empty( $_FILES[ $field_key ]['name'] ) ) {
+			$upload_args = [ 'file_key' => $field_key ];
+
+			// Without an explicit list, job_manager_upload_file() looks up the types allowed for this field.
 			if ( ! empty( $field['allowed_mime_types'] ) ) {
-				$allowed_mime_types = $field['allowed_mime_types'];
-			} else {
-				$allowed_mime_types = job_manager_get_allowed_mime_types();
+				$upload_args['allowed_mime_types'] = $field['allowed_mime_types'];
 			}
 
 			$file_urls       = [];
@@ -767,10 +768,7 @@ abstract class WP_Job_Manager_Form {
 			foreach ( $files_to_upload as $file_to_upload ) {
 				$uploaded_file = job_manager_upload_file(
 					$file_to_upload,
-					[
-						'file_key'           => $field_key,
-						'allowed_mime_types' => $allowed_mime_types,
-					]
+					$upload_args
 				);
 
 				if ( is_wp_error( $uploaded_file ) ) {
--- a/tests/php/tests/includes/abstracts/test_class.wp-job-manager-form.php
+++ b/tests/php/tests/includes/abstracts/test_class.wp-job-manager-form.php
@@ -1,0 +1,119 @@
+<?php
+/**
+ * Covers includes/abstracts/abstract-wp-job-manager-form.php.
+ *
+ * @package wp-job-manager
+ */
+
+require_once JOB_MANAGER_PLUGIN_DIR . '/includes/abstracts/abstract-wp-job-manager-form.php';
+
+/**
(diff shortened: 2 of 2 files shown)


tellyworth:
Reviewed this for possible inclusion in 2.4.7 — the keyed-lookup fix is correct and the failing-on-trunk tests prove the bug, but a few things surfaced that make it better suited to a follow-up round than a patch release:

- **The parity is incomplete:** `validate_fields()` (class-wp-job-manager-form-submit-job.php:~513) still gates its mime check on `! empty( $field['allowed_mime_types'] )`, so a field relying on the new implicit per-field lookup gets no mime validation on the posted-URL path (`current_<key>`, used for AJAX-uploaded/resubmitted files) — the two server entry points now disagree in a new way. The client-side `data-file_types` hint in file-field.php is a third diverging resolution path; a shared resolver would close all three at once.
- **Two public filter contracts shift silently:** `job_manager_upload_file_pre_upload` hooks can now receive `''` where an array has arrived since 1.25.2 (PHP 8 TypeError risk in third-party callbacks — consider still resolving the list into `$upload_args`), and `job_manager_mime_types` callbacks keyed on `'' === $field` stop matching. Both at least deserve an upgrade note.
- **Smaller items:** `! empty()` treats an explicit `allowed_mime_types => []` as 'use defaults' while `job_manager_upload_file()` would treat the same `[]` as reject-everything; and the keyed `company_logo` fallback list in wp-job-manager-functions.php uses `'jpg|jpeg|jpe'` keys while the field-level list uses separate keys — both feed the same filter, so callbacks behave differently depending on which fired.
- **Test hygiene:** the rejection test leaves `$job_manager_upload`/`$job_manager_uploading_file` set process-wide (the error path returns before the reset — itself a pre-existing bug), `expectException( Exception::class )` is satisfied by any converted notice (prefer `expectExceptionMessage`), and the uploaded/temp files aren't cleaned up (the sibling AJAX test unlinks its upload for this reason).

Suggest folding this into an upload-validation consolidation alongside #3052/#3091 rather than landing solo.

PR #3054, opened 2026-07-14
fix(submission): add form_id attribute to submit_job_form shortcode

### LINKED ISSUE #2236: Multiple Job submission forms
Requested in 5188773-zen

**Usecase**

The first job_submit shortcode lets the customer submit a job posting.

The second job submits a shortcode that lets the customer submit a Classified like a car or a book.

So the first form would have all the fields regarding jobs, like job title, and job description.

The second form would have all the fields regarding the classified like title, description, pick up / shipping possible, mint condition/used, etc.

faisalahammad:
Fixes #2236

The `[submit_job_form]` shortcode now accepts a `form_id` attribute. The active form id is passed as a 2nd argument to the `submit_job_form_fields` filter, so a site can render the form on multiple pages with different field sets by branching on the id:

```php
// Page A: [submit_job_form]
// Page B: [submit_job_form form_id="classifieds"]
add_filter( 'submit_job_form_fields', function ( $fields, $form_id ) {
    if ( 'classifieds' === $form_id ) {
        unset( $fields['job']['job_type'], $fields['job']['job_category'] );
    }
    return $fields;
}, 10, 2 );
```

The form id is round-tripped through a hidden field on both `job-submit.php` and `job-preview.php`, so POST validation uses the same field set that was rendered. The edit form (from `[job_dashboard]`) inherits the same plumbing for consistency.

POST-supplied `form_id` values are validated as a short slug (`[A-Za-z0-9_-]{1,32}`); anything else is discarded. Existing single-argument `submit_job_form_fields` callbacks keep working.

### Testing Instructions

1. Create a page with `[submit_job_form]` and configure it as the Submit Job Form page in WPJM settings. Submit a listing end-to-end (fill, preview, submit) and confirm it saves as before.
2. In a small mu-plugin, add the snippet from the description above, then create a second page with `[submit_job_form form_id="classifieds"]`.
3. Visit the second page. Confirm the `Job type` and `Job category` fields are absent and a hidden `form_id=classifieds` input is present.
4. Click Preview on the second page, view source, confirm the hidden `form_id=classifieds` input is still present, then click Submit Listing. The job saves (or goes to moderation) successfully.
5. Visit the original `[submit_job_form]` page again. Confirm `Job type` and `Job category` are still present and no `form_id` hidden input is emitted.

### Release Notes

* New `form_id` shortcode attribute for `[submit_job_form]`, allowing different field sets per form via the `submit_job_form_fields` filter.

* `submit_job_form_fields` now receives a 2nd argument: the active form id (an empty string when no `form_id` is set). Backward compatible.
* `templates/job-submit.php` and `templates/job-preview.php` echo a hidden `form_id` input when a form id is active. Bumped `@version` to 2.4.6.
* `includes/abstracts/abstract-wp-job-manager-form.php` gains a public `current_form_id` property on the form base class.

### Deprecated Code

*

### Screenshot / Video

*

### DIFF
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -74,6 +74,16 @@ abstract class WP_Job_Manager_Form {
 	public $form_name = '';
 
 	/**
+	 * Active form id (set from shortcode `[submit_job_form form_id="..."]`).
+	 *
+	 * Round-tripped via a hidden field so POST processing rebuilds the same field set.
+	 *
+	 * @access public
+	 * @var string
+	 */
+	public $current_form_id = '';
+
+	/**
 	 * Cloning is forbidden.
 	 */
 	public function __clone() {
@@ -149,6 +159,10 @@ abstract class WP_Job_Manager_Form {
 		$step_key = $this->get_step_key( $this->step );
 		$this->show_errors();
 		$this->show_messages();
+
+		if ( isset( $atts['form_id'] ) ) {
+			$this->current_form_id = sanitize_text_field( $atts['form_id'] );
+		}
 
 		if ( $step_key && is_callable( $this->steps[ $step_key ]['view'] ) ) {
 			call_user_func( $this->steps[ $step_key ]['view'], $atts );
--- a/includes/abstracts/abstract-wp-job-manager-form.php
+++ b/includes/abstracts/abstract-wp-job-manager-form.php
@@ -26,6 +26,16 @@ abstract class WP_Job_Manager_Form {
 	protected $fields = [];
 
 	/**
+	 * Cache key for `$fields`. Tracks the `current_form_id` value the
+	 * fields were built under so a 2nd `submit_job_form form_id="..."`
+	 * on the same page rebuilds instead of returning the 1st set.
+	 *
+	 * @access protected
+	 * @var string
+	 */
+	protected $fields_cache_key = '';
+
+	/**
 	 * Form action.
 	 *
 	 * @access protected
--- a/includes/class-wp-job-manager-shortcodes.php
+++ b/includes/class-wp-job-manager-shortcodes.php
@@ -99,6 +99,13 @@ class WP_Job_Manager_Shortcodes {
 	 * @return string|null
 	 */
 	public function submit_job_form( $atts = [] ) {
+		$atts = shortcode_atts(
+			[
+				'form_id' => '',
+			],
+			$atts,
+			'submit_job_form'
+		);
 		return $GLOBALS['job_manager']->forms->get_form( 'submit-job', $atts );
 	}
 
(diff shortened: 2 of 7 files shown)


donnchawp:

Thanks for this — the shortcode→template→POST plumbing is internally consistent and I couldn't
fault the escaping. A few things to resolve before it's mergeable:

**Needs fixing**

1. **`form_id` selects the field set but is entirely client-supplied.** In
   `submit_handler()` (`includes/forms/class-wp-job-manager-form-submit-job.php:739-741`) the
   posted `form_id` is format-validated only, with no binding to the page that rendered the form.
   The nonce proves the user may submit, not that the `form_id` matches what they were shown — so
   a submitter can post a different, less-restrictive `form_id` and skip fields the site intended
   to require (the Job-vs-Company-ad use case in the linked thread). Suggest folding `form_id`
   into the nonce action string so a mismatched value invalidates the nonce.
2. **The edit-job half is unreachable from real UI but still POST-reachable.** `[job_dashboard]`
   calls `get_form('edit-job')` with no `$atts` (`includes/class-job-dashboard-shortcode.php:375`),
   so the `isset($atts['form_id'])` branches never fire through the dashboard. A listing created
   under a reduced `form_id` then falls back to the **default** field set on edit — which may
   include required fields that were never collected, potentially blocking the save. Suggest
   persisting the originating `form_id` as post meta at submission and reading it server-side on
   edit.
3. `@since` tags use a hardcoded `2.4.6`
   (`class-wp-job-manager-form-submit-job.php:237`, `templates/job-preview.php:11`,
   `templates/job-submit.php:11`) — please use the `$$next-version$$` placeholder, as line 988 of
   that same file already does.
4. `init_fields()`'s `if ( $this->fields ) return;` cache ignores `form_id`, so two
   `[submit_job_form form_id="…"]` blocks on one page both render the first one's fields.
5. No test covers the edit-job path or asserts a malformed `form_id` is rejected.

faisalahammad:
@donnchawp All 5 issues resolved in [7c08b1f](https://github.com/faisalahammad/WP-Job-Manager/commit/7c08b1fbf08af85e5c7af8caa1bf9f33d8dfc63d):

1. **Nonce binding** \u2014 `form_id` appended to nonce action (`-form-{$form_id}`). Form rendered with `form_id="classifieds"` expects nonce minted for `submit-job-{id}-form-classifieds`; POSTing a different `form_id` fails nonce check. Added private `submit_nonce_action()` helper shared by both `output_submit_form_nonce_field()` and `check_submit_form_nonce_field()`.

2. **Edit-job meta round-trip** \u2014 `_form_id` post meta written at submission (both submit-job and edit-job `submit_handler()`). Edit-job constructor reads meta server-side and seeds `current_form_id` before `init_fields()`. POST round-trip (when present) still wins, so an explicit form_id on `[job_dashboard]` is not downgraded.

3. **@since/@version placeholders** \u2014 Hardcoded `2.4.6` replaced with `$$next-version$$` in all 3 sites: submit-job filter PHPDoc, `job-submit.php`, and `job-preview.php`.

4. **Fields cache scoped to form_id** \u2014 `init_fields()` skips early return when `current_form_id` differs from the cached `fields_cache_key`. Two `[submit_job_form form_id="..."]` blocks on one page each render their own field set.

5. **Test coverage** \u2014 4 new tests in `test_class.submit-job-form-id.php`: malformed form_id rejection, nonce-mismatch rejection (via `wp_die_handler` suppression), edit-job meta fallback, and cache invalidation on form_id change.

Additionally: stale `_form_id` deleted on default-form saves (CodeRabbit finding).

PHPCS passes clean across all 6 modified files.

PR #3056, opened 2026-07-15
feat: add job_manager_get_listings_include_category_children filter

### LINKED ISSUE #1483: Use Category Filter Type on All Category Selections
#### Steps to Reproduce
1. Create a parent/child category relationship.
2. Assign one listing to the parent category. Assign one listing to the child category.
3. Try to view just the parent category.

#### What I Expected

Only the listings assigned to the chosen parent category to appear.

#### What Happened Instead

The listings assigned to the chosen parent's children also appear.

trunk/master

#### Context / Source

https://github.com/Automattic/WP-Job-Manager/blob/master/wp-job-manager-functions.php#L104
https://github.com/Automattic/WP-Job-Manager/blob/master/includes/class-wp-job-manager-post-types.php#L384

The "Category Filter Type" explicitly checks for more than one category being selected. This is technically to be expected because the setting description says it's meant to adjust multiple category selections.

This leaves a gap in functionality when using a single category (even in multiselect mode).

I think the count for chosen categories should be removed and the filter type descriptions updated to better describe the affect it has on querying children.

rudberga:
Fixes #1483

- Add a new `job_manager_get_listings_include_category_children` filter around the
  `include_children` argument of the category `tax_query` in `get_job_listings()`.
- The default value preserves the current behavior exactly, so **nothing changes for
  any site unless it explicitly attaches a callback**. No settings need to change.
- A site that attaches a callback returning `false` gets strict (exact-term) matching:
  a listing assigned only to a child term no longer appears when its parent term is
  selected. This works for any category query, regardless of the "Category Filter
  Type" setting.

### Testing Instructions
No setting changes are required.

1. Create a parent category and a child category. Assign one listing to each.
2. On a `[jobs]` page, filter by the parent category only → **both** listings appear (default behavior, unchanged).
3. Add this callback (e.g. in a small plugin or theme `functions.php`):
   ```php
   add_filter( 'job_manager_get_listings_include_category_children', '__return_false' );
   ```
4. Filter by the parent category only → **only** the listing assigned directly to the parent appears; the child-term listing is excluded.

Automated: `make test` — see `test_get_job_listings_include_category_children_filter` in `tests/php/tests/test_class.wp-job-manager-functions.php` (asserts the default is preserved and that the filter can override it).

### Release Notes
Add a `job_manager_get_listings_include_category_children` filter to control whether category queries include child terms.

- New filter: `job_manager_get_listings_include_category_children( bool $include_children, string $operator, array $search_categories, array $args )`

### Deprecated Code
- None

----

| Plugin build for b1bfb73e5faa9f99725fdc4bde3cc9e4448f342a <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-3056-b1bfb73e.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/3056-b1bfb73e)             |

### DIFF
--- a/tests/php/tests/test_class.wp-job-manager-functions.php
+++ b/tests/php/tests/test_class.wp-job-manager-functions.php
@@ -419,6 +419,93 @@ class WP_Test_WP_Job_Manager_Functions extends WPJM_BaseTest {
 	}
 
 	/**
+	 * Forcing the filter to true under the AND operator adds each selected term's
+	 * descendants to the set a listing must match in full, so it narrows results
+	 * rather than widening them. When one selected term is an ancestor of another
+	 * the expanded set contains duplicates, WP_Tax_Query rejects the clause as
+	 * "inexistent_terms", and the query matches nothing at all.
+	 *
+	 * @since $$next-version$$
+	 * @covers ::get_job_listings
+	 */
+	public function test_get_job_listings_include_category_children_filter_in_and_mode() {
+		// tax_input is silently dropped by wp_insert_post() without this cap.
+		$this->assertTrue( current_user_can( get_taxonomy( \WP_Job_Manager_Post_Types::TAX_LISTING_CATEGORY )->cap->assign_terms ) );
+
+		$parents  = [];
+		$children = [];
+		foreach ( [ 'engineering', 'design' ] as $name ) {
+			$parents[ $name ]  = $this->factory->term->create(
+				[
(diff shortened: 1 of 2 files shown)


donnchawp (approved):
Looks good, ready to merge @rudberga !

PR #3060, opened 2026-07-15
Allow reusing a listing's existing logo on edit when authored by another user

donnchawp:
## What

Fixes a regression from #2995 (reported [here](https://github.com/Automattic/WP-Job-Manager/pull/2995#issuecomment-4985398582)). #2995 tightened company-logo attachment validation to reject a numeric attachment ID unless the current user owns it or can `edit_post` it. That broke the edit/re-save flow for listings whose logo was uploaded or replaced by a site admin: the attachment's `post_author` is the admin, so an employer editing **their own** listing hits **"Invalid attachment provided."** on save.

## Why it's safe

Carrying forward an attachment already bound to the listing being edited is not the same as binding an arbitrary foreign ID. By the time validation runs, `$this->job_id` is only set for a listing the current user passed `job_manager_user_can_edit_job()` on (or is mid-submitting via the draft/preview key). So this only permits reusing what is **already** attached to a listing the user is authorized to edit.

## Changes

`includes/forms/class-wp-job-manager-form-submit-job.php`
- Add `is_existing_listing_attachment( $attachment_id, $key )` — true when the ID matches the listing's persisted value for that field, resolved the same way the edit form pre-populates it (post thumbnail / `_company_logo` meta, or `_{key}` meta otherwise).
- `validate_attachment_ownership()` also allows the attachment when it is the listing's existing value.

The original security boundary is unchanged: a foreign attachment that is **not** the listing's saved logo is still rejected, even in edit mode.

## Testing

Adds regression coverage to `WP_Test_Submit_Job_Attachment_Ownership`, verified fail-before / pass-after:
- `test_existing_listing_logo_is_accepted_on_edit` — an admin-authored logo already set as the listing's featured image is accepted on edit. Without the fix this throws the exact reporter error.
- `test_foreign_attachment_still_rejected_on_edit` — a foreign attachment that is *not* the listing's saved logo is still rejected in edit mode (guards against `job_id` becoming a blanket bypass).

Full class passes (8 tests); `phpcs` clean.

Fixes #2995 (follow-up regression).

----

| Plugin build for e61ba7ec64983af0c2e5a50ca66cf8ad618a504e <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-3060-e61ba7ec.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/3060-e61ba7ec)             |

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -970,7 +970,11 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 				foreach ( array_filter( $file_urls ) as $file_url ) {
 					if ( is_numeric( $file_url ) ) {
 						$attachment_id = absint( $file_url );
-						if ( $attachment_id && ! $this->is_attachment_authorized_for_current_user( $attachment_id ) ) {
+						if (
+								$attachment_id
+								&& ! $this->is_attachment_authorized_for_current_user( $attachment_id )
+								&& ! $this->is_existing_listing_attachment( $attachment_id, $key )
+							) {
 							return new WP_Error( 'validation-error', __( 'Invalid attachment provided.', 'wp-job-manager' ) );
 						}
 					}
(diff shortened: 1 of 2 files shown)


Hastibe:
@donnchawp -- even with this fix implemented, we unfortunately just had a user run into the "Invalid attachment provided" error again (I also would suggest changing the error message to be more user-friendly -- we just started running a str_ireplace in our functions.php file to change the message to "Please re-upload (or remove) your logo, below" instead, so users can better understand what they can do to resolve it).

In this case, the error was encountered by the user when submitting a **new** job post with a saved logo that was owned by our admin in their job posting form. We verified this error that the user was encountering ourselves and noticed that -with this fix- relisting an expired job of theirs with the same logo owned by our admin did not result in the error (woo!). Unless something else is causing it, could this fix be expanded to also allow **new** job posts with a saved logo owned by an admin to also meet these new validation requirements (in addition to when a user edits, relists, renews, etc. a job post of their with a logo that our admin owns)?

Alternatively, could an option be added or snippet provided to allow disabling this new validation step altogether? I'm admittedly not clear on what the purpose of preventing a user from using another user's uploaded logo is since they could always upload one that isn't their own, too, of course--there must be a larger vulnerability concern here?

donnchawp:
@Hastibe thanks for the report. You're right. It should allow images that are uploaded by an admin already. The user is authorized to use those images. I'm going to create a new PR with a fix for this. And you're right, this fix came from a security report that we fixed.

PR #3065, opened 2026-07-17
feat: 'wp jm dedupe-logos' — collapse duplicate company-logo attachments

donnchawp:
## What

Retroactive companion to #3064 (runtime logo-upload dedup). Sites that ran for years before that fix accumulated large numbers of duplicate company-logo attachments — one new attachment per re-upload of the same logo. This adds a WP-CLI command to collapse the existing duplicates.

Context: forum report [Core bug: massive logo duplication / bloat server space](https://wordpress.org/support/topic/core-bug-massive-logo-duplication-bloat-server-space/) (~20k duplicate images; reporter confirmed they span 2018→present, i.e. long-tail accumulation, not a recent regression). #3064 stops new growth; this cleans up what's already there.

## Command

```
wp jm dedupe-logos                      # dry run (default): report only
wp jm dedupe-logos --report=plan.csv    # dry run, keeping the plan for review
wp jm dedupe-logos --live               # apply (prompts; --yes to skip)
wp jm dedupe-logos --owner=42 --live    # limit to one attachment owner
```

Note `--owner`, not `--user`: `--user` is a reserved WP-CLI global parameter that sets the acting user and never reaches the command.

**Run `--live` with the site in maintenance mode.** See Safety below for why.

## How it works

`WP_Job_Manager\Attachment_Deduplicator`:

1. Derives "logo" content signatures (`author:md5`) from attachments **currently referenced** as a logo — listing featured images (`_thumbnail_id`, across all listing statuses) and the `_company_logo` user default.
2. Groups **every image attachment those owners hold** by signature, keeping only signatures that match a referenced logo. This **folds in orphaned duplicate copies** that no longer reference anything (the bulk of the bloat).
3. Filters the group through `find_protected_ids()` — see Safety below.
4. Per group: keeps the **oldest** attachment as canonical, **re-points every reference to it, verifies the re-point landed**, and only then deletes the redundant copies.
5. Carries curated alt text / caption / description onto the canonical (as the oldest copy, it is usually the bare auto-generated upload).
6. Backfills `_wpjm_attachment_hash` on the canonical so the #3064 runtime dedup keeps de-duplicating future uploads against it.

### Safety

Identical content is **not** proof that an attachment is a logo — an owner may have uploaded the same image again for something this command knows nothing about. The candidate set is therefore treated as candidates only.

**Nothing is deleted while a reference this command cannot re-point still points at it.** `find_protected_ids()` makes one streamed pass over the meta rows whose *key* looks like it could hold an attachment reference, and reads each value in PHP rather than comparing it to the candidate IDs in SQL. That distinction is the whole check: `meta_value IN ( 12,34 )` is an exact match against a LONGTEXT column, so MySQL coerces a serialized ACF gallery to 0 (matching nothing) and a comma-separated `_product_image_gallery` to its first ID only — while the key looks covered. Reading the value instead recognises:

- a bare ID;
- a comma-separated list of them;
- an ID nested anywhere inside a serialized structure (an ACF gallery or repeater, `theme_mods_<theme>`'s `custom_logo`, a media widget);
- a URL or path, matched by filename — which is how the pre-1.24.0 `_company_logo` post meta, and content that prints a bare `` from a custom field, are seen at all.

Filenames are matched by *stem*, so a reference to `logo-150x150.png` protects `logo.png`. Post content is scanned the same way, for `wp-image-N`, block `{"id":N}` in all three shapes it is stored in (plain, JSON-escaped, and kses-rewritten to `"`), and plain URLs.

**References are moved before anything is deleted, and the move is verified.** `update_post_meta()` reports failure by return value, and core's `wp_delete_attachment()` deletes every `_thumbnail_id` row pointing at the attachment — so an unchecked write would turn a failed re-point into a listing that silently lost its logo, counted as a success. The run re-queries for surviving references after re-pointing and drops anything still referenced from the delete set, reporting it rather than deleting it.

Also:

- **Dry-run by default**, and `--live` confirms before touching anything. The dry run mutates nothing at all — the hash backfill now happens after the plan is complete, not during it.
- **`--report=<path>`** writes the full mapping to CSV before any mutation. On a 20k-attachment library terminal scrollback is not a record of an irreversible operation.
- **Per-owner only** (`author:hash`): identical logos belonging to different users are never merged, preserving the ownership boundary from #2995/#3060. **Guest uploads (`post_author` 0) are skipped entirely** — 0 is not an owner, as the runtime dedup already recognises.
- **Shared files are protected.** Two attachment rows can point at one `_wp_attached_file`; core's cleanup does no reference counting, so file deletion is vetoed when the path is still in use.
- **`force_delete` is off**, so sites with `MEDIA_TRASH` keep an undo.
- Two filters — `job_manager_dedupe_reference_meta_key_patterns` and `job_manager_dedupe_protected_attachment_ids` — so a site storing references under an unusual key can protect itself without patching the plugin. Both can only ever protect more; neither can widen the deletion set.

**What the sweep does not cover**, and why `--live` wants maintenance mode: protection is worked out once, before any deletion, and on the libraries this exists for that scan takes a while. A reference created after the scan and before the matching delete is not protected by it. The two handled keys are re-checked immediately before deleting; a new gallery entry, an edited page, or a logo set in the Customizer mid-run is not. Nothing here is multisite-aware either — run it per site with `wp --url=`.

### Performance

Sized for the libraries this exists for. Every pass **streams a page at a time** and pages by keyset (`WHERE id > ?`) rather than `OFFSET`, which over an unindexed scan re-reads and discards every earlier row on each page. Each page is primed, reduced, and released before the next is fetched; meta lookups, re-points, deletions and the owner list are chunked, and the object cache is released between batches.

The reference sweep is one pass over the meta rows whose key matches, intersected against the candidate set in PHP. `meta_value` has no index and the key patterns have a leading wildcard, so each of those is a full table scan — doing it once per batch of candidates, as the first version did, is what stops the command finishing on a large site.

Measured, not extrapolated. Fixture: N owners x 100 identical-bytes attachments each (distinct files on disk), a listing on every third one, and a `_company_logo` default per owner. Dry run, seeded and measured in separate processes so the seed's allocations do not set the peak. `wp-cli` plus WordPress is a 55.31 MB floor before the command starts.

| attachments | duplicates found | peak memory | over baseline | queries | queries before |
| ----------- | ---------------- | ----------- | ------------- | ------- | -------------- |
| 2,000       | 1,980            | 57.31 MB    | 2.00 MB       | 32      | 43             |
| 4,000       | 3,960            | 59.31 MB    | 4.00 MB       | 54      | 78             |
| 20,000      | 19,800           | 75.31 MB    | 20.00 MB      | 236     | 367            |

Memory over baseline grows at ~1 MB per 1,000 attachments, linearly. That residual is `$signature_cache` — one entry per image examined, held for the whole run — plus the duplicate ID list, and it is the thing to watch on a very large library rather than the content scan I had originally worried about. At 20,000 the whole run is 20 MB over an idle `wp` process, so the earlier "would not finish" concern is settled.

Query count scales with the number of *pages*, not with the number of duplicates: 236 queries for 19,800 duplicates. Against the previous implementation on the same fixture it is 36% fewer, which is the four-scans-per-batch reference sweep collapsing into one pass.

Two caveats on these numbers. Elapsed time is not meaningful here — the fixture is a 1x1 PNG, so `md5_file()` costs nothing and the tables are small enough that the full scans are cheap in a way they will not be on a real 20k-image library. And peak memory understates real attachment meta for the same reason.

One incidental finding from running the old code against the same fixture: it reported 1,979 duplicates and 1 "kept, referenced elsewhere", where the current code finds 1,980. The extra protection was spurious — `_wp_attached_file` matches the `%file%` key pattern, and MySQL coerced the path `2026/08/...` to the integer 2026, so attachment 2026 protected itself forever. That is the numeric-coercion failure mode the sweep rewrite removes, showing up on its own in a synthetic fixture.

## Scope / follow-ups

- **Logos only** (featured images + `_company_logo`), as agreed — the reported surface and lowest risk. Broadening to all file-field attachments is a possible follow-up.
- Depends on the `_wpjm_attachment_hash` meta key introduced in #3064.
- Multisite support, and re-checking protection per chunk rather than once per run, are both deliberately left out. The docblock says so.

## Testing

`WP_Test_Attachment_Deduplicator` (31 tests), including:

- merge + re-point featured images and the `_company_logo` user default;
- collapse orphaned duplicate copies (the reporter's real shape);
- an image is never deleted when it is referenced by: another post type's featured image, inline content, a serialized meta value, a comma-separated meta value, a legacy `_company_logo` URL, a resized filename in content, an escaped block id, a media-shaped meta key, term meta, `site_logo`, or the Customizer's `custom_logo`;
- a duplicate whose re-point could not be written is kept, not deleted;
- **expired** listings are re-pointed, not treated as orphans;
- identical logos of different users are not merged; guest-owned attachments are not merged;
- deleting a duplicate never removes a file the canonical still uses;
- curated alt text, caption and description survive; the canonical is backfilled with the runtime dedup hash, taken from the original upload rather than the `-scaled` copy;
- owner scoping follows attachment ownership, not listing authorship;
- a failed deletion is reported rather than counted as a success;
- the dry run reports the actual canonical → duplicates mapping, and mutates nothing;
- query count does not scale with the number of duplicates or owners;
- results are identical across batch boundaries, including a protected candidate landing mid-batch.

Also verified end-to-end via the actual CLI against a real database: seeded three identical logos on a published listing and an expired one, ran `--live`, confirmed both duplicates deleted, **both** listings re-pointed to the canonical, hash backfilled, and the canonical's file intact on disk. Full suite green; `phpcs` and `psalm` clean.

## Review history

The first implementation had several data-loss paths (non-logo references deleted unrepaired, expired listings orphaned, guests merged, shared files unlinked) and a scan that would not have finished on a 20k-attachment site. Those were fixed in 6a2ffdd..8198112.

A second review found two more, both fixed in 4318203:

1. The protection sweep matched `meta_value` exactly, so serialized and comma-separated references were invisible to it — the ACF-gallery and WooCommerce shapes, plus the Customizer logo and legacy `_company_logo` URLs. Rewritten as described in Safety above.
2. `repoint_all()` never checked whether its writes succeeded, so a failed re-point became a silently broken listing reported as a success. Now verified by re-query before deleting.

That review also flagged the postmeta sweep (four full table scans per batch) and `OFFSET` pagination as the real scaling problems, ahead of the content scan I had been worrying about. Both are fixed.

----

| Plugin build for c1d87f1621763e927f9626a928ab8577f916c726 <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/09/wp-job-manager-zip-3065-c1d87f16.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/09/3065-c1d87f16)             |

### DIFF
--- a/includes/class-attachment-deduplicator.php
+++ b/includes/class-attachment-deduplicator.php
@@ -19,11 +19,22 @@ if ( ! defined( 'ABSPATH' ) ) {
  * Two attachments are candidates for merging when they have identical file
  * contents and the same owner. Identical content is *not* on its own proof that
  * an attachment is a logo, though: an owner may have uploaded the same image
- * again for something this class knows nothing about. So no attachment is ever
+ * again for something this class knows nothing about. So an attachment is not
  * deleted while anything outside the two reference types handled here (listing
  * featured images and the `_company_logo` user default) still points at it —
  * see `find_protected_ids()`. Unrecognised reference means "keep", so the
  * failure mode is a duplicate left behind rather than a broken image.
+ *
+ * What that sweep can see has limits worth knowing before running it live:
+ *
+ * - It reads meta rows whose *key* looks like it could hold an attachment
+ *   reference, so a reference stored under a name with no media-ish word in it
+ *   is missed. `job_manager_dedupe_reference_meta_key_patterns` and
+ *   `job_manager_dedupe_protected_attachment_ids` exist for that.
+ * - It runs once, before any deletion. A reference created while the run is in
+ *   progress is not protected by it, other than the two handled keys, which are
+ *   re-checked immediately before deleting. Run `--live` in maintenance mode.
+ * - Nothing here is multisite-aware; run it per site with `wp --url=`.
  *
  * Attachments belonging to different users are never merged, mirroring the
  * runtime dedup boundary. Guest uploads (`post_author` 0) are skipped outright:
@@ -189,6 +200,21 @@ class Attachment_Deduplicator {
 	 * one, re-pointing every reference before deleting the redundant copies.
 	 * Dry-run by default; pass --live to apply.
 	 *
+	 * ## RUN THIS WITH THE SITE IN MAINTENANCE MODE
+	 *
+	 * Which attachments are safe to delete is worked out once, up front, and on the
+	 * libraries this exists for that scan takes a while — every image the logo owners
+	 * hold is hashed and every post's content is read. A reference created after that
+	 * scan and before the matching delete is not protected by it.
+	 *
+	 * Re-points are verified immediately before deleting, so a listing or user default
+	 * that changed in the meantime is left alone. Anything else — a new gallery entry,
+	 * a page edited to embed the image, a logo set in the Customizer — is not covered,
+	 * and deletion is not reversible unless the site defines MEDIA_TRASH.
+	 *
+	 * So: take a backup, put the site in maintenance mode, run it. On a live site the
+	 * dry run is safe at any time; --live is not.
+	 *
 	 * ## OPTIONS
 	 *
 	 * [--live]
@@ -248,8 +274,9 @@ class Attachment_Deduplicator {
 		$report_path = \WP_CLI\Utils\get_flag_value( $assoc_args, 'report', '' );
 
 		if ( ! $dry_run ) {
-			// Deletion is not reversible on sites without MEDIA_TRASH.
-			\WP_CLI::confirm( 'This will delete redundant logo attachments. Run without --live first to preview. Continue?', $assoc_args );
+			// Deletion is not reversible on sites without MEDIA_TRASH, and references
+			// created while the run is in progress are not protected by its scan.
+			\WP_CLI::confirm( 'This will delete redundant logo attachments, and is not reversible unless MEDIA_TRASH is on. Run without --live first to preview, and run this one with the site in maintenance mode. Continue?', $assoc_args );
 		}
 
 		// Discovery hashes every image the logo owners hold and scans post content, so
--- a/includes/class-attachment-deduplicator.php
+++ b/includes/class-attachment-deduplicator.php
@@ -294,6 +294,7 @@ class Attachment_Deduplicator {
 		// on the libraries this exists for it is minutes of silence otherwise.
 		\WP_CLI::log( 'Scanning for duplicate logos…' );
 
+		$started  = microtime( true );
 		$progress = null;
 		$report   = ( new self() )->run(
 			[
@@ -333,6 +334,17 @@ class Attachment_Deduplicator {
 		// What the run does not cover, counted: whether the site's remaining bloat
 		// is inside or outside this command's reach should be a number, not a guess.
 		\WP_CLI::log( sprintf( 'Not examined:            %d guest-owned image(s), %d image(s) matching no currently-referenced logo', $report['skipped_guest'], $report['skipped_unmatched'] ) );
+
+		// Elapsed time and peak memory are the two numbers a synthetic benchmark
+		// cannot supply; printing them makes every real-site run a data point.
+		$elapsed = microtime( true ) - $started;
+		\WP_CLI::log(
+			sprintf(
+				'Took %s; peak memory %s.',
+				$elapsed >= 60 ? sprintf( '%dm %ds', (int) ( $elapsed / 60 ), (int) $elapsed % 60 ) : sprintf( '%.1fs', $elapsed ),
+				size_format( memory_get_peak_usage( true ) )
+			)
+		);
 
 		// Print the mapping so a dry run can be audited, and a live run leaves a
 		// record of what was collapsed into what. Past a few screenfuls the CSV is
--- a/includes/class-attachment-deduplicator.php
+++ b/includes/class-attachment-deduplicator.php
@@ -16,9 +16,20 @@ if ( ! defined( 'ABSPATH' ) ) {
  * example by re-uploading the same logo on every submission) down to one,
  * re-pointing every reference before deleting the redundant copies.
  *
- * Scoped to logos only (listing featured images and the `_company_logo` user
- * meta) and to a single owner per group, mirroring the runtime dedup boundary:
- * attachments belonging to different users are never merged.
+ * Two attachments are candidates for merging when they have identical file
+ * contents and the same owner. Identical content is *not* on its own proof that
+ * an attachment is a logo, though: an owner may have uploaded the same image
+ * again for something this class knows nothing about. So no attachment is ever
+ * deleted while anything outside the two reference types handled here (listing
+ * featured images and the `_company_logo` user default) still points at it —
+ * see `find_protected_ids()`. Unrecognised reference means "keep", so the
+ * failure mode is a duplicate left behind rather than a broken image.
+ *
+ * Attachments belonging to different users are never merged, mirroring the
+ * runtime dedup boundary. Guest uploads (`post_author` 0) are skipped outright:
+ * 0 is not an owner, so unrelated submitters would otherwise collapse together.
+ *
+ * @since $$next-version$$
  *
  * @internal
  */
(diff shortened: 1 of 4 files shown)


donnchawp:
## Benchmark results, measured rather than extrapolated

The performance table in the description used to stop at 4,000 attachments and extrapolate to 20,000. I've now measured 20,000 directly, and measured the previous implementation (`81981122`) against the same fixture so the two are comparable. The description has been updated with this, but I'm repeating it here so it's in the review thread.

### Fixture

N owners, 100 identical-bytes attachments each, every one a distinct file on disk so the shared-file veto doesn't short-circuit the work. Bytes differ between owners, so each owner is one duplicate group instead of the whole site being one. A listing on every third attachment, plus a `_company_logo` default per owner, so the run has both referenced logos to derive signatures from and orphans to fold in.

Seeded and measured in separate processes. If you do it in one, the seed's own allocations set the peak and the memory number means nothing. `wp-cli` plus WordPress is a 55.31 MB floor before the command starts, hence the "over baseline" column.

Dry run in every case.

### Results

| attachments | duplicates found | peak memory | over baseline | queries | queries at `81981122` |
| ----------- | ---------------- | ----------- | ------------- | ------- | --------------------- |
| 2,000       | 1,980            | 57.31 MB    | 2.00 MB       | 32      | 43                    |
| 4,000       | 3,960            | 59.31 MB    | 4.00 MB       | 54      | 78                    |
| 20,000      | 19,800           | 75.31 MB    | 20.00 MB      | 236     | 367                   |

At 20,000 the whole run sits 20 MB above an idle `wp` process, so the "this won't finish" worry is settled. Growth is about 1 MB per 1,000 attachments and it's linear.

That residual is `$signature_cache`, one entry per image examined, held for the whole run. It's the thing to watch on a very large library. I'd been worrying about the `post_content` scan instead, which was the wrong thing: its peak is bounded by the page size, not by the number of posts.

Query count scales with the number of pages, not the number of duplicates. 236 queries for 19,800 duplicates. Against the old implementation on the same fixture it's 36% fewer, which is the four-scans-per-batch reference sweep collapsing into one pass.

### One thing that fell out of it

Running the old code over the same fixture reported 1,979 duplicates and 1 "kept, referenced elsewhere". The current code finds 1,980.

That extra protection was spurious. `_wp_attached_file` matches the `%file%` key pattern, and the old sweep compared `meta_value` to the candidate IDs in SQL, so MySQL coerced the path `2026/08/bench1-o0-1.png` to the integer 2026. Attachment 2026 protected itself, permanently.

It's a small thing in a synthetic fixture, but it's the numeric-coercion failure mode the rewrite is there to remove, and it turned up on its own without me constructing a case for it. On a real site with date-based upload paths, every attachment whose ID happens to match a year in use was quietly un-dedupable.

### Two caveats, please don't over-read the numbers

Elapsed time isn't meaningful here and I've left it out on purpose. The fixture is a 1x1 PNG, so `md5_file()` costs nothing, and the tables are small enough that the full table scans are cheap in a way they will not be on a real 20k-image library. Peak memory understates real attachment meta for the same reason.

So these numbers say the command's own bookkeeping doesn't blow up. They don't tell you how long it takes on the reporter's site. That still wants a timed dry run against a copy before anyone runs `--live` there.

I didn't commit the benchmark scripts, they were throwaway. The fixture is described above if you want to reproduce it.

tellyworth:
Pushed 7247de2f: the dry run (and live run) now counts what it does **not** examine, so the residue outside the command's reach is measured instead of guessed at.

The summary gains one line:

```
Not examined:            120 guest-owned image(s), 3401 image(s) matching no currently-referenced logo
```

backed by two new report keys, `skipped_guest` (post_author 0 — no owner to merge within) and `skipped_unmatched` (content matching no currently-referenced logo, so nothing anchors it as a logo duplicate). Context: in the [forum thread](https://wordpress.org/support/topic/core-bug-massive-logo-duplication-bloat-server-space/page/2/) the open question is how much of the reporter's bloat this tool will collapse versus deliberately leave alone. With these counts, one staging dry run answers that with numbers, and any follow-up (e.g. a review-based orphan report) can be sized against reality.

Costs one counter in the discovery pass plus two `COUNT(*)` queries — fixed overhead, which moved the query-budget test's deliberately-loose bound from 15 to 18. The `--report` CSV is unchanged on purpose: it's a machine-readable canonical ← duplicate mapping, and summary rows mixed into it would break anything consuming it (including piping IDs to `wp post delete`).

Two tests added (guest counts, unmatched counts incl. `--owner` scoping) and one extended; full suite green, phpcs and Psalm clean, verified end-to-end via the actual CLI.

PR #3082, opened 2026-07-29
Allow reusing a submitter's saved company logo on a new listing

donnchawp:
## What

Follow-up to #3060, [reported here](https://github.com/Automattic/WP-Job-Manager/pull/3060#issuecomment-5107970143): employers still hit **"Invalid attachment provided."** when submitting a **new** listing whose company logo was uploaded by a site admin. #3060 fixed the edit/relist flow; this fixes the new-submission flow.

## Why it happens

The form offers a logo back to the submitter along two different paths, and #3060 only taught the validation about one:

- **Editing a listing** (`$this->job_id` set) — `submit()` pre-populates from the listing's post thumbnail / `_company_logo` meta. Covered by #3060's `is_existing_listing_attachment()`.
- **A new listing** (no `job_id` yet) — `submit()` pre-populates the company fields from the submitter's own **`_company_logo` user meta**. Not covered: with `job_id` unset, `is_existing_listing_attachment()` returns `false`, so a saved logo authored by someone else is rejected.

#3060 also propagates the condition rather than only exposing it. Re-saving a listing whose logo an admin uploaded now succeeds, and `update_job_data()` writes that admin-owned attachment ID into the *employer's* `_company_logo` user meta. It is then offered back on their next new listing — and rejected. That is exactly the reporter's observation that relisting works but posting a new listing does not.

## Why it's safe

`is_saved_user_attachment()` is the user-meta counterpart to #3060's listing check, and rests on the same property: **it consults server-side state, never the request.**

`_company_logo` user meta has exactly one writer in the plugin — `update_job_data()` — which only runs after the existing validation has already passed. So the saved value records an attachment the submitter was previously authorised to use, even when they do not author it (an admin uploaded or replaced it on their listing). It is not a request-supplied ID, and nothing else in the plugin or its REST surface writes that meta key.

The allowance is scoped to exactly what `submit()` pre-populates and no wider: a new listing (no `job_id`), a logged-in submitter, the `company` group. Editing reads from the listing instead, and `is_existing_listing_attachment()` already covers that path.

The existing validation rules are otherwise unchanged: an attachment that is neither owned/editable, nor the listing's saved logo, nor the submitter's own saved value is still rejected.

## Changes

`includes/forms/class-wp-job-manager-form-submit-job.php`
- Add `is_saved_user_attachment( $attachment_id, $group_key, $key )` — true when the ID matches the current user's saved value for that field, mirroring the pre-population in `submit()` exactly (new listings only, logged-in submitters, `company` group only; guests have no saved value).
- `validate_attachment_ownership()` also allows the attachment when it is the submitter's saved value.
- Require the ID to still resolve to a live attachment before any of the three allowances applies. A stale ID (the admin deleted the media item) would otherwise pass on a saved value, then fail silently at the sink — `set_post_thumbnail()` ignores a dead ID — and be written straight back by `update_job_data()`, so the listing would publish with no logo and never self-heal. This also closes the same hole in `is_existing_listing_attachment()`'s meta branch: core clears `_thumbnail_id` when an attachment is deleted, but not `_company_logo` post meta.
- Make the rejection message recoverable. The rejected value is one the form itself offered back to the submitter, so a bare "Invalid attachment provided." leaves them with no way forward — it now names the field and says to re-upload or remove it. (The reporter was patching this with a `str_ireplace` in `functions.php`.) The malformed-URL throw site keeps the original string; it is a genuinely different condition.

## Testing

Adds coverage to `WP_Test_Submit_Job_Attachment_Ownership`. The first three were verified fail-before / pass-after:

- `test_saved_user_company_logo_is_accepted_on_new_submission` — the reported regression. Without the fix this throws the reporter's exact error.
- `test_deleted_saved_user_logo_is_rejected` — a saved logo whose attachment was deleted must error rather than silently publish a logo-less listing.
- `test_saved_user_logo_does_not_authorize_attachment_while_editing` — the allowance covers the new-submission path only.
- `test_foreign_attachment_still_rejected_on_new_submission` — the allowance is scoped to the submitter's actual saved value.
- `test_empty_saved_user_logo_does_not_authorize_foreign_attachment` — empty meta resolves to `0` and authorises nothing.
- `test_guest_submission_rejects_foreign_attachment` — the allowance never fires for logged-out submitters.
- `test_saved_user_meta_outside_company_group_does_not_authorize_attachment` — the `company` group guard, which had no coverage at all.
- `test_rejection_message_names_the_field_and_is_actionable` — locks in the message change.

Full class passes (19 tests). Full suite passes (615 tests; the 3 skips are the geocode tests needing a live API key). `phpcs` and Psalm clean.

Still uncovered, noted for follow-up: the draft-save path with a *valid* saved logo, and an end-to-end test of the priming chain (edit a listing with an admin-owned logo, assert the user meta is written, then submit a new listing). The two halves of that chain are currently tested separately and joined by a hand-written `update_user_meta()`.

----

| Plugin build for 068ca646868a03bfca9aea1bdf33da72b22656ee <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/07/wp-job-manager-zip-3082-068ca646.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/07/3082-068ca646)             |

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -981,18 +981,31 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 				foreach ( array_filter( $file_urls ) as $file_url ) {
 					if ( is_numeric( $file_url ) ) {
 						$attachment_id = absint( $file_url );
-						// The last two conditions are only safe because both consult server-side
+
+						if ( ! $attachment_id ) {
+							continue;
+						}
+
+						// The last two allowances are only safe because both consult server-side
 						// state, never the request: $this->job_id is set only for a listing the
 						// current user may edit (or is mid-submitting), and the user meta is
 						// written by this class only after this same check passed. They permit
 						// reusing a value already offered back to the submitter, NOT an arbitrary
 						// foreign ID — do not loosen either to accept request-supplied values.
-						if (
-								$attachment_id
-								&& ! $this->is_attachment_authorized_for_current_user( $attachment_id )
-								&& ! $this->is_existing_listing_attachment( $attachment_id, $key )
-								&& ! $this->is_saved_user_attachment( $attachment_id, $group_key, $key )
-							) {
+						//
+						// The existence check gates all three: a numeric ID that no longer resolves
+						// to an attachment (the media item was deleted) would otherwise pass on a
+						// stale saved value, fail silently at the sink — set_post_thumbnail()
+						// ignores a dead ID — and then be written straight back to the saved value,
+						// so the listing would publish with no logo and never self-heal.
+						$is_usable = 'attachment' === get_post_type( $attachment_id )
+							&& (
+								$this->is_attachment_authorized_for_current_user( $attachment_id )
+								|| $this->is_existing_listing_attachment( $attachment_id, $key )
+								|| $this->is_saved_user_attachment( $attachment_id, $group_key, $key )
+							);
+
+						if ( ! $is_usable ) {
 							// Tell the submitter how to recover. The rejected value is one the form
 							// offered them (a saved logo), so "invalid" alone leaves them stuck.
 							return new WP_Error(
@@ -1097,9 +1110,12 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 	protected function is_saved_user_attachment( $attachment_id, $group_key, $key ) {
 		$user_id = get_current_user_id();
 
-		// Mirrors the scope of the user-meta pre-population in submit(): logged-in
-		// submitters, company group only. Guests have no saved value to reuse.
-		if ( ! $user_id || 'company' !== $group_key ) {
+		// Mirrors the scope of the user-meta pre-population in submit(): a new listing
+		// (no job_id), a logged-in submitter, the company group only. Editing reads from
+		// the listing instead — is_existing_listing_attachment() covers that path — and
+		// guests have no saved value to reuse. Keeping this no broader than the
+		// pre-population it mirrors is what stops it becoming a general-purpose bypass.
+		if ( $this->job_id || ! $user_id || 'company' !== $group_key ) {
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


donnchawp:
@Hastibe - This will allow a user to use a company logo uploaded by an admin since they're already authorized to use it.

Hastibe:
Okay, thank you so much, @donnchawp -- I've overwritten the #3060 class-wp-job-manager-form-submit-job.php file I was using with [this one](https://github.com/Automattic/WP-Job-Manager/blob/296250b486705b9c5df5404daec021e734bfb2d5/includes/forms/class-wp-job-manager-form-submit-job.php) from here. Initial testing shows that it looks to be working as intended, and is much appreciated. 👍

donnchawp:
@Hastibe - thanks for testing that. There's been one more change since you grabbed that file and I think it matters for your sites, so it's worth overwriting it again with the [current version](https://github.com/Automattic/WP-Job-Manager/blob/068ca646868a03bfca9aea1bdf33da72b22656ee/includes/forms/class-wp-job-manager-form-submit-job.php).

Two things changed in 068ca64:

- The saved logo now has to still exist. In the version you have, if a submitter's saved _company_logo points at an attachment that was deleted, the ID passes validation, set_post_thumbnail() quietly ignores it, and then the dead ID gets written straight back to their user meta. The listing publishes with no logo and it never fixes itself. You mentioned orphaned metadata and broken image icons from your earlier cleanup, so I'd expect you to hit that one. It's now rejected with a message telling them to upload the logo again.
- The saved-logo allowance is limited to new listings. While editing, the value comes from the listing itself, and is_existing_listing_attachment() already covers that.

On the md5_file() point from the forum thread, I don't think that's what's biting you. By the time it runs, $attachment_url has already been rewritten to a local filesystem path (baseurl to basedir and so on), and there's a check that bails out if the result doesn't land under one of the site's own directories, so it never sees a URL. A saved logo posts back as a numeric ID anyway, and that path returns from validate_attachment_ownership() before any file is touched. What you're hitting is the ownership check that went out in 2.4.5.

Still working on the batched dedupe tool for the existing duplicates. I'll ping you when there's something to test.

Hastibe:
@donnchawp -- thanks for the heads-up, I appreciate it. Regarding the md5_file() point from the forum thread ([this one](https://wordpress.org/support/topic/core-bug-massive-logo-duplication-bloat-server-space/), I'm guessing?), I think you must be thinking of the user (ibiza69) who opened that thread (not me), which I just wanted to mention to avoid any confusion.

PR #3083, opened 2026-08-02
Prevent unauthorized job draft resumes

### LINKED ISSUE #1936: Preview job resuming with an other account.
**Describe the bug**
Hello,
Using the preview function when we post a job is resuming with a different account and with the same device (laptop). 
More details: 
When we are connected as for exemple " Alpha1" we click on post a job, we fill up all the necessary thing then We click on preview. a this moment if we disconnect and connect with an another account " Alpha2" And we click on post a job, There is a mention saying " You are editing an existing job" and job itself its the job i'v not finished with alpha1.
That curious. 

We are using:
Lastets versions of WPJOBManager and addons. 
Latest version of Wordpress
Latest version of Buddypresstheme and buddypressPlatforme.

benitoalba:
Fixes #1936

The resume cookie path checked the listing status and submitting key, but not whether the current visitor could edit the listing. On a shared device, a stale cookie could therefore resume another account's unfinished submission.

* Validate the submission cookie pair on `init` at priority 1, after job post types register and before submitted forms load.
* Clear cookies only when they are incomplete or reference an invalid ID, missing or non-job post, unsupported status, or mismatched submitting key.
* Ignore valid but currently unauthorized resumes without deleting their cookies, allowing an expired session to resume after the owner logs in again.
* Allow a valid guest-authored draft to resume after the visitor logs in; registered-user drafts still require `job_manager_user_can_edit_job()`.
* Retain status, key, post-type, and ownership checks in the submit form as defense in depth.
* Add regression coverage for both paid-listings flows, authentication transitions, ownership boundaries, and stale cookie variants.

### Testing Instructions

* `vendor/bin/phpunit --filter WP_Test_Submit_Job_Resume_Ownership` - 22 tests, 63 assertions.
* `vendor/bin/phpunit` - 629 tests, 2,236 assertions, with the 3 existing live-geocode tests skipped because no API key was configured.
* `vendor/bin/phpcs includes/class-wp-job-manager.php includes/forms/class-wp-job-manager-form-submit-job.php tests/php/tests/includes/test_class.submit-job-resume-ownership.php`.
* `php -l` on all three changed PHP files.
* Sent a real wp-env page request with an invalid job ID/key cookie pair and confirmed that the response included clearing `Set-Cookie` headers for both cookies before its redirect.

### Release Notes

* Prevent a job submission preview from being resumed by a different account on the same device without discarding a valid draft when authentication changes.

### DIFF
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -308,8 +308,8 @@ class WP_Job_Manager {
 		}
 
 		$job_id         = $has_job_id ? absint( $_COOKIE[ $job_id_cookie ] ) : 0;
-		$submitting_key = $has_job_key ? wp_unslash( $_COOKIE[ $job_key_cookie ] ) : '';
-		$job = $job_id ? get_post( $job_id ) : null;
+		$submitting_key = $has_job_key ? sanitize_text_field( wp_unslash( $_COOKIE[ $job_key_cookie ] ) ) : '';
+		$job            = $job_id ? get_post( $job_id ) : null;
 
 		if (
 			! $job instanceof WP_Post
--- a/includes/class-wp-job-manager.php
+++ b/includes/class-wp-job-manager.php
@@ -307,8 +307,9 @@ class WP_Job_Manager {
 			return;
 		}
 
-		$job_id         = $has_job_id ? absint( $_COOKIE[ $job_id_cookie ] ) : 0;
-		$submitting_key = $has_job_key ? sanitize_text_field( wp_unslash( $_COOKIE[ $job_key_cookie ] ) ) : '';
+		$job_id = $has_job_id ? absint( $_COOKIE[ $job_id_cookie ] ) : 0;
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- The submitting key is a bearer credential and must be compared verbatim.
+		$submitting_key = $has_job_key ? wp_unslash( $_COOKIE[ $job_key_cookie ] ) : '';
 		$job            = $job_id ? get_post( $job_id ) : null;
 
 		if (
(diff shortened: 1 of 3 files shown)


tellyworth:
Thanks for this @benitoalba. The bug is real and worth fixing. There are some gaps here though that will lead to data loss due to cookies being cleared: for example, someone who logs in or whose session expires midway through drafting a submission will lose work.

Also, I don't think the stale cookie cleanup works on a real page view, since `headers_sent()` will be true (it's always false in your tests).

I'd suggest rejecting via a mechanism that doesn't destroy cookies, and do the cookie validation in an early hook (like init) before headers are sent. Could you also add test coverage for a guest-to-login transition (guest saves a draft, logs in, resumes)?

benitoalba:
Thanks @tellyworth - I've addressed this feedback in the latest commits.

* Cookie validation now runs on `init` at priority 1, before headers and submitted forms.
* Only objectively stale cookie pairs are cleared. A valid draft that is temporarily unauthorized is ignored for that request while its cookies remain intact.
* Guest-authored drafts can resume after login, and registered-user drafts survive session expiry so the original owner can resume after logging back in.
* The form retains its status, key, post-type, and ownership checks as defense in depth.

I added coverage for both paid-listings flows, guest-to-login, session expiry/restoration, another logged-in user, invalid keys/statuses/posts, incomplete cookies, and empty cookie values. The focused suite passes with 22 tests and 63 assertions; the full suite passes with 629 tests and 2,236 assertions (3 existing live-geocode skips).

I also exercised a real wp-env page request with invalid cookies and confirmed that both clearing `Set-Cookie` headers are present in the response before the redirect.

Could you take another look when you have a chance? The GitHub Actions runs for the latest fork commit are awaiting maintainer approval.

PR #3092, opened 2026-08-11
Fix primary action button overflow on dashboard

### LINKED ISSUE #2829: Button overlap in Job Dashboard
**Describe the bug**
In some cases the button in the actions column will overlap another column

**To Reproduce**
1. Go to the Job Dashboard
2. Make the screen width about 930px
3. You can try adding Applications and applying to the job as well 
4. You can see the overlap

I think the problem is that the button is going outside of the parent flex container:

**Screenshots**

### LINKED ISSUE #2946: Primary action button on Job Dashboard overflows into adjacent column at narrow widths

**Describe the bug**

On the front-end Job Dashboard, the primary action button (`Continue Submission`, `Mark filled`, etc.) rendered in the Actions column is styled as a transparent outline button (`jm-ui-button--outline`) with `white-space: nowrap`. At narrower container widths the Actions column shrinks but the button cannot, so it overflows leftward and visually overlaps the Views / Date columns. Because the button has no background fill, the text behind it bleeds through and the dashboard looks broken.

**To Reproduce**

1. Open the Job Dashboard with at least one job in DRAFT state (so `Continue Submission` is the primary action) or ACTIVE (so `Mark filled` is the primary action).
2. View at a container width where the dashboard is rendered in its narrower layout (screenshot below taken at ~640 px).
3. Observe the outline button overlapping the Views column.

**Expected behavior**

The primary action should not overflow its column or overlap other column content. At widths where there isn't room for it, the action is still reachable via the three-dot menu and could be hidden, truncated, or stacked.

**Relevant code**

- Primary action rendering: `includes/class-job-dashboard-shortcode.php::the_primary_action()` — emits a `jm-ui-button--outline jm-ui-button--small` `<a>` into the Actions column.
- Row layout: `templates/job-dashboard.php:81-104` — Actions column shares the flex row with Title / Date / Views; all columns declare `flex: 1 1 calc(50% - var(--jm-ui-space-sm))` and rely on flex-shrink to fit.
- Column CSS: `assets/css/job-dashboard.scss` — `.jm-dashboard-job-column.actions` is `justify-content: flex-end` with no `min-width`; primary button has `white-space: nowrap` (via `.jm-ui-button*` base) and `flex-basis: fit-content`.

**Why ready-for-human, not agent**

The fix is a small CSS change but the design call isn't pre-decided:
- At what container width should the primary action be hidden/truncated/stacked?
- Hide vs truncate vs move-below-the-menu — each has different UX trade-offs.
- The dashboard already uses container queries elsewhere; the fix should align with the existing narrow-state breakpoints, which is a judgement call.

Once the design call is made, the implementation is ~10 lines of SCSS.

**Screenshot**

Image

faisalahammad:
Fixes #2946
Closes #2829

The primary action button on the front-end Job Dashboard (Continue Submission, Mark filled, etc.) is an outline button with `white-space: nowrap` and `flex-basis: fit-content`. At narrow container widths the Actions column shrinks, but the button cannot shrink with it, so it overflows to the left and overlaps the Views and Date columns. Because the button has no background fill, the text behind it bleeds through and the row looks broken.

This happens on both table variants: the small table at around 640px (#2946) and the large table at around 930px (#2829).

The fix lets the button shrink below its text width and truncates the inner label span with an ellipsis:

```scss
.jm-dashboard-action--primary {
	flex-basis: fit-content;
	white-space: nowrap;
	min-width: 0; /* allow the button to shrink below its text width */
	overflow: hidden;
	text-overflow: ellipsis;

	span {
		display: block; /* text-overflow does not apply to inline boxes */
		min-width: 0;
		overflow: hidden;
		text-overflow: ellipsis;
	}
}
```

The `span` needs `display: block`. `.jm-dashboard-action` sets `display: block` on the anchor and `job-dashboard.css` loads after `ui.css`, so the anchor is not the flex box `.jm-ui-button--outline` declares. That makes the inner `span` a non-replaced inline box, and CSS does not apply `overflow` or `text-overflow` to inline boxes. The `text-overflow` on the anchor is a fallback if a theme forces the span back to inline.

No PHP or template changes. The `<span>` wrapping the label already exists in `UI_Elements::button()`. The full action stays available in the three-dot overflow menu, which is built from the same `$job_actions` array, so truncating the standalone button never removes access to it.

This works for both table variants with a single rule, so no per-variant breakpoint is needed. The mobile layout (where the Actions column already wraps onto its own line) is not affected.

### Testing Instructions

1. Set up a Job Dashboard with at least one draft job (Continue Submission button) and one active job (Mark filled button).
2. Open the dashboard and narrow the browser window.
3. Small table (up to 4 columns): at around 640px container width, confirm the primary action button no longer overlaps the Views or Date columns. The label clips with an ellipsis.
4. Large table (5 or more columns): at around 930px container width, confirm the same behaviour (this is the #2829 case).
5. Open the three-dot overflow menu and confirm the full action is still listed and works.
6. Shrink to mobile width (under 540px small table, under 780px large table) and confirm no regression.

### Release Notes

* Fixed: the primary action button on the Job Dashboard no longer overlaps neighbouring columns at narrow widths. Long labels are truncated with an ellipsis, and the full action remains in the overflow menu.

*

### Deprecated Code

*

### Screenshot / Video

### DIFF
--- a/assets/css/job-dashboard.scss
+++ b/assets/css/job-dashboard.scss
@@ -138,8 +138,10 @@
 	white-space: nowrap;
 	min-width: 0; /* allow the button to shrink below its text width */
 	overflow: hidden;
+	text-overflow: ellipsis;
 
 	span {
+		display: block; /* text-overflow does not apply to inline boxes */
 		min-width: 0;
 		overflow: hidden;
 		text-overflow: ellipsis;
--- a/assets/css/job-dashboard.scss
+++ b/assets/css/job-dashboard.scss
@@ -136,6 +136,14 @@
 .jm-dashboard-action--primary {
 	flex-basis: fit-content;
 	white-space: nowrap;
+	min-width: 0; /* allow the button to shrink below its text width */
+	overflow: hidden;
+
+	span {
+		min-width: 0;
+		overflow: hidden;
+		text-overflow: ellipsis;
+	}
 }
 
 .jm-dashboard-action:where(:not(:hover):not(:focus)) {


tellyworth:
The ellipsis doesn't work for me. Here's what I get in Firefox, Safari is similar:


Screenshot 2026-08-31 at 1 31 48 pm
Screenshot 2026-08-31 at 1 31 33 pm

faisalahammad:
Thanks for testing. You were right, the ellipsis did nothing.

Cause: `.jm-dashboard-action` sets `display: block`, and `job-dashboard.css` is registered with `wp-job-manager-ui` as a dependency, so it loads after `ui.css`. Both selectors have the same specificity, so `display: block` wins over the `display: flex` that `.jm-ui-button--outline` declares. The anchor is then a block box, which makes the inner `<span>` a non-replaced inline box, and CSS does not apply `overflow` or `text-overflow` to inline boxes. So both span rules were dead and the label was just hard-clipped.

Fix in b70cdb81, still CSS only:

- `display: block` on the span, so it becomes a real box and the ellipsis applies. This also holds if the anchor ever resolves to `display: flex`, since flex items are blockified.
- `text-overflow: ellipsis` on the anchor as a fallback, in case a theme forces the span back to inline.

Please retest in Firefox and Safari at around 640px (small table) and around 930px (large table).

PR #3093, opened 2026-08-12
fix: show an error when a job dashboard duplicate fails

### LINKED ISSUE #2884: Enable Duplication for Active Status Job Listings
The current feature to Duplicate a job listing in the employer's job dashboard only works for Draft and Pending jobs. It does not work for Active or Expired jobs which seems to make little sense especially when the action remains visible for all job status'. 

If you click on Duplicate for an active job nothing happens. What could be more confusing? if it's not available at least hide it.

Please consider expanding the ability to duplicate a job to active and expired job listings.

faisalahammad:
Fixes #2884

The "Duplicate" action in the employer job dashboard already worked for listings in any status, including active (publish) and expired. The only real defect was a silent failure: when `job_manager_duplicate_listing()` returned 0, the handler fell through with no message and just reloaded the dashboard, which looked like the button did nothing.

This change makes that failure visible and actionable:

- `job_manager_duplicate_listing()` gains an optional `$wp_error` parameter. When it is `true` the function returns a `WP_Error` describing the failure instead of a bare `0`, so callers can tell a missing ID, a non-listing post, and a failed insert apart. The default stays `false` and still returns `0`, so the documented return contract is unchanged for existing callers.
- The internal `wp_insert_post()` call now passes `$wp_error = true` and bails out as soon as the insert fails. Previously the return value was never checked, so a failed insert let the function carry on and call `wp_set_object_terms( 0, ... )`, which writes orphan `term_relationships` rows against object ID 0. (The `update_post_meta( 0, ... )` calls on the same path were already no-ops, since `update_metadata()` bails on a falsy object ID.)
- The job dashboard now shows an error notice with a title and a next step when duplication fails, matching the error pattern used by the other dashboard actions (`mark_filled`, `mark_not_filled`, `relist`, `renew`):

  > **Could not duplicate {job title}**
  > The copy could not be saved. Please try again. If this keeps happening, contact the site administrator.

  The underlying `WP_Error` message is deliberately not shown to the employer, because WordPress core embeds `$wpdb->last_error` in its `db_insert_error` message and this notice renders on a public-facing page.

- The `job_manager_job_dashboard_success_message` filter is now skipped when the action failed. Without that guard a third-party filter returning text for the `duplicate` action would overwrite the error notice with a success notice.

A pre-existing test gap was also covered with regression tests against `job_manager_duplicate_listing()`.

### Testing Instructions

Install this branch on a test site.

Success path:

1. On the employer job dashboard (`[job_dashboard]` shortcode), open a row for an active (published) job.
2. Click "Duplicate". Confirm you are redirected to the submission page with the form pre-filled.
3. Repeat the same on a row for an expired job. Same pre-filled form, same redirect.
4. Repeat on a draft row to confirm draft duplication still works.

Failure path (this is what the PR adds). Force the underlying post insert to fail by dropping this into a mu-plugin:

```php
<?php
add_filter( 'wp_insert_post_empty_content', '__return_true' );
```

1. With the filter active, click "Duplicate" on any job in the dashboard.
2. Confirm the dashboard reloads showing the error notice titled "Could not duplicate {job title}", instead of doing nothing.
3. Remove the mu-plugin and confirm "Duplicate" works normally again.

Automated checks:

- PHPUnit: 613 tests, 2199 assertions, all pass (3 geocode tests skipped, they need a Google API key). Six tests cover `job_manager_duplicate_listing()`: duplicating from `publish` and from `expired` source statuses, the `0` return for an empty ID and for a non-listing post type, the matching `WP_Error` codes when `$wp_error` is `true`, and a forced insert failure asserting no listing is created and no term row is written against object ID 0. The last one was confirmed to fail without the early bail.
- `./vendor/bin/phpcs`: clean on both changed source files.

### Known limitation

The dashboard notice is carried across the redirect by `set_transient()`. On a site with no persistent object cache that is a `wp_options` write, so on a site whose database writes are failing the notice itself can fail to persist and the employer would still land on a clean dashboard with no feedback. That is a property of the existing dashboard notice mechanism, shared by every other action, and is not something this PR changes.

### Release Notes

Duplicating a job listing now shows an error notice explaining what to do next when the action fails, instead of silently reloading the dashboard.

None. `job_manager_duplicate_listing()` gains an optional `$wp_error` parameter; the default behaviour and return type are unchanged.

### Deprecated Code

None.

### Screenshot / Video

No visible UI change on the success path. The error notice only appears on the previously silent failure case.

### DIFF
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -484,7 +484,8 @@ class Job_Dashboard_Shortcode {
 						exit;
 					}
 
-					break;
+					// translators: Placeholder %s is the job listing title.
+					throw new \Exception( sprintf( __( 'Could not duplicate %s', 'wp-job-manager' ), wpjm_get_the_job_title( $job ) ) );
 				case 'relist':
 				case 'renew':
 				case 'continue':
--- a/includes/class-job-dashboard-shortcode.php
+++ b/includes/class-job-dashboard-shortcode.php
@@ -436,6 +436,8 @@ class Job_Dashboard_Shortcode {
 				throw new \Exception( __( 'Invalid ID', 'wp-job-manager' ) );
 			}
 
+			$action_failed = false;
+
 			switch ( $action ) {
 				case 'mark_filled':
 					// Check status.
@@ -477,15 +479,24 @@ class Job_Dashboard_Shortcode {
 						throw new \Exception( __( 'Missing submission page.', 'wp-job-manager' ) );
 					}
 
-					$new_job_id = job_manager_duplicate_listing( $job_id );
+					$new_job_id = job_manager_duplicate_listing( $job_id, true );
 
-					if ( $new_job_id ) {
+					if ( ! is_wp_error( $new_job_id ) && $new_job_id ) {
 						wp_safe_redirect( add_query_arg( [ 'job_id' => absint( $new_job_id ) ], job_manager_get_permalink( 'submit_job_form' ) ) );
 						exit;
 					}
 
-					// translators: Placeholder %s is the job listing title.
-					throw new \Exception( sprintf( __( 'Could not duplicate %s', 'wp-job-manager' ), wpjm_get_the_job_title( $job ) ) );
+					// Message.
+					$action_failed               = true;
+					$this->job_dashboard_message = Notice::error(
+						[
+							// translators: Placeholder %s is the job listing title.
+							'title'   => sprintf( __( 'Could not duplicate %s', 'wp-job-manager' ), wpjm_get_the_job_title( $job ) ),
+							'message' => __( 'The copy could not be saved. Please try again. If this keeps happening, contact the site administrator.', 'wp-job-manager' ),
+						]
+					);
+
+					break;
 				case 'relist':
 				case 'renew':
 				case 'continue':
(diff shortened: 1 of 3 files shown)


tellyworth:
> for example database write failure on that site

A site experiencing DB write failures has serious problems. This PR just produces a vague and un-actionable error message ("Could not duplicate..").

Error handling is good, but it needs to give the user an indication of what went wrong and what they should do next.

faisalahammad:
Fair point. Reworked it.

The notice now has a title and a next step:

> **Could not duplicate {job title}**
> The copy could not be saved. Please try again. If this keeps happening, contact the site administrator.

`job_manager_duplicate_listing()` takes an optional `$wp_error` parameter and returns a real `WP_Error`, so the failure paths are no longer flattened into a single `0`. The default stays `false` and still returns `0`, so existing callers do not change.

I left the raw `WP_Error` message out of the notice on purpose. Core puts `$wpdb->last_error` in its `db_insert_error` message, and this renders on a public page. Tell me if you would rather have it logged or gated behind `manage_options`.

Two things came out of this:

- A real bug on the same path. `wp_insert_post()` was called without the error flag and its return was never checked, so a failed insert kept going and called `wp_set_object_terms( 0, ... )`, writing orphan `term_relationships` rows. It bails now. Regression test included, confirmed failing without the fix.
- Guarded `job_manager_job_dashboard_success_message` so it is skipped when the action failed. Otherwise a site filtering `duplicate` would replace the error with a success notice.

One limitation, since it sits on your premise: the notice crosses the redirect in a transient, which is a `wp_options` write without a persistent object cache. On a site failing database writes the notice can fail too. That applies to every dashboard action today, so I left it out of this PR.

Testing instructions now use `wp_insert_post_empty_content` for a concrete repro.

PR #3094, opened 2026-08-12
fix: store scheduled listing dates in site timezone

### LINKED ISSUE #2879: Issues with Scheduled Job Listings and Date Handling
**Describe the bug**
A user in 9588796-zd-a8c has reported issues with date/time handling when scheduling job listings in non-UTC timezones.

**1. post_date Field Incorrectly Adjusted for Timezone**

Issue: When scheduling job listings, post_date in the database incorrectly includes the timezone offset, causing jobs to display at an incorrect local time (e.g., 11 AM instead of midnight in Australia/Sydney UTC+11).

Cause: strtotime($maybe_date_string) assumes the input is UTC+0 and applies an offset incorrectly.

Expected Behavior: post_date should store local time, and post_date_gmt should be the correct UTC equivalent.

Suggested Fix: Modify maybe_format_future_datetime() to append the correct timezone offset before processing.

**2. post_date and post_date_gmt Incorrectly Matching**

Issue: When a scheduled date is set, post_date and post_date_gmt store the same value instead of differentiating local and GMT times.

Cause: Both fields are set to maybe_formatted_date instead of calculating post_date_gmt relative to local time.

Expected Behavior:

post_date: Local scheduled time.

post_date_gmt: Correct UTC time.

Affected Code: wp-job-manager-form-submit-job.php ([lines 1086-1096](https://github.com/Automattic/WP-Job-Manager/blob/trunk/includes/forms/class-wp-job-manager-form-submit-job.php#L1086)).

**3. jQuery Date Picker Lacks Time Selection**

Issue: The date picker does not allow users to set a time for job listings, defaulting to 00:00 UTC+0, which leads to incorrect local scheduling.

Expected Behavior: A datetime picker should be used to select both date and time, ensuring proper scheduling.

Suggested Fix: Replace the current jQuery date picker with a datetime picker (jQuery DateTime.js). 

**To Reproduce**
1. Set timezone in general settings to UTC+11
2. Create a job listing and schedule it to be published on a future date.
3. Listing will be published at 11am (UTC+11) the next day instead of 00:00

**WordPress Environment**
- WordPress Version: 6.7.2
- PHP Version: 8.2.28

faisalahammad:
Fixes #2879

Scheduled job listings stored the same value in `post_date` and `post_date_gmt`, so on a non-UTC site the job published at the wrong wall-clock time (a UTC+11 site showed 11 AM instead of midnight). The scheduled date was also parsed with bare `strtotime()`, which assumes UTC and re-rendered the offset only in the local field.

This change parses the scheduled date in the site timezone and writes distinct local and GMT values:

- `format_scheduled_datetime()` replaces the old `maybe_format_future_datetime()` logic. The date is parsed with `DateTimeImmutable::createFromFormat( 'Y-m-d', ..., wp_timezone() )`, the time component is set explicitly, and the GMT value is computed by re-projecting the local instant onto UTC. Past, empty or invalid values fall back to the current time, matching the previous behavior.
- `apply_scheduled_date()` now writes `post_date` (site local) and `post_date_gmt` (UTC) from that pair instead of duplicating one string into both columns.
- The scheduled date field gained an optional HTML5 time input. A job can now be scheduled at an exact time, not just midnight. It only renders for the scheduled listing field, so existing date fields and submit-form integrations are unchanged unless they opt in.
- A new `get_posted_date_field()` handler merges the date and (optional) time inputs on submit.

### Testing Instructions

Environment: WordPress 6.4+, PHP 7.4+. Enable scheduled listings in WPJM settings, then set the site timezone to `Australia/Sydney` (UTC+11 in January/summer, UTC+10 in winter).

1. Submit a job and set the scheduled date to tomorrow, leaving the time empty. Expected: `post_date` is the local midnight (`00:00:00`), and `post_date_gmt` is the matching UTC instant (for Sydney in summer that is one day earlier at `13:00:00`). Before the fix `post_date` was 11 AM and both columns held the same value.
2. Submit another job with both a date and a time, for example tomorrow at 09:30. Expected: `post_date` stores `09:30:00`, not midnight.
3. Switch the site timezone to UTC and submit a job with a scheduled date. Expected: `post_date` equals `post_date_gmt`.
4. Set a manual fractional offset (for example UTC+5:30) and submit a job with a scheduled date. Expected: `post_date_gmt` is the correct projection (a `2030-01-02 00:00` local job stores `2030-01-01 18:30:00`).
5. Edit a scheduled job that has a time set. Expected: the date input shows only the date and the time input shows only the time, both pre-filled.
6. Submit a job with a past date, and one with the field left empty. Expected: the job publishes immediately in both cases.

Automated coverage: new `tests/php/tests/includes/test_class.wp-job-manager-form-submit-job.php` with 8 tests and 24 assertions covering UTC, UTC+11, fractional manual offset, explicit time, past date, empty, invalid, and invalid-time fallback paths.

### Release Notes

Fix scheduled job listings publishing at the wrong time on non-UTC sites, and allow scheduling at an exact time.

No new or updated actions or filters. The `date` field template gains an optional time input only when a field sets `enable_time`; the scheduled listing field is the only current user of it.

### Deprecated Code

No code is deprecated. The private `maybe_format_future_datetime()` method was replaced by `format_scheduled_datetime()` within the same class.

### Screenshot / Video

No screenshot: this is a backend date/time storage fix. The only visible change is an optional time input next to the scheduled date field.

### DIFF
--- a/includes/forms/class-wp-job-manager-form-submit-job.php
+++ b/includes/forms/class-wp-job-manager-form-submit-job.php
@@ -406,9 +406,10 @@ class WP_Job_Manager_Form_Submit_Job extends WP_Job_Manager_Form {
 		if ( get_option( 'job_manager_enable_scheduled_listings' ) ) {
 			$this->fields['job']['job_schedule_listing'] = [
 				'label'       => __( 'Scheduled Date', 'wp-job-manager' ),
-				'description' => __( 'Optionally set the date when this listing will be published.', 'wp-job-manager' ),
+				'description' => __( 'Optionally set the date and time when this listing will be published.', 'wp-job-manager' ),
 				'type'        => 'date',
 				'required'    => false,
+				'enable_time' => true,
 				'placeholder' => '',
 				'priority'    => '6.5',
 			];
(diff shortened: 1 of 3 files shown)


tellyworth:
This introduces some regressions that weren't previously covered by tests; I added some new tests in trunk to cover them. They're in the failing checks above, you'll see them locally if you rebase.

faisalahammad:
Thanks for the tests. Both regressions are fixed and the branch is rebased on trunk.

**1. Human-readable dates were rejected.** `format_scheduled_datetime()` parsed with `DateTimeImmutable::createFromFormat( 'Y-m-d', ... )`, so only ISO dates were accepted. Without JavaScript `datepicker.js` never creates the hidden ISO field, so the visible input posts its display text ("August 31, 2028") and those listings published immediately. It now parses with `new DateTimeImmutable( $value, wp_timezone() )` inside a try/catch. That accepts any `strtotime()`-parseable value like the old code, keeps the site timezone, and gives a date-only value midnight without the extra `setTime()` step. The manual `explode( ' ' )` time split is gone, since it broke on any date containing spaces.

**2. Required date fields never failed validation.** The new `get_posted_date_field()` read `$_POST` directly and skipped the `before_sanitize` callback that `get_posted_fields()` wires onto every field. That callback sets the `empty` flag, which `validate_fields()` checks. It now calls `get_posted_field()` for the date and appends the optional time after, so the callback and any custom sanitizer still run.

I also fixed a related problem in `templates/form-fields/date-field.php`. It split on any space and ran `substr( $value, 0, 10 )` for every date field, which truncated stored values. It now splits only when the field sets `enable_time`, and only on a real trailing `H:i`. The optional time input no longer copies `required` from the field.

Local run: 626 tests, 2215 assertions, 0 failures, 3 skipped (Geocode, needs an API key). PHPCS and Psalm are clean.

PR #3101, opened 2026-08-31
fix: emit JSON-LD with JSON escapes instead of HTML entities

### LINKED ISSUE #3027: JSON-LD structured data is HTML-escaped inside the script tag, corrupting the payload
Follow-up to #3026, which I deliberately kept out of that PR because it is security-adjacent and deserves its own review.

## Problem

`includes/class-wp-job-manager-post-types.php:1834` emits the JobPosting structured data like this:

```php
echo '<script type="application/ld+json">' . wpjm_esc_json( wp_json_encode( $structured_data ), true ) . '</script>';
```

`wpjm_esc_json()` (`wp-job-manager-functions.php:1818`) runs the encoded JSON through `_wp_specialchars( $json, ENT_NOQUOTES, 'UTF-8', true )` — note `$double_encode = true`.

The content of a `<script>` element is **raw text** per the HTML spec. Character references are not parsed there. So the HTML escaping does not get undone by the consumer: it is baked into the JSON payload.

## Evidence

Run against a real install:

```
=== What ships today, and can a JSON-LD consumer read it? ===
emitted : {"title":"Events &amp; Marketing coordinator","description":"Sales &lt;\/script&gt;&lt;script&gt;alert(1)&lt;\/script&gt; role"}
json_decode -> title = 'Events &amp; Marketing coordinator'
```

A conforming parser reads the literal string `Events &amp; Marketing coordinator`. Two consequences:

1. **Titles.** Any `&`, `'` or `"` in a job title reaches Google for Jobs as an entity. (#3026 improved this from `&amp;#038;` to `&amp;`, but could not fix it here.)
2. **Descriptions.** `description` legitimately carries HTML, and today every tag ships as `&lt;p&gt;`, `&lt;br /&gt;` and so on. This is arguably the more serious half — worth confirming what Google actually does with an escaped description before deciding how urgent this is.

## Proposed fix

Stop HTML-escaping the JSON. Ask `json_encode` to escape the dangerous characters *as JSON unicode escapes*, which parsers decode back to the correct characters:

```php
echo '<script type="application/ld+json">'
	. wp_json_encode( $structured_data, JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT )
	. '</script>';
```

Verified:

```
emitted : {"title":"Events & Marketing coordinator","description":"Sales <\/script><script>alert(1)<\/script> role"}
json_decode -> title = 'Events & Marketing coordinator'
round-trips exactly? YES

contains literal "<"        ? no
contains literal "</script" ? no
contains literal "<!--"     ? no
```

`JSON_HEX_TAG` is doing the security work. It matters for more than the obvious `</script>` breakout: a literal `<!--` followed by `<script` inside script raw text puts the tokenizer into the *script data escaped* states, where the next `</script>` no longer closes the element. Escaping `<` and `>` to `<` / `>` removes every one of those sequences. `JSON_HEX_AMP` is what actually fixes the reported corruption.

## Scope

- `wpjm_esc_json()` has exactly **one** production call site, the line above. Its `$html = false` branch (`ENT_QUOTES`) is dead code — nothing calls it with the default argument.
- The function is public, in `wp-job-manager-functions.php`, so an add-on could theoretically call it. I would leave it in place rather than remove it, and mark it `@deprecated $$next-version$$` if we want it gone.
- `tests/php/tests/includes/test_class.wp-job-manager-post-types.php:985` references it, but computes the expected value with the same call, so the assertion is tautological and will not catch a change. It needs replacing with an assertion that `json_decode()` of the emitted script content equals `wpjm_get_job_listing_structured_data()`.

## Acceptance criteria

- [ ] `json_decode()` of the emitted `<script type="application/ld+json">` content round-trips to the exact array returned by `wpjm_get_job_listing_structured_data()`, entities and all.
- [ ] Emitted script content contains no literal `<`, `</script`, or `<!--`.
- [ ] Regression test covers a title containing `& ' "` and a description containing `</script><script>`.
- [ ] The tautological assertion in `test_class.wp-job-manager-post-types.php` is replaced.

## Why `ready-for-human`

The change itself is small, but it removes an escaping layer from an output sink. The reasoning about script-raw-text tokenizer states above should be checked by a person rather than taken on trust.

tellyworth:
## What

Fixes #3027. The JobPosting structured data was run through `wpjm_esc_json()` (HTML entity escaping) before being echoed into its `<script type="application/ld+json">` block. Script-element content is raw text per the HTML spec — character references are never decoded there — so the escaping was baked into the payload: consumers read `Events &amp;amp; Marketing` titles and `&amp;lt;p&amp;gt;`-riddled descriptions.

## Fix

Encode with `wp_json_encode( $data, JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT )`, per the issue's proposal. The dangerous characters become JSON unicode escapes (`<` etc.), which JSON parsers decode back to the original characters — and the payload still cannot contain a literal `</script>`, so the security property is preserved at the right layer.

`wpjm_esc_json()` had exactly one caller (this one) and is deprecated with a pointer to the right approach; its phpcs `customEscapingFunctions` entry is left in place for back-compat.

## Tests

- New `test_output_structured_data_round_trips_without_html_entities`: extracts the emitted payload and asserts (a) it contains no literal `<` or `&`, (b) `json_decode` returns exactly the structured data array, (c) an ampersand in the company name survives as a plain ampersand, (d) description HTML survives as HTML.
- The existing `test_output_structured_data` expectation updated from the old (tautological) escaping to the new encoding.
- Full suite green locally: 616 tests, 3 known geocode skips.

Note: what the *title itself* contains (`&#038;` from the `the_title` filter chain) is upstream of emission and tracked by #3026's work — this PR makes emission faithful to whatever the structured data contains.

----

| Plugin build for 317eb454de905e2d049bed3031ebfadc9343318e <a href="#"></a> |
| 📦 [Download plugin zip](https://wpjobmanager.com/wp-content/uploads/2026/08/wp-job-manager-zip-3101-317eb454.zip)                       |
| ▶️ [Open in playground](https://wpjobmanager.com/playground/?core=2026/08/3101-317eb454)             |

### DIFF
--- a/includes/class-wp-job-manager-post-types.php
+++ b/includes/class-wp-job-manager-post-types.php
@@ -1831,7 +1831,11 @@ class WP_Job_Manager_Post_Types {
 		$structured_data = wpjm_get_job_listing_structured_data();
 		if ( ! empty( $structured_data ) ) {
 			echo '<!-- WP Job Manager Structured Data -->' . "\r\n";
-			echo '<script type="application/ld+json">' . wpjm_esc_json( wp_json_encode( $structured_data ), true ) . '</script>';
+			// Script-element content is raw text: HTML entities are never decoded there, so
+			// the payload must use JSON escapes (< etc.), not HTML entities. The HEX
+			// flags escape <, >, &, ' and " so the JSON cannot close the script element.
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- The JSON_HEX_* flags escape the payload for the script element.
+			echo '<script type="application/ld+json">' . wp_json_encode( $structured_data, JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT ) . '</script>';
 		}
 	}
 
--- a/tests/php/tests/includes/test_class.job-title-entity-decoding.php
+++ b/tests/php/tests/includes/test_class.job-title-entity-decoding.php
@@ -3,10 +3,11 @@
  * Tests that a job listing title is HTML-entity decoded before it is placed
  * into a plain-text context (email subjects, the plain-text email body).
  *
- * The JSON-LD structured data has the same defect but is not fixed here: the
- * emitted `<script type="application/ld+json">` block is re-escaped by
- * `wpjm_esc_json()` with `$double_encode = true`, so decoding the title alone
- * does not change what the consumer reads. See #3027.
+ * The JSON-LD structured data had the same defect, fixed separately in #3027:
+ * the emitted `<script type="application/ld+json">` block is now JSON-escaped
+ * (`wp_json_encode()` with the `JSON_HEX_*` flags) instead of HTML-escaped, so
+ * the payload round-trips; what the title itself contains is still governed by
+ * the storage/filter pipeline covered here.
  *
  * WordPress core hooks `wp_filter_kses()` onto `title_save_pre` for any user
  * without the `unfiltered_html` capability, so a listing submitted by a guest,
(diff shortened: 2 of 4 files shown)


tellyworth:
I checked the output with https://search.google.com/test/rich-results and it looks good to me, but I'm no expert.

PR #3116, opened 2026-09-15
feat: add date range filtering to [jobs] shortcode

### LINKED ISSUE #2306: Add Url query based on date range
This has been requested in 5466596-zen: 

>I wanted to reach out and ask question about how I would go about setting a date range if possible in a url query for the most recent listings on our site. But the tricky thing we would like to be to control the date range that the listings fall under. So an example would be all the listings from 08/01/2022 - 08/08/2022 would be what we want the url query to pull. 

We have [shortcodes](https://wpjobmanager.com/document/shortcode-reference/) to show the most recent jobs without filters and pagination, but we haven't implemented it with the date ranges selection feature yet. 

I noticed the [sMyles' Search and Filtering plugin](https://plugins.smyl.es/wp-job-manager-search-and-filtering) has this specific feature not deployed yet, as it's still in development: 

>Field Types in Development
Date Picker (Range & Single)

faisalahammad:
Fixes #2306

* Add `posted_after` and `posted_before` args to `get_job_listings()`. Both accept `Y-m-d` only and are applied as a native `WP_Query` `date_query` on `post_date` with `inclusive => true`, so both boundary days are included.
* Read the same two names from the URL in the `[jobs]` shortcode and from the AJAX request, and pass them through as data attributes, so a range survives pagination and filter updates.
* The range is added to the query args before the result cache hash, so each date range gets its own transient instead of sharing one with the unfiltered board.
* Malformed input fails closed and shows no listings, matching how the existing `author` filter handles bad input. A restriction was asked for, so it is honoured strictly or not at all, never silently widened to the full listing set.

Usage, both work the same way:

```
[jobs posted_after="2022-08-01" posted_before="2022-08-08"]
```

```
https://example.com/jobs/?posted_after=2022-08-01&posted_before=2022-08-08
```

Either bound can be supplied on its own. There is no filter-form UI for this, per the scope of the issue; sites that want date pickers keep using their own markup. The RSS link still drops the range, that is left as a separate follow-up.

### Testing Instructions

1. Create listings with known publish dates, at least one before the range, one on each boundary day, and one after. `wp post create --post_type=job_listing --post_date="2022-08-01 09:00:00"` then adjust the date for each.
2. Visit the job board URL with `?posted_after=2022-08-01&posted_before=2022-08-08`. Only the three listings inside the range are shown, including both boundary days.
3. Use only `?posted_after=2022-08-01`, then only `?posted_before=2022-08-08`. Each shows the expected half of the range.
4. Drop the `[jobs]` shortcode on a page with `posted_after` / `posted_before` attributes. Same result as the URL.
5. With pagination in use, click to page 2. The range is still applied.
6. If the theme or an extension enables job filters, change a filter or paginate. The AJAX request keeps the range.
7. Try `?posted_after=08/01/2022` and `?posted_after=2022-02-30`. Both show no listings rather than falling back to the full board.
8. Load the job board with no date params. The listings are unchanged from before this PR.

Unit tests cover the range including both boundary days, one-sided bounds, invalid and impossible dates, array-shaped input, and that two different ranges do not share a cache entry.

### Release Notes

* Add `posted_after` and `posted_before` to the `[jobs]` shortcode and to the job board URL query, so a listing date range can be filtered without custom code.

### Screenshot / Video

No visible UI change. This is a query parameter and shortcode attribute only, so there is nothing to screenshot.


Gnodesign:
Commenting to stay in the loop

